PR #139, opened 2015-01-01
Updated license in setup.py

jfsantos:
LICENSE.md states that the license for this project is now the ISC license, and commit messages also state that the project switched from GPL to ISC. This PR updates setup.py, so PyPI will (hopefully) show the correct license from now on.


bmcfee:
Thanks for catching that!

PR #146, opened 2015-01-15
CQT Enhancements

ebattenberg:
This adds:

1. An aliasing check when creating CQT filters
2. Changes the CQT oversampling/aggregation hop to be based on the shortest filter rather than the longest filter.
3. Adds 1/N Parseval normalization to CQT output.
4. Adds a `norm` argument to cqt and filters.constant_q in order to allow normalization using a different norm.
5. Does "sparsification" on fft_basis in order to exploit sparse ops.  (Without this there's no reason to do the processing in the frequency domain).

Let me know if any of these are useful.  I'm happy to revert any of them if a subset are useful.  Not sure what kind of lint requirements you have.  Also, I wasn't able to run the tests because I don't have matlab.  Why is Matlab a requirement for a Python library. ;)

### DIFF
--- a/librosa/core.py
+++ b/librosa/core.py
@@ -708,6 +708,7 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
     # Conjugate-transpose the basis
     fft_basis = np.fft.fft(basis, n=n_fft, axis=1).conj()
+    fft_basis = sparsify_fft_basis(fft_basis)
 
     n_octaves = int(np.ceil(float(n_bins) / bins_per_octave))
 
@@ -777,6 +778,33 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
     # Transpose magic here to ensure column-contiguity
     return np.ascontiguousarray(cqt_resp.T).T
+
+@cache
+def sparsify_fft_basis(fft_basis, sparse_limit=0.01):
+    '''
+    Return a sparse version of input fft_basis.
+
+    `sparse_limit` is the percentage of the original magnitude sum that can
+    be zeroed out for each filter.
+    '''
+    fft_basis = fft_basis.copy()
+
+    if sparse_limit > 0.0:
+        for i, kern in enumerate(fft_basis):
+
+            kern_mag = np.abs(kern)
+            band_sum = np.sum(kern_mag)
+            sorted_weights = np.sort(kern_mag)[::-1]
+            cumpct = np.cumsum(sorted_weights)/band_sum
+            cutoff_ind = np.where(cumpct >= (1-sparse_limit))[0][0]
+            cutoff_val = sorted_weights[cutoff_ind]
+            zero_inds = np.where(kern_mag < cutoff_val)
+            fft_basis[i][zero_inds] = 0.0
+
+        import scipy.sparse
+        fft_basis = scipy.sparse.csr_matrix(fft_basis)
+
+    return fft_basis
 
 
 @cache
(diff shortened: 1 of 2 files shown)


bmcfee:
These all sound like great changes. Thanks!

I'd rather merge this into the 0.4 development branch rather than master.  However, develop just underwent a heavy refactor over the past week, which I suspect will make auto-merging from this PR quite impossible.  I'll give it a shot later this week.


> Also, I wasn't able to run the tests because I don't have matlab. Why is Matlab a requirement for a Python library. ;)

The test fixtures should be pre-generated, so you shouldn't need matlab anymore.  (Not that we have any regression tests covering cqt right now anyway...)

ebattenberg:
Sounds good.  Keep me posted on the refactor.  I can help with the rebase.  I may have some more commits to add soon.

bmcfee:
> Not sure what kind of lint requirements you have.

This is buried in the wiki [here](https://github.com/bmcfee/librosa/wiki/Coding-style)

bmcfee:
Merged into develop.

bmcfee:
Hitting the points in order:

> An aliasing check when creating CQT filters

Good catch.

> Changes the CQT oversampling/aggregation hop to be based on the shortest filter rather than the longest filter.

Care to elaborate on this for the changelog?

> Adds 1/N Parseval normalization to CQT output.

Makes sense.

> Adds a norm argument to cqt and filters.constant_q in order to allow normalization using a different norm.

Good catch.

> Does "sparsification" on fft_basis in order to exploit sparse ops. (Without this there's no reason to do the processing in the frequency domain).

I'm not convinced this is totally necessary -- it's all dot products in the end, and still faster than time-domain convolution -- but it probably can't hurt.

ebattenberg:
>>Does "sparsification" on fft_basis in order to exploit sparse ops. (Without this there's no reason to do the processing in the frequency domain).

>I'm not convinced this is totally necessary -- it's all dot products in the end, and still faster than time-domain convolution -- but it probably can't hurt.


You're right that doing convolution in the frequency domain can be more efficient; however, in this case, we're not doing convolution but projection onto a basis functions (in the same way that computing the DFT is projecting a signal onto a basis of complex exponentials)

By Parseval's: sum_n x(n)y(n)* = 1/N sum_k X(k)Y(k)*

Computing the sums on either side of the above equation is the exact same time complexity (O(N)) but the one on the right additionally requires X(k) be computed from x(n) (O(NlogN)).  If the sparsity of X(k) is exploited then we get an efficiency win.  Otherwise, it's just an efficiency loss due to the unnecessary FFT.

ebattenberg:
>>Changes the CQT oversampling/aggregation hop to be based on the shortest filter rather than the longest filter.

>Care to elaborate on this for the changelog?

Sure.  The CQT basis functions are time limited by Hann windows of varying sizes.  When using a Hann window for the STFT, the largest hop size that is typically used is half of the window size.  Choosing a value larger than this will result in an undersampling of the time-frequency content.  Likewise, when choosing a hop size for a CQT basis function, the largest hop size should be half of the size of the window used for the basis function.  What was done previously when oversampling an octave before aggregation was to use half of the FFT size.  But the FFT size is chosen to be the smallest power of 2 that is greater than or equal to the *largest* basis function.  This means not only are the largest basis functions probably undersampled in time (e.g. a max basis function length of 184 yields an FFT size of 256 with oversampled hopsize of 128), but the smallest basis function is even more severely undersampled (e.g. a shortest basis function of length 50 with FFT size 256 and hop size 128).  

Unfortunately for efficiency reasons, a shortest basis function of length 50, should have a hop size no bigger than 25, and the smallest power of 2 not greater than 25 is 16, which is a lot smaller than 128.  This is something that should be taken into consideration when choosing very wide-band high-frequency filters.   I was considering exploring allowing the oversampled hopsize to be any integer (rather than limited to powers of 2), but I was having trouble getting the aggregate/sync procedure to work for that.  Doing that would have a significant efficiency savings for the highest octave.

PR #147, opened 2015-01-17
Further CQT Enhancements

ebattenberg:
This add a few things to the CQT:
1. Removes extra CQT frequency generated (I don't think there's a reason for an extra one).
2. Centers the filters with respect to the center of the FFT window.
3. Only operates on the portion of the spectrum corresponding to positive frequencies. (The complex filters are zero at negative frequencies anyway).  This save a bit of compute and memory.
4. Removes the Hann window from the STFT (We don't need an analysis window since we're not doing frequency analysis, we're just projecting onto basis functions).

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -124,10 +124,10 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         tuning = estimate_tuning(y=y, sr=sr)
 
     # First thing, get the fmin of the top octave
-    freqs = time_frequency.cqt_frequencies(n_bins + 1, fmin,
+    freqs = time_frequency.cqt_frequencies(n_bins, fmin,
                                            bins_per_octave=bins_per_octave)
 
-    fmin_top = freqs[-bins_per_octave-1]
+    fmin_top = freqs[-bins_per_octave]
 
     # Generate the basis filters
     basis, lengths = filters.constant_q(sr,
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -187,8 +187,9 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
         assert my_hop > 0
 
-        # Compute the STFT matrix
-        D = stft(my_y, n_fft=n_fft, hop_length=my_hop)
+        # Compute the STFT matrix. No window needed (ones=rectangular)
+        D = stft(my_y, n_fft=n_fft, hop_length=my_hop,
+                 window=np.ones)
 
         D = D.conj() # stft returns conjugate for some reason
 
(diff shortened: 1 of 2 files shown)


ebattenberg:
Actually, now that I think about it, centering the filters wrt the FFT window seems unnecessary.  I was trying to get the peak of the windows aligned, but they were already aligned when padding/centering to the max filter length.  

With respect to lower octaves, the alignment gets screwed up anyway.  Any reason to be concerned with alignment of the window peaks across octaves?

dpwe:
Yes, I'm very concerned about it. Also about gaps around the very short
windows (or do we already have very small hops?). What does the cqgram of
an impulse look like.

Thanks for all the work, Eric.

  DAn.

bmcfee:
> 1. Removes extra CQT frequency generated (I don't think there's a reason for an extra one).

Strange, I'll have to think about this one.

> 2. Centers the filters with respect to the center of the FFT window.

Excellent catch.  This does seem to resolve the drift issue at low octaves.

> Actually, now that I think about it, centering the filters wrt the FFT window seems unnecessary. I was trying to get the peak of the windows aligned, but they were already aligned when padding/centering to the max filter length.

The filters are properly centered at construction time, but were getting right-padded when passed into `fft` with `n=n_fft` (window rounded up to integral power of 2).  The resulting phase shift seems to have been responsible for low-octave drift.

(Incidentally, `pad_center` can operate on multidimensional arrays, so iterating is unnecessary.)

> 3. Only operates on the portion of the spectrum corresponding to positive frequencies. (The complex filters are zero at negative frequencies anyway). This save a bit of compute and memory.

Good catch.

> 4. Removes the Hann window from the STFT (We don't need an analysis window since we're not doing frequency analysis, we're just projecting onto basis functions).

Also good catch.

bmcfee:
> 1. Removes extra CQT frequency generated (I don't think there's a reason for an extra one).

Just to follow up: this is a hold-over from a previous implementation that was parameterized by `fmin` and `fmax`, rather than `fmin` and `n_bins`.  You are correct in that it's no longer necessary.

bmcfee:
> What does the cqgram of
> an impulse look like.

I cooked up an ipynb demonstrating before and after on impulses (okay, percussive component of the demo track): http://nbviewer.ipython.org/gist/bmcfee/067cfbff151905db7e4e

ebattenberg:
It looks like the version in develop still uses the conjugate version of the stft (since that is what stft produces by default).  I had a line taking the conjugate (of the conjugate), that I don't think made it to develop.

bmcfee:
> It looks like the version in develop still uses the conjugate version of the stft (since that is what stft produces by default). I had a line taking the conjugate (of the conjugate), that I don't think made it to develop.

It was there, but since we're only retaining magnitude, it's simpler to cancel the conj() in the basis instead.

ebattenberg:
Ahh, I see.  Groovy.

PR #152, opened 2015-01-21
Avoid final unused resampling in CQT

ebattenberg:
The output of the last resampling is unused, so skip it.  Probably not a big deal efficiency-wise because the last octave usually has so few samples, but it's a good idea just in case.


bmcfee:
Again, good catch. Thanks!

PR #153, opened 2015-01-22
Fix constant-Q frequency drift when rounding window size

ebattenberg:
This fixes constant-Q center frequency drift when rounding the window size up to an integer value.  It seems tolerable to allow the bandwidth of the filters to change with rounding as it's easiest to define the Hann window as an integer length.  However, it's pretty easy to fix the frequency rounding problem by using the floating point frequency when generating the complex exponential before windowing.  

Also, I removed the unnecessary check for window == None and the incorrect None means "no window is applied" line in the documentation.

### DIFF
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -425,7 +425,7 @@ def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, tuning=0.0,
             raise ValueError("Filter pass band lies beyond Nyquist")
 
         # Length of the filter
-        ilen = np.ceil(Q * sr / freq)
+        ilen = Q * sr / freq
         lengths.append(ilen)
 
         # Build the filter
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -364,8 +364,6 @@ def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, tuning=0.0,
 
     window : function or `None`
         Windowing function to apply to filters.
-        If `None`, no window is applied.
-
         Default: `scipy.signal.hann`
 
     resolution : float > 0 [scalar]
@@ -431,16 +429,15 @@ def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, tuning=0.0,
         lengths.append(ilen)
 
         # Build the filter
-        win = np.exp(Q * 1j * np.linspace(0, 2 * np.pi, ilen, endpoint=False))
+        sig = np.exp(1j*2*np.pi*freq*np.arange(ilen, dtype=float)/sr)
 
         # Apply the windowing function
-        if window is not None:
-            win = win * window(ilen)
+        sig = sig * window(ilen)
 
         # Normalize
-        win = util.normalize(win, norm=norm)
+        sig = util.normalize(sig, norm=norm)
 
-        filters.append(win)
+        filters.append(sig)
 
     max_len = max(lengths)
     if pad_fft:


ebattenberg:
Actually, we can better approximate the desired bandwidth by using a non-integer argument to the window function.  Both scipy.signal.hann and np.arange take non-integer arguments.  They round up for the resulting sequence length, but the window function uses floating point argument to calculate the period of the cosine window.

bmcfee:
Interesting.

The documentation for [scipy.signal.hann](http://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.signal.hann.html) doesn't mention this (note the type doc `"M : int"`).  But, looking at the source code, it seems to make sense.

On the other hand, the numpy documentation for [arange](http://docs.scipy.org/doc/numpy/reference/generated/numpy.arange.html) advises against doing this, and instead says to use linspace for consistent results.  It's not entirely clear what that means though.. any thoughts?

ebattenberg:
Yeah, it doesn't appear that non-int arguments are officially supported for either.  I just tried it, and it gave me the exact behavior I wanted without having to write my own non-int hann window.  If we want to do it correctly, it looks like we'll have to round up for the argument to arange and then roll our own window function which is pretty simple.  We can just rip out the guts of scipy.signal.hann.  I was trying to avoid doing that.  

Or we could just not worry about the bandwidth rounding problem.

ebattenberg:
Then what's your opinion on using hann with a float argument.  The underlying implementation could change and break our intended usage.  But I guess as long as there's a test for that, we'd be safe.

bmcfee:
It sounds good.  I'm actually not sure the behavior is unspecified.  arange seems unambiguous when it comes to floating point stop values and integer steps (modulo roundoff error).

scipy.signal.hann seems to be broken at the boundaries though:
```python
>>> scipy.signal.hann(5.01)
array([  0.000e+00,   4.978e-01,   1.000e+00,   5.065e-01,   7.423e-05,
         4.892e-01])
```
That last guy should be 0, not ~0.5.  It's trivial to correct this by setting `window[int(M):] = 0`, though maybe a pr to scipy would be in order...

ebattenberg:
Why do you think 4.892e-01 should be zero?

ebattenberg:
Good catch, I didn't see that.  I think a pr is a good idea.

bmcfee:
At least according to the docs, the definition is only valid for `0 <= n <= M-1`, which doesn't cover the last point.  Also it's weird to have a huge jump discontinuity at the edge of a window function.

dpwe:
I'm guessing it's a np.empty that doesn't even get written?  So you were
lucky it was only 4e-1?  Or maybe that's just too much of a coincidence.

  DAn.

bmcfee:
> I'm guessing it's a np.empty that doesn't even get written?

I don't think so.  It looks from the code like it's blindly applying the `cos(2 * pi * n / (M - 1))` beyond the domain of the function.  At the end of the array, `M-1 = 4.01` and `n=5` (from `arange`).

ebattenberg:
Yeah, it's just evaluating that periodic expression past the edge of the window.  Setting `win[int(M):] = 0` seems like a bit of a kludge.  Maybe we should write our own?  It's a really simple function.

bmcfee:
> Maybe we should write our own? It's a really simple function.

It is really simple, but I don't want to go into the rabbit hole of rewriting half of scipy.signal.  (What if someone wants to use a different window function?)

Instead, how about we stick an additional layer on top of calls to window functions with fractional length?  So instead of
```python
sig = sig * window(ilen)
```
we have
```python
sig = sig * fractional_window(window(ilen))
```
or, it could even be a decorator like `sig * fractional_window(window)(ilen)`.  The latter is a little less readable, but probably simpler to write because it will have access to the input and output of `window`.

bmcfee:
One more point in favor of a decorator: not all window functions behave similarly on fractional inputs.

For example:
```python
In [5]: len(scipy.signal.hann(5.1))
Out[5]: 6

In [6]: len(scipy.signal.boxcar(5.1))
Out[6]: 5
```

A decorator be used to trim or expand the window to a well-defined length.

bmcfee:
Here's [gist](https://gist.github.com/bmcfee/3ee8c1820a0cb246a803) implementing the wrapper decorator.

The convention is that the window length will always be `ceil(M)`, but values `window[floor(M):ceil(M)]` are forced to 0.  When `M` is integer-valued, the wrapper is effectively a no-op.

ebattenberg:
Cool.  I like your wrapper.  :)

Issue #154, opened 2015-01-22
Automatically choose fastest possible resampling quality in CQT


ebattenberg:
According to http://www.mega-nerd.com/SRC/api_misc.html#Converters, the resampling qualities "sinc_*" have 97 dB SNR and 97%, 90%, and 80% bandwidth for best, medium, and fastest, respectively.    It therefore shouldn't be too hard to estimate the minimum resampling bandwidth required given the distribution of CQT filters in an octave.  

I'm suggesting this because sinc_best can slow down the CQT by an order of magnitude.  It should only be used when absolutely necessary.

ebattenberg:
I just wrote some code that does this.  I'll send a pull request tomorrow.

ebattenberg:
One question though,  I'm using fc*(1+1/Q) as the upper boundary of the passband of the filter.  Is this correct?  Or should it be fc*(1+1/(2*Q))?  fc/Q is the bandwidth of the filter right?  So fc/(2*Q) is how far above the center frequency the passband goes?  I used fc*(1+1/Q) for the constant_q aliasing check in filters.constant_q (develop branch), so I'd have to modify that there as well, if it's fc*(1+1/(2*Q)).

PR #156, opened 2015-01-22
Add automatic res_type detection for CQT

ebattenberg:
See #154 .

This automatically detects the required resampling quality.  Using 'sinc_best' is really slow.  If we can get away with 'sinc_fastest' we should use it.

I have two concerns:
1. What is the upper passband limit of the highest frequency filter?  `fmax*(1+1/Q)` or `fmax*(1+0.5/Q)`?
2. What happens when scikits.samplerate isn't available?

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -15,6 +15,11 @@ from .. import util
 from ..feature.utils import sync
 
 
+BW_BEST = 0.97
+BW_MEDIUM = 0.9
+BW_FASTEST = 0.8
+
+
 @cache
 def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         bins_per_octave=12, tuning=None, resolution=2,
@@ -147,11 +152,11 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     filter_cutoff = fmax_t*(1 + 0.725 / Q) # 0.725 for Hann window
     nyquist = sr / 2.0
 
-    if filter_cutoff < 0.8*nyquist:
+    if filter_cutoff < BW_FASTEST*nyquist:
         res_type = 'sinc_fastest'
-    elif filter_cutoff < 0.9*nyquist:
+    elif filter_cutoff < BW_MEDIUM*nyquist:
         res_type = 'sinc_medium'
-    elif filter_cutoff < 0.97*nyquist:
+    elif filter_cutoff < BW_BEST*nyquist:
         res_type = 'sinc_best'
     else:
         res_type = 'sinc_best'
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -160,7 +160,24 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         res_type = 'sinc_best'
     else:
         res_type = 'sinc_best'
-        warnings.warn("Filter passband lies beyond resampling bandwidth")
+
+    if res_type == 'sinc_fastest' and audio._HAS_SAMPLERATE:
+
+        # How many times can we downsample by 2 before filtering?
+        # Requirements:
+        # filter_cutoff < BW*nyquist  # (BW is resampling bandwidth fraction)
+        # hop_length > 2**n_octaves
+
+        downsample_count1 = int(np.ceil(np.log2(BW_FASTEST * nyquist
+                                                / filter_cutoff)) - 1)
+        downsample_count2 = int(np.ceil(np.log2(hop_length) - n_octaves) - 1)
+        downsample_count = min(downsample_count1, downsample_count2)
+        if downsample_count > 0:
+            downsample_factor = 2**downsample_count
+            hop_length = int(hop_length/downsample_factor)
+            y = audio.resample(y, sr, sr/downsample_factor, res_type=res_type)
+            sr = sr/downsample_factor
+
 
 
     # Generate the basis filters
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -137,9 +137,28 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     if tuning is None:
         tuning = estimate_tuning(y=y, sr=sr)
 
-    # First thing, get the fmin of the top octave
-    fmin_t = cqt_frequencies(n_bins, fmin,
-                             bins_per_octave=bins_per_octave)[-bins_per_octave]
+    # First thing, get the freqs of the top octave
+    freqs = cqt_frequencies(n_bins, fmin,
+                            bins_per_octave=bins_per_octave)[-bins_per_octave:]
+
+    fmin_t = np.min(freqs)
+    fmax_t = np.max(freqs)
+
+    # Determine required resampling quality
+    Q = float(resolution) / (2.0**(1. / bins_per_octave) - 1)
+    filter_cutoff = fmax_t*(1+1.0/Q)
+    nyquist = sr/2.0
+
+    if filter_cutoff < 0.8*nyquist:
+        res_type = 'sinc_fastest'
+    elif filter_cutoff < 0.9*nyquist:
+        res_type = 'sinc_medium'
+    elif filter_cutoff < 0.97*nyquist:
+        res_type = 'sinc_best'
+    else:
+        ValueError("Highest frequency filter lies beyond Nyquist")
+
+
 
     # Generate the basis filters
     basis, lengths = filters.constant_q(sr,
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -16,7 +16,7 @@ from ..feature.utils import sync
 
 @cache
 def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
-        bins_per_octave=12, tuning=None, resolution=2, res_type='sinc_best',
+        bins_per_octave=12, tuning=None, resolution=2,
         aggregate=None, norm=2, sparsity=0.01):
     '''Compute the constant-Q transform of an audio signal.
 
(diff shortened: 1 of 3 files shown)


bmcfee:
> 1. What is the upper passband limit of the highest frequency filter? fmax*(1+1/Q) or fmax*(1+0.5/Q)?

Duno?  Maybe @dpwe can chime in?

> 2. What happens when scikits.samplerate isn't available?

If samplerate isn't available, `resample` falls back on `scipy.signal.resample` and ignores the `res_type` parameter.  So this set of changes should have no effect on that case.

librosa/core/constantq.py
```diff
@@ -137,9 +134,28 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
...
+    nyquist = sr/2.0
+
+    if filter_cutoff < 0.8*nyquist:
+        res_type = 'sinc_fastest'
+    elif filter_cutoff < 0.9*nyquist:
+        res_type = 'sinc_medium'
+    elif filter_cutoff < 0.97*nyquist:
+        res_type = 'sinc_best'
+    else:
+        ValueError("Highest frequency filter lies beyond Nyquist")
```
bmcfee: This exception seems redundant, since we should just hit a few lines down in `filters.constant_q` anyway.

... But that comparison is against `nyquist`, not `0.97 * nyquist`, so maybe they're not the same?  Maybe push up to `sinc_best` if `nyquist > filter_cutoff > 0.9 * nyquist`, but issue a warning if it's above `0.97 * nyquist`?

Also, if `_HAS_SAMPLERATE == False`, res_type has no effect, so it's not appropriate to throw an exception here.

ebattenberg:
I agree that the exception is redundant.

If the filter passband lies beyond `0.97*nyquist` (and we're using sksamplerate) then the cqt will have holes in it at the top of every octave.  Is that something to throw an error or a warning over?  I wonder what the resampling bandwidth is of scipy.resample.

dpwe:
I would be interested to see actual Cqt outputs of impulses and slow
exponential sine sweeps to reveal discontinuities between octaves.

Because the filters have broad roll-off, the top center frequency always
has to be substantially below nyquist. All we require is that each octave
covers a whole octave, but that can be from 0.3 to 0.6 Nyquist, say.  That
loses spectrum in the top octave, but I don't know how important that is.

And so, having the downsampling anti-aliasing below nyquist is ok in
principle (and inevitable in practice).

Also, to choose between fast and best resampling, I'd mostly want to look
at empirical aliasing noise measurements.  A swept sine ought to make alias
terms visible (on a wide, dB scale).

  DAn.

ebattenberg:
From #154, the SNRs and bandwidths are listed here: http://www.mega-nerd.com/SRC/api_misc.html#Converters

We could precisely measure the bandwidth and SNR, but I figure using the listed values above should be sufficient unless someone really wants to go through the trouble.

The next pullrequest I was planning on issuing involves downsampling before starting to descend through the octaves.  This can result in significant performance gains when the filters for the top octave would end up being large (meaning long FFTs) at the full sampling rate.  Another version of this would be doing two octaves of filters before downsampling.  This way we can afford to do the fastest resampling for subsequent octaves, resulting in a big speed gain.

ebattenberg:
Here are three examples with sinc fastest, medimum, and best, respectively.  I used sr=44100, fmin = 31Hz, bins_per_octave=12, and n_bins=114.  This gives a top center frequency of 21186, so it's pushing Nyquist.

figure_fastest
figure_medium
figure_best

ebattenberg:
With n_bin = 110 (a top center frequency of 17815), we get the following for fastest and best, respectively.  `0.8*nyquist` is 17640.  There is not an easily visible difference between the two, so I guess 0.8 is a bit of a conservative fraction of the bandwidth.  Do you think it's worth trying to get away with assuming more usable bandwidth when using "fastest"?

figure_fastest

figure_best

ebattenberg:
I ran some tests and came up with the approximate bandwidth of a Hann window.  I looked at the analytic expression for the DFT of the Hann window, and it wasn't too clean, so I went with a numerical approach.

I modeled the bandwidth over a range of widths as `bw = k*sr/ilen` where `ilen` is the window width and `k` is a constant we're solving for.  Using the 3dB cutoff of 0.707 to compute the bandwidth, this gives a `k` of around `1.45`.  

Using this, the upper boundary of the passband of a filter is `fc*(1+0.5*k/Q`  or `fc*(1+0.725/Q)`.

ebattenberg:
Ok, I kinda scrapped the whole previous idea of just setting the fastest possible resampling method.  

I added two new pre-processing steps that should allow significant efficiency gains.

1. Downsample the signal as much as possible before processing it.
  - As long as the pass band boundary of the top filter remains below `0.8*nyquist`, we can keep downsampling (using sinc_fastest).
  - This allows us to use smaller FFTs and is particularly helpful if the top filter is nowhere near nyquist (and will therefore be quite long).
2. If we are unable to use sinc_fastest, process two octaves before resampling.
  - This allows us to use sinc_fastest for subsequent octaves.


Any thoughts?  I'm going to post some timing examples to show the improvement.

ebattenberg:
For an example speedup of 1. (preemptive downsampling), I used:

```
sr = 44100
hop_length=1024
fmin=31
n_bins=80
tuning=0
sparsity=0.01
norm=1
```
With enhancement 1. disabled, the run takes 8.08 seconds.  With it enabled, 4.33 seconds.

ebattenberg:
For an example speedup of 2. (process two octaves first), I used:

```
sr = 44100
hop_length=1024
fmin=31
n_bins=114
tuning=0
sparsity=0.01
norm=1
```
With enhancement 2. disabled, the run takes 26.21 seconds. With it enabled, 10.42 seconds.

librosa/filters.py
```diff
@@ -449,7 +449,7 @@ def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, tuning=0.0,
     for i in np.arange(n_bins, dtype=float):
 
         freq = fmin * 2.0**(i / bins_per_octave)
-        if freq * (1 + 1.0 / Q) > sr / 2.0:
+        if freq * (1 + 0.725 / Q) > sr / 2.0: # 0.725 for Hann window
```
bmcfee: How difficult would it be to compute this value from the window function, rather than hard-code it?

bmcfee:
> With enhancement 1. disabled, the run takes 8.08 seconds. With it enabled, 4.33 seconds.
> ...
> With enhancement 2. disabled, the run takes 26.21 seconds. With it enabled, 10.42 seconds.

Just for a sanity check, do you have numerical comparison tests of the results with/without the speedup?

More generally, as a side note, we should come up with some way of unit-testing all this cqt business (but that probably should be a separate issue).

librosa/filters.py
ebattenberg: I don't think it would be worth computing it from scratch every time.  I like the dictionary idea.  What are the arguments in favor of supporting many different windows?  Do you know of anyone using something besides Hann?

bmcfee: > I don't think it would be worth computing it from scratch every time.

The only reason to do that would be for folks using home-made window functions.  I don't really expect this, but i'd rather not rule it out unless we have to.


> What are the arguments in favor of supporting many different windows?

Because we can/could.

> Do you know of anyone using something besides Hann?

Not off-hand.  I'm not entirely opposed to restricting it to be Hann-only, if it's really important, but I'd rather it be as flexible as we can get away with.


> I like the dictionary idea.

This would be pretty easy to pull off if we can come up with a list of cutoffs for each of the scipy windows.

ebattenberg:
I haven't done any numerical stability tests.  I skipped doing it for my first few PR's because my changes were actually changing the output (the oversampling hop, the center frequency rounding errors, etc.)  Now I guess it's worth doing.  

Something to consider is that with enhancement 2 when using norm=2, the results will be different than without it.  This is due to the fact that norm=2 causes visible banding between octaves.  This type of normalization doesn't allow for smooth transitions between the bottom of one octave and the top of the next.  With norm=1, you get smooth octave transitions.  So, with norm=2, you'll have banding between the last two octaves without enhancement 2, and no banding with it.  

I can do some informal tests with norm=1, but they won't be reproducible as unit tests unless we allow the speed enhancements to be disabled.  With norm=2, the output is dependent on how the octaves are processed.  Maybe that should be an issue.

Here are examples of the banding with both enhancements disabled.  Notice the banding in the norm=2 version.  
figure_3

ebattenberg:
Ok, here are some visual comparisons with the enhancements on and off.  

For enhancement 1 (which is only enabled when the top octave is far from Nyquist),
```
sr=44100
fmin=31
n_bins=80
bins_per_octave=12
norm=1
tuning=0
```

The deviation from reference is minor, but it is there.  

enh1

ebattenberg:
Ok, here are some visual comparisons with the enhancements on and off.  

For enhancement 2 (which is only enabled when the top octave is very close to Nyquist)
```
sr=44100
fmin=31
n_bins=112
bins_per_octave=12
norm=1
tuning=0
```

Again, some deviations (mostly at the boundary), but they are there.  

enh2-112

ebattenberg:
Ok, here are some visual comparisons with the enhancements on and off.  

Here's a more enlightening one for enhancement 2 that occurs when we get even closer to Nyquist
(Notice the addition of two extra bins).
```
sr=44100
fmin=31
n_bins=114
bins_per_octave=12
norm=1
tuning=0
```

Even using 'sinc_best' here causes significant banding in the reference version.  This causes significant deviation from the reference with enh2 enabled; however, I'd argue that the enh1 version is closer to the desire behavior (and it's much faster).  

enh2-114

ebattenberg:
And here's a version of the above with `norm=2`.  We have banding in both version.  In the ref, the banding is due to the cutoff of the resampler.  In enh2, the banding is due to the normalization method.  Notice than there is no banding between the top two octaves, because of what enh2 does.

enh2-norm2

bmcfee:
This all looks great!

I'm refactoring it a little bit to keep the main cqt function simple, but it should be merged by this afternoon.

ebattenberg:
Cool.  Thanks for cleaning it up.

PR #159, opened 2015-01-27
Hybrid-CQT

ebattenberg:
This adds a hybrid_cqt which is a combination of a pseduo-cqt and the full-cqt.  The full-cqt does oversampling and then aggregation when the hop_length is longer than half the shortest filter.  The hybrid-cqt avoids this computationally complex step by just doing a pseudo CQT for such filters.  (The pseudo-cqt being the phase invariant CQT that uses the both the magnitude of the filters and the magnitude of the stft.).  The filters that are longer than twice the hop_length are then carried out using the full-cqt.  

This is not meant to produce equivalent output to a full-cqt, but it is quite a bit faster when the highest octave would require a very small oversampling hop.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ pip-log.txt
 .Trashes
 ehthumbs.db
 Thumbs.db
+
+# Vim
+*.swp
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -229,8 +229,8 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
 @cache
 def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
-        bins_per_octave=12, tuning=None, resolution=2, res_type='sinc_best',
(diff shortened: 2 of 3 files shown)


ebattenberg:
Here's a visualization of the difference.  I'm still debating the best way to normalize the pseudo-cqt so that it is at a similar scale to the full-cqt.

```
sr=44100
hop_length=1024
fmin=31
bins_per_octave=12
n_bins=108
tuning=0
sparsity=0.01
norm=1
```

hybrid

ebattenberg:
Oh, and the most important different for the above is the timing:
Full-cqt = 0.32 seconds
Hybrid-cqt = 0.13 seconds

bmcfee:
Once again, looks great!

I'm going to refactor a little so it uses as much of the existing cqt's internals as possible, but it should merge later today.

bmcfee:
.. PS, @ebattenberg should really be in AUTHORS.md!

ebattenberg:
Great, thanks for the refactoring again.

Issue #150, opened 2015-01-20
Improve unit test coverage


bmcfee:
Currently, our tests are almost entirely regression against prior implementations.  But, as the library is growing and including a wide range of utility functions, we should have better coverage for correctness tests as well.

As of 5fcaac3beb4fd536604b7d99b02165a8919978d6, our test coverage looks as follows:

```
nosetests --with-coverage --cover-package=librosa 
.................................................................................................................................................................................................................................................................................................................................
Name                             Stmts   Miss  Cover   Missing
--------------------------------------------------------------
librosa                             15      0   100%   
librosa.beat                        83     52    37%   118-144, 236, 267-289, 296-299, 305-306, 314-350, 356-360, 366-377
librosa.core                         6      0   100%   
librosa.core.audio                 108     47    56%   20-23, 102, 115, 119, 123, 134, 146, 248-260, 313-331, 370, 462-492
librosa.core.constantq              47     33    30%   126-188, 195-201, 217-233
librosa.core.pitch                  92     78    15%   64-74, 123-142, 210-300, 361-409
librosa.core.spectrum              114     38    67%   122, 221, 225, 239-243, 260, 336, 339, 368, 471-513, 581-595, 639-641
librosa.core.time_frequency        124     67    46%   47-51, 95-99, 143-147, 190-192, 223, 259, 296, 345-375, 425-442, 473-474, 504-505, 544, 577, 709-710, 774-777, 826, 864-880
librosa.decompose                   46     30    35%   106-107, 213-254
librosa.display                    126    115     9%   75-122, 161-172, 273-407, 425-433
librosa.effects                     46     32    30%   47-56, 87-95, 126-134, 172-180, 225-231, 275-294
librosa.feature                      3      0   100%   
librosa.feature.spectral           136    104    24%   20-27, 96-113, 181-209, 251-290, 341-364, 398-400, 456-471, 520-532, 596-608, 681-702, 764-767, 831-837
librosa.feature.utils               49     38    22%   59-75, 145-166, 220-242
librosa.filters                     95     53    44%   50-58, 110, 296-322, 407-455, 498-516
librosa.onset                       46     20    57%   88-112, 200-206, 225
librosa.output                      39     28    28%   67-78, 113-115, 170-181, 210-226
librosa.segment                     63     45    29%   27-36, 89-117, 167-188, 254-262, 304-322
librosa.util                         4      0   100%   
[...]

PR #172, opened 2015-03-02
CQT Tests

ebattenberg:
I started writing the tests I promised in #150.  Right now it's mostly boilerplate.  I will update in my free moments this week.   What I plan to focus other than just code coverage is approximate equivalence between the different efficiency optimizations.   Let me know if you have any ideas for other things to test.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 '''Pitch-tracking and tuning estimation'''
+from __future__ import division
 
 import numpy as np
 
@@ -465,7 +466,7 @@ def __fft_filters(sr, fmin, bins_per_octave, tuning,
     basis *= lengths.reshape((-1, 1)) / n_fft
 
     # FFT and retain only the non-negative frequencies
-    fft_basis = np.fft.fft(basis, n=n_fft, axis=1)[:, :(n_fft / 2)+1]
+    fft_basis = np.fft.fft(basis, n=n_fft, axis=1)[:, :(n_fft // 2)+1]
 
     # normalize as in Parseval's relation, and sparsify the basis
     fft_basis = util.sparsify_rows(fft_basis / n_fft, quantile=sparsity)
--- a/tests/test_constantq.py
+++ b/tests/test_constantq.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python
+"""
+CREATED:2015-03-01 by Eric Battenberg <ebattenberg@gmail.com>
+unit tests for librosa core.constantq
+
+Run me as follows:
+    cd tests/
+    nosetests -v --with-coverage --cover-package=librosa
+"""
+
+
+from __future__ import print_function
+from __future__ import division
+
+# Disable cache
+import os
+try:
+    os.environ.pop('LIBROSA_CACHE_DIR')
+except:
+    pass
+
+import librosa
+import numpy as np
+
+from nose.tools import nottest, eq_, raises
+
+
+def test_cqt():
+
+    def __test(y, sr, hop_length, fmin, n_bins, bins_per_octave,
+               tuning, resolution, aggregate, norm, sparsity):
+
+        cqt_output = librosa.cqt(y,
+                                 sr=sr,
+                                 hop_length=hop_length,
+                                 fmin=fmin,
+                                 n_bins=n_bins,
+                                 bins_per_octave=bins_per_octave,
+                                 tuning=tuning,
+                                 resolution=resolution,
+                                 aggregate=aggregate,
+                                 norm=norm,
+                                 sparsity=sparsity)
+
+        assert cqt_output.shape[0] == n_bins
+
+        return cqt_output
+
+
+    sr = 11025
+
+
+    # Impulse train
+    y = np.zeros(int(10.0 * sr))
+    y[::sr] = 1.0
+
+
+    hop_length = 512
+    fmin = None
+    n_bins = 48
+    bins_per_octave = 12
+    tuning = None
+    resolution = 2
+    aggregate = None
+    norm = 1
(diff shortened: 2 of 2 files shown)


bmcfee:
> What I plan to focus other than just code coverage is approximate equivalence between the different efficiency optimizations.

Sounds great!  Thanks for doing this.

bmcfee:
I guess we actually need to fix the samplerate dependency on travis before the preemptive downsampling code can be tested, right?

At present, it installs through apt, but scikits.samplerate can't find it.

ebattenberg:
> I guess we actually need to fix the samplerate dependency on travis before the preemptive downsampling code can be tested, right?

Yup.  The tests are passing over here so far when I run them on my end.  (with samplerate installed).

ebattenberg:
And it looks like libsamplerate0-dev is the one I installed to get it working on my machine.  That's the one travis is apt-getting.  Hmm.

bmcfee:
Yeah, same on my end.  I blame the dogpile of package managers, in order of dumbness (apt, pip, conda)...

ebattenberg:
Other topic:  The tests helped me find a gnarly bug that breaks the early downsampling if `n_bins < bins_per_octave`.  I'm thinking we should support that use case even though it would be uncommon.  What are your thoughts?

bmcfee:
> n_bins < bins_per_octave. I'm thinking we should support that use case even though it would be uncommon

Sure.  I don't see any reason not to support generating cqt's that are less than one octave high.

ebattenberg:
> Sure. I don't see any reason not to support generating cqt's that are less than one octave high.

Also, this isn't exactly specific to the early downsampling.  It's just more apparent because we're hanging out near Nyquist more often because of it.  If you wanted a CQT with an `fmin` near enough to Nyquist that a whole octave won't fit, then you'll run into problems because `__fft_filters` always generates an entire octave of filters even if they aren't used.

bmcfee:
> Yeah, same on my end. I blame the dogpile of package managers, in order of dumbness (apt, pip, conda)...

I took another stab at trying to fix this, but no dice.  pip installs samplerate happily, and then the conda environment breaks things.

bmcfee:
After way too much prodding, I think this samplerate/conda issue can be resolved by hacking a custom numpy site.cfg for travis.  I'll take a crack at this later today; this note is mainly to myself.

bmcfee:
Building samplerate from source within conda seems to work, but only on python2: https://travis-ci.org/bmcfee/travis_test/builds/53989101

This should be fixed by [this PR](https://github.com/cournape/samplerate/pull/5), but it's not yet integrated in pypi.

bmcfee:
> This should be fixed by this PR, but it's not yet integrated in pypi.

indeed, this fixes it.

It seems unlikely that that PR will be merged any time soon, given the staleness of sksamplerate in pypi right now.  (Maybe we should encourage a proper fork?)

However, for testing, I think we can safely follow the build scheme in my test repo.

bmcfee:
samplerate tests now work on travis: https://coveralls.io/builds/2094075

ebattenberg:
Ok, I'm back on this.  Will try to get some passing tests written today.

bmcfee:
I'm digging through this one again, and it seems like the tests that should pass are failing due to too much early downsampling.

The first iteration of the grid sweep gives a parameter set like
```python
test_constantq.test_cqt(array([ 1.,  0.,  0., ...,  0.,  0.,  0.]), 11025, 512, None, 1, 12, 0, 1, None, 1, 0.01)
```
which should make one bin at 'C2'.  In doing so, it creates a full 12-bin filter bank with the following frequencies:
```
[ 32.70319566  34.64782887  36.70809599  38.89087297  41.20344461
  43.65352893  46.24930284  48.9994295   51.9130872   55.          58.27047019
  61.73541266]
```
and an early-downsampled `sr=86.1328125`.

Maybe we need to explicitly handle the case where `bins_per_octave < n_bins`  to avoid this sort of thing?

bmcfee:
Additionally, [this part](https://github.com/bmcfee/librosa/blob/develop/librosa/core/constantq.py#L453-L454) might be responsible, since it's forcing `filters.constant_q` to generate more filters than it needs.

ebattenberg:
Yeah. I'm aware. Just haven't had time to insert the logic.

bmcfee:
No problem.  I've pushed a couple of updates to this PR in https://github.com/bmcfee/librosa/tree/ebattenberg-cqt-tests , including the `n_filters` fix 9f5665cdbcafb5eb0c3c357081eb1766800c1e75 .

There's still something going wrong in computing the filter cutoff.  The Q factors within `cqt()` and `filters.constant_q` disagree substantially.  Not sure what's going on there...

ebattenberg:
FWIW, I'm hoping to start using this stuff in some of my work in the next
week, so hopefully I'll have an excuse to spend some more time on it and
finally get the tests passing. But I know I've promised that for a while. :)

bmcfee:
Coming back to this again.  It looks like early_downsampling is still being too aggressive.

I reduced the downsampling count calculation by 1, and this fixes most of our (currently failing) test cases.

bmcfee:
Also, my previous hack 9f5665c introduced a bug in the filter length calculation, since it uses `bins_per_octave` to decide how to space the filters.  Oops!

bmcfee:
> What I plan to focus other than just code coverage is approximate equivalence between the different efficiency optimizations. Let me know if you have any ideas for other things to test.

#196 contains bugfixes and tests for:

* error checking, size correctness
* approximate equivalence between cqt and hybrid_cqt (which implies pseudo_cqt)
* a simple "correctness" test where a pure tone is generated and recovered as a salient peak of the cqt, while ensuring that all incorrect frequency bins have sufficiently small magnitude.

Aside from filterbank construction (which is tested elsewhere), I think there's not much else to add here.  @ebattenberg : if this all sounds good, I'll close this PR and merge #196.

PR #173, opened 2015-03-03
Added chroma to tonnetz function

urinieto:
I have added the chroma to tonnetz functionality in features/utils.py.
It only accepts 12-dimensional chromagrams for now.
I have added a not very comprehensive test in test_features.py.
I have also fixed a tiny typo in chroma_stft docstring.


librosa/feature/utils.py
```diff
@@ -333,3 +333,93 @@ def sync(data, frames, aggregate=None, pad=True):
         start = end
 
     return data_agg
+
+
+@cache
+def chroma_to_tonnetz(chromagram, norm=np.inf):
```
bmcfee: Can we refactor this?  I suggest the following:

* move the filterbank construction (`phi` in your variables) into the `filters` submodule
* move chroma_to_tonnetz into `feature.spectral`.  The `util` submodule is for generic feature manipulation, not specific mappings.

librosa/feature/utils.py
```diff
@@ -333,3 +333,93 @@ def sync(data, frames, aggregate=None, pad=True):
...
+    tonnetz = np.zeros((6, t))
+
+    r1 = 1      # Fifths
+    r2 = 1      # Minor
+    r3 = 0.5    # Major
+
+    # Generate Transformation matrix
+    phi = np.zeros((6, 12))
+    for i in range(6):
+        for j in range(12):
```
bmcfee: It looks like this loop can be vectorized:

```python
if i < 2:
   phi[i, :] = r1 * fun(np.arange(12) * 7. * np.pi / 6))
...
```

At that point, it looks pretty simple to extend the support for high-res chroma input by changing the step size, eg:

```python
np.arange(0, 12, 12.0/n_chroma)
```
or even
```python
np.linspace(0, 12, n_chroma, endpoint=False)
```

I should note that we need to be careful about centering with high-res chroma though, so you might have to rotate that linspace left or right by half a semitone's worth of bins to get it to line up correctly.

librosa/feature/utils.py
```diff
@@ -333,3 +333,93 @@ def sync(data, frames, aggregate=None, pad=True):
...
+
+            if i < 2:
+                phi[i, j] = r1 * fun(j * 7 * np.pi / 6.)
+            elif i >= 2 and i < 4:
+                phi[i, j] = r2 * fun(j * 3 * np.pi / 2.)
+            else:
+                phi[i, j] = r3 * fun(j * 2 * np.pi / 3.)
+
+    # Do the transform to tonnetz
+    for i in range(t):
```
bmcfee: Can we vectorize this loop?

tests/test_features.py
```diff
@@ -433,3 +433,23 @@ def __make_data(coeffs, freq):
         freq = np.cumsum(np.abs(np.random.randn(1 + n_fft//2, 2)), axis=0)
         S = __make_data(coeffs, freq)
         yield __test, S, coeffs, freq
+
+
+def test_chroma_to_tonnetz():
```
bmcfee: Do you have any pre-computed data that we could use to build regression tests, as well as the general sanity checks already in place?

librosa/feature/utils.py
```diff
@@ -333,3 +333,93 @@ def sync(data, frames, aggregate=None, pad=True):
...
+
+    r1 = 1      # Fifths
+    r2 = 1      # Minor
+    r3 = 0.5    # Major
+
+    # Generate Transformation matrix
+    phi = np.zeros((6, 12))
+    for i in range(6):
+        for j in range(12):
+            fun = np.sin if i % 2 == 0 else np.cos
```
bmcfee: I think we can also simplify this a bit by replacing the anonymous, alternating function with a `np.pi/2` phase offset for the odd-numbered rows, and just using cosine all the way through.

Similarly, the different scaling fatcors `7/6, 3/2, 2/3` can be stacked up a vector and broadcast over the input.  I think this should make it possible to eliminate all loops from this filter construction.

bmcfee:
Looking good!  Is this ready to merge?

urinieto:
I still want to add some more tests. I will let you know when I add them.

bmcfee:
> I still want to add some more tests. I will let you know when I add them.

Ok then, I'll make a few more comments!

librosa/feature/spectral.py
```diff
@@ -1163,6 +1164,104 @@ def logfsgram(y=None, sr=22050, S=None, n_fft=4096,
...
+            5: Major y-axis
+
+    See Also
+    --------
+    librosa.feature.chroma_cqt
+        Compute a chromagram from a constat-Q transform.
+    librosa.feature.chroma_stft
+        Compute a chromagram from an STFT spectrogram or waveform.
+
+    Harte, C., Sandler, M., & Gasser, M. (2006). Detecting Harmonic Change in
```
bmcfee: numpydoc actually has provisions for citations.  If we use them, then references get collected and indexed properly.

Check out https://github.com/bmcfee/librosa/blob/develop/librosa/beat.py#L30-L38 for an example of how it works.

librosa/feature/spectral.py
```diff
@@ -28,7 +28,8 @@
            'mfcc',
            # Deprecated functions
            'logfsgram',
-           'chromagram']
+           'chromagram',
+           'tonnetz']
```
bmcfee: `tonnetz` should also be added to the feature submodule's docstring index [here](https://github.com/bmcfee/librosa/blob/develop/librosa/feature/__init__.py#L19)

librosa/feature/spectral.py
```diff
@@ -1163,6 +1164,104 @@ def logfsgram(y=None, sr=22050, S=None, n_fft=4096,
...
+            0: Fifth x-axis
+            1: Fifth y-axis
+            2: Minor x-axis
+            3: Minor y-axis
+            4: Major x-axis
+            5: Major y-axis
+
+    See Also
+    --------
+    librosa.feature.chroma_cqt
```
bmcfee: According to numpydoc style guidelines, function refs within the same submodule should be left bare, that is,

```
chroma_cqt
```
instead of
```
librosa.feature.chroma_cqt
```

This makes it easier for the hyperlinks at doc build.

(References into other submodules are fine, but you discard the package name.  So ```util.normalize``` instead of ```librosa.util.normalize```.)

librosa/feature/spectral.py
```diff
@@ -28,7 +28,8 @@
            'mfcc',
            # Deprecated functions
            'logfsgram',
-           'chromagram']
+           'chromagram',
+           'tonnetz']
```
bmcfee: Yikes!  move this one up from below the deprecation comment!

librosa/feature/spectral.py
```diff
@@ -1163,6 +1164,104 @@ def logfsgram(y=None, sr=22050, S=None, n_fft=4096,
...
+        Column-wise normalization.
+        See `librosa.util.normalize` for details.
+
+        If `None`, no normalization is performed.
+
+    Returns
+    -------
+    ton : np.ndarray [shape(6, t)]
+        Tonal centroid features for each frame.
+        Tonnetz dimensions:
```
bmcfee: What are 'x' and 'y' here?

I ask because I'm thinking it would be nice to add a tonnetz axis labeler to `display.specshow`, so you can do something like

```python
>>> ton = librosa.feature.tonnetz(y=y, sr=sr)
>>> librosa.display.specshow(ton, y_axis='tonnetz')
```
and get the right thing.

bmcfee:
Two more question, and feel free to ignore/punt/laugh at them both:

* Does the phase of the tonnetz basis depend on the phase of the chroma?  ie, if i rotate my chroma (say from C to A), does that change tonnetz?  Do we need to be aware of that?  Does your regression test look as expected?


* Does it make sense to map from cqt to tonnetz directly, or is it necessary to go through normalized chroma as an intermediate representation?

Since we vectorized the chroma->tonnetz mapping, it seems relatively trivial to replace `12` and `n_chroma` by `12 * c.shape[0] / bins_per_octave` and `bins_per_octave`.  The periodicity of the cosine basis should buy us height invariance for free.

librosa/feature/spectral.py
urinieto: Yeah, I actually thought about that.

'x' and 'y' represent the two dimensions of a circle. The tonal centroids are typically represented by a [torus](http://33.media.tumblr.com/c8a40bdbd7815016fd5ce6dc8983b8c2/tumblr_motw9jlB6I1qfadvwo1_r1_400.gif), which can be seen as three circles: fifths, minor, and major.

I believe there's no standard way of labeling the axis. In my dissertation I used (LaTeX syntax) 5_x, 5_y,  m3_x, m3_y, M3_x, M3_y. If you fancy this, I can add it to `librosa.display.specshow`.

urinieto:
My answers:

* The tonnetz will change depending on the order of the chroma. Their x-axis will still be the same (circles of fifths, minor thirds, and major thirds, in this order), but the circles will be _rotated_. Nevertheless, the same type of information should be encoded (you would just need to do some trigonometric magic to rotate the circles back to the desired position). Regression tests look as expected, however I have only compared them against my msaf function.

* No need to compute the chromagram, just the cqt should suffice. So yes, your idea should work.

urinieto:
Ok, I believe that with 0ad8dd013d1b97a6266ad00eb7c21e803303f62d we are ready to merge branches.

bmcfee:
whoa

bmcfee:
> No need to compute the chromagram, just the cqt should suffice. So yes, your idea should work.

Yeah, but upon further reflection, I don't think we really need to implement this directly.  An unnormalized chroma_cqt should get the same effect, and it's easy enough to compute.

Meanwhile: I made a few changes to the docs and add tonnetz axis display. cc1c624df671eb9060e6b7292b8f0ed19c608b27

Also, added you to the authors + updated changelog. a829b6e905e0fe4634efd0b66e4f5eb9e042251e

bmcfee:
BTW, your unit tests need to have an `assert` on `np.allclose()`.  Adding this line causes them all to fail.

Currently investigating and restructuring.

bmcfee:
Fixed it. e94552598cb99935f9acd81d281bcb935f18e3d6

urinieto:
Oops, sorry about that. And thanks for the merge (and all of the work you've put into this!).

So, now that I think about it, I'm not sure that normalizing tonnetz makes any sense at all. By doing so we are moving the points where the chords fall in the three circles (pushing them towards the unit circle), which I don't think it makes any sense at all (see Figure 4 of the [Harte paper](http://www.ofai.at/~martin.gasser/papers/oefai-tr-2006-13.pdf)).

In any case, the normalization parameter doesn't hurt, but we might probably not want to normalize by default.

bmcfee:
Indeed.  If they're actually cartesian coordinates on a circle, shouldn't the norm be 2 (if anything)?

bmcfee:
And yeah, I vote for removing normalization entirely.  If a user wants to normalize it post-hoc, that's easy enough (since it's the last step).  I'll make the change.

bmcfee:
Done.

urinieto:
Thanks titanic!

PR #182, opened 2015-04-08
Removed unused 'import os' from LibROSA demo.ipynb example

adrianholovaty:
Small thing here. I removed the "import os" from "LibROSA demo", as it's not used anywhere else in the file.


bmcfee:
Thanks.  I'll do a full pass through and audit of the example notebooks when drafting the 0.4 release, but this is a good reminder.

PR #189, opened 2015-05-08
added ability to change window type in chroma_stft and _spectogram

neomoha:
Hi Brian,

While calling the chroma_stft function I noticed that you cannot change the window type, say from the default hann to blackmannharris.

I thought it would be nice to have the ability to do that, so I forked your code and edited the two files that needed to be edited.

In case you think it's also useful for the master branch, here's a pull request.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -730,7 +730,7 @@ def perceptual_weighting(S, frequencies, **kwargs):
 
 
 @cache
-def _spectrogram(y=None, S=None, n_fft=2048, hop_length=512, power=1):
+def _spectrogram(y=None, S=None, window=None, n_fft=2048, hop_length=512, power=1):
     '''Helper function to retrieve a magnitude spectrogram.
 
     This is primarily used in feature extraction functions that can operate on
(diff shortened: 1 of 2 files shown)


bmcfee:
Hi Moha: thanks for this!

This seems useful, but I think I'd prefer to stick to a more streamlined pattern.  The feature functions are already pretty heavy in terms of call signature, and I'd prefer to not add unnecessary parameters.

You can get the desired results by doing the following:
```python
>>> D = librosa.stft(y, window=scipy.signal.blackmannharris)
>>> C = librosa.feature.chroma_stft(S=np.abs(D)**2)
```

This, in fact, is the entire point of supporting both time-series (`y, sr`) and spectrogram (`S`) inputs in the feature functions.

It might be a bit clumsier than having it all wrapped up inside `chroma_stft`, but it keeps the individual function simple and minimizes redundant functionality.  (It shouldn't be any more or less computationally efficient than your PR, modulo any memory allocation/scope creep.  If that really bothers you, I'd suggest writing a wrapper function that implements the above to limit the scope, so that things get garbage-collected as needed.)

Of course, if there are good arguments in favor of the PR as opposed to my recommendation, I'm all ears. :)

neomoha:
Hi Brian,

Thanks for your answer!

I do agree that your solution seems more streamlined. I tried it on a
sample audio and compared with my previous output.

It turns out, there is a root mean square deviation of 11.75. I've also
plotted the two approaches, which I'm attaching here.

It seems like the spectrogram approach outputs more energy overall. What do
you think?

M

neomoha:
Sorry, my bad, I wasn't passing the right parameters.

Please ignore my previous e-mail, this is what happens when you wake up
from a night out and start coding :-)

Cheers,

M.

bmcfee:
Ok, so everything's good?  (I'm guessing you passed energy when you were expecting power, or vice versa?)

Issue #128, opened 2014-11-22
Cropping and resampling leads to inconsistent results


dpwe:
I would like to be able to read and concatenate subranges with librosa.load, and have them be exactly what you'd get with a seemingly-equivalent single load, e.g.

> > > y1, sr = librosa.load(librosa.util.example_audio_file(), sr=40000, duration=0.1)
> > > y2, sr = librosa.load(librosa.util.example_audio_file(), sr=40000, duration=0.1, offset=0.1)
> > > y3, sr = librosa.load(librosa.util.example_audio_file(), sr=40000, duration=0.2, offset=0.0)
> > > print sum(np.append(y1, y2) != y3)
> > > 38

I was thinking "0" would be better.

.. but this has more serious challenges than the last sample, it also depends on the adjusting the starting sample due to offset to be consistent.

  DAn.

bmcfee:
Following up on this:

- Testing this bug should disable resampling (`sr=None`), just to eliminate all the irrelevant variables.
- It seems to be codec-dependent.  Using wav-file input, either stereo or mono, I get bit-level precision.
- With ogg and mp3, I still get failure as originally reported.  I suspect this is down to how the different codecs batch up frames, and it's really not something that I think we should invest in circumventing.

Any objections to closing this one?

bmcfee:
A bit more investigation:

- wav, au, and m4a give total agreement
- mp3 messes up only in the first frame; trimming everything past the first 1364 samples (at 44KHz) gets bit precision
- flac and ogg are messed up all the way through

In terms of audioread backend, I get:

- :+1: wav : MadFile
- :+1: au : MadFile
- :+1:  m4a : subprocess
- :-1: mp3 : MadFile
- :-1:  flac : subprocess
- :-1:  ogg : subprocess

So... it's a total mixed bag.  I still think we should punt this one, and maybe it will resolve itself through switching over to pysoundfile?

dpwe:
I agree with punting on this for now.  But I want to clear up a confusion.

The basic problem is that reading a sound file is not stateless in some
circumstances.  One obvious case is if reading the data involves some kind
of filtering, in which case the impulse response of that filter will make
the current samples depend on the past.  The reconstruction filter of
subband encoding schemes such as MP3 introduces some of this, so typically
you have to back up a few 10s of ms to "warm up" when seeking into such a
stream.  I suspect this is the somehow the source of the codec-specific
problems you report.

But the original problem I was raising was when reading sound files *with
[...]

PR #191, opened 2015-05-19
Fixed loading of audio files to make possible sample-accurate loading

romi1502:
There are two small modifications in the pull request: 
- the first one (l. 139) is a fix for a bug that appeared when s_start was a multiple of the size of the frame and that causes inaccurate duration for the loaded signal (loaded signal was in this case a frame too long). The bug can be reproduced by loading an audio file with an offset which is a multiple of the size of the frame (of the for loop l.121) when expressed in sample.
With ffmpeg backend for audioread, the following line returned a 1.023s long waveform x instead of a 1s long one (verified on both mac OS and Debian) :
x,sr = librosa.core.audio.load('test_signal.wav', sr=44100, offset=1024/44100., duration=1.0)
- the second modification (l.111 and 116) replaces ceil and floor by round for second to sample conversion which makes possible sample-accurate loading of audio files.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -108,12 +108,12 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
     with audioread.audio_open(os.path.realpath(path)) as input_file:
         sr_native = input_file.samplerate
 
-        s_start = int(np.floor(sr_native * offset)) * input_file.channels
+        s_start = int(np.round(sr_native * offset)) * input_file.channels
 
         if duration is None:
             s_end = np.inf
         else:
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks for this!

Would you mind making the following modifications:

* Pull against develop, not master
* Add tests to verify correctness both for shape and content.  For this, resampling and mono-conversion should be disabled so that it's purely a question of slicing.

We previously investigated this sort of thing in #128 , but were unable to completely resolve the issue due to differences in back-end codecs.  If you have any insight on this, or could test on different encodings, it would be much appreciated!

bmcfee:
Redundant with #192 .  Skipped.

PR #192, opened 2015-05-20
Fixed loading of audio files to make possible sample-accurate loading

romi1502:
Same content as #191 with unit test to check length and content correctness of loaded signals.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -56,6 +56,27 @@ def test_load():
         yield (__test, infile)
     pass
 
+def test_segment_load():
+    """
+    Test loading a segment. Check size accuracy
+    """
+
+    sample_len = 2003
+    fs = 44100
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -229,7 +250,7 @@ def __test(ref_power, clip):
             if six.callable(ref_power) or ref_power >= 0.0:
                 tf = __test
             else:
-                tf = raises(librosa.ParameterError)(__test)
+                tf = raises(ValueError)(__test)
```
bmcfee: I think this test is failing on travis because you reverted the exception type here (and also line 558 below).

romi1502: Hi Brian,
sorry about that, I commited the same files that I first commited in the master branch without taking care of these differences.
I just reset and redid the commits.

bmcfee:
This looks great.  Thanks again!

Issue #206, opened 2015-06-18
onset strength enhancements


bmcfee:
The current onset strength implementation provides basic spectral flux.  It's not too hard to generalize this to support the operations used in Böck and Widmer's [superflux](http://dafx13.nuim.ie/papers/09.dafx2013_submission_12.pdf) method.  Namely:

* [x] add a lag parameter for the spacing between differenced cells.  Instead of subtracting the previous cell (`lag=1`) go back further.  (This is `mu` in the reference paper.)
* [x] add a local vertical max (or other filtering operation) for the subtracted bin.
* [x] parameter-sweep to optimize beat tracking and onset detection scores

So, instead of doing
```python
flux = np.diff(S, axis=1)
```
we'd do
```python
S_filtered = scipy.ndimage.maximum_filter1d(S, filter_size, axis=0)
flux = S[:, lag:] - S_filtered[:, :-lag]
```
and then proceed as normal.

This would introduce two additional parameters:
- lag (default=1); int > 0
- max_window (default=3), int>0, or `None` to disable max-filtering

bmcfee:
resolved by merging #210

Issue #204, opened 2015-06-15
User-settable/gettable global default parameters


craffel:
Sorry  if this is addressed in another issue, I wasn't able to find reference to it although I remember it being discussed.  Now that most of the parameters are pretty standardized (`sr=22050`, `n_fft=2048`, `hop=n_fft/4`, etc), it would be cool if it was easy to retrieve these global default parameter values, or better, to set them by hand.  This would make user code potentially less verbose, but is not at all necessary.  It also could make user code slightly less error-prone, for example in the case where the user is using a sampling rate which is not 22050 - if they don't remember to update `n_fft` in every function call they make which does spectral processing, they'll have a bad time.  Is this something planned?  The main drawback that I see is that it could potentially make the documentation uglier.

bmcfee:
I like the idea, and it's something that's been in the back of my head for a while.

Do you know of any good way to do this, consistently, without wrecking the docstrings or call signatures?

craffel:
> Do you know of any good way to do this, consistently, without wrecking the docstrings or call signatures?

No, I don't.  The only thing I could think of is some kind of hack where you have a decorator which replaces keyword arguments with the config values based on their name... something like
```Python
defaults = {'sr': 22050}
def replace_defaults(foo):
    def replaced(*args, **kwargs):
        if 'sr' in kwargs:
            kwargs['sr'] = defaults['sr']
        return foo(*args, **kwargs)
    return replaced
```
Then you could still define a function with the normal defaults and it would show up as normal in docs/function code/etc., but they would be replaced by the global defaults.  Not proposing this as a good solution, but it's what I can think of OTOH.  Googling doesn't bring anything up.  The only other library I regularly use which has global defaults is Theano, but I don't think they're using them to set any keyword argument defaults.

bmcfee:
The decorator hack seems like the only viable approach here, but I doubt it can ever be made safe or dynamically consistent.

We might be able to use functools to mangle the docstrings/call signature when the function is first defined, kind of like how the `cache` object currently works.  I can see two downsides though:

* The defaults object would have to be constructed before all function definitions are executed.  This is possible, but tricky, and probably requires either environment variables or config files to pull off in a usable way.  Not that I'm against either of those things.

* The `defaults` dict would have to be frozen.  Otherwise, if it changes after the decorator executes, the  default variables (and, probably, mangled docstrings) are no longer consistent.

[...]

PR #210, opened 2015-06-22
[WIP] first pass at superflux onsets

bmcfee:
implementing #206 

Todo:

- [x] onset detection parameter sweep
- [x] beat tracking parameter sweep

### DIFF
--- a/librosa/onset.py
+++ b/librosa/onset.py
@@ -421,7 +421,8 @@ def onset_strength_multi(y=None, sr=22050, S=None, lag=1, max_size=3,
                           axis=0)
 
     # compensate for lag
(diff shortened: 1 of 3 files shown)


bmcfee:
At the default `sr=22050, hop_length=512`, the parameter sweep for onset detection maxes out (`f=0.749` on our dev set) with `lag=1` and `max_size=1`; that is, not using superflux extensions.

However, with `sr=44100, hop_length=256`, we can bump this up to `f=0.801` using `lag=4` and `max_size=1`.

This, to me, suggests that we leave the superflux extensions disabled by default, and otherwise retain our previous default parameter settings.  If, at some point in the future, we change the defaults (eg #204), we should revisit this decision.

PR #211, opened 2015-06-23
added CQT hop-length check to fix #208

### LINKED ISSUE #208: CQT bug when hop_length != 2^k
[paging @ebattenberg ]

When the hop isn't a power of two, the bottom octave(s) can fall out of alignment.

I came across this one while experimenting with different hop lengths in CQT.  Here's a [notebook](http://nbviewer.ipython.org/gist/bmcfee/c6aa2ab3d4823f69a4b5) to reproduce the error.

I think the error here is requiring `hop_length >= 2^k` (for `k >= num_octaves`).  Instead, 
it seems to me that the correct logic here would require hop_length to be an integer multiple of 2^k, and then require that `k >= num_octaves`.

Or perhaps there's a clever way to propagate the remainder of the hop after downsampling, rather than [flooring](https://github.com/bmcfee/librosa/blob/master/librosa/core/constantq.py#L227) as is currently done?

bmcfee:
Fixes #208 by adding a safety check for `hop_length % n * 2**n_octaves == 0`.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -81,7 +81,7 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     Raises
     ------
     ParameterError
-        If `hop_length < 2**(n_bins / bins_per_octave)`
+        If `hop_length` is not an integer multiple of `2**(n_bins / bins_per_octave)`
 
     See Also
     --------
@@ -131,9 +131,9 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     n_octaves = int(np.ceil(float(n_bins) / bins_per_octave))
 
     # Make sure our hop is long enough to support the bottom octave
-    if hop_length < 2**n_octaves:
-        raise ParameterError('Insufficient hop_length {:d} '
-                             'for {:d} octaves'.format(hop_length, n_octaves))
+    if np.mod(hop_length, 2**n_octaves) != 0:
+        raise ParameterError('hop_length must be an integer multiple of 2^{0:d} '
+                             'for {0:d}-octave CQT'.format(n_octaves))
 
     if fmin is None:
         # C2 by default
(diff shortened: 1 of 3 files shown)


librosa/core/constantq.py
```diff
@@ -131,9 +131,9 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     n_octaves = int(np.ceil(float(n_bins) / bins_per_octave))
 
     # Make sure our hop is long enough to support the bottom octave
-    if hop_length < 2**n_octaves:
-        raise ParameterError('Insufficient hop_length {:d} '
-                             'for {:d} octaves'.format(hop_length, n_octaves))
+    if np.mod(hop_length, 2**n_octaves) != 0 or hop_length < 2**n_octaves:
```
ebattenberg: Can one of these conditions be true and not the other?

bmcfee: Yes, `hop_length == 0` or `hop_length == - 2**n_octaves` would both pass if not for the second check.

ebattenberg: What I mean is can the second condition be true but not the first?

ebattenberg: Got it.  Maybe make that explicit?  `hop_length <= 0` ?

ebattenberg: I know the second condition was the original condition, so it's nice to keep that around though.

Either way, LGTM!

bmcfee: `hop_length == 0` right?  It's less than `2**n_octaves`, and has a remainder of 0.

bmcfee: Oh, I understand now.  `hop_length <= 0` and the original test are equivalent given the first condition, so it technically doesn't matter.

I don't have a strong preference for either flavor though.

ebattenberg:
Looks good!

PR #215, opened 2015-07-02
Address additional CQT hop_length concerns

ebattenberg:
This takes early-downsampling and two-octaves-before-downsampling into account when checking that `hop_length` constraints are met.  

I'm also assuming that `hop_length` only needs to be divisible by `2**(n_octaves-1)` since that is how many times we'd need to downsample.  Not sure if that's a valid assumption since the previous criterion was `hop_length >= 2**n_octaves`.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -226,7 +226,8 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         if i > 0:
             my_y = audio.resample(my_y, my_sr, my_sr/2.0, res_type=res_type)
             my_sr = my_sr / 2.0
-            my_hop = int(my_hop / 2.0)
+            assert my_hop % 2 == 0
+            my_hop //= 2
 
         # Compute a dynamic hop based on n_fft
         my_cqt = __variable_hop_response(my_y, n_fft,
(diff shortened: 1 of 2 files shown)


bmcfee:
These changes sound good.  It looks like the tests are failing by not throwing errors where they should; can you look into this?

ebattenberg:
> These changes sound good. It looks like the tests are failing by not throwing errors where they should; can you look into this?

Yeah.  Those travis logs are so dang long, I got bored of scrolling.  How do you quickly locate the failing tests?

bmcfee:
Click the "Download log" link to skip over the fancy formatting, then hit 'end'.

librosa/core/constantq.py
```diff
@@ -552,3 +551,19 @@ def __early_downsample(y, sr, hop_length, res_type, n_octaves,
         sr = sr // downsample_factor
```
ebattenberg: Should this be floating point division?

bmcfee: Good point; I suppose there's no reason to force sr to be integer-typed.

There may be points elsewhere in the library that check for this though.  I'll investigate and maybe start a separate issue for that.

ebattenberg:
K, I fixed the failing tests (removed one of them).  Also, I changed halving of the sample rate to floating point division in one place.

bmcfee:
Thanks!  This all looks great.

For the record: which test was removed and why?

ebattenberg:
> For the record: which test was removed and why?

The check for `ParameterError` in `hybrid_cqt` when the `hop_length` is invalid.  The `cqt` routine checks for this already, and it is what is called by `hybrid_cqt` anyway to do the stuff where hop_length matters.   Also, it's not trivial to figure out what should throw an error for hybrid_cqt since it only uses an actual cqt for a subset of the bins.

bmcfee:
I see.  Thanks!

bmcfee:
... is this ready to be merged, then?

ebattenberg:
Yes, sir.

PR #225, opened 2015-08-12
[wip] Sonify markers

### LINKED ISSUE #216: Enhancement: Add an audio marker to onsets like essentia.standard.AudioOnsetsMarker
Essentia has this handy feature, [AudioOnsetsMarker](http://essentia.upf.edu/documentation/reference/std_AudioOnsetsMarker.html), which adds a beep to the location of each onset. I've found the auditory cue as helpful as the visual cue, if not more so, for assessing the correctness of onsets.

For example, see the last cell of [this IPython notebook](http://musicinformationretrieval.com/onset_detection.html).

Perhaps it might look something like:

    onset_frames = librosa.onset.onset_detect(x, fs)
    x_with_beeps = librosa.onset.onset_marker(x, onset_frames)

bmcfee:
Porting the `clicks` from mir_eval to resolve #216

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -579,10 +579,12 @@ def clicks(times=None, frames=None, sr=22050, hop_length=512,
     >>> y_beats = librosa.clicks(frames=beats, sr=sr, length=len(y))
 
     >>> # Or use timing instead of frame indices
-    >>> y_beat_times = librosa.clicks(times=librosa.frames_to_time(beats, sr=sr), sr=sr)
+    >>> times = librosa.frames_to_time(beats, sr=sr)
+    >>> y_beat_times = librosa.clicks(times=times, sr=sr)
 
-    >>> # Or with a click frequency of 880 and a 500ms sample
-    >>> y_beat_times = librosa.clicks(frames=beats, sr=sr, click_freq=880, click_duration=0.5)
+    >>> # Or with a click frequency of 880Hz and a 500ms sample
+    >>> y_beat_times880 = librosa.clicks(times=times, sr=sr,
+    ...                                  click_freq=880, click_duration=0.5)
     """
 
     # Compute sample positions from time or frames
@@ -608,7 +610,11 @@ def clicks(times=None, frames=None, sr=22050, hop_length=512,
             raise ParameterError('click_freq must be strictly positive')
 
         angular_freq = 2 * np.pi * click_freq / float(sr)
-        click = np.logspace(0, -10, num=int(sr * click_duration), base=2.0)
+
+        click = np.logspace(0, -10,
+                            num=int(np.round(sr * click_duration)),
+                            base=2.0)
+
         click *= np.sin(angular_freq * np.arange(len(click)))
 
     # Set default length
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -10,11 +10,14 @@ import audioread
 import numpy as np
 import scipy.signal
 
+
+from .time_frequency import frames_to_samples, time_to_samples
 from .. import cache
 from .. import util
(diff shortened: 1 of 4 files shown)


bmcfee:
#216 Paging @stevetjoa and @craffel 

I think this is ready to be merged, but a once-over from the both of you would be nice.

I added a few enhancements to the mir_eval version from which it is cribbed:

- positions can be specified as either times or frames
- synthesis is additive, so you shouldn't get any pops with overlapping clicks
- frequency and duration are controllable.  You can still supply a sample buffer to use instead of a click/ping.
- click synthesis is slightly more efficient
- input validation and testing following our usual conventions

Let me know if I missed anything, or if something seems incorrect.  If not, I'll merge after getting the thumbs-up.

librosa/core/audio.py
```diff
@@ -521,6 +524,127 @@ def zero_crossings(y, threshold=1e-10, ref_magnitude=None, pad=True,
...
+        frequency (in Hz) of the default click signal.  Default is 1KHz.
+
+    click_duration : float > 0
+        duration (in seconds) of the default click signal.  Default is 100ms.
+
+    click : np.ndarray or None
+        optional click signal sample to use instead of the default blip.
+
+    length : int > 0
+        desired number of samples in the output signal,
```
craffel: Looks like this got cut off.  In `mir_eval`, it says 
```
    length : int
        desired number of samples in the output signal,
        defaults to times.max()*fs + click.shape[0] + 1
```

librosa/core/audio.py
```diff
@@ -521,6 +524,127 @@ def zero_crossings(y, threshold=1e-10, ref_magnitude=None, pad=True,
...
+        if click_freq <= 0:
+            raise ParameterError('click_freq must be strictly positive')
+
+        angular_freq = 2 * np.pi * click_freq / float(sr)
+
+        click = np.logspace(0, -10,
+                            num=int(np.round(sr * click_duration)),
+                            base=2.0)
+
+        click *= np.sin(angular_freq * np.arange(len(click)))
```
craffel: As a high-level comment, if you're going to offer parametrized click synthesis, that may be something useful to have as a separate function which `clicks` calls.  But, this is fine too.

craffel:
> synthesis is additive, so you shouldn't get any pops with overlapping clicks

Hm, that's an interesting change.  Weird things can happen when you additively combine these beeps too, though.  Just less weird I guess.

> click synthesis is slightly more efficient

How?

librosa/core/audio.py
bmcfee: Yeah, fair enough.  It's not obvious to me yet that it's worth factoring out, but I'll consider it.

librosa/core/audio.py
bmcfee: I cut that second line because it's not entirely correct now.  If `click` is not provided, it's a meaningless statement, and since we can support `frames` as well as `times`, it's not clear that this is the best way to express the duration.

Also, I dropped the +1; why was that there?

bmcfee:
> > click synthesis is slightly more efficient
> How?

In `mir_eval`, a 1-second sine wave is generated and then exponentially weighted.  This ends up making a bigger buffer than you need.  This version generates the weighting function first, and then samples the sine only for that duration.  It's not a big change, but it's something.

stevetjoa:
Looks great! I love the parameter options. Thanks a lot for doing this, and Colin, thanks for the original version in `mir_eval`.

librosa/core/audio.py
craffel: > I cut that second line because it's not entirely correct now.

You should change the comma to a period.

> Also, I dropped the +1; why was that there?

I don't remember.

Issue #221, opened 2015-07-12
tempo and rhythm enhancements


bmcfee:
[cc @mprockup]

The current functionality for tempo and rhythm is pretty basic; we should do something about that.

#210 improves the basic onset strength implementation.  From that, we can add a few interesting pieces, such as tempo-invariant rhythm descriptors via the Mellin scale transform (see [here](http://www.csd.uoc.gr/~hannover/MMILab-Andre_files/HolzapfelIEEEScale.pdf)), and some of the multi-band accent descriptors from @mprockup.

This will be a not-small job, and probably require many PRs.

For now, we should make a list of functions to be added.  Please revise and expand as needed in comments below.

- [x] #231 mellin scale transform
- [x] #226 multi-band onset aggregation.  This can probably be done by passing band markers into `onset_strength`.
- ~~DCT+median filtering?~~
- ~~local onset strength (onset.onset_strength + util.frame, plus some convenient defaults to make it line up with spectral features)~~
- [x] #247 local onset autocorrelation (short-time tempogram)
- [x] #251 fix onset alignment to match input spectrogram

dpwe:
May I take a stab at demystifying the Mellin transform?

The idea is to represent the autocorrelation on a log-lag time scale
(it has to be the autocorrelation, because we need a fixed zero-time
reference).  As with log-frequency, this mapping transforms a
*scaling* of the linear axis (i.e., a tempo change) to a *shift* in
the log-scale axis.  The Mellin transform then takes the Fourier
transform of this representation, whose magnitude, of course, is
invariant to shifts (excluding edge effects).

A log-lag axis poses problems for small time intervals, just as a
log-frequency axis poses problems for low frequencies.

A complete invariance to tempo invariance raises the risk of
inadvertently matching things that have very large differences in
tempo, which is probably not the perceptual equivalence we're hoping
for.  But I guess choosing the actual range of log-lag values that are
sampled (i.e., determining the impact of the edge effects) can control
this.

  DAn.

dpwe:
By which I meant to say: I think log-lag autocorrelation is a
generally useful and intuitive representation, much like log-frequency
spectrum.

  DAn.

bmcfee:
multi-band onsets are now in place after merging #229.

bmcfee:
Is there a canonical parameterization for the (discrete) scale transform?

[...]

PR #231, opened 2015-08-19
Fast Mellin Transform

bmcfee:
#221

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -857,7 +857,7 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='slinear', beta=0.5, over_sample=2, axis=
         raise ParameterError('y must be finite everywhere')
 
     # original grid
-    x = np.arange(n)
+    x = np.arange(0, n)
 
     # build the interpolator
     f_interp = scipy.interpolate.interp1d(x, y, kind=kind, axis=axis)
@@ -867,11 +867,11 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='slinear', beta=0.5, over_sample=2, axis=
     x_exp = np.logspace(np.log(t_min) / log_base,
                         np.log(n - 1) / log_base,
                         num=n_fmt,
-                        endpoint=True,
+                        endpoint=False,
                         base=base)
 
     # Clean up any rounding errors at the boundaries of the interpolation
-    x_exp = np.unique(np.clip(x_exp, t_min, n - 1))
+    x_exp = np.clip(x_exp, t_min, x[-1])
 
     # Resample the signal
     y_res = f_interp(x_exp)
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -886,7 +886,7 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='slinear', beta=0.5, over_sample=1, axis=
     elif n_fmt < 3:
         raise ParameterError('n_fmt must be a positive integer')
     else:
-        log_base = np.log(n_fmt - 1) - np.log(n_fmt - 2)
+        log_base = (np.log(n_fmt - 1) - np.log(n_fmt - 2)) / over_sample
 
     if not np.all(np.isfinite(y)):
         raise ParameterError('y must be finite everywhere')
@@ -901,11 +901,12 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='slinear', beta=0.5, over_sample=1, axis=
     # build the new sampling grid
     # exponentially spaced between t_min/n and 1
     # we'll go one past where we need, and drop the last sample
+    n_over = int(np.ceil(over_sample))
     x_exp = np.logspace((np.log(t_min) - np.log(n)) / log_base,
                         0,
-                        num=n_fmt + 1,
+                        num=n_fmt + n_over,
                         endpoint=False,
-                        base=base)[:-1]
+                        base=base)[:-n_over]
 
     # Clean up any rounding errors at the boundaries of the interpolation
     x_exp = np.clip(x_exp, float(t_min) / n, x[-1])
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -760,7 +760,7 @@ def perceptual_weighting(S, frequencies, **kwargs):
 
 
 @cache
-def dst(x, lag_min=1, n_bins=128, delta_c=1.0, axis=-1, aggregate=None):
+def dst(x, lag_min=1, n_bins=None, delta_c=None, axis=-1, aggregate=None):
     """The discrete scale transform (DST) of a signal x.
 
     Parameters
@@ -771,11 +771,14 @@ def dst(x, lag_min=1, n_bins=128, delta_c=1.0, axis=-1, aggregate=None):
     lag_min : int > 0
         The minimum sample lag (in samples)
 
-    n_bins : int > 0
-        The number of scale transform bins to use
+    n_bins : int > 0 or None
+        The number of scale transform bins to use.
+        If None, then `n_bins = ceil(n * log(n))` is taken,
+        where `n = x.shape[n]`
 
-    delta_c : float > 0
-        The spacing between scale bins
+    delta_c : float > 0 or None
+        The spacing between scale bins.
+        If None, then `delta_c = pi / log(1 + n / lag_min)` is taken.
 
     axis : int
         The axis along which to transform `x`
@@ -830,13 +833,19 @@ def dst(x, lag_min=1, n_bins=128, delta_c=1.0, axis=-1, aggregate=None):
     >>> plt.tight_layout()
     """
 
-    if n_bins < 1:
-        raise ParameterError('n_bins must be a positive integer')
+    n = x.shape[axis]
 
     if lag_min < 1:
         raise ParameterError('lag_min must be a positive integer')
 
-    if delta_c <= 0:
+    if n_bins is None:
+        n_bins = np.ceil(n * np.log(n))
+    elif n_bins < 1:
+        raise ParameterError('n_bins must be a positive integer')
+
+    if delta_c is None:
+        delta_c = np.pi / np.log(1 + float(n_bins) / lag_min)
+    elif delta_c <= 0:
         raise ParameterError('delta_c must be strictly positive')
 
     # build the lag-sampled differential of x
@@ -846,7 +855,7 @@ def dst(x, lag_min=1, n_bins=128, delta_c=1.0, axis=-1, aggregate=None):
             slices[axis] = slice(0, None, lag_min)
             x_agg = x[slices]
         else:
-            x_agg = util.sync(x, np.arange(0, x.shape[axis], lag_min),
+            x_agg = util.sync(x, np.arange(0, n, lag_min),
                               axis=axis, aggregate=aggregate)
     else:
         x_agg = x
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -777,7 +777,7 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='slinear', beta=0.5, over_sample=1, axis=
     t_min : float > 0
         The minimum time spacing (in samples)
 
-    n_fmt : int > 0 or None
+    n_fmt : int > 2 or None
         The number of scale transform bins to use.
         If None, then `n_bins = over_sample * ceil(n * log((n-1)/t_min))` is taken,
         where `n = y.shape[axis]`
@@ -803,7 +803,7 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='slinear', beta=0.5, over_sample=1, axis=
     Raises
     ------
     ParameterError
-        if `n_fmt < 1` or `t_min <= 0`
+        if `n_fmt < 2` or `t_min <= 0`
         or if `y` is not finite
         or if `y.shape[axis] < 3`.
 
@@ -875,21 +875,23 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='slinear', beta=0.5, over_sample=1, axis=
     if t_min <= 0:
         raise ParameterError('t_min must be a positive number')
 
-    log_base = np.log(n - 1) - np.log(n - 2)
-    base = np.exp(log_base)
 
     if n_fmt is None:
         if over_sample < 1:
             raise ParameterError('over_sample must be >= 1')
 
+        log_base = np.log(n - 1) - np.log(n - 2)
         n_fmt = int(np.ceil(over_sample * (np.log(n - 1) - np.log(t_min)) / log_base))
 
-    elif n_fmt < 1:
+    elif n_fmt < 3:
         raise ParameterError('n_fmt must be a positive integer')
+    else:
+        log_base = np.log(n_fmt - 1) - np.log(n_fmt - 2)
 
     if not np.all(np.isfinite(y)):
         raise ParameterError('y must be finite everywhere')
 
+    base = np.exp(log_base)
     # original grid: signal covers 0 to 1
     x = np.linspace(0, 1, num=n, endpoint=False)
 
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -6,6 +6,7 @@ import numpy as np
 import scipy.fftpack as fft
 import scipy
 import scipy.signal
+import scipy.interpolate
 import six
 
 from . import time_frequency
@@ -17,7 +18,7 @@ __all__ = ['stft', 'istft', 'magphase',
            'ifgram',
            'phase_vocoder',
            'logamplitude', 'perceptual_weighting',
-           'dst']
+           'fmt']
 
 
 @cache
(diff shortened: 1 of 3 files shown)


bmcfee:
... needs tests

bmcfee:
I think this one is feature-complete.  Prior to merging, a little code-review from @dpwe and/or @mprockup would be appreciated. :)  

Some commentary:

* As mentioned elsewhere #221 , this PR implements the method of [DeSena and Rocchesso](http://asp.eurasipjournals.com/content/pdf/1687-6180-2007-089170.pdf), and not the Williams and Zaluba method.  If comparing the two, your mileage may vary.  Anecdotally, this method seems both faster and more stable at the low-order coefficients.
* I was having a lot of difficulty getting the sampling scheme described in the paper to work out, that is, produce approximately equivalent and aligned scale spectra for time-stretched signals.  I modified the scheme a little so that rather than sample the interval `[0, n-1]` (as described in the paper), we first normalize the time axis to cover `[0, 1]`.  This way, you can ensure that two time-stretched variations of a signal will yield identical (relative) sampling coordinates; without that guarantee, subsequent approximations due to function interpolation may diverge dramatically (eg, shfiting the spectrum, which is bad!).
* The unit tests cover the basics: error handling, output shape, etc.  There's one set of "correctness" tests that generates a single-cycle sine wave, stretches it, and compares the resulting scale coefficients for numerical precision.  Better interpolation modes (linear < quadratic < cubic) yield lower tolerance thresholds, as one might hope.
* The `over_sample` parameter seems to cause more harm than good; I get the most accurate results in the stretch tests when `over_sample=1` (DeSena suggest 2 for critical sampling in fft).  This might come down to rounding errors, but it makes me a little nervous.  Anyone have thoughts on this?
* Inferring `n_fmt` from the sampling formula works, but doesn't generate consistent results for stretched signals.  Forcing n_fmt on a stretched signal does work though, and should be encouraged IMO.  (Incidentally, `n_fmt=512` looks nice for local onset strength autocorrelations.)
* Because of the modifications to the sampling strategy, I'm not sure we can implement a proper inverse transform.  I can try, but I'd rather not sink the effort unless there's need for it; so... do we want/need this?  I can't really imagine a use case, apart from cool-factor.
* The default `t_min` is set to 0.5 (samples).  D&R don't report results for different `t_min` values, but half a sample combined with linear interpolation makes sense to me, since it would let you recover `y[0]` by extrapolation.

bmcfee:
Quick note: putting a bit more thought into explicit setting of `n_fmt`: the exponential sampling base should depend on the target sampling rate, not the rate inferred from the length of the input signal (as derived in D&R).

This should be a quick fix later this afternoon.

bmcfee:
> It's also possible that there's a boundary condition problem in determining the position of the final sample, which can cause some strangeness. Not 100% sure I understand this yet.

The boundary condition problem boils down to clipping the interpolation grid.  There's no problem generating the correct sampling points, but they then have to be clipped to fit within the range accepted by the interpolators.  (I added this clipping mostly to deal with floating point cruft, but it's a good sanity check anyway.)  This is normally fine, but combining this with `over_sampling > 1` can cause some of the tests to fail.

On the plus side, the `over_sampling` parameter now works properly (with tests).

bmcfee:
As I've mentioned elsewhere, I think this is ready to merge, but an extra set of eyes would be helpful before pulling the trigger.

dpwe:
OK, I'm listening.  But I'm not familiar with the source papers, so it's
not the easiest thing for me to review.

I'll take a look.

  DAn.

bmcfee:
hooray

In all seriousness... yeah, I know this stuff isn't particularly easy to review, so really anything would be greatly appreciated!

The algorithm itself is pretty simple: exponential resampling following by FFT.  Of course, the devil's in the details of getting the resampling exactly right.  The unit tests cover this (up to a point), so mainly I'm just hoping for an extra set of eyes to make sure it looks sane and there are no glaring oversights.

bmcfee:
Note: i expanded the docs and comments to clarify a few of the more mysterious-looking bits of the code.

dpwe:
I read the paper and looked over the code.

I guess I would trust this more if we had inverse FMT and it worked ;-)

I'm not sure why you implemented beta rather than only supporting the scale
transform.  I have no idea if your implementation is right for beta other
than 0.5.  And I'm not at all sure it's strictly correct even in that case:
My understanding of the scale transform lines up with eqn 9. of De Sena and
Rocchesso i.e. we resample on an exponential grid, then take the Fourier
transform - no windowing by exp(beta t) required.

Now I look in more detail at that part of the paper, I think it's not quite
internally consistent: If g(t) = f(e^t) for the scale transform, which is
supposed to be beta=0.5, then it can't also be g(t) = f(e^t) exp(beta t)
for general beta.  It should be something like f(e^t) exp((beta - 0.5) t)
at a guess.

This is just a weighting of the time axis, but it is kinda nasty.  Do your
tests work without it?

Oh, reading more closely, equation (9) is with beta = 0, so it's what it
says.  In the intro it says beta=0.5 gives the "scale transform", but
beta=0 gives "a compression/expansion invariant transform", which is sort
of what we want (well, it's only the magnitude that is invariant), but
maybe only up to a scale factor.

It seems clear to me that FT{f(e^t)} will show the desired
dilation-becomes-shift-becomes-phase-twist behavior, but so will FT{a^t
f(e^t)}.  The a^t term looks like something you include to ensure
convergence, but convergence is not an issue for our finite range.

I concur that x_exp**beta is what you want for (e^{\beta t}).

Maybe that was the source of your problems before you renormalized the time
axis - you were windowing with something that got as big as e^128 or
something which doesn't lead to happy floating point operations.  Probably
just running with beta=0 (or dropping the beta term) would just make things
better.  As it stands, we're still exponentially weighting the tail of the
sequence: why is that something we'd want to do?

I'm disturbed that you had to use np.clip, but I guess this is just
rounding at the lower edge? I'm not sure how np.logspace works; I would
have thought it could get both ends pretty tight.  Are you saying that you
were trying to run the logspace down to log(0.5)

It also makes no sense to me that oversampling makes things worse.  It
makes a lot of sense to me that it makes things better, as declared in the
paper.

I kind of like your argument for t_min = 0.5, but that last octave from
t_min=1 to t_min=0.5 is costing you a lot of samples.  And you shouldn't be
using linear interpolation (because we need band limited interpolation), so
you can't really extrapolate to zero.  I see how you can, but you shouldn't
even want to IMO.

Actually I just tried it with n=128, and an octave is less than 100
samples, so it's not such a big deal.  Also, the exponential window for
beta=0.5 is not such a big deal either (down weights the very smooth values
near 0 by about 0.1), although I still don't think it's helping any.

So I think it's all good.  The only part I'm unhappy with is the np.clip; I
can't see how we could get anywhere near to exceeding 0..N-1/N given that
we're deliberately starting some distance away from zero, and that we're
only going up to one step before 1.0.  I suspect you don't actually need
this any more.

BTW, I think inverseFMT is pretty simple - you just set up the interpolator
using x_exp, and sample at x.  I don't know what it's useful for, but it's
not intimidating.

Hope this is what you were wanting.  Sorry that it's a bit rambling.  I
didn't think about what happens with multidimensional input, but I trust
you to get that right.
  DAn.

bmcfee:
Thanks for all this!  Responses inline:

> I read the paper and looked over the code.
> 
> I guess I would trust this more if we had inverse FMT and it worked ;-)
> 
> I'm not sure why you implemented beta rather than only supporting the scale
> transform.  I have no idea if your implementation is right for beta other
> than 0.5.  And I'm not at all sure it's strictly correct even in that case:
> My understanding of the scale transform lines up with eqn 9. of De Sena and
> Rocchesso i.e. we resample on an exponential grid, then take the Fourier
> transform - no windowing by exp(beta t) required.
> 

I implemented the full mellin transform simply because it's no additional work, so we may as well go as general as possible.

The relevant equation is 16, not 9.

> Now I look in more detail at that part of the paper, I think it's not quite
> internally consistent: If g(t) = f(e^t) for the scale transform, which is
> supposed to be beta=0.5, then it can't also be g(t) = f(e^t) exp(beta t)
> for general beta.  It should be something like f(e^t) exp((beta - 0.5) t)
> at a guess.
> 

The paper states ```g(t) = f(e^t)``` for Mellin with ```p= -jc``` (ie, ```beta==0```), and more generally ```g(t) = f(e^t) (e^t)^beta```.

> This is just a weighting of the time axis, but it is kinda nasty.  Do your
> tests work without it?
> 
> Oh, reading more closely, equation (9) is with beta = 0, so it's what it
> says.  In the intro it says beta=0.5 gives the "scale transform", but
> beta=0 gives "a compression/expansion invariant transform", which is sort
> of what we want (well, it's only the magnitude that is invariant), but
> maybe only up to a scale factor.
> 
> It seems clear to me that FT{f(e^t)} will show the desired
> dilation-becomes-shift-becomes-phase-twist behavior, but so will FT{a^t
> f(e^t)}.  The a^t term looks like something you include to ensure
> convergence, but convergence is not an issue for our finite range.
> 
> I concur that x_exp**beta is what you want for (e^{\beta t}).
> 

Ok... Should I interpret that last statement as "this looks okay now"?

> Maybe that was the source of your problems before you renormalized the time
> axis - you were windowing with something that got as big as e^128 or
> something which doesn't lead to happy floating point operations.  Probably
> just running with beta=0 (or dropping the beta term) would just make things
> better.  As it stands, we're still exponentially weighting the tail of the
> sequence: why is that something we'd want to do?
> 
> I'm disturbed that you had to use np.clip, but I guess this is just
> rounding at the lower edge? I'm not sure how np.logspace works; I would
> have thought it could get both ends pretty tight.  Are you saying that you
> were trying to run the logspace down to log(0.5)

It's rounding at both edges, but yes.  It shouldn't be an issue, but the scipy interpolators are finnicky with domain boundaries.  In my preliminary tests, I was seeing sample points at `0.499999999997` (or somesuch) that throw errors for being less than `t_min = 0.5`.  Similar things happen that the right edge.

> 
> It also makes no sense to me that oversampling makes things worse.  It
> makes a lot of sense to me that it makes things better, as declared in the
> paper.
> 

I agree that it should work, and it does, but it also involves using a finer sampling grid, and it's therefore more likely to get slightly out of alignment when using two differently scaled test samples.  For long enough test signals, it does work out fine, but using these in the unit tests causes everything to slow to a crawl.

> I kind of like your argument for t_min = 0.5, but that last octave from
> t_min=1 to t_min=0.5 is costing you a lot of samples.  And you shouldn't be
> using linear interpolation (because we need band limited interpolation), so
> you can't really extrapolate to zero.  I see how you can, but you shouldn't
> even want to IMO.

Sure.  I just use linear as an example of how one might reconstruct `t=0`; the whole idea here is to be as agnostic as possible to the interpolation mode, since it's scipy's job to get that right.  In my testing though, I don't see much qualitative difference between linear and quadratic.

> 
> Actually I just tried it with n=128, and an octave is less than 100
> samples, so it's not such a big deal.  Also, the exponential window for
> beta=0.5 is not such a big deal either (down weights the very smooth values
> near 0 by about 0.1), although I still don't think it's helping any.
> 

I think the windowing is necessary for the magnitude scale-invariance property to hold.  Is this not proved in Cohen'93?

> So I think it's all good.  The only part I'm unhappy with is the np.clip; I
> can't see how we could get anywhere near to exceeding 0..N-1/N given that
> we're deliberately starting some distance away from zero, and that we're
> only going up to one step before 1.0.  I suspect you don't actually need
> this any more.

Yeah, the clip makes me nervous, but for different reasons.  Taking it out does break things because of rounding.  I'm only hesitant about clipping because I can imagine some bizarre corner case where the same point gets sampled twice and everything gets thrown off by one.  I think this would require an absurdly small `t_min` setting though.

> 
> BTW, I think inverseFMT is pretty simple - you just set up the interpolator
> using x_exp, and sample at x.  I don't know what it's useful for, but it's
> not intimidating.
> 

It's conceptually simple, but reconstructing `f(0)` is non-trivial.  I didn't want to have that issue block the merge of this PR, especially since the inverse transform doesn't have a specific application driving it.

> Hope this is what you were wanting.  Sorry that it's a bit rambling.  I
> didn't think about what happens with multidimensional input, but I trust
> you to get that right.

Totally.  Thanks again!

Multidimensional input is handled in the tests and works as expected.  (You can only transform one axis at a time, so we don't have a 2d-FMT.  I'm okay with that.)

>   DAn.

dpwe:
So I was thinking about it more, and now I understand why we have
beta=0.5: because of the stretch in the x-axis, to preserve the total
energy, it's necessary to downscale the amplitude in inverse
proportion to the (square root) of the time dilation, hence alpha^t.
So, sorry for objecting to that.

> Ok... Should I interpret that last statement as "this looks okay now"?

Not that statement, but it does look OK now.

> I'm disturbed that you had to use np.clip, but I guess this is just
> rounding at the lower edge? I'm not sure how np.logspace works; I would
> have thought it could get both ends pretty tight. Are you saying that you
> were trying to run the logspace down to log(0.5)
>
> It's rounding at both edges, but yes. It shouldn't be an issue, but the
> scipy interpolators are finnicky with domain boundaries. In my preliminary
> tests, I was seeing sample points at 0.499999999997 (or somesuch) that throw
> errors for being less than t_min = 0.5. Similar things happen that the right
> edge.

But the *interpolator* isn't imposing t_min = 0.5.  The interpolator
has a point at t = 0, so for any sensible t_min, we'll never get
within rounding error of zero.  Why are you hanging on to the clip?

>> It also makes no sense to me that oversampling makes things worse. It
>> makes a lot of sense to me that it makes things better, as declared in the
>> paper.
>
> I agree that it should work, and it does, but it also involves using a finer
> sampling grid, and it's therefore more likely to get slightly out of
> alignment when using two differently scaled test samples. For long enough
> test signals, it does work out fine, but using these in the unit tests
> causes everything to slow to a crawl.

Really?  I see that oversampling caused you pain in sampling because
my test was 701 points with oversample=1 and 1401 with oversample=2; I
didn't follow the details, but I hate those end-point issues in
resampling.

But I don't understand your argument at all.  You're saying that in
your test cases you're trying to match the FMTs of different-lengthen
signals, so this makes it hard to line up?  What's the scenario in
which anything could ever be off-by-one?

Oh, is it because we always start at the last point, and scale
backwards from there?

How about you enforce oversampling to be integer?  I don't think
people really want irrational oversampling.

>> BTW, I think inverseFMT is pretty simple - you just set up the interpolator
>> using x_exp, and sample at x. I don't know what it's useful for, but it's
>> not intimidating.
>
> It's conceptually simple, but reconstructing f(0) is non-trivial. I didn't
> want to have that issue block the merge of this PR, especially since the
> inverse transform doesn't have a specific application driving it.

Agreed, and agreed to not block the PR.  But reconstructing for t <
t_min is a weird thing to get hung up on - leaving those points as
zero seems completely fine to me.

  DAn.

bmcfee:
> But the *interpolator* isn't imposing t_min = 0.5.  The interpolator
> has a point at t = 0, so for any sensible t_min, we'll never get
> within rounding error of zero.  Why are you hanging on to the clip?

Derp; yes, you're right.  It was only the right edge that had the clipping issue.  I guess we could drop the bottom-end clipping and still be fine, but the right-edge will always be a nuisance.



> Oh, is it because we always start at the last point, and scale
> backwards from there?
> 

Yes, exactly.

> How about you enforce oversampling to be integer?  I don't think
> people really want irrational oversampling.

I'm not sure that would fix anything, since the number of samples is determined by rounding anyway (see here)[https://github.com/bmcfee/librosa/pull/231/files#diff-25de824aab720f5de88a5f1182d98ce6R900].

librosa/core/spectrum.py
```diff
@@ -759,6 +761,197 @@ def perceptual_weighting(S, frequencies, **kwargs):
...
+    >>> plt.legend(frameon=True)
+    >>> plt.axis('tight')
+    >>> plt.subplot(1, 2, 2)
+    >>> plt.semilogy(np.abs(scale1), label='Original')
+    >>> plt.semilogy(np.abs(scale2), label='Stretched')
+    >>> plt.xlabel('scale coefficients')
+    >>> plt.title('Scale transform magnitude')
+    >>> plt.legend(frameon=True)
+    >>> plt.axis('tight')
+    >>> plt.tight_layout()
```
jakevdp: Does doing these calls here screw up doctests? I've had issues with that before. Not sure if you're running those in this project

jakevdp:
Beautiful code. I want to read-up on the Mellin transform a bit and then give it a closer look.

librosa/core/spectrum.py
bmcfee: We're not currently using doctests, but the documentation does build properly.

It's a separate issue to get our doc builds running on sphinx 1.3 and reinstitute doctests properly.

jakevdp:
Reading through the paper... one observation:

>  The Hermite spline is a third-degree spline with each polynomial of the spline in Hermite form.

The current default in the code is a linear spline (``slinear``). Perhaps you should default to cubic instead?

librosa/core/spectrum.py
```diff
@@ -759,6 +761,197 @@ def perceptual_weighting(S, frequencies, **kwargs):
 
 
 @cache
+def fmt(y, t_min=0.5, n_fmt=None, kind='slinear', beta=0.5, over_sample=1, axis=-1):
+    """The fast Mellin transform (FMT) [1]_ of a signal y.
```
jakevdp: I'd say explicitly a "uniformly-sampled signal y"

librosa/core/spectrum.py
```diff
@@ -759,6 +761,197 @@ def perceptual_weighting(S, frequencies, **kwargs):
...
+        IEEE Transactions on Signal Processing 41, no. 12 (1993): 3275-3292.
+
+    Parameters
+    ----------
+    y : np.ndarray, real-valued
+        The input signal(s).  Can be multidimensional.
+        The target axis must contain at least 3 samples.
+
+    t_min : float > 0
+        The minimum time spacing (in samples)
```
jakevdp: Some pointer on how to choose this value could be helpful.

librosa/core/spectrum.py
```diff
@@ -759,6 +761,197 @@ def perceptual_weighting(S, frequencies, **kwargs):
...
+    n_over = int(np.ceil(over_sample))
+    x_exp = np.logspace((np.log(t_min) - np.log(n)) / log_base,
+                        0,
+                        num=n_fmt + n_over,
+                        endpoint=False,
+                        base=base)[:-n_over]
+
+    # Clean up any rounding errors at the boundaries of the interpolation
+    # The interpolator gets angry if we try to extrapolate, so clipping is necessary here.
+    x_exp = np.clip(x_exp, float(t_min) / n, x[-1])
```
jakevdp: I'm not sure that clipping is the best choice here... shouldn't this be a mask instead? i.e. only use values that you actually can constrain, rather than (effectively) extrapolating with a horizontal line.

librosa/core/spectrum.py
bmcfee: :+1:

librosa/core/spectrum.py
bmcfee: I'm glad that you picked up on this point, since it's certainly the most head-scratching part of this implementation.

Dan and I went back and forth on this in the PR thread earlier.  I think we all agree that clipping is generally terrible.  The reason that it's included here is to resolve floating point precision errors at the interpolation boundaries.  It's not that we're extrapolating exactly, it's that the sample point generator is fuzzing slightly out of the original sample domain --- ``[t_min, (n-1)/n]`` --- and we need to lock it back down.

Using a mask (throwing out these points, if I understand correctly) seems incorrect here, since it would throw off the sample count.  Or maybe there's a more clever way to do this?

jakevdp:
I think this looks good, but it's hard for me to evaluate whether it's algorithmically correct. I think the sine-wave tests are a start, but there may be some global issue that it doesn't catch.

One thing you might do is write a test utility which makes use of a known smooth function, so that you can evaluate the result without the spline resampling step. Comparing this "continuous" transform with the discrete may be a more robust check.

Another strategy perhaps: from the paper, it looks like the inverse transform is relatively straightforward to implement based on the forward transform – perhaps implementing that and doing a number of forward-backward identity tests would be more convincing!

Overall, though, the code and algorithm look solid.

librosa/core/spectrum.py
bmcfee: [summarizing offline convo]

It's probably a good sanity check to wrap the clip in an if statement to see if it's meaningful.  It may be an effective no-op, but it would make the code more readable.

bmcfee:
> The current default in the code is a linear spline (slinear). Perhaps you should default to cubic instead?

Good call, now implemented.  I haven't noticed much difference in the results empirically, but I agree that it's saner to default to high-quality rather than high-speed.

bmcfee:
> One thing you might do is write a test utility which makes use of a known smooth function, so that you can evaluate the result without the spline resampling step. Comparing this "continuous" transform with the discrete may be a more robust check.

I've actually implemented the reverse of this (offline): I pulled out the resampling code as a self-contained function and verified that it produces the same grid points (to numerical precision) on two scalings of a signal (assuming `n_fmt` is set appropriately).  Since the grid generation depends only on the signal's duration and other sampling parameters (but not the signal itself), this seems like a decent sanity check to me.

Assuming the sample grid construction is correct, the remainder of `fmt` is straightforward (sample + fft).

I feel comfortable with merging this now.

Thanks @jakevdp and @dpwe for all the feedback!

PR #235, opened 2015-08-24
Add squared window normalization to istft

r9y9:
This work was done when I noticed that the current istft didn't ensure (almost) perfect reconstruction in general.

In this PR, 

- implement Griffin's signal reconstruction algorithm for better istft
- remove magic window scaling in istft
- add tests that ensure istft is able to reconstruct signals with various n_fft, hop_length and window functions (in scipy.signal)
- remove trailing whitespaces (my editor automatically does)

The algorithm is based on the following paper:
D. W. Griffin and J. S. Lim, "Signal estimation from modified short-time Fourier transform,"
IEEE Trans. ASSP, vol.32, no.2, pp.236–243, Apr. 1984.
http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.306.7858

NOTE: 
~~For now, some of tests in `test_istft` are failing (at least in my local environment) with my changes. To fix inconsistency between librosa.core.istft and istft in matlab, I looked into `tests/makeTestData.m` to figure out how ground truth data is generated, but I couldn't find istft implementation (`istft` function is provided by external Toolbox?) and still not sure how. I guess istft tests or ground truth data should be updated. Any suggestions are really appreciated.~~

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -188,9 +188,19 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window=None,
 def istft(stft_matrix, hop_length=None, win_length=None, window=None,
           center=True, dtype=np.float32):
     """
-    Inverse short-time Fourier transform.
+    Inverse short-time Fourier transform (ISTFT).
 
-    Converts a complex-valued spectrogram `stft_matrix` to time-series `y`.
+    Converts a complex-valued spectrogram `stft_matrix` to time-series `y`
+    by minimizing the mean squared error between `stft_matrix` and STFT of
+    `y` as described in [1]_.
+
+    In general, window function, hop length and other parameters should be same
+    as in stft, which mostly leads to perfect reconstruction of a signal from
+    unmodified `stft_matrix`.
+
+    .. [1] D. W. Griffin and J. S. Lim,
+        "Signal estimation from modified short-time Fourier transform,"
+        IEEE Trans. ASSP, vol.32, no.2, pp.236–243, Apr. 1984.
 
     Parameters
     ----------
@@ -203,12 +213,13 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
 
     win_length  : int <= n_fft = 2 * (stft_matrix.shape[0] - 1)
         When reconstructing the time series, each frame is windowed
+        and each sample is normalized by the sum of squared window
         according to the `window` function (see below).
 
         If unspecified, defaults to `n_fft`.
 
     window      : None, function, np.ndarray [shape=(n_fft,)]
-        - None (default): use an asymmetric Hann window * 2/3
+        - None (default): use an asymmetric Hann window
         - a window function, such as `scipy.signal.hanning`
         - a user-specified window vector of length `n_fft`
 
@@ -288,7 +299,7 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
         y[sample:(sample + n_fft)] = y[sample:(sample + n_fft)] + ytmp
         ifft_window_sum[sample:(sample + n_fft)] += ifft_window_square
 
-    # Normalize by window
+    # Normalize by sum of squared window
     approx_nonzero_indices = ifft_window_sum > util.SMALL_FLOAT
     y[approx_nonzero_indices] /= ifft_window_sum[approx_nonzero_indices]
 
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -93,7 +93,7 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window=None,
               3.189e-04 -0.000e+00j,  -5.961e-06 -0.000e+00j],
            [  2.441e-03 +2.884e-19j,   5.145e-02 -5.076e-03j, ...,
              -3.885e-04 -7.253e-05j,   7.334e-05 +3.868e-04j],
-          ..., 
+          ...,
            [ -7.120e-06 -1.029e-19j,  -1.951e-09 -3.568e-06j, ...,
              -4.912e-07 -1.487e-07j,   4.438e-06 -1.448e-05j],
            [  7.136e-06 -0.000e+00j,   3.561e-06 -0.000e+00j, ...,
@@ -255,8 +255,7 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
 
     if window is None:
         # Default is an asymmetric Hann window.
-        # 2/3 scaling is to make stft(istft(.)) identity for 25% hop
-        ifft_window = scipy.signal.hann(win_length, sym=False) * (2.0 / 3)
+        ifft_window = scipy.signal.hann(win_length, sym=False)
 
     elif six.callable(window):
         # User supplied a windowing function
@@ -275,7 +274,10 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
     ifft_window = util.pad_center(ifft_window, n_fft)
 
     n_frames = stft_matrix.shape[1]
-    y = np.zeros(n_fft + hop_length * (n_frames - 1), dtype=dtype)
+    expected_signal_len = n_fft + hop_length * (n_frames - 1)
+    y = np.zeros(expected_signal_len, dtype=dtype)
+    ifft_window_sum = np.zeros(expected_signal_len, dtype=dtype)
+    ifft_window_square = ifft_window * ifft_window
 
     for i in range(n_frames):
         sample = i * hop_length
@@ -283,7 +285,12 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
         spec = np.concatenate((spec.conj(), spec[-2:0:-1]), 0)
         ytmp = ifft_window * fft.ifft(spec).real
 
-        y[sample:(sample+n_fft)] = y[sample:(sample+n_fft)] + ytmp
+        y[sample:(sample + n_fft)] = y[sample:(sample + n_fft)] + ytmp
+        ifft_window_sum[sample:(sample + n_fft)] += ifft_window_square
+
+    # Normalize by window
+    approx_nonzero_indices = ifft_window_sum > util.SMALL_FLOAT
+    y[approx_nonzero_indices] /= ifft_window_sum[approx_nonzero_indices]
 
     if center:
         y = y[int(n_fft // 2):-int(n_fft // 2)]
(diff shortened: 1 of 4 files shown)


bmcfee:
Thanks for the PR!  Your changes look pretty reasonable to me, so we should focus on getting the tests into shape.

The regression tests for istft are based on @dpwe's implementation [here](http://www.ee.columbia.edu/~dpwe/resources/matlab/pvoc/istft.m).  However, if the new implementation is precise enough to test on recovering synthetic signals (eg white noise or a sine tone), that'd be fine too.

It looks like you have a float-typed array index somewhere that numpy, which is causing severe [log bloat](https://s3.amazonaws.com/archive.travis-ci.org/jobs/76957069/log.txt) on travis.

Otherwise, I'll throw in a couple of line comments for specific points.

librosa/core/spectrum.py
```diff
@@ -283,7 +285,11 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
         spec = np.concatenate((spec.conj(), spec[-2:0:-1]), 0)
         ytmp = ifft_window * fft.ifft(spec).real
 
-        y[sample:(sample+n_fft)] = y[sample:(sample+n_fft)] + ytmp
+        y[sample:(sample + n_fft)] = y[sample:(sample + n_fft)] + ytmp
+        ifft_window_sum[sample:(sample + n_fft)] += ifft_window_square
+
+    # Normalize by window
+    y[ifft_window_sum != 0] /= ifft_window_sum[ifft_window_sum != 0]
```
bmcfee: A couple of points here:

1) The index set ``ifft_window_sum != 0`` should only be computed once
2) This normalization isn't numerically stable.  Instead, it should perhaps be replaced by ``ifft_window_sum < util.SMALL_FLOAT``.  (If ``ifft_window_sum`` can go negative, which it shouldn't, replace it by ``np.abs(ifft_window_sum)``.)

tests/test_core.py
```diff
@@ -290,6 +290,45 @@ def __test(infile):
...
+            x, n_fft=n_fft, hop_length=hop_length, window=window)
+        x_reconstructed = librosa.core.istft(
+            S, hop_length=hop_length, window=window)
+
+        L = min(len(x), len(x_reconstructed))
+        x = np.resize(x, L)
+        x_reconstructed = np.resize(x_reconstructed, L)
+
+        # NaN should not happen
+        assert not np.any(np.isnan(x_reconstructed))
```
bmcfee: ``np.all(np.isfinite(..))`` might be a better choice, in case the normalization causes an underflow to ``+-Inf`` that wouldn't be caught by ``NaN``.  ``isfinite`` should catch both.

tests/test_core.py
```diff
@@ -290,6 +290,45 @@ def __test(infile):
...
+
+    test_file = 'data/test1_44100.wav'
+    x, sr = librosa.load(test_file, sr=None, mono=True)
+
+    for window_func in [bartlett, hann, hamming, blackman, blackmanharris]:
+        for n_fft in [512, 1024, 2048, 4096]:
+            win = window_func(n_fft, sym=False)
+            symwin = window_func(n_fft, sym=True)
+            for hop_length_denom in six.moves.range(2, 10):
+                hop_length = n_fft / hop_length_denom
```
bmcfee: ``n_fft // hop_length_denom`` to ensure integer division.  I think this is the cause of the log bloat.

r9y9:
Thanks for the review! I fixed the problems you pointed out and added tests for white noise and sine wave as well as real music signal. Also I removed `test_stft` that caused test failure, since I think new tests are sufficient enough; it ensures istft is able to recover signals fairly precisely with a wide variety of n_fft, hop_length and window functions, which are not tested enough previously.

A few visualization examples can be found on a notebook [here](https://gist.github.com/r9y9/303cc5d95767c6b301e1#file-istft-tests-ipynb). 

If the changes seems to be good for you, I'm going to remove WIP from the title.

r9y9:
Removed debugging print and added `np.random.seed` before white noise generation. Force pushed.

tests/test_core.py
```diff
@@ -267,27 +267,53 @@ def test_magphase():
...
+    for x, atol in [(x1, 1.0e-6), (x2, 1.0e-7), (x3, 1.0e-7)]:
+        for window_func in [bartlett, hann, hamming, blackman, blackmanharris]:
+            for n_fft in [512, 1024, 2048, 4096]:
+                win = window_func(n_fft, sym=False)
+                symwin = window_func(n_fft, sym=True)
+                for hop_length_denom in six.moves.range(2, 10):
+                    hop_length = n_fft // hop_length_denom
+                    yield (__test, x, n_fft, hop_length, win, atol)
+                    yield (__test, x, n_fft, hop_length, symwin, atol)
+
```
bmcfee: For full test coverage, there should also be a test that passes ``window_func`` directly as a function to ``stft`` and ``istft``, rather than the pre-computed windows we have currently.

bmcfee:
> I fixed the problems you pointed out and added tests for white noise and sine wave as well as real music signal. Also I removed test_stft that caused test failure, since I think new tests are sufficient enough; it ensures istft is able to recover signals fairly precisely with a wide variety of n_fft, hop_length and window functions, which are not tested enough previously.
> 

Looks good to me!

> A few visualization examples can be found on a notebook here.
> 
> If the changes seems to be good for you, I'm going to remove WIP from the title.

Sure.  @dpwe or @craffel care to chime in a second (third?) opinion on this?

Otherwise, please update the changelog and authors list to include this contribution.  Barring any objections, I think we'll be merge-able.  Thanks again!

craffel:
The tests look sane to me.  Is this change basically just adding normalization by the window square?  That makes sense to me for analysis/reconstruction tasks where the window is the same on analysis and reconstruction, but isn't that the wrong thing to do when the analysis and reconstruction windows are different?  Not that that's a terribly common case or anything

dpwe:
You might want to test it on some actual analyze-modify-synthesize
examples to see if it's still doing what you want.  The reason not to
divide out a non-flat overall window response, but instead to try to
construct a set of windows that will sum to a constant factor, is to
avoid having to assume that you know precisely the window has been
applied.

I guess for STFT domain process that is only scaling magnitudes this
should be pretty safe.  But things like the phase vocoder that adjust
phases are more challenging.

  DAn.

r9y9:
> The tests look sane to me. Is this change basically just adding normalization by the window square?

Yes.

> isn't that the wrong thing to do when the analysis and reconstruction windows are different? Not that that's a terribly common case or anything

Though I think I need to survey some theory of istft, as far as I remember, istft basically assumes the same window and hop length used in stft, so I guess it might be perhaps wrong. However, that's wrong too without squared window normalization, at least in the sense of perfect reconstruction from unmodified STFT? I think we should add a mention that same hop_length and window function are recommended in stft and istft.

>  The reason not to divide out a non-flat overall window response, but instead to try to construct a set of windows that will sum to a constant factor, is to avoid having to assume that you know precisely the window has been applied.

hm, I'm not completely sure the reason why not to divide by the window response. I'm sorry if it's due to my misunderstanding, even if not to, there is an implicit assumption that the same window is used in both analysis and synthesis? e.g. hann window with quarter hop length in stft assumes 2/3 scaled hann window in istft for perfect reconstruction. I would appreciate if you (or someone else) could elaborate on this.

I tested a couple of magnitude scaling and phase vocoding. The examples can be found at:
http://nbviewer.ipython.org/gist/r9y9/4a604032594e3f9855ee. 

Note that since `IPython.display.Audio` automatically normalizes gain of signals, the effect of magnitude scaling is not able to confirm from audio on the notebook, but it works good as we can see the visualization of the scaled waveform. As for the phase vocoding, it sounds good to me. Any thoughts?

tests/test_core.py
r9y9: FIxed. Thanks for your comment!

craffel:
> I think we should add a mention that same hop_length and window function are recommended in stft and istft.

Yes.

bmcfee:
> I tested a couple of magnitude scaling and phase vocoding. The examples can be found at:
> http://nbviewer.ipython.org/gist/r9y9/4a604032594e3f9855ee.

Quick note on this: it looks like you're exponentiating ``S_phase`` after separating via ``magphase``.  This is unnecessary since ``S_phase = exp(1.j * np.angle(D))``.  I didn't hear any artifacts when listening, but this should probably just be ``S = scale * S_mag * S_phase`` (or just ``S *= scale``).

r9y9:
@bmcfee Ah, my fault. `S *=scale` makes sense, thanks. Updated the notebook.

bmcfee:
LGTM!

To-dos:

- [ ] update docstring wrt matching stft/istft parameters
- [ ] udpate authors.md, changelog

r9y9:
I clarified the PR title, rebased to the master, squashed my two commits into one in order to avoid test failure of the first commit, and clarified the commit message.

r9y9:
I revisited the theory of the istft described in the Griffin's paper. I will add a few comments below.

tl:dr
Squared window normalization ensures minimizing the mean squared error between STFT of a given signal and STFT of the estimated signal in theory, rather than perfect reconstruction.

> That makes sense to me for analysis/reconstruction tasks where the window is the same on analysis and reconstruction, but isn't that the wrong thing to do when the analysis and reconstruction windows are different? 

@craffel  As far as I can tell, squared window normalization makes sense even if different window are used in analysis and synthesis, in the sense of minimizing the mean squared error between STFT of a signal to be analyzed and STFT of the estimated signal, though it doesn't ensure perfect reconstruction in general.

Squared window normalization ensures minimizing the mean squared error of STFT and STFT of estimated signal, however, it doesn't mean perfect reconstruction is always achieved. As mentioned in the Griffin's paper, for nonzero window functions (same in analysis and synthesis) and maximum hop length (=1), it ensures perfect reconstruction as well, but for hop length (>1), it's more restricted( I'm afraid that I'm still not sure what exactly the condition is).

As being said, the tests ensure almost perfect reconstruction for reasonable settings (n_fft, hop_length and type of window functions) that are not supported in the previous istft, and the squared window normalization is mathematically derived explicitly, so I think it's good to have in librosa.

r9y9:
I'm going to update docstrings, author.md and contributions.md if there's  no critical objections.

craffel:
Ok, I'm sold.

bmcfee:
> As being said, the tests ensure almost perfect reconstruction for reasonable settings (n_fft, hop_length and type of window functions) that are not supported in the previous istft, and the squared window normalization is mathematically derived explicitly, so I think it's good to have in librosa.

> I'm going to update docstrings, author.md and contributions.md if there's no critical objections.

:+1:

r9y9:
Updated docstrings, authors.md and changelog.md.

bmcfee:
Thanks @r9y9 !

This looks great. Are you planning anything else here, or shall I merge?

r9y9:
I think it's ready to merge:)

r9y9:
Thank you for merging this!

bmcfee:
Thanks for the contribution!

Issue #242, opened 2015-08-30
unit tests for piptrack


bmcfee:
Somehow, this one slipped through the cracks.  (There was implicit integration testing via cqt, but it should still be tested independently.)

bmcfee:
fixed by #243

PR #243, opened 2015-08-31
tests for piptrack

bmcfee:
#242

### DIFF
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -229,7 +229,10 @@ def piptrack(y=None, sr=22050, S=None, n_fft=4096, hop_length=None,
     if hop_length is None:
         hop_length = int(n_fft // 4)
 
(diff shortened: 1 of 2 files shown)


craffel:
Looks good to me.  

> This is only difficult for the combination of low frequencies and short windows, where the (default) fft bin center may be more than one cent off from the target. (For example, at 22050 Hz, n_fft=1024, the closest bin center to 110hz is off by about 3 cents.) The rest of the cases are easy, but included for completeness.

Right, but it still passes these cases because it's interpolating correctly, right?  You're just saying that the other test cases are more trivial?

bmcfee:
> Right, but it still passes these cases because it's interpolating correctly, right? You're just saying that the other test cases are more trivial?

Yes; the simple cases are "correct" even when the returned values are default fft bin centers.

PR #247, opened 2015-09-03
[WIP] tempogram

bmcfee:
as mentioned in #221 

This PR implements localized onset strength autocorrelation, following a similar strategy to [Grosche et al, 2010](http://resources.mpi-inf.mpg.de/MIR/tempogramtoolbox/2010_GroscheMuellerKurth_TempogramCyclic_ICASSP.pdf) (section 3.3).  The only significant difference between the method reported there and this implementation (with default settings) is that we apply the window function to the onset strength envelope before autocorrelation, and they apply it afterwards.  Their method can be recovered by square-rooting the window function `window=hann(win_length, sym=False)**0.5`.

Additional modifications:

- [x] Introduce a new submodule `feature.rhythm`
  - [x] `feature.rhythm.tempogram`
  - [x] unit tests
- [x] Slight refactoring of `onset` and `feature` to avoid circular imports
- [x] New specshow axis mode `tempo` for drawing tempograms.


bmcfee:
This should be feature-complete now.

One question for the peanut gallery.  The tempogram display mode (```librosa.display.specshow(tempogram, y_axis='tempo')```) is currently linear in BPM with log-spaced ticks.  Would it make more sense to log-warp the image itself and use linear ticks, as is done for log-frequency visualization?

It wouldn't be difficult to do, but it might not look so great given the lower initial resolution of tempogram images.  It would require a little more refactoring of the display module, but nothing too complex.

PR #263, opened 2015-10-10
Adapt numpy 1.10.0  change of `numpy.pad`

r9y9:
`pad_length` must be of integral type (ref [NumPy 1.10.0 Release Notes#](https://github.com/numpy/numpy/blob/master/doc/release/1.10.0-notes.rst#nppad-supports-more-input-types-for-pad_width-and-constant_values)), otherwise it throws `TypeError`.

This PR should fix the test failure that I had with numpy 1.10.0:

```
$ nosetests -s -v tests/test_filters.py
... 
======================================================================
ERROR: test_filters.test_constant_q(11025, array([ 130.81278265]), 24, 24, 0.25, 2, False, 2)
----------------------------------------------------------------------
  File "/home/ryuichi/anaconda/envs/py35/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/ryuichi/Dropbox/python/librosa/tests/test_filters.py", line 186, in __test
    norm=norm)
  File "/home/ryuichi/Dropbox/python/librosa/librosa/filters.py", line 605, in constant_q
    for filt in filters])
  File "/home/ryuichi/Dropbox/python/librosa/librosa/filters.py", line 605, in <listcomp>
    for filt in filters])
  File "/home/ryuichi/Dropbox/python/librosa/librosa/util/utils.py", line 291, in pad_center
    return np.pad(data, lengths, **kwargs)
  File "/home/ryuichi/anaconda/envs/py35/lib/python3.5/site-packages/numpy/lib/arraypad.py", line 1315, in pad
    raise TypeError('`pad_width` must be of integral type.')
TypeError: `pad_width` must be of integral type.
```

### DIFF
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -83,7 +83,7 @@ def dct(n_filters, n_input):
     >>> dct_filters
     array([[ 0.031,  0.031, ...,  0.031,  0.031],
            [ 0.044,  0.044, ..., -0.044, -0.044],
-           ..., 
+           ...,
            [ 0.044,  0.044, ..., -0.044, -0.044],
            [ 0.044,  0.044, ...,  0.044,  0.044]])
 
@@ -143,7 +143,7 @@ def mel(sr, n_fft, n_mels=128, fmin=0.0, fmax=None, htk=False):
     >>> melfb
     array([[ 0.   ,  0.016, ...,  0.   ,  0.   ],
            [ 0.   ,  0.   , ...,  0.   ,  0.   ],
-           ..., 
+           ...,
            [ 0.   ,  0.   , ...,  0.   ,  0.   ],
            [ 0.   ,  0.   , ...,  0.   ,  0.   ]])
 
@@ -153,7 +153,7 @@ def mel(sr, n_fft, n_mels=128, fmin=0.0, fmax=None, htk=False):
     >>> librosa.filters.mel(22050, 2048, fmax=8000)
     array([[ 0.  ,  0.02, ...,  0.  ,  0.  ],
            [ 0.  ,  0.  , ...,  0.  ,  0.  ],
-           ..., 
+           ...,
            [ 0.  ,  0.  , ...,  0.  ,  0.  ],
            [ 0.  ,  0.  , ...,  0.  ,  0.  ]])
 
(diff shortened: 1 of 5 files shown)


bmcfee:
Thanks for this!  I haven't had time to check in on the latest numpy release yet.

It's probably worth a quick audit to see if there are other calls to ``np.pad`` that would be affected.

librosa/filters.py
```diff
@@ -599,7 +599,7 @@ def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, tuning=0.0,
     if pad_fft:
         max_len = int(2.0**(np.ceil(np.log2(max_len))))
     else:
-        max_len = np.ceil(max_len)
+        max_len = int(np.ceil(max_len))
```
bmcfee: I think this cast might be better implemented from within `util.pad_center` instead of in the calling function.

r9y9: I believe this PR is a minimum fix for the test failure, but yes, it might be better implemented in `util.pad_center`, though it might affect functions that use `util.pad_center`.

Originally I thought two ways to fix:

1. Casting `size` (which is the second argument in `util.pad_center`)  to be `int` type regardless of the type of `size` (i.e., `size = int(size)`)
2. Throwing an error if `size` is not a integral type (that current implementation does)

My decision was the second one: throwing `TypeError` when float-valued `size` is supplied, since the first one might causes unexpected results if it doesn't assume the `size` to be only int-valued.

I guess `size` is assumed to be integer type, but I din't realized that what exactly it does. Can I ask you what do you think how to fix this in `util.pad_center`? Thank you for your feedback!

bmcfee: > since the first one might cause unexpected results if it doesn't assume the size to be only int-valued.

I think this is fine, since the docstring for `pad_center` explicitly states that `size` should be an integer type.



> Can I ask you what do you think how to fix this in util.pad_center?

I would do just as you suggest: ``size = int(size)``.  I wouldn't worry too much about being bullet-proof here, since the `util` submodule is primarily for use by other parts of librosa.  As long as those parts do the right thing, and the behavior is well-documented, I'm happy.

r9y9:
OK, I have checked all codes that use `np.pad` and added casts to avoid potential `TypeError` in `np.pad`.

bmcfee:
This looks great to me.  Do you anticipate any other changes, or is it ready to merge?

r9y9:
I think it's ready to merge :)

bmcfee:
Thanks again!

PR #268, opened 2015-10-16
travis-ci setup for python 3.5

bmcfee:


bmcfee:
Added bonus, we support python 3.5 now!

Issue #202, opened 2015-06-08
Conda packaging


bmcfee:
Providing a conda distribution might save some headaches at install time, especially when it comes to codec dependencies.

bmcfee:
If we end up making a binstar channel, it would be helpful to include libsamplerate as well.

carlthome:
:+1:

stefan-balke:
hey @bmcfee,

I started to play around with anaconda and the packaging.
Here is a first rough starter for libsamplerate.
The plan is to use this to offer the binaries for all platforms at binstar.

https://github.com/stefan-balke/conda-libsamplerate

I still had some compilation issues there so I forked the repo to make a single change.
Now it seems to build...

stefan-balke:
BTW: Can we solve the issue with scikits.samplerate?

I filed an issue here such that the pypi version gets updated.
https://github.com/cournape/samplerate

bmcfee:
That issue is already fixed in the fork; it just needs to be merged into the main distribution.

stefan-balke:
Let's see if David is moving, otherwise we could also file a new version on pypi, couldn't we?

stefan-balke:
So Erik accepted my **MAJOR** :) PR
Conda should build it now with this recipe.

@bmcfee should we start a channel on binstar?
I think it needs to be an organization and then you can upload packages to it.

bmcfee:
> @bmcfee should we start a channel on binstar?
> I think it needs to be an organization and then you can upload packages to it.

Yeah, that makes sense.  Maybe we should punt this til after #240, by which i mean, finally do #240?  That sounds like a fine post-ismir-deadline project to me.

stefan-balke:
Yeah, sounds good.

Any opinions on what packages should be be covered by the channel?
* librosa
* libsamplerate might be the most important since it made the most trouble so far

What else?

bmcfee:
* scikits.samplerate
* libsndfile and/or libav/ffmpeg

stefan-balke:
Seen that?
https://github.com/bastibe/libsndfile-binaries

It's also here: https://github.com/erikd/libsndfile

I can try to get the builds running and write a conda recipe.

Any opinion about those recipes? Should they be as basic as possible?

And, where should they live? 
https://github.com/conda/conda-recipes ?

stefan-balke:
> And, where should they live? 

Looked up other libs - they usually have a repo for all the needed recipes.
[...]

PR #277, opened 2015-11-05
Docs for OS X

### LINKED ISSUE #276: Better installation instructions
Since the same issues keep popping up again and again ( #270 #275 #219 #261 ), it seems like we must have some shortcomings in the documentation.

In particular, the following points should be addressed explicitly:

- audioread backends
- scikit-samplerate
- general installation issues on osx / pointers to anaconda install instructions

Does anyone out there, preferably a mac user with direct experience, want to chip in on this?

stefan-balke:
Fixes #276

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,14 +41,37 @@ Alternatively, you can download or clone the repository and use `easy_install` t
 
 ```
 unzip librosa.zip
-easy_install librosa
+pip install -e librosa
 ```
 or
 ```
 git clone https://github.com/bmcfee/librosa.git
-easy_install librosa
+pip install -e librosa
 ```
 
+By calling `pip list` you should see `librosa` now as an installed pacakge:
+```
+librosa (0.x.x, /path/to/librosa)
+```
+
+## Installation on OS X
+
+### libsamplerate
+
+In order to use scipy with libsamplerate, we recommend to use homebrew (http://brew.sh).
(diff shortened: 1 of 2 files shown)


bmcfee:
This looks great so far, thanks!

Can we also include this as a topic in the documentation, rather than (in addition to?) the readme?

Also, do we really recommend using brew?  Most people I've talked to recommend using (ana)conda for managing python environments.  Unless you're talking about using brew only for libsamplerate, in which case, the docs should be more explicit on that point.

stefan-balke:
I personally use brew for my Python setup but this is more a matter of taste.
Anaconda is fine, too.
If many people use this, we might add `conda install` to `pip install`?

bmcfee:
> If many people use this, we might add conda install to pip install?

See #202 :)  So.. not yet.

The problem here is that there aren't conda distributions for libsamplerate or ffmpeg.  (ffmpeg should be a last-ditch backend though; gstreamer is a much better option since it doesn't invoke a shell.)

bmcfee:
Is there anything left to do on this one, or is it merge-ready?

stefan-balke:
I think as long as Anaconda is not compatible with all dependencies - this is good for now.

bmcfee:
Ok.  I'll punt other updates into #202 and merge this one then.

Thanks again!

Issue #165, opened 2015-02-05
inverse CQT


bmcfee:
(cc @ebattenberg @dpwe)

Seems like we ought to be able to do this, if we preserve phase in the cqt.

The main barrier comes from variable_hop, which uses oversampling and aggregation to handle large hop-sizes with small filters.  This works fine in magnitude, but I don't see a clean way to make it work out in the complex domain.

Perhaps we can add a `dtype` parameter to cqt: if the user requests a complex-type output, we preserve phase, and throw an exception on variable-hop aggregation.  Otherwise, we let it through in the current state.

Are there other barriers to icqt that I'm not seeing at the moment?

dpwe:
Typically you need to have a very high frame rate (close to the original
sampling rate, depending on the Q and the top frequency) to preserve all
the information in the highest bin.

What do we provide as the phase for under sampled wide-bandwidth bins?

  DAn.

ebattenberg:
I don't have any experience doing an inverse CQT.  As far as I understand it, the inverse transform is approximate (not exactly invertible).  Some things to consider that would affect the quality of the approximation are:

1. The sparsity threshold for the frequency kernels.
2. The hop-size.  I not sure how this would work with any type of over-sampling aggregation.  I also don't think the pseudo CQT (in the hybrid-cqt) could be inverted.  Other than those cases, for bands that are sampled at a small enough hop-size, does there need to be some sort of normalization to account for the fact that some bands would be oversampled (compared to the typically half overlapping hop-size of the invertible lapped transforms like the MDCTs).   

I'm also picturing trying to invert the CQT of an impulse and it seems like it would difficult to avoid extensive smearing in the lower bands of the reconstruction.

dpwe:
I think it's the high frequencies that are the limitation.  If you have
constant Q, you have bandwidth increasing with frequency.  A semitone CQ
bank might have a bandwidth of 6%, or a Q of 16.  That means if you want
50% overlap of time windows, you can only afford to downsample by 8, i.e.,
for 16 kHz data, you need a frame rate of ~2 kHz (0.5 ms hop).

Of course, that's only for the top bin; you could certainly save space by
downsampling by 2 for every octave, but it's much more complicated to
describe and manage.

It's not enough to be able to analyze and reconstruct.  What we want is
analyze, then *modify*, then reconstruct.  You can fake this with pseudo
CQT (spectrogram blurred into CQ bins) if you re-expand it out (pseudo
[...]

PR #279, opened 2015-11-09
Further CQT enhancements

bmcfee:
The current cqt implementation returns only magnitude.  If we're going to support inverse CQT #165 , phase will be critical for good reconstruction.

This PR is the first step in this direction.

It's currently implemented as API-compatible with the existing CQT; complex-output is enabled by setting `real=False`.  In 0.5, we should remove this parameter and make it always return complex-valued transforms.

As a side issue, this PR fixes `util.sync` to retain the `dtype` of the input data.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -21,7 +21,7 @@ __all__ = ['cqt', 'hybrid_cqt', 'pseudo_cqt']
 @cache
 def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         bins_per_octave=12, tuning=None, resolution=2,
-        aggregate=None, norm=1, sparsity=0.01):
+        aggregate=None, norm=1, sparsity=0.01, real=True):
     '''Compute the constant-Q transform of an audio signal.
 
     This implementation is based on the recursive sub-sampling method
@@ -81,7 +81,8 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     Raises
     ------
     ParameterError
-        If `hop_length` is not an integer multiple of `2**(n_bins / bins_per_octave)`
+        If `hop_length` is not an integer multiple of
+        `2**(n_bins / bins_per_octave)`
 
     See Also
     --------
@@ -129,7 +130,6 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
     # How many octaves are we dealing with?
     n_octaves = int(np.ceil(float(n_bins) / bins_per_octave))
-
 
     if fmin is None:
         # C2 by default
@@ -186,7 +186,8 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                          hop_length,
                                          min_filter_length,
                                          fft_basis,
-                                         aggregate)
+                                         aggregate,
+                                         real)
 
         # Convolve
         cqt_resp.append(my_cqt)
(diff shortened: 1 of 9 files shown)


librosa/core/constantq.py
```diff
@@ -565,4 +573,3 @@ def __num_two_factors(x):
         x //= 2
 
     return num_twos
```
bmcfee: I feel like this could be made much more efficient by counting the trailing zeros in the binary representation.  Maybe not necessary though.

At any rate, there should be an initial cast to int.

bmcfee:
Paging @ebattenberg for feedback.  Not urgent, but I reordered some of the `abs` operations in a not-entirely-equivalent way.  This may effect the accuracy of pseudo and hybrid cqt, though whether it's improved or not, I can't say just yet.

ebattenberg:
Changing the Pseudo CQT to take phase into account changes it into not a pseudo CQT but an undersampled (in time) CQT.

bmcfee:
Fair enough.  Is this really a meaningful distinction, or is there a reference implementation/description of pcqt that we adhere to?

dpwe:
It's a big deal. Imagine the "impulse response" corresponding to a
high-frequency, wide-bandwidth CQT bin. It's a little windowed (complex)
sinusoid whose window might be 1ms or shorter. Now tell me the phase of
this kernel over a 10ms window - there could be several entire instances of
the kernel within the window, each with unrelated phases. So a single phase
value doesn't really mean anything.

If you do a single calculation against the complex Fourier coefficients,
you're most likely taking the inner product against one particular
positioning of the IR, which will invite severe time-aliasing (under
sampling).

Taking the average magnitude, the total energy coming through the filters
defined by the impulse response over any length of window, does make sense.

  DAn.

bmcfee:
> If you do a single calculation against the complex Fourier coefficients,
> you're most likely taking the inner product against one particular
> positioning of the IR, which will invite severe time-aliasing (under
> sampling).

Ok, this makes sense.  Thanks.

Backing up a bit: maybe it's just silly to offer complex mode for pseudo-cqt in the first place?  I suppose that anyone that cares enough about phase (eg for later cqt inversion or what-have-you) would just use the full cqt anyway.  Does that seem reasonable to everyone else?

ebattenberg:
> Backing up a bit: maybe it's just silly to offer complex mode for pseudo-cqt in the first place? I suppose that anyone that cares enough about phase (eg for later cqt inversion or what-have-you) would just use the full cqt anyway. Does that seem reasonable to everyone else?

Yeah.  The Pseudo CQT is an alternative to averaging and downsampling the CQT for higher frequency bands.  It should be more efficient but is not exactly equivalent.  Though I don't know how well inverting the CQT will go if you average and downsample in time to get down to the desired hop size.

ebattenberg:
> Is this really a meaningful distinction, or is there a reference implementation/description of pcqt that we adhere to?

I just googled "pseudo CQT" to get the exact definition, but just got links to librosa. ;)

bmcfee:
> I just googled "pseudo CQT" to get the exact definition, but just got links to librosa

I guess the spirit of my question was more: is "pseudo CQT" a precise term, or do we have some wiggle room in its definition?  At any rate, DAn's comment makes sense, and I'm happy to keep pcqt as magnitude-only.

bmcfee:
I don't think there's anything left to do on this one.  I have a working (but terribly slow) inverse-cqt implemented in a notebook, but I don't think perfecting inverse-cqt should hold up complex-cqt.

Does anyone object to merging?

bmcfee:
In working out #302 , it occurred to me that we should probably rename the `resolution` parameter for cqt functions, because: A) it's not all that descriptive, and B) it conflicts with `resolution` in tuning, which has a much more precise interpretation.  Since this PR has other cqt changes, we may as well consolidate.

Do folks on this thread have any thoughts on a better name for `resolution`?  I'd hesitate to use something like `q_factor` since the resolution parameter is just one contributing factor to `Q` here, and I don't want to be misleading.  Eric H suggested something like `width`.  Maybe `filter_scale`?

(And don't worry, we have a warning shim that will provide backward compatibility now, so the old argument name will continue to work.)

bmcfee:
> I don't think there's anything left to do on this one. I have a working (but terribly slow) inverse-cqt implemented in a notebook, but I don't think perfecting inverse-cqt should hold up complex-cqt.

Fast inverse CQT implementation is [here](https://gist.github.com/3db6ce4321157a2bb479) ; gist won't store audio buffers, so you'll have to download it and run it on your own favorite example.  I think this basically demonstrates that the complex cqt is doing the right thing.

(Caveats abound: icqt sounds lousy for resolution > 1, and frequency over-sampling actually degrades quality due to using filter^H instead of a proper basis inversion.  SNR is also rather low, mostly i suspect due to LPF inherent in this cqt implementation, but to my ear it sounds pretty good.)

bmcfee:
Final to-dos:

- [ ] decide on new name for `resolution`
- [ ] implement deprecation warning for `real=True`

Issue #291, opened 2015-12-15
display.frequency_ticks


bmcfee:
Following the discussion [here](https://groups.google.com/forum/#!topic/librosa/dkJzIo_yydA): we should have logic to specify (and automatically determine) the units on frequency axes: mHz, Hz, KHz, MHz, GHz?

drepetto:
I think those four are sufficient. I guess the trick will be figuring out which range is appropriate. It's common to see labels in the 1000s of mHz range. It's similar to the situation with current measurements in Amps, where a power supply will say 1500 mA. So I'm not sure how to judge when to bump up to the next range. Maybe an optional flag specifying a specific unit?

bmcfee:
> Maybe an optional flag specifying a specific unit?

Sure; I'd probably do something like ``time_ticks()`` and have ``fmt in [None, 'mHz', 'Hz' ...]`` with ``None` corresponding to automatic scaling.

As for range boundaries, how about:
* [0, 10Hz) -> mHz
* [10Hz, 10KHz) -> Hz
* [10KHz, 10MHz) -> KHz
* [10MHz, 10GHz) -> MHz
* [10GHz, infty) -> GHz?

drepetto:
Sure, those look like reasonable ranges. I doubt anyone would use,
say, 7000hHz, but 1500hMz is reasonable. So at the low end I'd
probably end up setting units explicitly most of the time. But those
look like good defaults.

.d

bmcfee:
hHz? :)

One more question: would it be preferable to have all ticks on the same unit scale, or append the unit to each tick?

The former appeals to my sense of consistency.

The latter would be much more useful when working on log-frequency axes (as I usually do).

I'd rather not support both if we don't have to, since the interface could easily get too complicated.

drepetto:
Sorry, mHz!

Yes, one unit per axis makes most sense to me. Maybe switching to
scientific notation where it makes sense, like a num_zeroes
threshold...

.d

stefan-balke:
Just a remark:

KHz
should be

kHz. Here is a list: https://en.wikipedia.org/wiki/Hertz

bmcfee:
> KHz should be kHz.

:+1:

Issue #284, opened 2015-11-30
Fix display tests for matplotlib 1.5


bmcfee:
Apparently version 1.5 moved the tests submodule, which breaks our display tests.  This should be an easy fix.

See also: [this thread](https://github.com/matplotlib/matplotlib/issues/5487)

bmcfee:
It seems like the image comparison test code has been refactored a bit, and we won't need to maintain a copy of it within librosa.

Once mpl 1.5.1 is released, we should re-implement the image comparison tests.

bmcfee:
Fixed by merging #299

PR #295, opened 2016-01-05
Frequency ticks

bmcfee:
Implements #291


review-ninja:
<a href="https://app.review.ninja/bmcfee/librosa/pull/295" target="_blank">ReviewNinja

bmcfee:
This one probably isn't merge-able until we reimplement the display tests in #284 .

In order to preserve the previous default behavior, I added a `freq_fmt` parameter to `specshow` that gets passed through to `frequency_ticks`.  Should a similar parameter be added to override time ticks?

bmcfee:
#284 is resolved now, so this PR is no longer blocked.

The tests currently fail because the frequency ticks now include additional precision (eg, ``11025.00`` instead of ``11025``).

The added precision seems like clutter most of the time, at least with linear frequency axes.  As @drepetto mentioned in the thread [here](https://github.com/bmcfee/librosa/issues/291#issuecomment-165489518), it may only make sense to use scientific notation where appropriate, and keep integer formatting otherwise.  Anyone care to suggest a sensible policy for this?

drepetto:
It seems like you'd only need fractional frequency labels when you have a pretty small frequency range, right? Maybe there's some clever way of using the ratio of display pixels to frequency bands or something to decide. 

And, of course, if you display frequencies < 1.0 then the fractional part should always be turned on.

Or it could just all be controlled by a couple flags (fractional, scientific). I don't know if flag proliferation is a problem or not...

best,
douglas

dpwe:
Doesn't replacing the `{:0.2f}` with `{:.4g}` give us pretty much what we want?  Suppresses trailing zeros and decimal points, switches to scientific notation for numbers that are too big.
```
>>> for case in [0.00000012345, 0.00012345, 0.12345, 12.345, 123.450, 1234.5678, 12345.6, 123456]:
...   print "{:.4g}".format(case)
... 
1.235e-07
0.0001234
0.1235
12.35
123.5
1235
1.235e+04
1.235e+05
```

drepetto:
Dan,

that sorta works, except that it'll make the most common plot frequency plot ranges, like 0-22050Hz, pretty ugly, since it switches to scientific notation at 10k. I think it makes sense to stay with one number format for an axis, no?

best,
douglas

dpwe:
Shouldn't 0..22050 render as 0..22.05 kHz?

  DAn.

bmcfee:
Thanks @dpwe , I had completely forgotten about the ``g`` format code.

This also revealed a bug/thinko in the frequency tickers where the ticks were forced to type ``int`` prior to display.  The latest commit fixes this (so that fractional ticks can occur) and uses `{:.5g}` as the default format.  Switching from 4 to 5 digits retains the expected formatting for commonly used nyquist frequencies (11025, 22050).

I've updated the test fixtures accordingly.



> I think it makes sense to stay with one number format for an axis, no?

I generally agree with that, except for when you have log axes, at which point scientific is the only thing that really makes sense.  Switching to g formatting here seems to do the right thing in the cases I've tried.  @drepetto how about you check out this PR branch, give it a spin, and let me know what you think?

dpwe:
On Sun, Jan 31, 2016 at 9:02 AM, Brian McFee <notifications@github.com>
wrote:

> Thanks @dpwe <https://github.com/dpwe> , I had completely forgotten about
> the g format code.
>
Haha.  I spent half an hour trying to craft regexps that would delete
trailing zeros for strings that included a decimal point before I actually
checked the "format mini language" page.

  DAn.

bmcfee:
Other to-do on this:

- [ ] full coverage on freq ticker test
- [ ] add analogous `time_fmt` option to display

bmcfee:
I'm ready to stick a fork in this one, provided @drepetto is happy.

bmcfee:
Gah! I did not mean to merge that already. Serves me right fit checking pr status on my phone.

drepetto:
Okay, I'll take a look tonight. Thanks.

drepetto:
Yep, works great now for the 1Hz data that started the thread. Thanks so much!

douglas

PR #299, opened 2016-01-29
fixed mpl tests for 1.5.1

bmcfee:
Implements #284


bmcfee:
Note: The updated tests pass on 1.5.1, but not earlier versions of mpl. The differences are primarily in the text alignment (1.5.0) and api (1.4.3).

Qualitatively, there is no difference, and the library itself should work with older versions of mpl.

Issue #272, opened 2015-10-25
add versions to requirements


bmcfee:
From #270 discussion: there really should be minimum version numbers on the requirements in setup.py.

- [x] audioread >= 2.0
- [x] numpy >= 1.8.0
- [x] scipy >= 0.13.0
- [x] sklearn >= 0.14.0
- [x] matplotlib >= 1.4 ?
- [x] joblib >=0.7
- [x] decorator >= 3.0
- [x] six >= 1.3

bmcfee:
Resolved by merging #300

PR #300, opened 2016-02-01
added versions to dependencies in setup.py

bmcfee:
Implements #272


bmcfee:
Note: I tried to do a fairly rigorous job of finding minimal version number for each dependency based on each one's changelog, but this may not be exact.  Maintenance releases are generally not counted here, only API compatibility, so YMMV!

Issue #283, opened 2015-11-30
HPSS enhancement


bmcfee:
[paging @cortexelus]

The method of [Dreidger et al.](http://www.terasoft.com.tw/conf/ismir2014/proceedings/T110_127_Paper.pdf) extends Fitzgerald's HPSS algorithm to support harmonic-percussive-residual separation.  It's a relatively straightforward modification to the algorithm: rather than generating the mask by a [``harm > perc``](https://github.com/bmcfee/librosa/blob/master/librosa/decompose.py#L307), we instead require a margin of separation ``harm > margin * perc``, where ``margin >= 1``.  (``margin == 1`` recovers the existing algorithm.)

A couple of notes:

- I don't think we need the ``epsilon`` parameter from section 2 of the paper if we compare directly rather than the ratio.  That is, we do ``harm > perc * margin`` instead of ``harm / (perc + epsilon) > margin``.
- Should we change the API to return a residual component?  Or keep the same API, and take the convention ``R = D - (H + P)``?
- It might be useful to have different margin parameters for the harmonic and percussive mask calculations.  We can support this by a tuple-or-scalar interface similar to the ``kernel_size`` parameter.

bmcfee:
> Should we change the API to return a residual component? Or keep the same API, and take the convention R = D - (H + P)?

On this point, I think we should preserve the existing API. That is, compute H and P using margins, and if the user wants the residual, they have to construct it manually.  This should be documented with an example though.

bmcfee:
... and one more to be extra clear, I don't think we need to implement the iterative procedure here.  It's simple enough to do as an add-on, but the core method should just stick to the one-shot separation with tunable margins.

PR #304, opened 2016-02-13
Harmonic-Percussive-Residual Source Separation 

Cortexelus:
As discussed in #283 

### Updates
- `decompose.hpss` takes an optional margin argument (float or tuple), returns (H, P) according to [ Dreidger et al.](http://www.terasoft.com.tw/conf/ismir2014/proceedings/T110_127_Paper.pdf) Includes additional examples in comments.
- `effects.percussive` takes an optional margin argument (float)
- `effects.harmonic` takes an optional margin argument (float)
- `effects.residual(y, margin=(2.0, 3.0))` returns the residual component between the margins
- `examples/hprss.py` performs a demo of h-p-r source separation

---
This change is [Reviewable](https://reviewable.io/reviews/bmcfee/librosa/304)

### DIFF
--- a/librosa/decompose.py
+++ b/librosa/decompose.py
@@ -373,4 +373,6 @@ def hpss(S, kernel_size=31, power=2.0, mask=False, margin=1.0):
         except DivideByZero:
             raise ParameterError("Power is too large. Try using power between 1 and 10.")
 
+
+
     return ((S * mask_harm) * phase, (S * mask_perc) * phase)
--- a/librosa/decompose.py
+++ b/librosa/decompose.py
@@ -181,20 +181,27 @@ def decompose(S, n_components=None, transformer=None, sort=False, fit=True, **kw
 
 
 @cache
-def hpss(S, kernel_size=31, power=2.0, mask=False):
+def hpss(S, kernel_size=31, power=2.0, mask=False, margin=1.0):
     """Median-filtering harmonic percussive source separation (HPSS).
 
-    Decomposes an input spectrogram `S = H + P`
+    If margin = 1.0, decomposes an input spectrogram `S = H + P`
     where `H` contains the harmonic components,
     and `P` contains the percussive components.
 
-    This implementation is based upon the algorithm described by [1]_.
+    If margin > 1.0, decomposes an input spectrogram `S = H + P + R` 
+    where `R` contains residual components not included in `H` or `P`. 
+
+    This implementation is based upon the algorithm described by [1]_ and [2]_.
 
     .. [1] Fitzgerald, Derry.
         "Harmonic/percussive separation using median filtering."
         13th International Conference on Digital Audio Effects (DAFX10),
         Graz, Austria, 2010.
 
+    .. [2] Driedger, Müller, Disch.
+        "Extending harmonic-percussive separation of audio."
+        15th International Society for Music Information Retrieval Conference (ISMIR 2014),
+        Taipei, Taiwan, 2014.
 
     Parameters
     ----------
@@ -205,10 +212,9 @@ def hpss(S, kernel_size=31, power=2.0, mask=False):
         kernel size(s) for the median filters.
 
         - If scalar, the same size is used for both harmonic and percussive.
-        - If iterable, the first value specifies the width of the
+        - If tuple, the first value specifies the width of the
           harmonic filter, and the second value specifies the width
           of the percussive filter.
-
 
     power : float >= 0 [scalar]
         Exponent for the Wiener filter when constructing mask matrices.
@@ -221,6 +227,13 @@ def hpss(S, kernel_size=31, power=2.0, mask=False):
     mask : bool
         Return the masking matrices instead of components
 
+    margin : float or tuple (margin_harmonic, margin_percussive)
+        margin size(s) for the masks (as described in [2]_)
+
+        - If scalar, the same size is used for both harmonic and percussive.
+        - If tuple, the first value specifies the margin of the
+          harmonic mask, and the second value specifies the margin
+          of the percussive mask.    
 
     Returns
     -------
@@ -282,6 +295,20 @@ def hpss(S, kernel_size=31, power=2.0, mask=False):
            [ 1.,  1., ...,  1.,  1.],
            [ 1.,  1., ...,  1.,  1.]])
 
+
+    Separate into harmonic/percussive/residual components by using a margin > 1.0
+
+    >>> H, P = librosa.decompose.hpss(D, margin=3.0)
+    >>> R = D - (H+P)
+    >>> y_harm = librosa.core.istft(H)
+    >>> y_perc = librosa.core.istft(P)
+    >>> y_resi = librosa.core.istft(R)
+
+
+    Get a more isolated percussive component by widening its margin 
+
+    >>> H, P = librosa.decompose.hpss(D, margin=(1.0,5.0))
+
     """
 
     if np.iscomplexobj(S):
@@ -296,6 +323,17 @@ def hpss(S, kernel_size=31, power=2.0, mask=False):
         win_harm = kernel_size[0]
         win_perc = kernel_size[1]
 
+    if np.isscalar(margin):
+        margin_harm = margin
+        margin_perc = margin
+    else:
+        margin_harm = margin[0]
+        margin_perc = margin[1]
+
+    # margin minimum is 1.0
+    if margin_harm < 1 or margin_perc < 1: 
+        raise ParameterError("Margins must be >= 1.0")
+
     # Compute median filters. Pre-allocation here preserves memory layout.
     harm = np.empty_like(S)
     harm[:] = median_filter(S, size=(1, win_harm), mode='reflect')
@@ -304,11 +342,13 @@ def hpss(S, kernel_size=31, power=2.0, mask=False):
     perc[:] = median_filter(S, size=(win_perc, 1), mode='reflect')
 
     if mask or power < util.SMALL_FLOAT:
-        mask_harm = (harm > perc).astype(float)
-        mask_perc = 1 - mask_harm
+        mask_harm = (harm >  perc * margin_harm).astype(float)
+        mask_perc = (perc >= harm * margin_perc).astype(float)
         if mask:
             return mask_harm, mask_perc
+
     else:
+
         perc = perc ** power
         zero_perc = (perc < util.SMALL_FLOAT)
         perc[zero_perc] = 0.0
@@ -317,12 +357,22 @@ def hpss(S, kernel_size=31, power=2.0, mask=False):
         zero_harm = (harm < util.SMALL_FLOAT)
         harm[zero_harm] = 0.0
 
-        # Find points where both are zero, equalize
-        harm[zero_harm & zero_perc] = 0.5
-        perc[zero_harm & zero_perc] = 0.5
+        # For margin==1, the residual component must be zero, so we split zeros evenly.
+        if margin_harm == 1 and margin_perc == 1:
+            harm[zero_harm & zero_perc] = 0.5
+            perc[zero_harm & zero_perc] = 0.5
 
-        # Compute harmonic mask
-        mask_harm = harm / (harm + perc)
-        mask_perc = perc / (harm + perc)
+        try: 
+            # Compute harmonic mask
+            mask_harm = harm / (harm + perc * margin_harm**power)
+            mask_harm[np.isnan(mask_harm)] = 0
+
+            # Compute percussive mask
+            mask_perc = perc / (perc + harm * margin_perc**power)
+            mask_perc[np.isnan(mask_perc)] = 0
+        except DivideByZero:
+            raise ParameterError("Power is too large. Try using power between 1 and 10.")
+
+
 
     return ((S * mask_harm) * phase, (S * mask_perc) * phase)
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -36,7 +36,7 @@ from . import decompose
 from . import util
 from .util.exceptions import ParameterError
 
-__all__ = ['hpss', 'harmonic', 'percussive', 'residual',
+__all__ = ['hpss', 'harmonic', 'percussive',
            'time_stretch', 'pitch_shift',
            'remix']
 
@@ -68,7 +68,6 @@ def hpss(y, **kwargs):
     --------
     harmonic : Extract only the harmonic component
     percussive : Extract only the percussive component
-    residual : Extract only the residual component
     librosa.decompose.hpss : HPSS on spectrograms
 
 
@@ -114,7 +113,6 @@ def harmonic(y, **kwargs):
     --------
     hpss : Separate harmonic and percussive components
     percussive : Extract only the percussive component
-    residual : Extract only the residual component
     librosa.decompose.hpss : HPSS for spectrograms
 
     Examples
@@ -141,11 +139,7 @@ def harmonic(y, **kwargs):
     return y_harm
 
 
-<<<<<<< HEAD
 def percussive(y, **kwargs):
-=======
-def percussive(y, margin=1.0):
->>>>>>> b914064cf904ea1b18473044c32ab5fc29cde599
     '''Extract percussive elements from an audio time-series.
 
     Parameters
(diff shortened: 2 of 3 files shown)


bmcfee:
Thanks for doing this!  It all looks pretty good so far.  A few comments:

- `effects.residual` isn't the most informative name, since `residual` could refer to any source-separation artifact.  I'm also not sure that it makes that much sense to include it as an effect, rather than adding margin parameters to the existing harmonic and percussive effects (as you've already done).  I suggest either renaming this, or removing it and giving an example of how to construct residuals from hpss.
- if we're going to add `margin` to the effects wrappers, maybe it's better to just provide a `**kwargs` so that all parameters (power, window) can be passed through?
- Coverage went down a bit; it looks like there's no test with `margin` as a tuple parameter.  Can you add that case?

examples/hprss.py
```diff
@@ -0,0 +1,86 @@
...
+
+import argparse
+import sys
+import librosa
+
+
+def hprss_demo(input_file, output_harmonic, output_percussive, output_residual):
+    '''HPRSS demo function.
+
+    :parameters:
```
bmcfee: Can we change this to numpydoc style docstrings? !fix

librosa/decompose.py
```diff
@@ -221,6 +227,13 @@ def hpss(S, kernel_size=31, power=2.0, mask=False):
     mask : bool
         Return the masking matrices instead of components
 
+    margin : float or tuple (margin_harmonic, margin_percussive)
+        margin size(s) for the masks (as described in [2]_)
+
+        - If scalar, the same size is used for both harmonic and percussive.
+        - If iterable, the first value specifies the margin of the
```
bmcfee: 'iterable' isn't quite right here, because we use explicit indices (0 and 1) rather than `next`ing.  I'd just leave it as tuple.

librosa/decompose.py
```diff
@@ -322,7 +357,7 @@ def hpss(S, kernel_size=31, power=2.0, mask=False):
         perc[zero_harm & zero_perc] = 0.5
 
         # Compute harmonic mask
-        mask_harm = harm / (harm + perc)
-        mask_perc = perc / (harm + perc)
+        mask_harm = (harm >  perc * margin_harm).astype(float)
```
bmcfee: This looks like it can only produce binary mask matrices, rather than wiener filtering soft-masks.

I think what you need to do here is rework the mask logic so that a third soft-mask (`res`) captures all energy not assigned to `harm` or `perc`.  Then you would get `mask_harm = harm / (harm + perc + res)`

bmcfee:
Note: I changed the milestone on this to 0.4.3.  I hope that's ok by you @Cortexelus!

librosa/decompose.py
bmcfee: @Cortexelus any thoughts on this?

Cortexelus: I need to fix this. 
1. Complete test coverage
2. Eliminate effects.residual, doesn't makes sense.
3. **kwargs is a good idea.
4. Fix soft masks 

I also implemented an alternative noise/sinusoidal separation using phase acceleration thresholds based on [[1]](http://vud.org/papers/ApelACMC2014.pdf). Do you want me to add this to librosa too?

This was part of my larger experiment: a phase vocoder time stretch that treats noise, harmony, and percussion components differently. It seems:
 - *Percussion* needs to be unstretched, phases unchanged.
 - *Harmony/sinusoidal* just needs to respect ∆phase
 - *Noise* The more it stretches , the more it becomes sinusoidal, which produces artifacts. The noise synthesis method in the paper (white noise filtered by smoothed magnitude spectrum) is OK, it avoids the artifacts, but the fidelity of texture is lost. I could try statistical texture methods next. Thoughts?

bmcfee: > I also implemented an alternative noise/sinusoidal separation using phase acceleration thresholds based on [1]. Do you want me to add this to librosa too?

That'd be awesome.  Let's make it a separate PR though.

stefan-balke: Hex @Cortexelus,

you know about the TSM toolbox from Jonathan?
He's following a similar strategy: PV for harmonic signal and Overlap-Add for transients.
I think that's also what the rubberband people are doing.

https://www.audiolabs-erlangen.de/resources/MIR/TSMtoolbox/

bmcfee:
Circling back on this.

I'd like to push out 0.4.3 in the next few weeks -- specifically, a week ahead of echo nest analyze disappearing, which puts us at May 24.

Any chance we can get this one finalized before then?

Cortexelus:
> This looks like it can only produce binary mask matrices, rather than wiener filtering soft-masks.
> I think what you need to do here is rework the mask logic so that a third soft-mask (res) captures all energy not assigned to harm or perc. Then you would get mask_harm = harm / (harm + perc + res)

Dreidger uses a binary mask. I'm unclear on how to extend the logic to soft-masks. 

Let me see if I can work it out:

Given a time-freq bin, suppose median filter `[** power]` gives us `harm=0.5` and `perc=0.5` 
If using binary mask, with `margin=1.1`, bin would be `100% residual`
If using soft-masks, with `margin=1` (no residual), bin would be `50% harm, 50% perc` 
If using soft-masks, with `margin=1.1` bin would be.. what? 

* Margin is always >= 1.0
* The larger the `margin_harm`:
     - Less likely the bin is `harm`, more likely `res` (binary mask)
     - Less `harm` it is, the more its `harm` value moves over to its `res` value 
* The masks should sum to 1.0

Let's try 
* normalize `harm` to its proportion of `harm + perc`
* divide `harm` by its margin.
* add what we take away to `res` 

If we divide `harm=0.5` and `perc=0.5` by `margin=1.1`, we get . . . `45% harm, 45% perc, 10% res`

If we divide `harm=0.5` and `perc=0.5` by `margin=2`, we get . . . `25% harm, 25% perc, 50% res`

OK seems to make sense to me so far. So a general formula might be:

```
mask_harm = (harm / (harm + perc)) / margin_harm
mask_perc = (perc / (harm + perc)) / margin_perc
mask_res = 1 - mask_harm - mask_perc
```
Let's test it.

If we divide `harm=2` and `perc=8` by `margin=(2,1)`, 
mask_harm = 2/10 / 2 = 0.1
mask_perc = 8/10 / 1 = 0.8
mask_res = 0.1

If we divide `harm=2` and `perc=8` by `margin=(1,2)`, we get . . . 
mask_harm = 2/10 / 1 = 0.2
mask_perc = 8/10 / 2 = 0.4
mask_res = 0.4

If we divide `harm=5` and `perc=5` by `margin=(1,10)`, we get . . . 
mask_harm = 5/10 / 1 = 0.5
mask_perc = 5/10 / 10 = 0.05
mask_res = 0.45

If we divide `harm=1` and `perc=9` by `margin=(5,5)`, we get . . . 
mask_harm = 1/10 / 5 = 0.02
mask_perc = 9/10 / 5 = 0.18
mask_res = 0.8

If we divide `harm=1` and `perc=9` by `margin=(2,2)`, we get . . . 
mask_harm = 1/10 / 2 = 0.05
mask_perc = 9/10 / 2 = 0.45
mask_res = 0.5

Seems to hold up to how I would expect it to behave. OK. Time to test the audio. .

bmcfee:
> Dreidger uses a binary mask. I'm unclear on how to extend the logic to soft-masks.

Yeah, you're right; the paper isn't entirely clear about how to do this with softmasks.

I'm not sure I follow your derivation though.  What if we break it into smaller pieces?

First do harmonic vs non-harmonic:  since the comparison point is ``perc * margin_harm``, you would get
```
mask_harm = harm**p / (harm**p + (perc * margin_harm)**p)
```
then do percussive vs non-percussive:
```
mask_perc = perc**p / (perc**p + (harm * margin_perc)**p)
```
Whatever's left over must be residual, right?  This reduces to the residual-free case when the margins are 1, and to the binary residual case when `p=0`.

After doing some algebra (which I would type out, but it's messy and not having tex in github is really hurting me right now) you can prove that this produces a valid soft-mask, ie: ``mask_harm + mask_perc <= 1``.

bmcfee:
PS: just tried out the above, and it sounds pretty good!

Cortexelus:
Thanks for uploading a notebook gist. 

I tried audio tests at `margin=10` for binary mask, your soft-mask, my soft-mask. I conclude your formula behaves closer to how I would expect it to behave than mine. Compared to the binary mask, yours produces similar audio separation for similar margin values. 

> This reduces to the residual-free case when the margins are 1

Correct
```python
hpss(S, margin=1, power=2)
...
np.allclose(mask_harm + mask_perc, np.ones_like(mask_harm)) # true 
```

> This reduces . . . to the binary residual case when p=0. 

Not true.

p=0 always sets mask_harm to 0.5
```python
mask_harm = harm**p / (harm**p + (perc * margin_harm)**p)
mask_harm = 1 / (1 + 1) = 0.5
```

Indeeed, removing ` or power < util.SMALL_FLOAT:` from [304 HRPSS Notebook](https://gist.github.com/bmcfee/862b3a592f4459af913b01508875c319) and testing:

```python
# HPSS with residual, power = 0.0 
Hrp0, Prp0 = hpss(S, margin=10, power=0.0)
print np.allclose(Hrp0,Prp0) # True
```
```python
print mask_harm
[[ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 ..., 
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]]
[[ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 ..., 
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]
 [ 0.5  0.5  0.5 ...,  0.5  0.5  0.5]]
```

I think you meant when `p=very large`. Audible tests at `p=10` sound very similar to the binary masks.

`p=30` and above I get divide by zero errors at 
```python
mask_harm = harm / (harm + perc * margin_harm**power)
```

bmcfee:
> Not true.
> 
> p=0 always sets mask_harm to 0.5

Er, in the limiting case it should reduce to producing binary masks.  The reason for that range check is to prevent numerical underflow... but you raise a valid point that there should maybe be a similar check for power too large (numerical overflow).

Cortexelus:
It seems:
Larger p, harder the filter.  p=large approaches the binary mask.
Lower p, softer the filter.  p=0 does no separation.

bmcfee:
Derp, yeah, you're totally right.

bmcfee:
It seems like we have this one pretty well sorted out at a high level.

Before merging, we (meaning @Cortexelus :grin: ) should do the following:
- [ ] convert example's docstrings to numpydoc format (see line comment)
- [ ] implement a test that checks for magnitude inputs, varying `power`, and `margin>1`, `np.all(H + P <= S)`
- [ ] implement a test like the above but for mask output, check `not np.any(H & P)`
- [ ] add a safety check that `margin` inputs are at least 1.  Throw a `ParameterError` if not.  Add a test to cover this case.
- [ ] Remove commented code blocks, pylint
- [ ] squash noisy commits

stefan-balke:
After chatting with Jonathan, this is how he ended up doing it (Matlab code...):

```
maskHarm = magSpecHarm ./ (magSpecHarm + sepFactor*magSpecPerc + eps);
maskPerc = magSpecPerc ./ (sepFactor*magSpecHarm + magSpecPerc + eps);
maskRem  = ones(size(magSpec)) - maskHarm - maskPerc;
```

bmcfee:
@stefan-balke thanks for verifying.  That's more or less equivalent to the Weiner filter formulation I listed above in the notebook, except that we can avoid the epsilon smoothing by using comparison operators (for masks).

BTW: I updated the [notebook](https://gist.github.com/bmcfee/862b3a592f4459af913b01508875c319) with a couple of minor corrections and visualization of the example.

bmcfee:
BTW, I'm in the process of migrating the example scripts / notebooks into a separate [librosa_gallery](https://github.com/bmcfee/librosa_gallery) repository.  This way, we can have a self-contained source for examples with a broader variety of example clips.

I mention this because this PR currently includes an example script, but I imagine we'll probably just absorb that demo into the audio effects notebook before pushing.  So if you want to drop the demo script from this PR to simplify things, that's totally fine (mildly encouraged).

Cortexelus:
Since soft mask approaches binary mask as power becomes large, it appears to make sense to revert to a binary mask in the instances of numerical overflow.

```python
# Compute harmonic mask
mask_harm = harm / (harm + perc * margin_harm**power)
# Compute percussive mask
mask_perc = perc / (perc + harm * margin_perc**power)

# Find instances of numeric overflow, use binary mask instead
# because soft mask approaches binary mask as power becomes large
mask_overflow = np.isnan(mask_perc) | np.isnan(mask_harm) | np.isinf(mask_perc) | np.isinf(mask_harm)
mask_harm[mask_overflow] = (harm >  perc * margin_harm).astype(float)
mask_perc[mask_overflow] = (perc >= harm * margin_perc).astype(float)
```

But the `mask_harm[mask_overflow]` bit is throwing errors:

```
  File "/usr/local/Cellar/numpy/1.10.4/libexec/nose/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/cj/Dropbox/_PROJECTS/librosa_c/tests/test_decompose.py", line 80, in __hpss_test
    H, P = librosa.decompose.hpss(D, kernel_size=w, power=p, mask=mask, margin=margin)
  File "/Users/cj/anaconda/lib/python2.7/site-packages/librosa-0.4.2-py2.7.egg/librosa/decompose.py", line 388, in hpss
    mask_harm[mask_overflow] = (harm >  perc * margin_harm).astype(float)
ValueError: boolean index array should have 1 dimension
```

This is a strange numpy error, because the previous boolean index array works just fine:
```
zero_harm = (harm < util.SMALL_FLOAT)
harm[zero_harm] = 0.0
```

1) Does this make sense high-level?
2) What's the right syntax?

Cortexelus:
These boolean index arrays are the same type! 

```
mask_overflow    ndarray     1025x198: 202950 elems, type `bool`, 202950 bytes (198 kb)
zero_harm        ndarray     1025x198: 202950 elems, type `bool`, 202950 bytes (198 kb)
zero_perc        ndarray     1025x198: 202950 elems, type `bool`, 202950 bytes (198 kb)
```

Cortexelus:
Oh I need to do 
```
mask_harm[mask_overflow] = (harm[mask_overflow] > perc[mask_overflow] * margin_harm).astype(float)
mask_perc[mask_overflow] = (perc[mask_overflow] >= harm[mask_overflow] * margin_perc).astype(float)
```

Okay! Now it works, all the tests pass, and it can take large powers.

bmcfee:
> mask_overflow = np.isnan(mask_perc) | np.isnan(mask_harm) | np.isinf(mask_perc) | np.isinf(mask_harm)

you can simplify this to:

```python
~(np.isfinite(mask_perc) & np.isfinite(mask_harm))
```

not sure I understand what'd going on though?  I'd prefer to fail loudly (ie raise an exception) rather than silently revert to a binary mask.

Cortexelus:
> not sure I understand what'd going on though? I'd prefer to fail loudly (ie raise an exception) rather than silently revert to a binary mask.

A single value if harm ** power could overflow and the whole thing would fail loudly, even if the rest of it were fine. I recall powers of 6 failing some examples. This prevents us from exploring the full range of soft masks all the way into the hard mask.  My thought was: if it's true that in the limiting case large power approaches a binary mask, we could just replace the nonfinite values with the result of the binary mask. 

What do you think?

bmcfee:
I think the power issue in wiener filtering is a distinct enough issue that were should handle it as a separate PR.

bmcfee:
Update: please see #356 and [this gist](https://gist.github.com/c9b3775b3eb65946c39771ef22d8c999) which implements a prototype.

For now, let's finish up this PR first, and then I'll merge in the new softmask op and make the necessary changes.

bmcfee:
Closing since it's redundant with #355

PR #306, opened 2016-02-17
CQT updates

bmcfee:
Updates to CQT re: #165

- [x] added a `scale` parameter to `resample` so that output signals have the same energy after resampling
- [x] set `real=False` in `cqt` by default

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -22,7 +22,7 @@ __all__ = ['cqt', 'hybrid_cqt', 'pseudo_cqt']
 @cache
 def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         bins_per_octave=12, tuning=None, filter_scale=1,
-        aggregate=None, norm=2, sparsity=0.01, real=False,
+        aggregate=None, norm=1, sparsity=0.01, real=False,
         resolution=util.Deprecated()):
     '''Compute the constant-Q transform of an audio signal.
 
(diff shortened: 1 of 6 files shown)


librosa/core/constantq.py
```diff
@@ -338,6 +342,10 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                   'filter_scale', filter_scale,
                                   '0.4.2', '0.5.0')
 
+    # If filter_scale is None, auto-tune it based on frequency resolution
+    if filter_scale is None:
+        filter_scale = (2.0**(1./bins_per_octave) - 1.0)/(2.0**(1./12) - 1.0)
```
ejhumphrey: With the same thing happening here and at L481, might be worth a helper function? lots of literals...

librosa/filters.py
```diff
@@ -571,6 +571,10 @@ def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, tuning=0.0,
                                   'filter_scale', filter_scale,
                                   '0.4.2', '0.5.0')
 
+    # If filter_scale is None, auto-tune it based on frequency resolution
+    if filter_scale is None:
+        filter_scale = (2.0**(1./bins_per_octave) - 1.0)/(2.0**(1./12) - 1.0)
```
ejhumphrey: image

ejhumphrey:
looks good to this guy ... the CI fails seem tiny (`test_display.test_cqt_hz`, `test_display.test_tonnetz`)

bmcfee:
> the CI fails seem tiny

yeah, just image comparison tests.  I'll fix those after the rest of the dust settles.

librosa/filters.py
bmcfee: Yeah. This might actually just take a small refactor.  Ideally, it should be isolated to cqt_filters, but the value is actually needed in a few places.

bmcfee:
killing this PR, as it's basically obsolete/redundant with #347 #351

PR #308, opened 2016-02-17
fix resampling scale for CQT

### LINKED ISSUE #307: CQT and resampling normalization: bug or feature?
Tagging @ejhumphrey @ebattenberg @dpwe 

I've mentioned this in a few threads ( #165 #302 ), but wanted to get a focused discussion on this point ahead of 0.4.2.

The CQT method assumes that downsampling the signal and keeping the filter-bank fixed at one octave is a good approximation of applying the full filter-bank to the original signal.  This assumption only holds if the energy of the downsampled signal `y_hat` is scaled to match that of the original `y` after resampling, otherwise, you get exponentially worse approximations going down the octaves. The resamplers used in librosa (either `scipy.signal.resample` or `scikits.samplerate` sinc-interpolation) do not re-scale the output signal, and hence lead to banding artifacts in the resulting spectrograms.

My question for you lot: should we consider this a bug, and fix it in the next minor revision?  (It's at least a thinko.)  Or leave it as is to preserve backwards compatibility?  (FWIW: I think you should be able to simulate the scaling artifacts easily in post-processing, if we really feel the need to support old code.)

I have a PR out #306 that I was planning for 0.5.0 which includes a fix for this, among other changes.  However, if we decide to tag this issue as a bug, I can easily migrate the fix up to here.

bmcfee:
This PR isolates the resampling scale fix from #306 and should resolve #307.

This does change the behavior of CQT, as compared to 0.4.1 and earlier.  To recover the previous behavior, each octave should be divided by `2.0**k` where `k` index the octaves from top (0) to bottom (n_octaves - 1).


bmcfee:
Since we're already breaking (feature value) backwards compatibility here, we may as well fix the default `filter_scale` parameter while we're at it.  The old behavior is still recoverable by calling with `filter_scale=2`, but we'll now prefer better time resolution by default.

librosa/core/audio.py
```diff
@@ -203,7 +203,7 @@ def to_mono(y):
 
 
 @cache
-def resample(y, orig_sr, target_sr, res_type='sinc_best', fix=True, **kwargs):
+def resample(y, orig_sr, target_sr, res_type='sinc_best', fix=True, scale=False, **kwargs):
```
ejhumphrey: quick double check: does `scale=False` make sense here? I expected `True`, but maybe that'd be sane for `rescale`... just a thought

bmcfee: `scale=False` is to preserve backwards compatibility for user-facing functions.  Most resampling applications do not need to preserve energy scale if you interpret them as interpolating the original signal at a different rate.  Scale is only really important for comparing filter response before and after resampling, as we do in CQT.

Issue #312, opened 2016-03-07
Sparsify recurrence_matrix and friends


bmcfee:
The functions in the `segment` module use dense (ndarray) representations of recurrence matrices.  This obviously doesn't scale for high sample rates and/or long tracks.

We should transition these toward inherently sparse implementations, with optional dense conversion.

stefan-balke:
What do you have in mind?

With scipy.sparse?

Conversion via .todense()?

bmcfee:
Yeah, exactly.

The underlying implementation would have to change substantially, since it's currently based on sorting the distance matrix.

Since we already have sklearn as a dependency (for agglomerative clustering and decomposition methods), it probably makes sense to just punt down to [NearestNeighbors](http://scikit-learn.org/stable/modules/generated/sklearn.neighbors.NearestNeighbors.html) and repack the output in a sparse matrix format  (or just call `kneighbors_graph`).

The tricky parts here will be making the recurrence<->lag functions operate cleanly with sparse inputs.

bmcfee:
... the other tricky part is preserving diagonal suppression.  I'm actually not sure how to do this without changing the semantics.

The current implementation does the following: for each `i`, find `k` nearest neighbors `j` such that `|i-j| >= w`.

Two ways we could make this work with sklearn:

- find `k+2w-1` neighbors, then prune out any within `w` of `i`, and take the `k` smallest remaining.
- implement a custom distance function that returns `np.inf` when points are too close in time.

Both of these are potentially slow, though I think I prefer the first one.

bmcfee:
Upon further thought, option 2 above won't work because it breaks metric properties (and therefore, efficient search data structures), so we're stuck with option 1.

Using `mode='distance'` in `kneighbors_graph` should make it relatively easy to prune out the extraneous links, since the distance is readily available.

PR #313, opened 2016-03-07
sparse recurrence. implements #312

bmcfee:
This PR replaces the recurrence matrix construction by a call out to sklearn's nearest neighbors object.

It also adds a parameter `sparse` which, if set to `True`, returns a sparse representation of the recurrence matrix.

The `lag_to_recurrence` and `recurrence_to_lag` functions now automatically support sparse input/output as well.

A new utility function `roll_sparse` is introduced, which extends `np.roll` to work on sparse matrices.

### DIFF
--- a/librosa/segment.py
+++ b/librosa/segment.py
@@ -57,7 +57,7 @@ __all__ = ['recurrence_matrix',
 
 @cache
 def recurrence_matrix(data, k=None, width=1, metric='euclidean',
-                      sym=False, dense=True, axis=-1):
+                      sym=False, sparse=False, axis=-1):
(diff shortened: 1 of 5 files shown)


bmcfee:
Should we add support for sparse lag<->recurrence conversion?

I took a stab at this earlier, and doing it properly (ie, space-efficiently) is somewhat tricky. The problem here ultimately comes down to the difference between the slicing semantics of matrices and arrays, and I couldn't work out an elegant solution.

bmcfee:
This one's ready for CR, if anyone feels up for it @stefan-balke @urinieto

Issue #318, opened 2016-04-04
Feature Normalization


stefan-balke:
I would like to add a threshold to the feature normalization.

Explanation taken from here: https://www.audiolabs-erlangen.de/content/resources/MIR/chromatoolbox/2011_MuellerEwert_ChromaToolbox_ISMIR.pdf

> To avoid random energy distributions occurring during pas- sages of very low energy (e. g., passages of silence before the actual start of the recording or during long pauses), we replace a chroma vector x by the uniform vector of norm one in case |x|p falls below a certain threshold...

Anyone against adding a this parameter to the `normalize()` function?

bmcfee:
Sounds good, though I'd like to generalize it a little.

Currently, the logic [here](https://github.com/bmcfee/librosa/blob/master/librosa/util/utils.py#L631) looks like:

```
length[length < SMALL_FLOAT] = 1.0
return S / length
```
that is, frames with very small norm are left unnormalized.

I think two generalizations make sense:

1. User-specified threshold for this decision, instead of our current arbitrary SMALL_FLOAT constant
2. User-specified fill-value.
    - if `None`, then leave the frame as is (current behavior)
    - if `0`, then zero out the frame
    - if anything non-zero, fill with a constant and normalize (chroma toolbox behavior)

How does that sound?

stefan-balke:
> How does that sound?

Good to me. Standard values should be the ones used at the moments so old algorithms don't break...

stefan-balke:
Let's postpone this to 0.5 since it might not be compatible any longer...

bmcfee:
Agreed.  I think we'll want to change the api to reverse the interpretation of axes.

bmcfee:
For posterity: the semantics of `util.normalize` are such that `Xn = normalize(X, axis=a)` results in `scipy.linalg.norm(Xn, axis=a) == 1` (excepting thresholded/small-value cases).  That is, `a` is the axis being summed over in the norm computation, and not the axis where indexing into it means that the resulting slice has unit norm.  This is a bit counter-intuitive from the perspective of numpy ufuncs, but consistent with the behavior of `scipy.linalg.norm`.

The documentation is updated in #471, but this seemed to be a point of confusion previously.

bmcfee:
Implemented as #471

Issue #12, opened 2013-06-26
improved chroma features


bmcfee:
Include some other chroma variants, as described in [Mueller and Ewert, 2011](http://ismir2011.ismir.net/papers/PS2-8.pdf):
- [x] Pitch
- [x] Chroma
- [x] CENS
- [ ] CRP
- [ ] LBC [Ni, Y.; McVicar, M.; Santos-Rodriguez, R.; De Bie, T., 2012](http://arxiv.org/pdf/1107.4969)

stefan-balke:
Is this still of any interest? I have something for the CENS features.

vahndi:
CENS and CRP features would be useful imo

bmcfee:
Sure.  I'm for it as long as we can provide numerical regression tests against the chroma toolbox implementations, as is done for the METlab features.

When I last looked into this, it seemed like the chroma toolbox code wasn't factored in such a way as to make this easy, but it shouldn't be impossible either.

stefan-balke:
Okay, the chroma toolbox delivers "pitch features" (logarithmic freq. axis with 12 bands per octave) using filterbanks.
For numerical regression tests, we could go with CQT features from librosa.
Load them in Matlab and calculate the chroma features.
I think that might be the hardest part to compare.

I'm not too familiar with CQT and the comparison to the filters used in the chroma toolbox.

bmcfee:
> Okay, the chroma toolbox delivers "pitch features" (logarithmic freq. axis with 12 bands per octave) using filterbanks.

Do you mean filters applied to STFT?

> For numerical regression tests, we could go with CQT features from librosa.

I'm guessing it'll be easier to go the other direction: save the intermediate representation (ie spectrogram) from chroma toolbox and make that the input to the python implementation.  That way, you're comparing apples to apples and isolating the filter and post-processing instead of the low-level dsp.  This is how most of the rastamat regression tests are set up

Of course, the best option would be to perfectly simulate the chroma toolbox operations all the way through, that way you don't have to modify any matlab code.  I got this to work out with the metlab code ([see here](https://github.com/bmcfee/librosa/blob/master/tests/test_met_features.py#L29-L41)), though it wasn't exactly easy.

stefan-balke:
> Do you mean filters applied to STFT?
No.
The chroma toolbox takes the time domain.
Loads filter coefficients for all musical pitches.
And convolves it with the time domain signal.

For the chroma toolbox the following pipeline is used :
[...]

PR #320, opened 2016-04-07
[CR] CENS-Features

stefan-balke:
Hi all,

I started to implement CENS features as provided by the Chroma Toolbox.
A first step to provide some Chroma variants as mentioned in #12.

It is the same to an absolute tolerance of 0.1 - so not ready yet. Next, I will look into the the feature normalization #318 to decrease the error between the implementations.

One problematic step will remain. In the Chroma Toolbox they used:
`f_feature_stat = (upfirdn(f_feature',stat_window,1,downsampSmooth))';`
to smooth and resample the feature sequence.

I tried to simulate this by
```
import scikits.samplerate
chroma_cens = scikits.samplerate.resample(chroma_quant.T, 1.0/downsample_smooth, 'sinc_best').T
```
which I certainly should not do this way.

Maybe you guys have some comments on this.

Things to do before code review:
- [x] Add more tests with resampling (typical CENS features: 9/2, 21/5, 41/10)


librosa/feature/spectral.py
```diff
@@ -939,6 +940,93 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+
+    for cur_quant_step_idx, cur_quant_step in enumerate(quant_steps):
+        chroma_quant += (chroma > cur_quant_step) * quant_weights[cur_quant_step_idx]
+
+    # Apply temporal smoothing
+    win = np.hanning(win_len_smooth)
+    win /= np.sum(win)
+
+    for cur_band_idx, cur_band in enumerate(chroma_quant):
+        chroma_quant[cur_band_idx] = np.convolve(cur_band, win, mode='same')
```
stefan-balke: Is there a better way to filter a matrix row-wise? I only know of `scipy.signal.convolve`

librosa/feature/spectral.py
```diff
@@ -939,6 +940,93 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+
+    # Apply resampling
+    if downsample_smooth == 1:
+        chroma_cens = chroma_quant
+    else:
+        import scikits.samplerate
```
stefan-balke: This is bad.

tests/makeCTData.m
```diff
@@ -0,0 +1,35 @@
...
+
+function f_cqt = testCQT(output_path)
+    f_audio = wav_to_audio('', 'data/', 'test1_44100.wav');
+    f_cqt = audio_to_pitch_via_FB(f_audio);
+    save(fullfile(output_path, 'features-CT-cqt.mat'), 'f_cqt');
+end
+
+function testCENS(f_cqt, output_path)
+    parameter.winLenSmooth = 41;
+    parameter.downsampSmooth = 1;
```
stefan-balke: For the moment, I avoid the resampling by setting the downsampling factor to 1.

bmcfee:
> which I certainly should not do this way.
> 
> Maybe you guys have some comments on this.

If you're going to do this by resampling, you should use core.resample() instead of directly calling external libraries.

Is resampling really the right thing to do here?  I'll cop to not knowing exactly what the operation you're doing here is, but it sounds like an interpolation.  Why not use scipy's interpolation routines, and let the user specify the mode (probably cubic spline by default).

librosa/feature/spectral.py
bmcfee: You can use a 2d-convolution with the filter reshaped to `(1, FILTER_SIZE)`.

Also, window functions should be generic / provided by scipy.signal, not numpy.

stefan-balke:
> Is resampling really the right thing to do here? I'll cop to not knowing exactly what the operation you're doing here is, but it sounds like an interpolation. Why not use scipy's interpolation routines, and let the user specify the mode (probably cubic spline by default).

They do this here:
`f_feature_stat = (upfirdn(f_feature',stat_window,1,downsampSmooth))';`

I think the Matlab help makes it clearer:

>  upfirdn  Upsample, apply a specified FIR filter, and downsample a signal.
>     upfirdn(X,H,P,Q) is a cascade of three systems applied to input signal X:
>           (1) Upsampling by P (zero insertion).  P defaults to 1 if not 
>               specified.
>           (2) FIR filtering with the filter specified by the impulse response 
>               given in H.
>           (3) **Downsampling by Q (throwing away samples).**  Q defaults to 1 if not 
>               specified.
>     upfirdn uses an efficient polyphase implementation.

Well, throwing away samples doesn't sound that hard :)

librosa/feature/spectral.py
stefan-balke: > You can use a 2d-convolution with the filter reshaped to (1, FILTER_SIZE).

Tried this one:
`chroma_quant = sp.signal.convolve2d(chroma_quant, np.tile(win, (1, chroma_quant.shape[0])), mode='same')
`

but results in this:

screen shot 2016-04-08 at 10 04 39
(Top is from Chroma Toolbox, bottom from librosa)

bmcfee: Your 2d conv looks strange to me -- why the tiling?

stefan-balke: Thought the kernel has to be repeated for each row to be filtered.
Fixed it.

librosa/util/utils.py
```diff
@@ -627,8 +632,19 @@ def normalize(S, norm=np.inf, axis=0):
...
         raise ParameterError('Unsupported norm: {}'.format(repr(norm)))
 
+    # get the vectors-ids smaller than the threshold
+    small_values_idx = length < threshold
+
     # Avoid div-by-zero
-    length[length < SMALL_FLOAT] = 1.0
+    length[small_values_idx] = 1.0
+
+    if axis == 1:
```
stefan-balke: Is there a way to index an array based on the `axis` parameter?

Otherwise we need an `if` for each axis to get the indexing right.

bmcfee: My usual trick for this goes like the following:

```python
idx = [slice(None)] * ndim
idx[axis] = target_axis_indexing_slice
f(x[idx])
```

You can see this in action [here](https://github.com/bmcfee/librosa/blob/master/librosa/util/utils.py#L343-L351)

bmcfee:
> They do this here:

Ok, so they're actually doing a standard resampling.  In that case, yeah, punting to librosa.resample will do something comparable.

librosa/util/utils.py
stefan-balke: Mind-boggling but seems to work :)

stefan-balke:
> Ok, so they're actually doing a standard resampling. In that case, yeah, punting to librosa.resample will do something comparable.

No strong opinion on this. At the moment, I just skip columns.

How close do we want to get to the Chroma Toolbox?

What is necessary to finish this PR up?

bmcfee:
> No strong opinion on this. At the moment, I just skip columns.
> 
> How close do we want to get to the Chroma Toolbox?
> 

Usually in this situation, I try to do regression tests against pre-computed outputs from the reference material.  For long feature pipelines, I break it up into stages so that the parts can be tested independently -- i think that makes sense here.  For example, if we can slice off the cqt computation from chroma toolbox, and verify that our cqt->cens transformation matches the reference output, then that's good enough.

In terms of "matching the output", that can be up to interpretation.  Is there any configuration that bypasses the resampling operation?  If we can test for precision without resampling, and then just match shape with resampling, i suspect that's probably okay.  I'd like to see the differences before making a call on it though.

> What is necessary to finish this PR up?

Tests, code review, docs.  :smile:

stefan-balke:
BTW: The normalize function only operates on 2D matrices - we should probably make dimension check first. It will produce an output for n-dims, but it's nonsense...

bmcfee:
> BTW: The normalize function only operates on 2D matrices - we should probably make dimension check first. It will produce an output for n-dims, but it's nonsense...

It should work on arbitrary dimensions -- all the norm calculation uses ufuncs with a target axis, and the normalization operation should be a valid broadcast.  If it's failing outside of 2d, something is broken.

librosa/util/utils.py
```diff
@@ -627,8 +632,26 @@ def normalize(S, norm=np.inf, axis=0):
...
+
     # Avoid div-by-zero
-    length[length < SMALL_FLOAT] = 1.0
+    length[small_values_idx] = 1.0
+
+    # init empty slice object
+    slices = [slice(None)] * S.ndim
+
+    # set slice for target-axis
+    target_axis_idx = int(np.logical_not(axis))
```
stefan-balke: If we want to support n-dims, we need to introduce another parameter here to specify the axis where the we want to replace values.

bmcfee: > If we want to support n-dims, we need to introduce another parameter here to specify the axis where the we want to replace values.

I would expect the semantics to be that slices with low norm get set to the threshold on all axes other than `axis`.

bmcfee:
This PR is failing tests because of a broken comparison test in the display module.

I just fixed this upstream, so if you `git rebase master`, it should be fixed.

stefan-balke:
I tracked the error down to the use of Hanning in Matlab.

```
hanning(N) returns the N-point symmetric Hanning window in a column
    vector. ** Note that the first and last zero-weighted window samples
    are not included**.
```

Which would mean that I add a 2 to our smoothing and it works. Shall we follow Matlab?

bmcfee:
I'm not sure I follow your question.  Can you link to the part of the code (in the PR) where this is relevant?

librosa/feature/spectral.py
```diff
@@ -939,6 +942,90 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+    quant_steps = [0.4, 0.2, 0.1, 0.05]
+    quant_weights = [0.25, 0.25, 0.25, 0.25]
+
+    chroma_quant = np.zeros_like(chroma)
+
+    for cur_quant_step_idx, cur_quant_step in enumerate(quant_steps):
+        chroma_quant += (chroma > cur_quant_step) * quant_weights[cur_quant_step_idx]
+
+    # Apply temporal smoothing
+    win = scipy.signal.hann(win_len_smooth)
```
stefan-balke: Here I use a hann window. In Matlab, they leave out the zeros at both sides...

bmcfee: Hrm.  For your application here (convolution), the zeros shouldn't matter.

Are you sure the problem isn't symmetric vs asymmetric hann windows?  I seem to recall matlab's hann matching `scipy.signal.hann(n, sym=False)`  (defualt is `sym=True`).

stefan-balke:
> Hrm. For your application here (convolution), the zeros shouldn't matter.

Right, but as far as I can see this, e.g. hann(41) in Matlab is the same as hann(43) in python.
I'll verify this tomorrow, don't have a Matlab at hand right now...

librosa/feature/spectral.py
stefan-balke: Right, but as far as I can see this, e.g. hann(41) in Matlab is the same as hann(43) in python.
I'll verify this tomorrow, don't have a Matlab at hand right now...

bmcfee: They'll be the same for odd window lengths, but may differ for even.

stefan-balke: Damn, I'm mixing up things between `hann()` and `hanning()`- sorry for the confusion.

Chroma-Toolbox: Uses `hanning(N)` (Matlab leaves out the zeros)
Python equivalent would be: `scipy.signal.hanning(N+2, sym=False)`

I won't change it in the implementation, but doing the unit-testing would mean that I would test the reference Chroma-Toolbox-CENS with 41 tap smoothing window against a 43 tap librosa implementation.

bmcfee: In scipy at least, `hanning` is an alias for `hann`, but not to be confused with `hamming`.  Is matlab different?



> I won't change it in the implementation, but doing the unit-testing would mean that I would test the reference Chroma-Toolbox-CENS with 41 tap smoothing window against a 43 tap librosa implementation.

This seems reasonable.

stefan-balke: >  Is matlab different?

Yep, here is an example:

```
>> hann(5)'

ans =

         0    0.5000    1.0000    0.5000         0

>> hanning(5)'

ans =

    0.2500    0.7500    1.0000    0.7500    0.2500
```

bmcfee:
Hey @stefan-balke -- it looks like your rebase somehow moved my commits into this branch.  This makes it hard for me to review the new bits you introduce in this branch.

Would you mind squashing them out?

stefan-balke:
> Would you mind squashing them out?

Yeah, thanks to @nils-werner, I figured this out ---somehow.

bmcfee:
Awesome. Thanks!

Is this ready for comments/review then?

stefan-balke:
> Is this ready for comments/review then?

No, give me the rest of the week and I'll add the docs. Also, I'm not sure how to test the normalization stuff, maybe someone has an idea to extent the existing unit-tests.

bmcfee:
Ok, no problem.  Just flag this when it's ready and I'll come back to it.

stefan-balke:
> Ok, no problem. Just flag this when it's ready and I'll come back to it.

@bmcfee, added the docs so it's ready for a first review–still missing the unit test for the normalization though.

librosa/feature/spectral.py
```diff
@@ -950,6 +953,136 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
     return chroma
 
 
+def chroma_cens(y=None, sr=22050, C=None, hop_length=512, fmin=None,
+                threshold=0.001, tuning=None, n_chroma=12,
+                n_octaves=7, bins_per_octave=None, cqt_mode='full', window=None,
+                win_len_smooth=41, downsample_smooth=10):
+    r'''Computes the chroma variant "Chroma Energy Normalized" (CENS), following [1].
```
bmcfee: Should be "[1]_" for sphinx to build the link properly.

librosa/feature/spectral.py
```diff
@@ -950,6 +953,136 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+        (Setting to 1 keeps original feature rate.)
+        Default: 10
+
+    Returns
+    -------
+    chroma_cens : np.ndarray [shape=(n_chroma, t)]
+        The output cens-chromagram
+
+    See Also
+    --------
```
bmcfee: chroma_cqt, ...

librosa/feature/spectral.py
```diff
@@ -950,6 +953,136 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+
+    Examples
+    --------
+
```
bmcfee: Add an example that illustrates the difference between chroma_cqt and chroma_cens on a 10sec clip

librosa/feature/spectral.py
```diff
@@ -950,6 +953,136 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+    '''
+
+    cqt_func = {'full': cqt, 'hybrid': hybrid_cqt}
+
+    if bins_per_octave is None:
+        bins_per_octave = n_chroma
+
+    # Build the CQT if we don't have one already
+    if C is None:
```
bmcfee: Is this part necessary?  We never use C again after this.  Why pass this computation along to chroma_cqt?

librosa/feature/spectral.py
```diff
@@ -950,6 +953,136 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+                        fmin=fmin,
+                        bins_per_octave=bins_per_octave,
+                        n_octaves=n_octaves, window=window)
+
+    # L1-Normalization
+    chroma = util.normalize(chroma, norm=1, axis=0,
+                            mode='fill_constant', threshold=threshold)
+
+    # Quantize amplitudes
+    quant_steps = [0.4, 0.2, 0.1, 0.05]
```
bmcfee: magic numbers should be either parametrized (see, eg., spectral_contrast) or made into CONSTANT (all-caps).

librosa/feature/spectral.py
```diff
@@ -950,6 +953,136 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+
+    chroma_quant = np.zeros_like(chroma)
+
+    for cur_quant_step_idx, cur_quant_step in enumerate(quant_steps):
+        chroma_quant += (chroma > cur_quant_step) * quant_weights[cur_quant_step_idx]
+
+    # Apply temporal smoothing
+    # hanning is tapered to 0, thus throwing away those samples
+    # taking a plus 2 to retain the wished smoothing size
+    win = scipy.signal.hanning(win_len_smooth + 2, sym=False)[1:-1]
```
bmcfee: Since you're using this window for convolution, and the first/last entries are 0, is it necessary to  trim them out here?

If it doesn't hurt, I'd say leave them in, since the code is clearer and there shouldn't be a noticeable difference in efficiency.

librosa/feature/spectral.py
```diff
@@ -950,6 +953,136 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+    win /= np.sum(win)
+    win = np.atleast_2d(win)
+
+    chroma_cens = scipy.signal.convolve2d(chroma_quant, win, mode='same')
+
+    # Apply downsampling
```
bmcfee: Is there a reason to decimate here instead of resample properly?

More generally, I know this is part of how chroma toolbox works, but I think we should consider removing it from the core chroma calculation for two reasons:

1. Downsampling applies generally to any feature analysis, and should be generic.
2. In general, features should be computed at the frame-rate specified by the user through `hop_length`.  This way, all feature outputs are on the same time index, so concatenation/aggregation/etc work seamlessly.

librosa/feature/spectral.py
```diff
@@ -950,6 +953,136 @@ def chroma_cqt(y=None, sr=22050, C=None, hop_length=512, fmin=None,
...
+    if downsample_smooth > 1:
+        # chroma_cens = resample(chroma_cens, orig_sr=1, target_sr=1/downsample_smooth)
+        chroma_cens = chroma_cens[:, ::downsample_smooth]
+
+    # L2-Normalization
+    chroma_cens = util.normalize(chroma_cens, norm=2, axis=0,
```
bmcfee: norm is pinned to 2 here ; shouldn't it be `norm` as provided by the user?

librosa/util/utils.py
```diff
@@ -627,8 +636,26 @@ def normalize(S, norm=np.inf, axis=0):
...
+        # set slice for target-axis
+        target_axis_idx = int(np.logical_not(axis))
+        slices[target_axis_idx] = np.squeeze(small_values_idx)
+
+        if mode == 'fill_zeros':
+            # zero-out vectors
+            S[slices] = 0
+        if mode == 'fill_constant':
+            # fill with unit vectors
+            S[slices] = normalize(np.ones([S.shape[axis], 1]), norm=norm, axis=axis)[0]
```
bmcfee: This is a destructive operation.  If you're going to modify S in-place, do so on a copy.

librosa/util/utils.py
```diff
@@ -627,8 +636,26 @@ def normalize(S, norm=np.inf, axis=0):
...
     # Avoid div-by-zero
-    length[length < SMALL_FLOAT] = 1.0
+    length[small_values_idx] = 1.0
+
+    if mode and np.sum(small_values_idx):
+        # init empty slice object
+        slices = [slice(None)] * S.ndim
+
+        # set slice for target-axis
+        target_axis_idx = int(np.logical_not(axis))
```
bmcfee: I don't understand this bit.  Can you explain in comments?

It seems like this will limit the function to only work on 2d arrays, whereas before it worked for arbitrary dimension (even if the comments did not say as much).  I'd like to not lose that functionality if we don't have to.

tests/test_features.py
```diff
@@ -10,14 +15,6 @@
 except KeyError:
     pass
 
-import matplotlib
```
bmcfee: Does the entire test suite still run if you cut this bit out?  I remember having trouble with the image comparison tests due to import-order/test discovery order messing up the mpl configuration.

bmcfee:
> so it's ready for a first review

Great!  I went through and put a bunch of comments in-place, and it looks pretty good so far.  I have a bunch of style comments, and a couple of possible technical gotchas.

librosa/util/utils.py
stefan-balke: Exactly, it's limited to two dimensions... :/
Any idea of how it should behave in higher dimensions?
In 3d, should it set a matrix to zeros? I'm clueless about this...

tests/test_features.py
stefan-balke: No problems here.

librosa/feature/spectral.py
stefan-balke: > Is there a reason to decimate here instead of resample properly?

Matlab's `upfirdn` is doing it this way but the goal was to match chroma toolbox behaviour...

> In general, features should be computed at the frame-rate specified by the user 

Sounds reasonable. Changed it.

librosa/util/utils.py
bmcfee: > Any idea of how it should behave in higher dimensions?

The short answer is that the entire slice that you'd be normalizing should be set.

For instance, if you have a 3d array, and you're normalizing along `axis=1`, then 
``S[:, i, :]`` should be normalized all as one "vector".  If this sounds strange, imagine that you first permuted the 0 and 1 axes, so that you have ``S_permute[i, :, :]``, and then flatten the last two axes so that you just have ``S_permute_flat[i, :]``.  Any normalization you do to ``S`` should be equivalent to the corresponding vector-wise normalization on the flattened version.

In terms of what this means for your code, I still don't fully understand the implementation here.

What I would expect is that for each index ``i`` where the length is below the threshold, you do  something like
```python
index = [slice(None)] * S.ndim

i = too_short_indices

for i in too_short_indices:
    index[axis] = i
    S[tuple(index)] = #filler
```
Maybe this assignment can be vectorized so you don't need the loop?  At any rate, the logical_not seems strange, and will not generalize to high-dimensions.

tests/test_features.py
stefan-balke: > For instance, if you have a 3d array, and you're normalizing along axis=1, then 
> S[:, i, :] should be normalized all as one "vector". If this sounds strange, imagine that you first permuted the 0 and 1 axes, so that you have S_permute[i, :, :], and then flatten the last two axes so that you just have S_permute_flat[i, :]. Any normalization you do to S should be equivalent to the corresponding vector-wise normalization on the flattened version.

Okay, let's work this out. @nils-werner and me bent our heads around this for a couple of hours with the result that librosa is doing something different than you explain - or we are plain dumb :)

Example:

```python
S = np.zeros((3, 4, 5))
S[:, :, :] = 3.0
S[0, 0, 0] = 12.0

B = librosa.util.normalize(S, norm=np.inf, axis=2)
print(B)
```
yields

```
[[[ 1.    0.25  0.25  0.25  0.25]
  [ 1.    1.    1.    1.    1.  ]
  [ 1.    1.    1.    1.    1.  ]
  [ 1.    1.    1.    1.    1.  ]]

 [[ 1.    1.    1.    1.    1.  ]
  [ 1.    1.    1.    1.    1.  ]
  [ 1.    1.    1.    1.    1.  ]
  [ 1.    1.    1.    1.    1.  ]]

 [[ 1.    1.    1.    1.    1.  ]
  [ 1.    1.    1.    1.    1.  ]
  [ 1.    1.    1.    1.    1.  ]
  [ 1.    1.    1.    1.    1.  ]]] 
```

What we would expect based on your explanation is:

```
[[[ 1.    1.    1.    1.    1.  ]
  [ 0.25  1.    1.    1.    1.  ]
  [ 0.25  1.    1.    1.    1.  ]
  [ 0.25  1.    1.    1.    1.  ]]

 [[ 0.25  1.    1.    1.    1.  ]
  [ 0.25  1.    1.    1.    1.  ]
  [ 0.25  1.    1.    1.    1.  ]
  [ 0.25  1.    1.    1.    1.  ]]

 [[ 0.25  1.    1.    1.    1.  ]
  [ 0.25  1.    1.    1.    1.  ]
  [ 0.25  1.    1.    1.    1.  ]
  [ 0.25  1.    1.    1.    1.  ]]]
```

What do you think?

bmcfee: Option 3: librosa is right and I am dumb?  It's totally possible that I forgot how normalize works.

Shouldn't max-normalization along `axis=2` be equivalent to `S / S.max(axis=2, keepdims=True)`?  If so, then the first example is doing it right.

Re-reading the code (in master), everything looks as I'd expect it to according to numpy broadcast rules.

nils-werner: Depends on what you understand as "normalization along `axis=2`". We understood it as

 1. Iterate over selected axis
 2. Get elements in all other dimensions (e.g. column of a matrix or (N-1)D slice of an ND tensor)
 3. Normalize all these values

All values are touched exactly once.

The code below implements that in `normalize_axis0` and compares it to your
solution:

```python
import numpy


def normalize_axis0(S):
    # S is of dimension n

    # Iterate over axis 0
    for i in range(S.shape[0]):

        # S[i, ...] is of dimension n-1
        S[i, ...] /= numpy.max(S[i, ...])

    return S

A = numpy.ones((3, 3)) * 3.0
A[0, 0] = 12.0

print A / numpy.max(A, axis=0, keepdims=True)
print "-----"
print normalize_axis0(A)

print "----------"

A = numpy.ones((3, 3, 3)) * 3.0
A[0, 0, 0] = 12.0

print A / numpy.max(A, axis=0, keepdims=True)
print "-----"
print normalize_axis0(A)
```

In the 2D case those are interchangeable if you disregard the transposition error:

```
[[ 1.    1.    1.  ]
 [ 0.25  1.    1.  ]
 [ 0.25  1.    1.  ]]
-----
[[ 1.    0.25  0.25]
 [ 1.    1.    1.  ]
 [ 1.    1.    1.  ]]
```

In the 3D case however they differ.

Your code only regards a 1D slice (one "french fry") in the tensor, our implementation changes
a 2D slice (a whole "slice of bread").

```
[[[ 1.    1.    1.  ]
  [ 1.    1.    1.  ]
  [ 1.    1.    1.  ]]

 [[ 0.25  1.    1.  ]
  [ 1.    1.    1.  ]
  [ 1.    1.    1.  ]]

 [[ 0.25  1.    1.  ]
  [ 1.    1.    1.  ]
  [ 1.    1.    1.  ]]]
-----
[[[ 1.    0.25  0.25]
  [ 0.25  0.25  0.25]
  [ 0.25  0.25  0.25]]

 [[ 1.    1.    1.  ]
  [ 1.    1.    1.  ]
  [ 1.    1.    1.  ]]

 [[ 1.    1.    1.  ]
  [ 1.    1.    1.  ]
  [ 1.    1.    1.  ]]]
```

bmcfee: Yeah, I think you're right -- I had it backwards.

nils-werner: With that in mind you can also see what the "transposition error" is: You're not iterating along `axis` but, in the 2D case, along the other one. `axis` is then the operation direction inside the iteration loop.

That's also the reason why [the documented behaviour](https://github.com/bmcfee/librosa/blob/master/librosa/util/utils.py#L594) is so unexpected. NumPy usually does it the other way round (`axis=0` are rows). In the 2D case you simply swap an `axis=0` with an `axis=1` and waste no further thought on it.

In the 3D case however it really messes with your understanding of what is happening :-)

Changing this however will be a compatibility breaking change...

stefan-balke: Okay, seems that we are all on the same page now. This axis-business is something I can't do all day long.

I just tried the cens feature tests with the master-branch normalization and it passes so I would suggest we close this one and figure the normalization out in a separate PR?
@nils-werner and I can sit together tomorrow and propose some new normalization which then might break the way the axis along which to normalize is specified to enable higher dimensions.

@bmcfee what do you think?

Alternative: Only allow 2d normalization.

stefan-balke:
So, this one is IMHO ready to merge. Should I squash the commits since the contribution isn't exactly worth 13 commits :)

tests/test_features.py
bmcfee: Yeah, sorry for the confusion.  It's been a long time since I wrote that function, and my head has been in more standard numpy dimensional conventions more recently.



> I just tried the cens feature tests with the master-branch normalization and it passes so I would suggest we close this one and figure the normalization out in a separate PR?

Just to be clear, none of the previous normalization behavior has changed, right?

Otherwise, I think we're in good shape.  There are still some unresolved line comments (docstrings mainly) that I'd like to see resolved before merging.

stefan-balke: > Just to be clear, none of the previous normalization behavior has changed, right?

Right.

> Otherwise, I think we're in good shape. There are still some unresolved line comments (docstrings mainly) that I'd like to see resolved before merging.

Did a double-check. Fixed all your comments. Saw anything else?

librosa/util/utils.py
bmcfee: !resolved

librosa/feature/spectral.py
bmcfee: !completed

librosa/feature/spectral.py
bmcfee: !resolved

librosa/feature/spectral.py
bmcfee: !completed

librosa/feature/spectral.py
bmcfee: !completed

librosa/feature/spectral.py
bmcfee: !fixed

librosa/feature/spectral.py
bmcfee: !resolved

librosa/feature/spectral.py
bmcfee: !fixed

librosa/util/utils.py
bmcfee: !resolved

librosa/util/utils.py
bmcfee: !fixed

tests/test_features.py
bmcfee: !completed

librosa/feature/spectral.py
bmcfee: !resolved

librosa/feature/spectral.py
bmcfee: "See Also" links are still missing here. !fix

librosa/feature/spectral.py
bmcfee: !resolved

bmcfee:
LGTM; the only remaining thing is the missing "See also" references.  Should point to `chroma_cqt` and anything else you feel is relevant to understanding the parameters of the algorithm.

librosa/feature/spectral.py
stefan-balke: !completed

Issue #298, opened 2016-01-26
DTW


stefan-balke:
Hey there,

I recently did a DTW implementation based on Meinard's book with some Cython speed up for the dynamic programming.
As this is not yet reflected in librosa, I wondered if we could do a module for music syncronization.

Best
Stefan

sonovice:
That would be wonderful! I would love to see that.

bmcfee:
(Looping in @craffel and @dpwe)

We've had some previous discussion of DTW in #41, and more or less concluded that: 1) dtw is independent of the other functionality in librosa, 2) there are already many implementations available, so we'd do best to not reinvent the wheel.  (I'd also throw in 3: that adding a cython dependency for one feature may unnecessarily complicate the distribution process, but that's a minor issue.)

I'm happy to reconsider these points, especially if we can provide a robust implementation that has demonstrable advantages.

craffel:
> We've had some previous discussion of DTW in #41, and more or less concluded that: 1) dtw is independent of the other functionality in librosa, 2) there are already many implementations available, so we'd do best to not reinvent the wheel. (I'd also throw in 3: that adding a cython dependency for one feature may unnecessarily complicate the distribution process, but that's a minor issue.)

Agreed.

> I recently did a DTW implementation based on Meinard's book with some Cython speed up for the dynamic programming.

You should use numba, not Cython. It's more readable, and usually a bit faster.  Here's my numba implementation:
https://github.com/craffel/djitw/blob/master/djitw/dtw.py

bmcfee:
> You should use numba, not Cython. It's more readable, and usually a bit faster.

I generally agree with the readability bit, but isn't numba even worse for packaging due to the llvm dependency?

craffel:
> I generally agree with the readability bit, but isn't numba even worse for packaging due to the llvm dependency?

Not with conda, it has always been installable easily and immediately (including llvm dependencies) with `conda install numba`.  Without conda, it used to be (about a year ago) that installing the right version of llvm and llvmlight took a lot of command line bullshittery, but it's much better now.

bmcfee:
> Not with conda,

Sure.  But since librosa isn't distributed via conda (see #202) the dependency chain is more complicated when pip installing.  Not that this is a deal-breaker, but we already get enough complaints about the install process that I'd rather not complicate it any more than necessary.

stefan-balke:
Okay, sure, I get your points and agree.

BUT:
[...]

PR #323, opened 2016-04-18
[CR] DTW

stefan-balke:
Hey all,

I mainly took @craffel dijtw source code and merged it with mine.
As dicussed in #298, we want the following features:
- [x] Arbitrary step sizes
- [x] Additive or multiplicative local weights for the steps
- [x] Subsequence (so it can be used for matching)
- [x] Global path constraints (e.g. Sakoe-Chiba band etc.)
- [x] make numba optional (cf. Brian's comment)
- [x] test backtracking explicitly
- [x] plot D + wp (in the examples)
- ~~Gullying~~

After finishing the features (which are in dijtw), we need more tests and some final notebook with benchmarking the implementation against "vanilla-vanilla" dtw.

---
This change is [Reviewable](https://reviewable.io/reviews/bmcfee/librosa/323)

### DIFF
--- a/.travis_dependencies.sh
+++ b/.travis_dependencies.sh
@@ -11,7 +11,7 @@ conda_create ()
     conda update -q conda
     conda config --add channels pypi
     conda info -a
-    deps='pip numpy scipy pandas requests nose coverage numpydoc matplotlib sphinx scikit-learn seaborn'
+    deps='pip numpy scipy nose coverage matplotlib scikit-learn'
 
     conda create -q -n $ENV_NAME "python=$TRAVIS_PYTHON_VERSION" $deps
     conda update --all
@@ -21,11 +21,9 @@ src="$HOME/env/miniconda$TRAVIS_PYTHON_VERSION"
 if [ ! -d "$src" ]; then
     mkdir -p $HOME/env
     pushd $HOME/env
-    
+
         # Download miniconda packages
         wget http://repo.continuum.io/miniconda/Miniconda-3.16.0-Linux-x86_64.sh -O miniconda.sh;
-        # Install libsamplerate
-        apt-get source libsamplerate
 
         # Install both environments
         bash miniconda.sh -b -p $src
@@ -33,16 +31,14 @@ if [ ! -d "$src" ]; then
         export PATH="$src/bin:$PATH"
         conda_create
 
-        pushd libsamplerate-*
-            ./configure --prefix=$src/envs/$ENV_NAME
-            make && make install
-        popd
-
         source activate $ENV_NAME
 
-        pip install git+https://github.com/bmcfee/samplerate.git
         pip install python-coveralls
-            
+
+        if [ "$ENABLE_NUMBA" = true ]; then
+            conda install numba
+        fi
+
         source deactivate
     popd
 else
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@ python:
     - "3.4"
     - "3.5"
 
+env:
+    - ENABLE_NUMBA=false
+    - ENABLE_NUMBA=true
+
 before_install:
     - bash .travis_dependencies.sh
     - export PATH="$HOME/env/miniconda$TRAVIS_PYTHON_VERSION/bin:$PATH";
--- a/librosa/core/__init__.py
+++ b/librosa/core/__init__.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """
-Core IO and DSP 
+Core IO and DSP
 ===============
 
 Audio processing
@@ -76,6 +76,16 @@ Pitch and tuning
     pitch_tuning
     piptrack
 
+Dynamic Time Warping
+--------------------
+.. autosummary::
+    :toctree: generated/
+
+    dtw
+    calc_accu_cost
+    backtracking
+    band_mask
+
 """
 
 from .time_frequency import *  # pylint: disable=wildcard-import
@@ -83,6 +93,6 @@ from .audio import *  # pylint: disable=wildcard-import
 from .spectrum import *  # pylint: disable=wildcard-import
 from .pitch import *  # pylint: disable=wildcard-import
 from .constantq import *  # pylint: disable=wildcard-import
+from .dtw import *  # pylint: disable=wildcard-import
 
 __all__ = [_ for _ in dir() if not _.startswith('_')]
-
--- a/librosa/core/dtw.py
+++ b/librosa/core/dtw.py
@@ -1,0 +1,341 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""Sequence Alignment with Dynamic Time Warping."""
+
+import numpy as np
+from scipy.spatial.distance import cdist
+from ..util.decorators import optional_jit
+from ..util.exceptions import ParameterError
+
+__all__ = ['dtw', 'fill_off_diagonal']
+
+
+def fill_off_diagonal(x, radius, value=0):
+    """Sets all cells of a matrix to a given ``value``
+    if they lie outside a constraint region.
+    In this case, the constraint region is the
+    Sakoe-Chiba band which runs with a fixed ``radius``
+    along the main diagonal.
+    When ``x.shape[0] != x.shape[1]``, the radius will be
+    expanded so that ``x[-1, -1] = 1`` always.
+
+    ``x`` will be modified in place.
+
+    Parameters
+    ----------
+    x : np.ndarray [shape=(N, M)]
+        Input matrix, will be modified in place.
+    radius : float
+        The band radius (1/2 of the width) will be
+        ``int(radius*min(x.shape))``.
+    value : int
+        ``x[n, m] = value`` when ``(n, m)`` lies outside the band.
+
+    Examples
+    --------
+    >>> x = np.ones((8, 8))
+    >>> global_constraints(x, 0.25)
+    >>> x
+    array([[1, 1, 0, 0, 0, 0, 0, 0],
+           [1, 1, 1, 0, 0, 0, 0, 0],
+           [0, 1, 1, 1, 0, 0, 0, 0],
+           [0, 0, 1, 1, 1, 0, 0, 0],
+           [0, 0, 0, 1, 1, 1, 0, 0],
+           [0, 0, 0, 0, 1, 1, 1, 0],
+           [0, 0, 0, 0, 0, 1, 1, 1],
+           [0, 0, 0, 0, 0, 0, 1, 1]])
+    >>> x = np.ones((8, 12))
+    >>> global_constraints(x, 0.25)
+    >>> x
+    array([[1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0],
+           [1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
+           [0, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
+           [0, 0, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0],
+           [0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 0, 0],
+           [0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 0],
+           [0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1],
+           [0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1]])
+    """
+    nx, ny = x.shape
+
+    # Calculate the radius in indices, rather than proportion
+    radius = int(round(radius*np.min(x.shape)))
+
+    nx, ny = x.shape
+    offset = np.abs((x.shape[0] - x.shape[1]))
+
+    if nx < ny:
+        idx_u = np.triu_indices_from(x, k=radius+offset)
+        idx_l = np.tril_indices_from(x, k=-radius)
+    else:
+        idx_u = np.triu_indices_from(x, k=radius)
+        idx_l = np.tril_indices_from(x, k=-radius-offset)
+
+    # modify input matrix
+    x[idx_u] = value
+    x[idx_l] = value
+
+
+def dtw(X, Y, dist='euclidean', step_sizes_sigma=None,
+        weights_add=None, weights_mul=None, subseq=False, backtrack=True,
+        global_constraints=False, band_rad=0.25):
+    '''Dynamic time warping (DTW).
+
+    This function performs a DTW and path backtracking on two sequences.
+    We follow the nomenclature and algorithmic approach as described in [1].
+
+    .. [1] Meinard Mueller
+           Fundamentals of Music Processing — Audio, Analysis, Algorithms, Applications
+           Springer Verlag, ISBN: 978-3-319-21944-8, 2015.
+
+    Parameters
+    ----------
+    X : np.ndarray [shape=(K, N)]
+        audio feature matrix (e.g., chroma features)
+
+    Y : np.ndarray [shape=(K, M)]
+        audio feature matrix (e.g., chroma features)
+
+    dist : str
+        Identifier for the cost-function as documented
+        in `scipy.spatial.cdist()`
+
+    step_sizes_sigma : np.ndarray [shape=[n, 2]]
+        Specifies allowed step sizes as used by the dtw.
+
+    weights_add : np.ndarray [shape=[n, ]]
+        Additive weights to penalize certain step sizes.
+
+    weights_mul : np.ndarray [shape=[n, ]]
+        Multiplicative weights to penalize certain step sizes.
+
+    subseq : binary
+        Enable subsequence DTW, e.g., for retrieval tasks.
+
+    backtrack : binary
+        Enable backtracking in accumulated cost matrix.
+
+    global_constraints : binary
+        Applies global constraints to the cost matrix ``C`` (Sakoe-Chiba band).
+
+    band_rad : float
+        The Sakoe-Chiba band radius (1/2 of the width) will be
+        ``int(radius*min(C.shape))``.
+
+    Returns
+    -------
+    D : np.ndarray [shape=(N,M)]
+        accumulated cost matrix.
+        D[N,M] is the total alignment cost.
+        When doing subsequence DTW, D[N,:] indicates a matching function.
+
+    wp : np.ndarray [shape=(N,2)]
+        Warping path with index pairs.
+        Each row of the array contains an index pair n,m).
+        Only returned when ``backtrack`` is True.
+
+    Raises
+    ------
+    ParameterError
+        If you are doing diagonal matching and Y is shorter than X
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import matplotlib.pyplot as plt
+    >>> y, sr = librosa.load(librosa.util.example_audio_file(), offset=10, duration=15)
+    >>> X = librosa.feature.chroma_cens(y=y, sr=sr)
+    >>> noise = np.random.rand(X.shape[0], 200)
+    >>> Y = np.concatenate((noise, noise, X, noise), axis=1)
+    >>> D, wp = librosa.dtw(X, Y, subseq=True)
+    >>> f, axarr = plt.subplots(2)
+    >>> axarr[0].imshow(D, cmap=plt.get_cmap('gray_r'),
+    ...                 origin='lower', interpolation='nearest', aspect='auto')
+    >>> axarr[0].set_title('Database Excerpt')
+    >>> axarr[0].plot(wp[:, 1], wp[:, 0], label='Optimal path', marker='o', color='r')
+    >>> axarr[0].set_xlim([0, Y.shape[1]])
+    >>> axarr[0].set_ylim([0, X.shape[1]])
+    >>> axarr[0].legend()
+    >>> axarr[1].plot(D[-1, :] / wp.shape[0])
+    >>> axarr[1].set_xlim([0, Y.shape[1]])
+    >>> axarr[1].set_ylim([0, 2])
+    >>> axarr[1].set_title('Matching Function')
+    '''
+    # Default Parameters
+    if step_sizes_sigma is None:
+        step_sizes_sigma = np.array([[1, 1], [0, 1], [1, 0]])
+    if weights_add is None:
+        weights_add = np.array([0, 0, 0])
+    if weights_mul is None:
+        weights_mul = np.array([1, 1, 1])
+
+    # take care of dimensions
+    X = np.atleast_2d(X)
+    Y = np.atleast_2d(Y)
+
+    # if diagonal matching, Y has to be longer than X
+    # (X simply cannot be contained in Y)
+    if np.array_equal(step_sizes_sigma, np.array([[1, 1]])) and (X.shape[1] > Y.shape[1]):
+        raise ParameterError('For diagonal matching: Y.shape[1] >= X.shape[1]')
+
+    max_0 = step_sizes_sigma[:, 0].max()
+    max_1 = step_sizes_sigma[:, 1].max()
+
+    # calculate pair-wise distances
+    C = cdist(X.T, Y.T, dist)
+
+    if global_constraints:
+        # Apply global constraints to the cost matrix
+        fill_off_diagonal(C, band_rad, value=np.inf)
+
+    # initialize whole matrix with infinity values
+    D = np.ones(C.shape + np.array([max_0, max_1])) * np.inf
+
+    # set starting point to C[0, 0]
+    D[max_0, max_1] = C[0, 0]
+
+    if subseq:
+        D[max_0, max_1:] = C[0, :]
+
+    D_steps = np.empty(D.shape, dtype=np.int)
+
+    # calculate accumulated cost matrix
+    D, D_steps = calc_accu_cost(C, D, D_steps,
+                                step_sizes_sigma,
+                                weights_mul, weights_add,
+                                max_0, max_1)
+
+    # delete infinity rows and columns
+    D = D[max_0:, max_1:]
+    D_steps = D_steps[max_0:, max_1:]
+
+    if backtrack:
+        if subseq:
+            # search for global minimum in last row of D-matrix
+            wp_end_idx = np.argmin(D[-1, :]) + 1
+            wp = backtracking(D_steps[:, :wp_end_idx], step_sizes_sigma)
+        else:
+            # perform warping path backtracking
+            wp = backtracking(D_steps, step_sizes_sigma)
+
+        return D, np.asarray(wp, dtype=int)
+    else:
+        return D
+
+
+@optional_jit(nopython=True)
+def calc_accu_cost(C, D, D_steps, step_sizes_sigma,
+                   weights_mul, weights_add, max_0, max_1):
+    '''Calculate the accumulated cost matrix D.
+
+    Use dynamic programming to calculate the accumulated costs.
+
+    Parameters
+    ----------
+    C : np.ndarray [shape=(N, M)]
+        pre-computed cost matrix
+
+    D : np.ndarray [shape=(N, M)]
+        accumulated cost matrix
+
+    D_steps : np.ndarray [shape=(N, M)]
+        steps which were used for calculating D
+
+    step_sizes_sigma : np.ndarray [shape=[n, 2]]
+        Specifies allowed step sizes as used by the dtw.
+
+    weights_add : np.ndarray [shape=[n, ]]
+        Additive weights to penalize certain step sizes.
+
+    weights_mul : np.ndarray [shape=[n, ]]
+        Multiplicative weights to penalize certain step sizes.
+
+    max_0 : int
+        maximum number of steps in step_sizes_sigma in dim 0.
+
+    max_1 : int
+        maximum number of steps in step_sizes_sigma in dim 1.
+
+    Returns
+    -------
+    D : np.ndarray [shape=(N,M)]
+        accumulated cost matrix.
+        D[N,M] is the total alignment cost.
+        When doing subsequence DTW, D[N,:] indicates a matching function.
+
+    D_steps : np.ndarray [shape=(N,M)]
+        steps which were used for calculating D.
+
+    See Also
+    --------
+    dtw
+    '''
+    for cur_n in range(max_0, D.shape[0]):
+        for cur_m in range(max_1, D.shape[1]):
+            # accumulate costs
+            for cur_step_idx, cur_w_add, cur_w_mul in zip(range(step_sizes_sigma.shape[0]),
+                                                          weights_add, weights_mul):
+                cur_D = D[cur_n-step_sizes_sigma[cur_step_idx, 0],
+                          cur_m-step_sizes_sigma[cur_step_idx, 1]]
+                cur_C = cur_w_mul * C[cur_n-max_0, cur_m-max_1]
+                cur_C += cur_w_add
+                cur_cost = cur_D + cur_C
+
+                # check if cur_cost is smaller than the one stored in D
+                if cur_cost < D[cur_n, cur_m]:
+                    D[cur_n, cur_m] = cur_cost
+
+                    # save step-index
+                    D_steps[cur_n, cur_m] = cur_step_idx
+
+    return D, D_steps
+
+
+@optional_jit(nopython=True)
+def backtracking(D_steps, step_sizes_sigma):
+    '''Backtrack optimal warping path.
+
+    Uses the saved step sizes from the cost accumulation
+    step to backtrack the index pairs for an optimal
+    warping path.
+
+
+    Parameters
+    ----------
+    D_steps : np.ndarray [shape=(N, M)]
+        Saved indices of the used steps used in the calculation of D.
+
+    step_sizes_sigma : np.ndarray [shape=[n, 2]]
+        Specifies allowed step sizes as used by the dtw.
+
+    Returns
+    -------
+    wp : list [shape=(N,)]
+        Warping path with index pairs.
+        Each list entry contains an index pair
+        (n,m) as a tuple
+
+    See Also
+    --------
+    dtw
+    '''
+    wp = []
+    # Set starting point D(N,M) and append it to the path
+    cur_idx = (D_steps.shape[0]-1, D_steps.shape[1]-1)
+    wp.append((cur_idx[0], cur_idx[1]))
+
+    # Loop backwards.
+    # Stop criteria:
+    # Setting it to (0, 0) does not work for the subsequence dtw,
+    # so we only ask to reach the first row of the matrix.
+    while cur_idx[0]:
+        cur_step_idx = D_steps[(cur_idx[0], cur_idx[1])]
+
+        # save tuple with minimal acc. cost in path
+        cur_idx = (cur_idx[0]-step_sizes_sigma[cur_step_idx][0],
+                   cur_idx[1]-step_sizes_sigma[cur_step_idx][1])
+
+        # append to warping path
+        wp.append((cur_idx[0], cur_idx[1]))
+
+    return wp
--- a/librosa/util/decorators.py
+++ b/librosa/util/decorators.py
@@ -55,3 +55,16 @@ def deprecated(version, version_removed):
         return func(*args, **kwargs)
 
     return decorator(__wrapper)
+
+
+'''Define the optional_jit decorator
+   If numba is importable, use numba.jit.
+   Else create a no-op decorator.
+'''
+try:
+    from numba.decorators import jit as optional_jit
+except ImportError:
+    def optional_jit(*_, **__):
+        def __wrapper(func, *args, **kwargs):
+            return func(*args, **kwargs)
+        return decorator(__wrapper)
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ setup(
         'resampy >= 0.1.2'
     ],
     extras_require={
-        'docs': ['numpydoc', 'sphinx!=1.3.1', 'sphinx_rtd_theme']
+        'docs': ['numpydoc', 'sphinx!=1.3.1', 'sphinx_rtd_theme'],
+        'numba': ['numba >= 0.25', ]
     }
 )
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -20,11 +20,9 @@ import glob
 import numpy as np
 import scipy.io
 import six
-import warnings
-
+from nose.tools import eq_, raises
 import matplotlib
 matplotlib.use('Agg')
-from nose.tools import nottest, eq_, raises
 
 
 # -- utilities --#
@@ -56,6 +54,7 @@ def test_load():
     for infile in files('data/core-load-*.mat'):
         yield (__test, infile)
     pass
+
 
 def test_segment_load():
 
@@ -136,7 +135,6 @@ def test_resample_stereo():
         # Check that we're within one sample of the target length
         target_length = y.shape[-1] * sr_out // sr_in
         assert np.abs(y2.shape[-1] - target_length) <= 1
-
 
     y, sr_in = librosa.load('data/test1_44100.wav', mono=False, sr=None, duration=5)
 
@@ -572,7 +570,6 @@ def test_piptrack_errors():
 
 def test_piptrack():
 
-
     def __test(S, freq):
         pitches, mags = librosa.piptrack(S=S, fmin=100)
 
@@ -585,7 +582,6 @@ def test_piptrack():
         # We should be within one cent of the target
         assert np.all(np.abs(np.log2(recovered_pitches) - np.log2(freq)) <= 1e-2)
 
-
     sr = 22050
     duration = 3.0
 
@@ -597,7 +593,6 @@ def test_piptrack():
             S = np.abs(librosa.stft(y, n_fft=n_fft, center=False))
 
             yield __test, S, freq
-
 
 
 def test_estimate_tuning():
@@ -739,7 +734,6 @@ def test_clicks():
         elif click is not None:
             assert len(y) == nmax + len(click)
 
-
     test_times = np.linspace(0, 10.0, num=5)
 
     # Bad cases
@@ -761,7 +755,7 @@ def test_clicks():
 
 
 def test_fmt_scale():
-    # This test constructs a single-cycle cosine wave, applies various axis scalings, 
+    # This test constructs a single-cycle cosine wave, applies various axis scalings,
     # and tests that the FMT is preserved
 
     def __test(scale, n_fmt, over_sample, kind, y_orig, y_res, atol):
@@ -815,8 +809,8 @@ def test_fmt_scale():
 
         for kind in ['slinear', 'quadratic', 'cubic']:
             for n_fmt in [None, 64, 128, 256, 512]:
-                for os in [1, 2, 3]:
-                    yield __test, scale, n_fmt, os, kind, y_orig, y_res, atol[kind]
+                for cur_os in [1, 2, 3]:
+                    yield __test, scale, n_fmt, cur_os, kind, y_orig, y_res, atol[kind]
 
                 # Over-sampling with down-scaling gets dicey at the end-points
                 yield __test, 1./scale, n_fmt, 1, kind, y_res, y_orig, atol[kind]
--- a/tests/test_dtw.py
+++ b/tests/test_dtw.py
@@ -1,0 +1,91 @@
+import librosa
+import numpy as np
+
+
+def test_dtw_global():
+    # Example taken from:
+    # Meinard Mueller, Fundamentals of Music Processing
+    X = np.array([[1, 3, 3, 8, 1]])
+    Y = np.array([[2, 0, 0, 8, 7, 2]])
+
+    gt_D = np.array([[1., 2., 3., 10., 16., 17.],
+                     [2., 4., 5., 8., 12., 13.],
+                     [3., 5., 7., 10., 12., 13.],
+                     [9., 11., 13., 7., 8., 14.],
+                     [10, 10., 11., 14., 13., 9.]])
+
+    mut_D, _ = librosa.dtw(X, Y)
+
+    assert np.array_equal(gt_D, mut_D)
+
+
+def test_dtw_global_diagonal():
+    # query is a linear ramp
+    X = np.linspace(0.1, 1, 10)
+    Y = X
+
+    gt_wp = list(zip(list(range(10)), list(range(10))))[::-1]
+
+    mut_D, mut_wp = librosa.dtw(X, Y, subseq=True, dist='cosine',
+                                step_sizes_sigma=np.array([[1, 1]]),
+                                weights_mul=np.array([1, ]))
+
+    assert np.array_equal(np.asarray(gt_wp), np.asarray(mut_wp))
+
+
+def test_dtw_subseq():
+    # query is a linear ramp
+    X = np.linspace(0, 1, 100)
+
+    # database is query surrounded by noise
+    noise_len = 200
+    noise = np.random.rand(noise_len)
+    Y = np.concatenate((noise, noise, X, noise))
+
+    _, mut_wp = librosa.dtw(X, Y, subseq=True)
+
+    # estimated sequence has to match original sequence
+    # note the +1 due to python indexing
+    mut_X = Y[mut_wp[-1][1]:mut_wp[0][1]+1]
+    assert np.array_equal(X, mut_X)
+
+
+def test_dtw_fill_off_diagonal_8_8():
+    # Case 1: Square matrix (N=M)
+    mut_x = np.ones((8, 8))
+    librosa.fill_off_diagonal(mut_x, 0.25)
+
+    gt_x = np.array([[1, 1, 0, 0, 0, 0, 0, 0],
+                     [1, 1, 1, 0, 0, 0, 0, 0],
+                     [0, 1, 1, 1, 0, 0, 0, 0],
+                     [0, 0, 1, 1, 1, 0, 0, 0],
+                     [0, 0, 0, 1, 1, 1, 0, 0],
+                     [0, 0, 0, 0, 1, 1, 1, 0],
+                     [0, 0, 0, 0, 0, 1, 1, 1],
+                     [0, 0, 0, 0, 0, 0, 1, 1]])
+
+    assert np.array_equal(mut_x, gt_x)
+    assert np.array_equal(mut_x, gt_x.T)
+
+
+def test_dtw_fill_off_diagonal_8_12():
+    # Case 2a: N!=M
+    mut_x = np.ones((8, 12))
+    librosa.fill_off_diagonal(mut_x, 0.25)
+
+    gt_x = np.array([[1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0],
+                     [1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
+                     [0, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
+                     [0, 0, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0],
+                     [0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 0, 0],
+                     [0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 0],
+                     [0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1],
+                     [0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1]])
+
+    assert np.array_equal(mut_x, gt_x)
+
+    # Case 2b: (N!=M).T
+    mut_x = np.ones((8, 12)).T
+    librosa.fill_off_diagonal(mut_x, 0.25)
+
+    assert np.array_equal(mut_x, gt_x.T)


craffel:
Nice.  I can take a look at this in May.

bmcfee:
I know this is still WIP and commenting at this stage is annoying, but can we not make numba a requirement?

[This comment](https://github.com/bmcfee/librosa/issues/298#issuecomment-175069435) provides a workaround for optional jit compilation, so that the same code should work with or without numba.

bmcfee:
This PR is failing tests because of a broken comparison test in the display module.

I just fixed this upstream, so if you git rebase master, it should be fixed.

stefan-balke:
> This comment provides a workaround for optional jit compilation, so that the same code should work with or without numba.

Done.

Should this go into 0.4.3? I think it could go but if we decide that the function interfaces are fine. Putting more features into it later is easy then.

bmcfee:
> Should this go into 0.4.3? I think it could go but if we decide that the function interfaces are fine. Putting more features into it later is easy then.

If you think we can do it quickly (ie next couple of weeks), then sure.  If not, we can continue to punt it.

bmcfee:
> If you think we can do it quickly (ie next couple of weeks), then sure. If not, we can continue to punt it.
>  Hide all checks

@stefan-balke Any thoughts on this?  I'm fine either way, but if you don't have time to finish it soon, let's punt.

stefan-balke:
Since this is planned to be joint work with @craffel I'd prefer to have it ready for the next version after the his phd thesis is handed in.
@craffel any objections?

craffel:
Yeah, still interested in helping, and I will have more bandwidth to help later this month.

bmcfee:
Ok, I've reassigned #298 to 0.5 then.

stefan-balke:
fyi: This is on my next week's list.

stefan-balke:
Alright, I got most of the things done.

### Gullying
@craffel can we talk about the gullying?
Couldn't we also do it with the same idea as subsequence DTW?
(Initialize the first row of the accumulated cost matrix D with the first row of the C matrix instead of init with the cumsum of C's first row?)

In the case of allowing gullying at the beginning 3 frames, we could simply init D's first row with:

`[C[0,0], C[0,1], C[0,2] ....proceed with cumsum(C)... ]`

and for the ending frames in a similar way.

### Display function
@bmcfee with the rewrite of the display module, I'm confused where to put the visualizations.

I need two things:
 * Plot the cost matrix
 * Add the warping path to the matrix

Could look like this:
```
fig = plt.figure()
ax = fig.add_subplot(111)
imax = ax.imshow(C, cmap=plt.get_cmap('gray_r'),
                 origin='lower', interpolation='nearest', aspect='auto')

x_val = [x[1] for x in p_adj]
y_val = [x[0] for x in p_adj]
ax.plot(x_val, y_val, marker='o', color='r')
ax.autoscale(tight=True)
plt.ylabel('$X$')
plt.xlabel('$Y$')
cbar = fig.colorbar(imax)
cbar.ax.set_ylabel('Cost')
ax.set_title('Cost Matrix $C$ with an optimal warping path')
```

Should I do own functions for this or should it derive from anything in the display module?

bmcfee:
> Should I do own functions for this or should it derive from anything in the display module?

I don't think we necessarily need special case display functions for this.  What's wrong with the following (matching your notation)?
```python
>>> librosa.display.specshow(C)
>>> plt.plot(p_adj[:, 0], p_adj[:, 1], label='Optimal path')
```
Or, if you want time axes:
```python
>>> librosa.display.specshow(C, x_axis='time', y_axis='time')
>>> p_time = librosa.frames_to_time(p_adj)
>>> plt.plot(p_time[:, 0], p_time[:, 1], label='Optimal path')
```

stefan-balke:
> I don't think we necessarily need special case display functions for this. What's wrong with the following (matching your notation)?

Nothing, `specshow` just sounded wrong. :)
Will put it into the examples then.

craffel:
> @craffel can we talk about the gullying?
Couldn't we also do it with the same idea as subsequence DTW?

No, I don't think so. I don't think we're on the same page. The gully simply constrains the possible end points. For example, if the distance matrix is of shape (100, 200) and the gully was .9 then the path would have to end any of (90, 199) through (99, 199) or (99, 180) through (99, 199). Does that make sense? I have definitions of it more rigorously in any of my papers which use DTW. So, it's something I usually enforce at traceback time, not in the distance matrix, although I'd expect you could also do it by filling portions of the final row and column of the matrix with infinity. I don't see how it could be accomplished by modifying the first row or column because in my understanding it is independent of the first row and column, it's only a constraint on the final point. BTW, this is something I picked up (and maybe reinterpreted/bastardized) from @dpwe, not sure where he got it from.

craffel:
Also, thanks for all of this! I will spend some time checking it out in the near future.

stefan-balke:
Hey Colin,

>  So, it's something I usually enforce at traceback time

Yep, thought about this and got to the same conclusion :)

However, I wonder if we couldn't accomplish this by restricting the subsequence DTW somehow.

Here a small sketch:

img_2108

So for me, gullying live somewhat in between subsequence and global DTW but I'll think about it a little bit more...

bmcfee:
Could we get some equations (or pointers/screenshots to relevant papers) to describe what "gullying" means?

If @stefan-balke's drawing (c) is accurate, it seems like you could enforce this by, as @craffel says, filling the top row (outside the red box) and right column (outside the green box) by infinity, and the rest fixes itself.

craffel:
Firs things first, what I originally wrote

>  if the distance matrix is of shape (100, 200) and the gully was .9 then the path would have to end any of (90, 199) through (99, 199) or (99, 180) through (99, 199)

is wrong for subsequence DTW, I wrote it in haste.   For subsequence DTW, it should be "the path would have to end any of (90, 199) through (99, 199) or (99, 90) through (99, 199)".

> Here a small sketch:

Your sketch looks reasonable to me.

> Could we get some equations (or pointers/screenshots to relevant papers) to describe what "gullying" means?

As I said above, I have definitions in my papers which use DTW (and they all say basically the same thing, and it's something I picked up from @dpwe, not sure where he got it from.  Here's one from ISMIR 2015 (note this definition is not for subsequence DTW):

gully

For subsequence DTW it's `g*min(M, N) <= p[L] <= N, g*min(M, N) <= q[L] <= M`.  (also in retrospect I should have defined p and q to be natural numbers, not real numbers, but oh well).  @dpwe's code/description is here: http://www.ee.columbia.edu/ln/rosa/matlab/alignmidi/

stefan-balke:
Anyone in for a CR? We can add gullying when @craffel has more cycles..

bmcfee:
I'm happy to CR after it passes on travis.

stefan-balke:
Oops, it was passing in 3.7. I think we need numba in the dependencies...Tried to add it but it seems it's not used in Travis.
```
Obtaining file:///home/travis/build/librosa/librosa
Requirement already satisfied (use --upgrade to upgrade): audioread>=2.0.0 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): numpy>=1.8.0 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): scipy>=0.13.0 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): scikit-learn>=0.14.0 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): matplotlib>=1.5 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): joblib>=0.7.0 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): decorator>=3.0.0 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): six>=1.3 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): resampy>=0.1.2 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): python-dateutil in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from matplotlib>=1.5->librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): pytz in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from matplotlib>=1.5->librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): cycler in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from matplotlib>=1.5->librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): pyparsing!=2.0.4,>=1.5.6 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from matplotlib>=1.5->librosa==0.5.0.dev0)
Requirement already satisfied (use --upgrade to upgrade): Cython>=0.21 in /home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages (from resampy>=0.1.2->librosa==0.5.0.dev0)
Installing collected packages: librosa
  Running setup.py develop for librosa
```

.travis_dependencies.sh
```diff
@@ -11,7 +11,7 @@ conda_create ()
     conda update -q conda
     conda config --add channels pypi
     conda info -a
-    deps='pip numpy scipy pandas requests nose coverage numpydoc matplotlib sphinx scikit-learn seaborn'
+    deps='pip numpy scipy pandas requests nose coverage numpydoc matplotlib sphinx scikit-learn seaborn numba'
```
stefan-balke: Here I add numba.

bmcfee:
> .Tried to add it but it seems it's not used in Travis.

I just wiped the image cache and restarted the build.  Hopefully that should do it.

But yes, numba needs to be installed via conda to get llvm packages; pip alone won't do it.

stefan-balke:
> But yes, numba needs to be installed via conda to get llvm packages; pip alone won't do it.

True fact.

stefan-balke:
```
test_dtw.test_dtw_global ... ok
test_dtw.test_dtw_global_diagonal ... ok
test_dtw.test_dtw_subseq ... ok
```

Here we go....

stefan-balke:
> EDIT: scratch that; the problem is that when numba can be imported, it returns the jit wrapper as a proper decorator. But when numba can't be imported, it returns a bool, which doesn't work as a decorator.

Yup, but I'm doing something similar. I'm importing the decorator. If it's available - use it - otherwise return the function and do the normal Python thingy....

stefan-balke:
Ready for a first CR.

bmcfee:
Reviewed 1 of 1 files at r2, 1 of 4 files at r9.
Review status: 2 of 8 files reviewed at latest revision, 13 unresolved discussions.

---

*[setup.py, line 44 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKifx-WqCcSPuLrAXmu:-KKifx-WqCcSPuLrAXmv:-1791857362) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/setup.py#L44)):*
> ```Python
>     ],
>     extras_require={
>         'docs': ['numpydoc', 'seaborn', 'sphinx_rtd_theme'],
> ```

Looks like a regression here since the display rewrite was merged.  Can you rebase master?

---

*[examples/dtw.ipynb, line 1 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKigXFfY2g6YdKX8x3Z:-KKigXFfY2g6YdKX8x3_:-62904888) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/examples/dtw.ipynb#L1)):*
> ```
> {
> ```

Can you remove this file?  Example notebooks are now housed in the `librosa_gallery`repo.

---

*[librosa/core/__init__.py, line 96 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKigmFTy1I4f3k18B0f:-KKigmFTy1I4f3k18B0g:1455688091) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/__init__.py#L96)):*
> ```Python
> from .pitch import *  # pylint: disable=wildcard-import
> from .constantq import *  # pylint: disable=wildcard-import
> from .dtw import *  # pylint: disable=wildcard-import
> ```

I'm not sure this belongs under core.  What are the arguments for/against moving this up to a top-level submodule?

---

*[librosa/core/dtw.py, line 13 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihEsffHzUzu4_Tujv:-KKihEsffHzUzu4_Tujw:1814494449) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L13)):*
> ```Python
> 
> @optional_jit
> def band_mask(radius, mask):
> ```

This seems inefficient.  How about allocating an all-ones (boolean) matrix and then using `tril_indices`/`triu_indices` to zero out the triangles?

---

*[librosa/core/dtw.py, line 81 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKih_0TImnvuBWwwsYc:-KKih_0TImnvuBWwwsYd:767472258) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L81)):*
> ```Python
> def dtw(X, Y,
>         dist='euclidean',
>         step_sizes_sigma=np.array([[1, 1], [0, 1], [1, 0]]),
> ```

Don't use mutable data types as default parameters

---

*[librosa/core/dtw.py, line 136 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiitFy4rQzKz__qwTe:-KKiitFy4rQzKz__qwTf:-1968089942) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L136)):*
> ```Python
>         When doing subsequence DTW, D[N,:] indicates a matching function.
> 
>     wp : list [shape=(N,)]
> ```

This should be cast to ndarray, dtype=int

---

*[librosa/core/dtw.py, line 173 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihsVmLyf3CSLD1KuS:-KKihsVmLyf3CSLD1KuT:659661466) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L173)):*
> ```Python
>     Y = np.atleast_2d(Y)
> 
>     # if diagonal matching, Y has to be longer than Y
> ```

Y longer than Y?

---

*[librosa/core/dtw.py, line 175 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihvh61xOEjIlIAIkc:-KKihvh61xOEjIlIAIkd:-1871950178) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L175)):*
> ```Python
>     # if diagonal matching, Y has to be longer than Y
>     # (X simply cannot be contained in Y)
>     if np.array_equal(step_sizes_sigma, np.array([[1, 1]])) & \
> ```

use parentheses, not backslash continuation

---

*[librosa/core/dtw.py, line 190 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiiAE0YsPkVO-lIw_s:-KKiiAE0YsPkVO-lIw_t:-191455909) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L190)):*
> ```Python
> 
>         # set cost to infinity where the mask is True
>         C[my_mask == 0] = np.inf
> ```

It seems inefficient to construct a full mask matrix here.  It would be cheaper to do this in-place by using `tril_indices`/`triu_indices` to get the index sets, and then directly assigning into `C[idx] = np.inf`.

---

*[librosa/core/dtw.py, line 209 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiigpHY3nuo92wC6ro:-KKiigpHY3nuo92wC6rp:626777217) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L209)):*
> ```Python
>                                 max_0, max_1)
> 
>     # delete infinity rows and columns
> ```

Is this necessary?

---

*[librosa/core/dtw.py, line 277 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKijKd5QXMfOV-pOTJU:-KKijKd5QXMfOV-pOTJV:-149396661) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L277)):*
> ```Python
>         for cur_m in range(max_1, D.shape[1]):
>             # loop over all step sizes
>             for cur_step_idx in range(step_sizes_sigma.shape[0]):
> ```

This would be clearer if you iterated over `zip(step_sizes_sigma, weights_add, weights_mul)` instead of indices 

---

*[librosa/core/dtw.py, line 327 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKijuBrnjGU4Xeonou8:-KKijuBsw1dQV-8H4YPy:95777420) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L327)):*
> ```Python
>     wp.append(cur_idx)
> 
>     while True:
> ```

This and the exit condition of the loop could be combined: `while cur_idx[0]:`

---

*[librosa/util/decorators.py, line 60 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKik7ESzB22KSNG-o3j:-KKik7ESzB22KSNG-o3k:1787205406) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/util/decorators.py#L60)):*
> ```Python
> 
> 
> def optional_jit(func):
> ```

This implementation makes it impossible to pass parameters to jit.

I think you could do it a little differently:

```python
try:
    from numba.decorators import jit as optional_jit
    optional_jit = jit
except ImportError:

    def optional_jit(**_):
        def __wrapper(func, *args, **kwargs):
            return func(*args, **kwargs)
        return decorator(__wrapper)
```
This way, you can still do things like
```python
@optional_jit(target='gpu')
def myfunc(...)
```
and it will do the right thing whether or not numba is installed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Review status: 2 of 8 files reviewed at latest revision, 14 unresolved discussions.

---

*[.travis_dependencies.sh, line 14 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiu92hWPVY0YCCX7_D:-KKiu92iWHJcG3XggErl:-24333973) ([raw file](https://github.com/librosa/librosa/blob/2f49646eb6089bf9e97df9871044c98491364b77/.travis_dependencies.sh#L14)):*
> ```Shell
>     conda config --add channels pypi
>     conda info -a
>     deps='pip numpy scipy pandas requests nose coverage numpydoc matplotlib sphinx scikit-learn seaborn'
> ```

seaborn should be removed ; again, regression from master

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/__init__.py, line 96 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKigmFTy1I4f3k18B0f:-KKm7Gdg7mYxhe74KTtY:-209195556) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/__init__.py#L96)):*
> I'm not sure this belongs under core.  What are the arguments for/against moving this up to a top-level submodule?
</details>
Thought that DTW is more a technique, rather than an application. Example: decompose uses the technique NMF or a to be created module `score_alignment.py` uses DTW. But not having a strong opinion here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 81 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKih_0TImnvuBWwwsYc:-KKm8Ck2W1hQ1QRtPUPF:-754364307) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L81)):*
> Don't use mutable data types as default parameters
</details>
Rather having a parameter for each direction? I doubt that this is practical.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 173 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihsVmLyf3CSLD1KuS:-KKm8WtdGJeN7N--NtPB:-499146009) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L173)):*
> Y longer than Y?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[setup.py, line 44 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKifx-WqCcSPuLrAXmu:-KKm7BBE42OtO91n8G4-:-499146009) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/setup.py#L44)):*
> Looks like a regression here since the display rewrite was merged.  Can you rebase master?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[examples/dtw.ipynb, line 1 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKigXFfY2g6YdKX8x3Z:-KKm711X7lizLiccWfSP:-499146009) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/examples/dtw.ipynb#L1)):*
> Can you remove this file?  Example notebooks are now housed in the `librosa_gallery`repo.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 136 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiitFy4rQzKz__qwTe:-KKm9GJYUqO-c_m9iPxJ:-499146009) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L136)):*
> This should be cast to ndarray, dtype=int
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 175 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihvh61xOEjIlIAIkc:-KKm9b0bp9oStkEKzGfC:-1966989700) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L175)):*
> use parentheses, not backslash continuation
</details>
Done. Still fits if I put it in one line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 209 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiigpHY3nuo92wC6ro:-KKmA0tyOi27IRembB-0:67632193) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L209)):*
> Is this necessary?
</details>
These are the padded columns and rows which were added to keep the looping simple.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 327 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKijuBrnjGU4Xeonou8:-KKmAwc8TkKfdeg8n2CD:-499146009) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L327)):*
> This and the exit condition of the loop could be combined: `while cur_idx[0]:`
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/util/decorators.py, line 60 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKik7ESzB22KSNG-o3j:-KKmBuAUnooGnZ9_Yf2f:-1079102887) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/util/decorators.py#L60)):*
> This implementation makes it impossible to pass parameters to jit.
> 
> I think you could do it a little differently:
> 
> ```python
> try:
>     from numba.decorators import jit as optional_jit
>     optional_jit = jit
> except ImportError:
> 
>     def optional_jit(**_):
>         def __wrapper(func, *args, **kwargs):
>             return func(*args, **kwargs)
>         return decorator(__wrapper)
> ```
> This way, you can still do things like
> ```python
> @optional_jit(target='gpu')
> def myfunc(...)
> ```
> and it will do the right thing whether or not numba is installed.
</details>
Sorry, don't get you here. Is this still a separate decorator then?

What does `optional_jit = jit` mean if you import `optional_jit` from `numba`?

I think you have a particular plan here so please could you post the whole function?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[.travis_dependencies.sh, line 14 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiu92hWPVY0YCCX7_D:-KKm7CxxH3I3UBUUZ-sn:-499146009) ([raw file](https://github.com/librosa/librosa/blob/2f49646eb6089bf9e97df9871044c98491364b77/.travis_dependencies.sh#L14)):*
> seaborn should be removed ; again, regression from master
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[.travis_dependencies.sh, line 14 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiu92hWPVY0YCCX7_D:-KKmEk4LPMqZPbtWYR09:-628858489) ([raw file](https://github.com/librosa/librosa/blob/2f49646eb6089bf9e97df9871044c98491364b77/.travis_dependencies.sh#L14)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Done.
</details>
Actually this one is still in the master branch:
https://github.com/librosa/librosa/blob/master/.travis_dependencies.sh#L14

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Review status: 2 of 7 files reviewed at latest revision, 9 unresolved discussions.

---

*[setup.py, line 44 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKifx-WqCcSPuLrAXmu:-KKmw4bFeerBOYbAmmFh:-2096614165) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/setup.py#L44)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Done.
</details>
numba should still be an extras_require here and not a core dependency, since it's not really pip-installable.

---

*[librosa/core/dtw.py, line 81 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKih_0TImnvuBWwwsYc:-KKmwJzrsUA0lNppRqYl:693799606) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L81)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Rather having a parameter for each direction? I doubt that this is practical.
</details>
No, the usual way would be to set the default as `None` and then construct the default value inside the function.

It doesn't really matter in this case since the values are read-only within the function, but it's good style to avoid objects as defaults.

---

*[librosa/core/dtw.py, line 209 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiigpHY3nuo92wC6ro:-KKmx6QfT5uzFQdHhQ6Q:1525437177) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L209)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> These are the padded columns and rows which were added to keep the looping simple.
</details>
Okay.  I guess this slice ought to be cheap?  Can you verify that this only constructs a view and not a copy of `D`?

---

*[librosa/core/dtw.py, line 174 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKmwpF2B7FeVaIUak1T:-KKmwpF2B7FeVaIUak1U:311574035) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L174)):*
> ```Python
>     # (X simply cannot be contained in Y)
>     if np.array_equal(step_sizes_sigma, np.array([[1, 1]])) & X.shape[1] >= Y.shape[1]:
>         raise AssertionError('For diagonal matching: M >= N')
> ```

Don't use AssertionErrors in production code; those should be reserved for testing.

Instead, use a more descriptive exception class that can be caught directly.  This one looks like a `ParameterError` to me.

---

*[librosa/util/decorators.py, line 60 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKik7ESzB22KSNG-o3j:-KKmxPFxwhKUVZUdD2gZ:-1863415316) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/util/decorators.py#L60)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Sorry, don't get you here. Is this still a separate decorator then?
> 
> What does `optional_jit = jit` mean if you import `optional_jit` from `numba`?
> 
> I think you have a particular plan here so please could you post the whole function?
</details>
I think that code block I posed above is the whole function?

In more detail,here's what I had in mind:

1. If numba's installed, `optional_jit` should just be `numba.jit`.
2. If numba's not installed, `optional_jit` should be a no-op that leaves code untouched.  The tempting thing to do here is just say `optional_jit = lambda x : x`, but then you can't have additional arguments in the decorator call.  The dummy wrapper i put above lets you have arguments, but then throws them out, so the decorated function is untouched.  Using `decorator` here preserves the call signature and docstrings in the right way.

Thinking a bit more about it though, you probably want dummy positional arguments as well:
```python
def optional_jit(*_, **__):
    ...
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Review status: 2 of 7 files reviewed at latest revision, 12 unresolved discussions.

---

*[librosa/core/__init__.py, line 96 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKigmFTy1I4f3k18B0f:-KKnG92Ohzu6lEiHeUi8:933903501) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/__init__.py#L96)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Thought that DTW is more a technique, rather than an application. Example: decompose uses the technique NMF or a to be created module `score_alignment.py` uses DTW. But not having a strong opinion here.
</details>
Sure, that's reasonable enough.  I just want to make sure we think it through completely, since the core submodule is already getting pretty huge.

---

*[librosa/core/dtw.py, line 174 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKmwpF2B7FeVaIUak1T:-KKnGawBGscmY-LvUEcE:-1158845451) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L174)):*
> Don't use AssertionErrors in production code; those should be reserved for testing.
> 
> Instead, use a more descriptive exception class that can be caught directly.  This one looks like a `ParameterError` to me.
</details>
Additionally, `M` and `N` are not meaningful parameters here.  Use `X.shape[1]` and `Y.shape[1]` instead

---

*[librosa/core/dtw.py, line 221 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKnGq64CiNmpycx6mVy:-KKnGq64CiNmpycx6mVz:-1035149732) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L221)):*
> ```Python
>         wp = None
> 
>     return D, np.asarray(wp, dtype=int)
> ```

This will fail when `backtrack==False` because you can't case `None` to int.

I think it's probably best to only return `D` when the path is not requested; this is consistent with other numpy styles (eg, svd works this way).

---

*[librosa/core/dtw.py, line 336 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKnHEWiyJVLBI9Z_9dC:-KKnHEWiyJVLBI9Z_9dD:-1092335112) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L336)):*
> ```Python
>         # save tuple with minimal acc. cost in path
>         cur_idx = (cur_n-step_sizes_sigma[cur_step_idx][0],
>                    cur_m-step_sizes_sigma[cur_step_idx][1])
> ```

I think all of this can be vectorized without splitting `cur_idx` into `cur_n, cur_m`.  It would also make the code a bit easier to read:

```python
cur_step_idx = D_steps[cur_idx]
cur_idx = cur_idx - step_sizes_sigma[cur_step_idx]
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Tangential discussion here:

If we're going to introduce an optional dependency for numba, we should really set up the build matrix to run with and without numba installed.  (Really, we should have been doing this all along with the other optional dependencies, but since they're all gone now, it's somewhat moot.)

There are instructions for doing this kind of thing [here](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix).  I think it makes sense to just set this up as a boolean environment variable, and then have the `.travis_dependencies` script either install numba or not.  This will result in a 3x2 build matrix.

@stefan-balke can you add that into this PR?

stefan-balke:
> @stefan-balke can you add that into this PR?

Yup.

stefan-balke:
*[librosa/core/dtw.py, line 174 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKmwpF2B7FeVaIUak1T:-KKo19mZ4y3Qc6nytCxN:1455218469) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L174)):*
> Additionally, `M` and `N` are not meaningful parameters here.  Use `X.shape[1]` and `Y.shape[1]` instead
</details>
M and N are due to the reference to Meinard's book. I tried to keep it consistent for all functions here. Maybe I better define both variables in the docs?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Review status: 1 of 8 files reviewed at latest revision, 12 unresolved discussions.

---

*[librosa/core/dtw.py, line 174 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKmwpF2B7FeVaIUak1T:-KKo1x8safpHOqW-3-1K:173721098) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L174)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> M and N are due to the reference to Meinard's book. I tried to keep it consistent for all functions here. Maybe I better define both variables in the docs?
</details>
I think it's still clearer to refer directly to the input variables and not define new notation.  Users should not have to refer to a book to decode an exception message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 174 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKmwpF2B7FeVaIUak1T:-KKo2RNFPsgzVXMnWUB3:2046083656) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L174)):*
> I think it's still clearer to refer directly to the input variables and not define new notation.  Users should not have to refer to a book to decode an exception message.
</details>
Oh, you mean the exception message...didn't get that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 209 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiigpHY3nuo92wC6ro:-KKo3GSxvSxbXsdQafgx:-205375782) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L209)):*
> Okay.  I guess this slice ought to be cheap?  Can you verify that this only constructs a view and not a copy of `D`?
</details>
Wouldn't that be the standard behaviour? How could I verify that?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 221 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKnGq64CiNmpycx6mVy:-KKo3dsCk0Gynm06al63:-499146009) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L221)):*
> This will fail when `backtrack==False` because you can't case `None` to int.
> 
> I think it's probably best to only return `D` when the path is not requested; this is consistent with other numpy styles (eg, svd works this way).
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Reviewed 2 of 3 files at r13.
Review status: 3 of 8 files reviewed at latest revision, 11 unresolved discussions.

---

*[.travis_dependencies.sh, line 14 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiu92hWPVY0YCCX7_D:-KKo4kfzJMyYmNCXbt6A:-2090506562) ([raw file](https://github.com/librosa/librosa/blob/2f49646eb6089bf9e97df9871044c98491364b77/.travis_dependencies.sh#L14)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Actually this one is still in the master branch:
> https://github.com/librosa/librosa/blob/master/.travis_dependencies.sh#L14
</details>
Really?  That's weird.

Apparently so is libsamplerate O_o

I'll make a separate PR to clean that stuff out.  Thanks for catching it.

---

*[librosa/core/dtw.py, line 209 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiigpHY3nuo92wC6ro:-KKo5PqEeG0tOGrZHVwB:188886802) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L209)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Wouldn't that be the standard behaviour? How could I verify that?
</details>
It would be when you're slicing one axis at a time, i'm not sure if the double slice triggers a copy though, since it breaks stride continuity at the row boundaries.

You can test this by looking at `D.flags['OWNDATA']` which should be `False` if it's a view (good) and `True` if it's a copy (bad).

---

*[librosa/core/dtw.py, line 174 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKmwpF2B7FeVaIUak1T:-KKo59MLvct_xGp4_oBO:-1497031300) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L174)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Oh, you mean the exception message...didn't get that.
</details>
Also I think you want an `and` (logical conjunction) not `&` (bitwise conjunction) here.

pylint ought to catch this sort of thing for you &lt;/hint&gt; :grin:

---

*[librosa/core/dtw.py, line 221 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKnGq64CiNmpycx6mVy:-KKo5vOTVtZVgCbY3W4Y:-294915522) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L221)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Done.
</details>
:+1: now you could simplify the branching logic here so we only check `backtrack` once.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 336 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKnHEWiyJVLBI9Z_9dC:-KKoB8uJAPBwXNDaT1jV:1482202178) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L336)):*
> I think all of this can be vectorized without splitting `cur_idx` into `cur_n, cur_m`.  It would also make the code a bit easier to read:
> 
> ```python
> cur_step_idx = D_steps[cur_idx]
> cur_idx = cur_idx - step_sizes_sigma[cur_step_idx]
> ```
</details>
Tried to fix it. Not convincing yet imho...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 221 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKnGq64CiNmpycx6mVy:-KKoBhdSdK-EuEkh0g-b:1102498225) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L221)):*
> :+1: now you could simplify the branching logic here so we only check `backtrack` once.
</details>
Oops, you're right :) - Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 209 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiigpHY3nuo92wC6ro:-KKoC8p-dhzh61OvYfaA:-1775979053) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L209)):*
> It would be when you're slicing one axis at a time, i'm not sure if the double slice triggers a copy though, since it breaks stride continuity at the row boundaries.
> 
> You can test this by looking at `D.flags['OWNDATA']` which should be `False` if it's a view (good) and `True` if it's a copy (bad).
</details>
Nice one. It's `False` - so all good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 174 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKmwpF2B7FeVaIUak1T:-KKoCe6pJsrm8LEPM7nT:-1602766771) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L174)):*
> Also I think you want an `and` (logical conjunction) not `&` (bitwise conjunction) here.
> 
> pylint ought to catch this sort of thing for you &lt;/hint&gt; :grin:
</details>
Okay, not mine (Atom + linter). :/

But unveiled a logic error...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 81 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKih_0TImnvuBWwwsYc:-KKoFfTqJlWGJCpJfLX1:-499146009) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L81)):*
> No, the usual way would be to set the default as `None` and then construct the default value inside the function.
> 
> It doesn't really matter in this case since the values are read-only within the function, but it's good style to avoid objects as defaults.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 277 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKijKd5QXMfOV-pOTJU:-KKoMNZ16kuzoBHuCijF:-175059366) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L277)):*
> This would be clearer if you iterated over `zip(step_sizes_sigma, weights_add, weights_mul)` instead of indices 
</details>
Tried that as well but I need to keep the `cur_step_idx` for the step-matrix...so it's a mix :(

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Review status: 3 of 8 files reviewed at latest revision, 9 unresolved discussions.

---

*[librosa/core/dtw.py, line 277 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKijKd5QXMfOV-pOTJU:-KKoMdSI8tcpXy1LPhge:2136537809) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L277)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Tried that as well but I need to keep the `cur_step_idx` for the step-matrix...so it's a mix :(
</details>
Oh, I hadn't noticed that before.

---

*[librosa/core/dtw.py, line 336 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKnHEWiyJVLBI9Z_9dC:-KKoMU-5xpGNtsuFEI60:-1766212163) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L336)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Tried to fix it. Not convincing yet imho...
</details>
"not convincing" or "does not work"?

---

*[librosa/core/dtw.py, line 179 \[r15\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKoM-_qSvEVaCQ2auZf:-KKoM-_qSvEVaCQ2auZg:485367348) ([raw file](https://github.com/librosa/librosa/blob/e403967dec7e26c44c789e60170c1c4999ae3d7e/librosa/core/dtw.py#L179)):*
> ```Python
>     # (X simply cannot be contained in Y)
>     if np.array_equal(step_sizes_sigma, np.array([[1, 1]])) and (X.shape[1] > Y.shape[1]):
>         raise ValueError('For diagonal matching: Y.shape[1] >= X.shape[1]')
> ```

librosa uses a custom exception class (`ParameterError`) for this kind of error (mal-formed inputs).  This makes it easier to isolate improper usage of the library from unexpected internal errors.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 336 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKnHEWiyJVLBI9Z_9dC:-KKoNWUAvGbzU3RX9Erx:848386400) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L336)):*
> "not convincing" or "does not work"?
</details>
It works, but the casting is now ugly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 179 \[r15\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKoM-_qSvEVaCQ2auZf:-KKoOjOrRAY5p35f6FOI:-1497627346) ([raw file](https://github.com/librosa/librosa/blob/e403967dec7e26c44c789e60170c1c4999ae3d7e/librosa/core/dtw.py#L179)):*
> librosa uses a custom exception class (`ParameterError`) for this kind of error (mal-formed inputs).  This makes it easier to isolate improper usage of the library from unexpected internal errors.
</details>
This explains why I didn't find it in the Python docs :)

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Review status: 3 of 8 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[librosa/core/dtw.py, line 336 \[r12\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKnHEWiyJVLBI9Z_9dC:-KKoZjNUyGhyp8KFf-qE:-868067669) ([raw file](https://github.com/librosa/librosa/blob/0ed698ce372567844da411f7e2b680d9c49e27d1/librosa/core/dtw.py#L336)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> It works, but the casting is now ugly.
</details>
oh, I see.  nevermind then.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/util/decorators.py, line 60 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKik7ESzB22KSNG-o3j:-KKrNbaxhvSO492wLdg3:-499146009) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/util/decorators.py#L60)):*
> I think that code block I posed above is the whole function?
> 
> In more detail,here's what I had in mind:
> 
> 1. If numba's installed, `optional_jit` should just be `numba.jit`.
> 2. If numba's not installed, `optional_jit` should be a no-op that leaves code untouched.  The tempting thing to do here is just say `optional_jit = lambda x : x`, but then you can't have additional arguments in the decorator call.  The dummy wrapper i put above lets you have arguments, but then throws them out, so the decorated function is untouched.  Using `decorator` here preserves the call signature and docstrings in the right way.
> 
> Thinking a bit more about it though, you probably want dummy positional arguments as well:
> ```python
> def optional_jit(*_, **__):
>     ...
> ```
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 13 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihEsffHzUzu4_Tujv:-KKrc05ll9veJ_x7pZTe:1108700745) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L13)):*
> This seems inefficient.  How about allocating an all-ones (boolean) matrix and then using `tril_indices`/`triu_indices` to zero out the triangles?
</details>
I'd like to ping @craffel here since it's his code but he's still busy. Maybe it's due to `numba`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 190 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiiAE0YsPkVO-lIw_s:-KKrcXcC76uE88Yi6IzV:2101406251) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L190)):*
> It seems inefficient to construct a full mask matrix here.  It would be cheaper to do this in-place by using `tril_indices`/`triu_indices` to get the index sets, and then directly assigning into `C[idx] = np.inf`.
</details>
see above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Reviewed 1 of 2 files at r6, 1 of 3 files at r13, 1 of 2 files at r17.
Review status: 6 of 8 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[librosa/core/dtw.py, line 13 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihEsffHzUzu4_Tujv:-KKsLMSt9zoEpEbh3wZV:-1027344695) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L13)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> I'd like to ping @craffel here since it's his code but he's still busy. Maybe it's due to `numba`?
</details>
I'm not even sure we need this function at all.

It only gets used in one place, and that's explicitly to zero out the triangles of a given matrix.

I would much rather have a function that runs in-place to do this, eg,

```python
def band_mask(x, k, value):

    # not suer about k here, might need to be radius+1
    idx_u = np.triu_indices_from(x, k=k)
    idx_l = np.tril_indices_from(x, k=k)
    x[idx_u] = value
    x[idx_l] = value
```

this ought to be more memory efficient than the current numba code, and just as fast since it's using numpy builtins.  The main benefit though is that it's much less code, and the functionality is obvious.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Review status: 6 of 8 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[.travis_dependencies.sh, line 14 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiu92hWPVY0YCCX7_D:-KKsUUdUR5_-Vmdwtsl7:-28547246) ([raw file](https://github.com/librosa/librosa/blob/2f49646eb6089bf9e97df9871044c98491364b77/.travis_dependencies.sh#L14)):*
> Really?  That's weird.
> 
> Apparently so is libsamplerate O_o
> 
> I'll make a separate PR to clean that stuff out.  Thanks for catching it.
</details>
I fixed this in master; please rebase

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[.travis_dependencies.sh, line 14 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiu92hWPVY0YCCX7_D:-KKyfIM_wV6r2urTXw2k:-499146009) ([raw file](https://github.com/librosa/librosa/blob/2f49646eb6089bf9e97df9871044c98491364b77/.travis_dependencies.sh#L14)):*
> I fixed this in master; please rebase
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 13 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihEsffHzUzu4_Tujv:-KKyftnF6HsLIg0jpk1x:-687760316) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L13)):*
> I'm not even sure we need this function at all.
> 
> It only gets used in one place, and that's explicitly to zero out the triangles of a given matrix.
> 
> I would much rather have a function that runs in-place to do this, eg,
> 
> ```python
> def band_mask(x, k, value):
> 
>     # not suer about k here, might need to be radius+1
>     idx_u = np.triu_indices_from(x, k=k)
>     idx_l = np.tril_indices_from(x, k=k)
>     x[idx_u] = value
>     x[idx_l] = value
> ```
> 
> this ought to be more memory efficient than the current numba code, and just as fast since it's using numpy builtins.  The main benefit though is that it's much less code, and the functionality is obvious.
</details>
Looks good, will try it out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Reviewed 1 of 1 files at r18.
Review status: 6 of 8 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 13 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihEsffHzUzu4_Tujv:-KN6wtJHvW8JjB3WES8E:1320045255) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L13)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Looks good, will try it out.
</details>
Finally found some time to do it. Still not convinced by the if-else condition. Any thoughts?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Review status: 5 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[librosa/core/dtw.py, line 13 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihEsffHzUzu4_Tujv:-KN7TzUSE9OJlrLw8yB7:2127325512) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L13)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> Finally found some time to do it. Still not convinced by the if-else condition. Any thoughts?
</details>
What do you mean by if-else condition?

---

*[tests/test_dtw.py, line 69 \[r19\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KN7U7OtsDXqGeWTQS5I:-KN7U7OtsDXqGeWTQS5J:-275341941) ([raw file](https://github.com/librosa/librosa/blob/c0290fa6ee913ae9c261b811a81bbb7d2387e3de/tests/test_dtw.py#L69)):*
> ```Python
>     assert np.array_equal(mut_x, gt_x)
> 
>     # Case 2: N!=M
> ```

This should be a separate test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[librosa/core/dtw.py, line 13 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKihEsffHzUzu4_Tujv:-KNBIMEmDe-37zQX0bmm:1203392194) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L13)):*
> What do you mean by if-else condition?
</details>
Checking for the "longer" axis. Would love to have something like `np.roll_to_biggest_axis()` but it's okay I guess.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
I squashed the commits. This one is ready to go merge.

bmcfee:
Reviewed 1 of 1 files at r10.
Review status: 6 of 8 files reviewed at latest revision, 5 unresolved discussions.

---

*[librosa/core/dtw.py, line 190 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiiAE0YsPkVO-lIw_s:-KNCivtIAP2xMDb9dj7z:749340018) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L190)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> see above
</details>
looks like `my_mask` is no longer used here.

Can you pylint this and track down any other unused variables / icky stuff?

---

*[tests/test_dtw.py, line 69 \[r19\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KN7U7OtsDXqGeWTQS5I:-KNCidKBEGwd1R2QGk_g:1542667906) ([raw file](https://github.com/librosa/librosa/blob/c0290fa6ee913ae9c261b811a81bbb7d2387e3de/tests/test_dtw.py#L69)):*
> This should be a separate test.
</details>
There should also be a test on the transpose of this, just to make sure our corner cases are correct.  Otherwise looks good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
Done.

bmcfee:
Review status: 6 of 8 files reviewed at latest revision, 7 unresolved discussions.

---

*[librosa/core/dtw.py, line 14 \[r21\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KNDgLRLv12qt_hNyExc:-KNDgLRLv12qt_hNyExd:-174340457) ([raw file](https://github.com/librosa/librosa/blob/7fce28c9dc3cb89f8550e23e8864135a4190e6e5/librosa/core/dtw.py#L14)):*
> ```Python
> 
> def band_mask(x, radius, value=0):
>     """Construct band-around-diagonal mask (Sakoe-Chiba band).  When
> ```

This doesn't actually create a band anymore; it fills values outside the band.  Maybe we can find a more appropriate name and/or description?

---

*[librosa/core/dtw.py, line 28 \[r21\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KNDgTVyHTGBGeYM0K46:-KNDgTVyHTGBGeYM0K47:497412784) ([raw file](https://github.com/librosa/librosa/blob/7fce28c9dc3cb89f8550e23e8864135a4190e6e5/librosa/core/dtw.py#L28)):*
> ```Python
>         ``int(radius*min(mask.shape))``.
>     value : int
>         Replacement-value.
> ```

This is unclear.  Suggest "`x[i,j] = value` when `(i, j)` lies outside the band"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
I called the band now ``global_constraints`` which is a bit more general.

bmcfee:
Review status: 6 of 8 files reviewed at latest revision, 5 unresolved discussions.

---

*[librosa/core/dtw.py, line 190 \[r10\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KKiiAE0YsPkVO-lIw_s:-KNHk0vgx5Q5dXpNrQES:775170352) ([raw file](https://github.com/librosa/librosa/blob/8f7ef212c78cb3ba3028e6e7568e4607f789c540/librosa/core/dtw.py#L190)):*
> looks like `my_mask` is no longer used here.
> 
> Can you pylint this and track down any other unused variables / icky stuff?
</details>
It's bad luck to have a variable name (`global_constraints`) the same as a function within scope.

Given the options, I'd prefer to change the function name (again, sorry).  The name "global_constraints" doesn't convey much information.  How about `fill_off_diagonal`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
Did I miss anything? Otherwise this one is ready.

bmcfee:
Review status: 6 of 8 files reviewed at latest revision, 4 unresolved discussions.

---

*[librosa/core/dtw.py, line 7 \[r23\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KNb1H7jnOuC_t-UA4dn:-KNb1H7jnOuC_t-UA4do:1399087098) ([raw file](https://github.com/librosa/librosa/blob/a0e402d13478a9163dcdc472ba8739bdcd14d10c/librosa/core/dtw.py#L7)):*
> ```Python
> import numpy as np
> from scipy.spatial.distance import cdist
> from librosa.util.decorators import optional_jit
> ```

Use local module names:
```python
from ..util.decorators import optional_jit
```

---

*[tests/test_dtw.py, line 69 \[r19\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KN7U7OtsDXqGeWTQS5I:-KNb1qxlUXgl2hzSCZEd:-1639676152) ([raw file](https://github.com/librosa/librosa/blob/c0290fa6ee913ae9c261b811a81bbb7d2387e3de/tests/test_dtw.py#L69)):*
> There should also be a test on the transpose of this, just to make sure our corner cases are correct.  Otherwise looks good.
</details>
There still needs to be a test for the transpose (12x8)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

stefan-balke:
*[tests/test_dtw.py, line 69 \[r19\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KN7U7OtsDXqGeWTQS5I:-KNWJp6_WNQ1apspeC25:-458649260) ([raw file](https://github.com/librosa/librosa/blob/c0290fa6ee913ae9c261b811a81bbb7d2387e3de/tests/test_dtw.py#L69)):*
> There still needs to be a test for the transpose (12x8)
</details>
It's in line 88 or you want a separate test? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Reviewed 1 of 2 files at r23, 1 of 1 files at r24.

---

*[tests/test_dtw.py, line 69 \[r19\]](https://reviewable.io:443/reviews/librosa/librosa/323#-KN7U7OtsDXqGeWTQS5I:-KNbDnvCFlOzxwoBh18o:153347722) ([raw file](https://github.com/librosa/librosa/blob/c0290fa6ee913ae9c261b811a81bbb7d2387e3de/tests/test_dtw.py#L69)):*
<details><summary><i>Previously, stefan-balke (Stefan Balke) wrote…</i></summary>
> It's in line 88 or you want a separate test? 
</details>
ohhh i see.  I prefer small tests, but this is fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/323)*

bmcfee:
Merged.  Thanks @stefan-balke !

dpwe:
I got all excited only to find I had been unassigned!

  DAn.

Issue #322, opened 2016-04-09
Add resampy support to resample()


bmcfee:
Once [resampy](https://github.com/bmcfee/resampy) has a stable release, we should switch over to use that instead of (in addition to) scipy.resample.

This should add resampy as a proper dependency to librosa.

We can then deprecate the scikits.samplerate options.

bmcfee:
Fixed by merging #324

PR #324, opened 2016-04-20
[WIP] Resampy integration

bmcfee:
This PR implements #322.

- All resampling operations are switched over to use resampy by default
- CQT now uses resampy's `kaiser_best` and `kaiser_fast` for its downsampling
- audio.bandwidth parameters have been updated to match resampy's pre-computed roll-off frequencies
- the `resample` implementation has been simplified

Question for the other contributors (paging @ebattenberg @stefan-balke ):

Should we deprecate scikits.samplerate support?  I'm leaning toward yes, unless there's a strong argument in favor of keeping it going forward.  If we do deprecate it, I'll add the appropriate warning code in this PR.

TODO:

- [x] after resampy 0.1 release, remove force-install from travis and update setup.py requirements
- [x] deprecate scikits.samplerate?

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,16 +55,6 @@ librosa (0.x.x, /path/to/librosa)
 
 ### Hints for OS X
 
-#### libsamplerate
-
-In order to use *scipy* with *libsamplerate*, you can use *homebrew* (http://brew.sh)
-for installation:
-```
-brew install libsamplerate
-```
-
-The Python bindings are installed via `pip install scikits.samplerate`.
-
 #### ffmpeg
 
 To fuel `audioread` with more audio-decoding power, you can install *ffmpeg* which
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -10,6 +10,7 @@ import audioread
 import numpy as np
 import scipy.signal
 import scipy.fftpack as fft
+import resampy
 
 
 from .time_frequency import frames_to_samples, time_to_samples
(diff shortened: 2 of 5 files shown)


ebattenberg:
Other than easier install, how does resampy compare to libsamplerate in terms of quality and speed?

bmcfee:
> Other than easier install, how does resampy compare to libsamplerate in terms of quality and speed?

It's a bit faster -- maybe 10%, in my ad-hoc experiments -- and ever-so-slightly lower quality for now.  The high-quality mode has a slightly lower roll-off frequency than src's best (0.946 vs 0.97), and a little more aliasing down around -110dB.  The "fast" mode is comparable to src, but with a higher roll-off (0.85 vs 0.8).  See [this issue discussion](https://github.com/bmcfee/resampy/issues/4) for details.

On the plus side, the API makes it quite easy to drop in new filter designs, so later versions may close that gap back down.

Aside from simplifying the installation, which is a huge win from my perspective, it also has a non-viral license.  This means we can make it a proper dependency without having to deal with any dual licensing headaches.

stefan-balke:
As I still feel the pain of trying to build libsamplerate I'm get my hands up for resampy!
I think most of the time we take the resampling for granted and treat it like a black box. This package gets the thing back on the table which is good.

ebattenberg:
Yeah, it seems like resampy has a lot of benefits at little to no cost.  (Other than the coding you've done, Brian).  I support it replacing libsamplerate and whatever we use when libsamplerate isn't available.

bmcfee:
> I support it replacing libsamplerate and whatever we use when libsamplerate isn't available.

I'll take that as a ratifying vote :smile: will add in the deprecation warnings before merging.

PR #347, opened 2016-05-01
[CR] fix hybrid_cqt scaling. should resolve #341

### LINKED ISSUE #341: hybrid_cqt magnitude scaling is inconsistent
Hi,
With @CarmineCella we've tried running the CQT on synthetic sounds, such as white noise (`np.random.randn`) as well as Diracs.
It looks like the `cqt` function, which relies on the pyramid algorithm of Schoerkuber and Klapuri, is not appropriately scaled in magnitude across octaves.

Below is the code to compute the CQT of a Dirac and the corresponding figure:

```python
import librosa
import numpy as np
x = np.zeros(32768)
x[16384] = 1.0
sr = 22050.0
c = librosa.cqt(x, sr=sr)
librosa.display.specshow(c)
```

cqt_dirac

Am I mistaken to think that the magnitudes should stay equal across octaves instead ?

bmcfee:
This fix attempts to resolve the scale discrepancy between the pseudo and full components of a hybrid_cqt.

For hop_lengths that are powers of 2, the response to an impulse is approximately constant, and the scale correction is smooth.

For odd-multiple hop-lengths, the full CQT spectrum decays but the pseudo-cqt spectrum is flat.  In these cases, we see a jump discontinuity.  I'm not sure this can be resolved directly here.

The new unit tests check the continuity of the spectrum in response to an impulse, and the total (logarithmic) deviation from the median energy.

Some additional CR from @ebattenberg and/or @dpwe would be appreciated before merging.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -385,6 +385,21 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                    filter_scale=filter_scale,
                                    norm=norm,
                                    sparsity=sparsity)
+
+        # Compute the scaling factor for the pseudo-cqt
+        Q = float(filter_scale) / (2.0**(1. / bins_per_octave) - 1)
+
+        filter_cutoff = (freqs[-(n_bins_pseudo+1)] *
+                         (1 + filters.window_bandwidth('hann') / Q))
+
+        n_octaves = int(np.ceil(float(n_bins - n_bins_pseudo) / bins_per_octave))
+
+        downsample_count = __early_downsample_count(sr/2,
+                                                    filter_cutoff,
+                                                    hop_length,
+                                                    n_octaves)
+
+        my_pseudo_cqt *= 2.0**(downsample_count - 1)
         cqt_resp.append(my_pseudo_cqt)
 
     n_bins_full = int(np.sum(~pseudo_filters))
@@ -586,20 +601,26 @@ def __variable_hop_response(y, n_fft, hop_length, min_filter_length,
     return my_cqt
 
 
-def __early_downsample(y, sr, hop_length, res_type, n_octaves,
-                       nyquist, filter_cutoff):
-    '''Perform early downsampling on an audio signal, if it applies.'''
 
-    if res_type != 'kaiser_fast':
-        return y, sr, hop_length
+def __early_downsample_count(nyquist, filter_cutoff, hop_length, n_octaves):
+    '''Compute the number of early downsampling operations'''
 
     downsample_count1 = int(np.ceil(np.log2(audio.BW_FASTEST * nyquist /
                                             filter_cutoff)) - 1)
     num_twos = __num_two_factors(hop_length)
     downsample_count2 = max(0, num_twos - n_octaves + 1)
-    downsample_count = min(downsample_count1, downsample_count2)
 
-    if downsample_count > 0:
+    return min(downsample_count1, downsample_count2)
+
+
+def __early_downsample(y, sr, hop_length, res_type, n_octaves,
+                       nyquist, filter_cutoff):
+    '''Perform early downsampling on an audio signal, if it applies.'''
+
+    downsample_count = __early_downsample_count(nyquist, filter_cutoff,
+                                                hop_length, n_octaves)
+
+    if downsample_count > 0 and res_type == 'kaiser_fast':
         downsample_factor = 2.0**(downsample_count)
 
         assert hop_length % downsample_factor == 0
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -3,9 +3,9 @@
 '''Pitch-tracking and tuning estimation'''
 from __future__ import division
 
+from warnings import warn
 import numpy as np
 import scipy.fftpack as fft
-from warnings import warn
 
 from . import audio
 from .time_frequency import cqt_frequencies, note_to_hz
@@ -193,18 +193,16 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     if res_type != 'kaiser_fast':
 
         # Do the top octave before resampling to allow for fast resampling
-        fft_basis, n_fft, _ = __fft_filters(sr, fmin_t,
-                                            n_filters,
-                                            bins_per_octave,
-                                            tuning,
-                                            filter_scale,
-                                            norm,
-                                            sparsity)
+        fft_basis, n_fft, _ = __cqt_filter_fft(sr, fmin_t,
+                                               n_filters,
+                                               bins_per_octave,
+                                               tuning,
+                                               filter_scale,
+                                               norm,
+                                               sparsity)
 
-        my_cqt = __fft_response(y, n_fft, hop_length, fft_basis)
-
-        # Convolve
-        cqt_resp.append(my_cqt)
+        # Compute the CQT filter response and append it to the stack
+        cqt_resp.append(__cqt_response(y, n_fft, hop_length, fft_basis))
 
         fmin_t /= 2
         fmax_t /= 2
@@ -222,13 +220,13 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                              .format(n_octaves - 1, n_octaves))
 
     # Now do the recursive bit
-    fft_basis, n_fft, _ = __fft_filters(sr, fmin_t,
-                                        n_filters,
-                                        bins_per_octave,
-                                        tuning,
-                                        filter_scale,
-                                        norm,
-                                        sparsity)
+    fft_basis, n_fft, _ = __cqt_filter_fft(sr, fmin_t,
+                                           n_filters,
+                                           bins_per_octave,
+                                           tuning,
+                                           filter_scale,
+                                           norm,
+                                           sparsity)
 
     my_y, my_sr, my_hop = y, sr, hop_length
 
@@ -248,11 +246,9 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
             my_sr /= 2.0
             my_hop //= 2
 
-        # Compute a dynamic hop based on n_fft
-        my_cqt = __fft_response(my_y, n_fft, my_hop, fft_basis)
+        # Compute the cqt filter response and append to the stack
+        cqt_resp.append(__cqt_response(my_y, n_fft, my_hop, fft_basis))
 
-        # Convolve
-        cqt_resp.append(my_cqt)
 
     return __trim_stack(cqt_resp, n_bins, real)
 
@@ -361,7 +357,7 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
     if n_bins_pseudo > 0:
         fmin_pseudo = np.min(freqs[pseudo_filters])
-        my_pseudo_cqt = pseudo_cqt(y, sr,
+        cqt_resp.append(pseudo_cqt(y, sr,
                                    hop_length=hop_length,
                                    fmin=fmin_pseudo,
                                    n_bins=n_bins_pseudo,
@@ -369,23 +365,19 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                    tuning=tuning,
                                    filter_scale=filter_scale,
                                    norm=norm,
-                                   sparsity=sparsity)
-
-        cqt_resp.append(my_pseudo_cqt)
+                                   sparsity=sparsity))
 
     if n_bins_full > 0:
-        my_cqt = np.abs(cqt(y, sr,
-                            hop_length=hop_length,
-                            fmin=fmin,
-                            n_bins=n_bins_full,
-                            bins_per_octave=bins_per_octave,
-                            tuning=tuning,
-                            filter_scale=filter_scale,
-                            norm=norm,
-                            sparsity=sparsity,
-                            real=False))
-
-        cqt_resp.append(my_cqt)
+        cqt_resp.append(np.abs(cqt(y, sr,
+                                   hop_length=hop_length,
+                                   fmin=fmin,
+                                   n_bins=n_bins_full,
+                                   bins_per_octave=bins_per_octave,
+                                   tuning=tuning,
+                                   filter_scale=filter_scale,
+                                   norm=norm,
+                                   sparsity=sparsity,
+                                   real=False)))
 
     return __trim_stack(cqt_resp, n_bins, True)
 
@@ -468,15 +460,11 @@ def pseudo_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     if tuning is None:
         tuning = estimate_tuning(y=y, sr=sr)
 
-    fft_basis, n_fft, _ = __fft_filters(sr,
-                                        fmin,
-                                        n_bins,
-                                        bins_per_octave,
-                                        tuning,
-                                        filter_scale,
-                                        norm,
-                                        sparsity,
-                                        hop_length=hop_length)
+    fft_basis, n_fft, _ = __cqt_filter_fft(sr, fmin, n_bins,
+                                           bins_per_octave,
+                                           tuning, filter_scale,
+                                           norm, sparsity,
+                                           hop_length=hop_length)
 
     fft_basis = np.abs(fft_basis)
 
@@ -487,8 +475,8 @@ def pseudo_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     return fft_basis.dot(D)
 
 
-def __fft_filters(sr, fmin, n_bins, bins_per_octave, tuning,
-                  filter_scale, norm, sparsity, hop_length=None):
+def __cqt_filter_fft(sr, fmin, n_bins, bins_per_octave, tuning,
+                     filter_scale, norm, sparsity, hop_length=None):
     '''Generate the frequency domain constant-Q filter basis.'''
 
     basis, lengths = filters.constant_q(sr,
@@ -506,7 +494,7 @@ def __fft_filters(sr, fmin, n_bins, bins_per_octave, tuning,
     if hop_length is not None and n_fft < 2.0**(1 + np.ceil(np.log2(hop_length))):
         n_fft = int(2.0 ** (1 + np.ceil(np.log2(hop_length))))
 
-    # normalize by inverse length to compensate for phase invariance
+    # re-normalize bases with respect to the FFT window length
     basis *= lengths[:, np.newaxis] / float(n_fft)
 
     # FFT and retain only the non-negative frequencies
@@ -535,7 +523,7 @@ def __trim_stack(cqt_resp, n_bins, real):
     return C
 
 
-def __fft_response(y, n_fft, hop_length, fft_basis):
+def __cqt_response(y, n_fft, hop_length, fft_basis):
     '''Compute the filter response with a target STFT hop.'''
 
     # Compute the STFT matrix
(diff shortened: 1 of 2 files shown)


bmcfee:
After introducing the fix, we get the following peak-energy curves in response to an impulse.

```python
x = np.zeros(22050)
x[11025] = 1.0
sr = 22050.0
cf = librosa.cqt(x, sr=sr, hop_length=192)
ch = librosa.hybrid_cqt(x, sr=sr, hop_length=192)
```

image

For `hop_length=64` we get

image

For `512` we get

image

bmcfee:
To do: resolve the global scaling to match full cqt.

bmcfee:
Also note: the dip in the above plots for the full cqt region is due to the impulse not being centered on any frame.

bmcfee:
Now that we have hybrid_cqt computing reasonably continuous outputs on impulse inputs, I wanted to sort out the occasional discrepancy in total magnitude between `hybrid_cqt` and `cqt`.

Here's a simple test function to compare whether the two are operating in the same range:
```python
def hvftest(sr, hop_scale):
    
    hop_length = 64 * hop_scale

    x = np.zeros(sr)

    center = int(np.floor(len(x) / (2 * float(hop_length))) * hop_length)

    x[center] = 1.0

    cf = np.max(np.abs(librosa.cqt(x, sr=sr, hop_length=hop_length, real=False, tuning=0)))
    ch = np.max(librosa.hybrid_cqt(x, sr=sr, hop_length=hop_length, tuning=0))
    
    return cf / ch
```
It generates a 1-second input with a frame-centered impulse near the middle.  I then sweep over sampling rates and `hop_scale`; for this experiment, I'm keeping the number of octaves pinned to 7 (84 bins), so the hop_length has to be an integer multiple of `2**(7-1) = 64`.  Sweeping over the scale reveals where things go wrong:
```python
for sr in [11025, 16000, 22050, 32000, 44100]:
    print('sr={}'.format(sr))
    for hop_scale in range(1, 33):
        tv = hvftest(sr, hop_scale)
        if np.abs(np.log2(tv)) >= 1:
            print('\tFAIL: hop_scale={:2d}, scale error={:.3g}'.format(hop_scale, tv))
    print('-----')
```
With the following output
```
sr=11025
-----
sr=16000
	FAIL: hop_scale= 2, scale error=0.485
	FAIL: hop_scale= 4, scale error=0.484
	FAIL: hop_scale= 6, scale error=0.484
	FAIL: hop_scale= 8, scale error=0.484
	FAIL: hop_scale=12, scale error=0.485
	FAIL: hop_scale=14, scale error=0.485
	FAIL: hop_scale=16, scale error=0.485
	FAIL: hop_scale=22, scale error=0.485
	FAIL: hop_scale=24, scale error=0.485
	FAIL: hop_scale=26, scale error=0.485
	FAIL: hop_scale=28, scale error=0.485
	FAIL: hop_scale=30, scale error=0.485
	FAIL: hop_scale=32, scale error=0.485
-----
sr=22050
-----
sr=32000
	FAIL: hop_scale= 4, scale error=0.487
	FAIL: hop_scale= 8, scale error=0.487
	FAIL: hop_scale=12, scale error=0.488
	FAIL: hop_scale=16, scale error=0.487
	FAIL: hop_scale=24, scale error=0.488
	FAIL: hop_scale=28, scale error=0.487
	FAIL: hop_scale=32, scale error=0.487
-----
sr=44100
-----
```
These scale-discrepancies are always coming from the bottom end of the spectrum, where both cqt methods should be operating in `full` mode.  Example for `sr=16000, hop_scale=2`:

image

image

Even though hcqt is continuous here, I'm inclined to think that `cqt` is scaled correctly and `hcqt` is off by a power of two.  However, it's not obvious to me how to tell from the hop length parameter when this should happen.  Any thoughts?

bmcfee:
Doing more digging: the notches at the high end in vanilla cqt are coming from `variable_hop_response` using `np.mean` as aggregation with complex inputs.

Is there a sane way to get rid of `variable_hop_response`?

bmcfee:
Update: scale discrepancy from CQT vs HCQT comes down to the basis construction / fft length.  More on this as it develops.

bmcfee:
I'm beginning to suspect that resolving this entire issue would be easier if `pseudo_cqt` also did early downsampling.  Is there any glaring reason why this would be a bad idea?

dpwe:
I don't understand the issue.  "Early downsampling" is when you're not
trying to get any information out of the top part of your spectrum, so
you downsample to push the nyquist closer to the top of what you want
to get before doing anything?  Sounds OK.  How often do we not care
about the top end?  I suppose when trying to extract chroma we
typically don't care about components above 2 kHz?

  DAn.

bmcfee:
> I don't understand the issue.  "Early downsampling" is when you're not
> trying to get any information out of the top part of your spectrum, so
> you downsample to push the nyquist closer to the top of what you want
> to get before doing anything?  Sounds OK.  How often do we not care
> about the top end?  I suppose when trying to extract chroma we
> typically don't care about components above 2 kHz?

There are a few issues floating around, but the main thing is that `hybrid_cqt`'s scale is occasionally off by a factor of 2 from `cqt`, which I believe can be traced down to a combination of a) early downsampling in `cqt` (but not `pseudo_cqt`) and b) re-scaling of filters when hop-length is too large relative to the filter width.

I'm currently correcting for (a) by calculating the amount of early downsampling that `cqt` will do, and re-scaling the pseudo-cqt output so that the results are approximately continuous.  This seems like a brutal hack -- pseudo_cqt should be scaling things correctly in the first place.

WRT the "top end", this is a parameter of the algorithm, (`fmax = fmin * 2**(n_bins / bins_per_octave)`), so we know exactly when we don't care.  For cqt inversion, it's probably good if we can get the top end as correct as possible, but this doesn't really matter for hybrid/pseudo since those don't produce phase information anyway, and their entire purpose (as I understand it) is to capture high-frequency energy at long hop lengths.

dpwe:
> WRT the "top end", this is a parameter of the algorithm, (fmax = fmin *
> 2**(n_bins / bins_per_octave)), so we know exactly when we don't care.

.. with the caveat that this is the *center* of the top bin, but CQT
systems often have low Q and may have significant gain an octave above
that center frequency.

> For
> cqt inversion, it's probably good if we can get the top end as correct as
> possible, but this doesn't really matter for hybrid/pseudo since those don't
> produce phase information anyway, and their entire purpose (as I understand
> it) is to capture high-frequency energy at long hop lengths.

OK.  But the magnitude response to an impulse should not have ripples
across frequency, and, in a given channel, should vary smoothly, with
a single maximum as the impulse is delayed in time.  I'm not clear
where we are on that.

I'm still intrigued by the idea of an invertible CQT, at least one
that doesn't have variable frame rates for different octaves.  With a
fixed frame rate, I think it's rarely something you'd actually use.
Guess I should read the papers.

  DAn.

bmcfee:
> OK.  But the magnitude response to an impulse should not have ripples
> across frequency, and, in a given channel, should vary smoothly, with
> a single maximum as the impulse is delayed in time.  I'm not clear
> where we are on that.

That's fixed.  The problem was using over-sample and aggregate on complex responses, rather than directly sampling the complex responses.  Full CQT gets the smooth decay that you describe, hybrid-cqt stays flat.



> I'm still intrigued by the idea of an invertible CQT, at least one
> that doesn't have variable frame rates for different octaves.  With a
> fixed frame rate, I think it's rarely something you'd actually use.
> Guess I should read the papers.

Not sure how to parse that -- are you intrigued by variable or fixed frame rates?  The frame rate here isn't really variable; it's scaled in proportion to the recursive downsampling so that time indices line up properly.

bmcfee:
36ea61 fixes the second HCQT-vs-CQT scaling bug.  The culprit was [this line](https://github.com/bmcfee/librosa/pull/347/commits/36ea6148c1bb4372d8a85dfadfcaf70a1e35717e#diff-fe8c64d882784094d6447d1ca79159b1L533), which added an extraneous `n_fft` normalization to the filter construction.  This is now correctly compensated for by the implicit basis rescaling within `cqt`, as well as the explicit renormalization post-padding in `__fft_filters`.

As of this point, all of my internal tests are passing:

```python
for sr in [11025, 16384, 22050, 32768, 44100]:
    print('sr={}'.format(sr))
    for hop_scale in range(1, 33):
        tv = hvftest(sr, hop_scale)
        if np.abs(np.log2(tv)) >= 1:
            print('hop_scale={:2d}, scale error={:.3g}'.format(hop_scale, tv))
    print('-----')
```
yields
```
sr=11025
-----
sr=16384
-----
sr=22050
-----
sr=32768
-----
sr=44100
-----
```

This PR also implements spectral continuity and variance tests for both cqt and hybrid_cqt in response to frame-centered impulse signals.  There's still a small amount of discontinuity in the hybrid cqt spectrum at the full->pseudo boundary, with pseudo having about a 3% relative boost in amplitude (compard to full in the same frequency range). I'm not sure if this is down to windowing artifacts or what.

A consequence of this PR is that the overall scale of CQT outputs is shifted considerably.  I'm not sure if we should try to revert that for backwards compatibility (which sounds like a headache), or just call it a bug fix.  Any thoughts?

Either way, CR from @ebattenberg would be much appreciated.

bmcfee:
FWIW: here's a link to [inverse cqt](https://gist.github.com/bmcfee/3db6ce4321157a2bb479) tested out on the fixes implemented in this PR.  Sadly, gist doesn't host Audio widgets, but the output sounds good to me (modulo the obvious low-pass filter).

dpwe:
I dunno.  You're not revealing the problems in your CQTgram because of
the intrinsic down weighting of high-frequency.  Try pre-emphasis of
the plots:

freqs = librosa.cqt_frequencies(C.shape[0], fmin=fmin,
                            bins_per_octave=bins_per_octave)
plt.figure(figsize=(12, 6))
plt.subplot(1,2,1)
librosa.display.specshow(librosa.logamplitude(np.dot(np.diag(freqs**2),
C**2), ref_power=1),
                         bins_per_octave=int(12*over_sample),
                         y_axis='cqt_hz', cmap='magma')
plt.colorbar(orientation='horizontal')
plt.subplot(1,2,2)
librosa.display.specshow(librosa.logamplitude(np.dot(np.diag(freqs**2),
C2**2), ref_power=1),
                         bins_per_octave=int(12*over_sample),
                         y_axis='cqt_hz', cmap='magma')
plt.colorbar(orientation='horizontal')
plt.tight_layout()

.. then you see that the icqt isn't just stopping at some frequency,
it's got progressive and serious amplitude loss above 500 Hz.

Including bins above 72 (i.e. above 2 kHz) definitely sounds "crunchy"
to me, i.e. seriously distorted.

Try this:
# Resynthesize from just the top 2 octaves of the CQT
bins_per_octave = int(12 * over_sample)
fmin = librosa.note_to_hz('C1')
y2 = 80*icqt(np.dot(np.diag(np.arange(96) >= 72), C), sr=sr,
hop_length=librosa['hop_length'],
          bins_per_octave=bins_per_octave,
          filter_scale=res_factor,
          fmin=fmin)
trange = np.arange(11*sr, 11*sr + 2000)
plt.plot(np.ascontiguousarray(y2)[trange], alpha=0.5,
label='Reconstructed', color='g')

.. then the period-128 discontinuous time windowing is pretty
apparent.  That's the crunchy, I guess - you can hear the amplitude
modulation at the frame rate when the frame rate becomes too long for
the reconstruction "grains".

  DAn.

bmcfee:
> You're not revealing the problems in your CQTgram because of
> the intrinsic down weighting of high-frequency.

Can we move this discussion to the #165 thread?  I'd rather keep this one focused on hcqt /continuity bug-fixing.  I just threw in the link to icqt because you had mentioned it earlier.

dpwe:
ha ha.  I'm just trying to exercise your coping strategies for
disruptive OS collaborators.

bmcfee:
open source development

bmcfee:
I added a couple of more cleanups:

- Set `real=None` by default, which triggers the warning and then acts as `real=True`.  This way, you can call with `real=True` to disable the warning.
- Removed all assertions.  Assertions are bad practice in live code because they don't provide a useful exception type.  The assertions we had were already covered by exceptions anyway.  The one exception (yuk yuk) to that being the second nyquist check, which is implicit if the `audio.BW_FASTEST > 0.5` anyway (and it always is).

Still waiting on CR for this, if anyone has the time.  If not, I feel pretty good about merging anyway.

librosa/core/constantq.py
```diff
@@ -62,9 +63,9 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
```
dpwe: L168/L172: Should index with -n_filters: instead of -bins_per_octave: ?

bmcfee: No: we're slicing off the frequencies of the top octave.

dpwe: But if n_bins < bins_per_octave, does cqt_frequencies still return a whole octave?

bmcfee: In that case, numpy array indexing will do the right thing.  For instance, say we have 36 bpo and n_bins = 12:

```python
In [2]: x = np.arange(12)

In [3]: x[-12:]
Out[3]: array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11])

In [4]: x[-36:]
Out[4]: array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11])

```

dpwe: I agree it works as written, my point was readability.  I guess I'm still reeling from wondering what n_filters is going to do (the plot thickens...)

bmcfee: I'm not sure how readability could be improved for that particular op; it's already commented and correctly does what it claims.  I'm inclined to leave it as is.

librosa/core/constantq.py
```diff
@@ -172,46 +176,32 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
...
-    elif filter_cutoff < audio.BW_BEST * nyquist:
-        res_type = 'kaiser_best'
     else:
         res_type = 'kaiser_best'
 
-    cqt_resp = []
-
-    len_orig = len(y)
-
     y, sr, hop_length = __early_downsample(y, sr, hop_length,
```
dpwe: AFAICT _early_downsample is a no-op for kaiser_best.  The whole kaiser_fast/kaiser_best logic is making my head hurt.  I think I get it now - if the top octave is close to nyquist, we need to be a little more precise in downsampling.  But the way it all folds together now seems convoluted.  It would be easier to understand if I could quickly tell the "normal case", then see the "kaiser_best" exception as its own block.

librosa/core/constantq.py
```diff
@@ -62,9 +63,9 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         Filter scale factor. Small values (<1) use shorter windows
         for improved time resolution.
 
-    aggregate : None or function
-        Aggregation function for time-oversampling energy aggregation.
-        By default, `np.mean`.  See `librosa.util.sync`.
+    aggregate : None or function [DEPRECATED]
```
dpwe: So aggregate does nothing now.  Should we even support it?  Won't code that references it be broken anyway?

librosa/core/constantq.py
dpwe: So instead of 181 et seq., we test if the upper edge is above some factor relative to our threshold, and if so do some special-case kaiser_best-type stuff, but for the common case we fall through and proceed as usual.

dpwe: And we don't have res_type at all, just explicit 'kaiser_best' in the special case, 'kaiser_fast' in the common case.

librosa/core/constantq.py
```diff
@@ -172,46 +176,32 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
...
 
-        # Compute a dynamic hop based on n_fft
-        my_cqt = __variable_hop_response(y, n_fft,
-                                         hop_length,
-                                         min_filter_length,
-                                         fft_basis,
-                                         aggregate)
+        my_cqt = __fft_response(y, n_fft, hop_length, fft_basis)
 
         # Convolve
```
dpwe: You say # Convolve in two places, but both times the convolution actually happened on the preceding line, and doesn't need a comment anyway.

librosa/core/constantq.py
```diff
@@ -259,15 +246,10 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                                res_type=res_type,
                                                scale=True)
             my_sr /= 2.0
-            assert my_hop % 2 == 0
             my_hop //= 2
 
         # Compute a dynamic hop based on n_fft
```
dpwe: And this comment is now wrong.  If you don't think __fft_response is self-explanatory, you should give the function a better name.  No need to comment.

dpwe: And this comment is now wrong.  If you don't think __fft_response is self-explanatory, you should give the function a better name.  No need to comment.

librosa/core/constantq.py
```diff
@@ -369,9 +351,12 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                          filter_scale=filter_scale)
 
     # Determine which filters to use with Pseudo CQT
-    pseudo_filters = lengths < 2*hop_length
+    # These are the ones that fit within 2 hop lengths after padding
+    pseudo_filters = 2.0**np.ceil(np.log2(lengths)) < 2 * hop_length
```
dpwe: I think I preferred the old test. What matters is the actual window, not the length of the padded carrier, no?

librosa/core/constantq.py
```diff
@@ -522,17 +503,17 @@ def __fft_filters(sr, fmin, n_bins, bins_per_octave, tuning,
     # Filters are padded up to the nearest integral power of 2
     n_fft = basis.shape[1]
 
-    if hop_length is not None and n_fft < 2 * hop_length:
-        n_fft = int(2.0 ** (np.ceil(np.log2(2 * hop_length))))
+    if hop_length is not None and n_fft < 2.0**(1 + np.ceil(np.log2(hop_length))):
+        n_fft = int(2.0 ** (1 + np.ceil(np.log2(hop_length))))
 
     # normalize by inverse length to compensate for phase invariance
```
dpwe: mysterious comment, mysterious scaling.

librosa/core/constantq.py
bmcfee: Good catch, will fix.

librosa/core/constantq.py
bmcfee: The old test was missing filters in some edge cases, leading to strong discontinuities.

I think what matters is how __fft_response is going to behave when provided that set of filters as input; in that sense, the two tests should match up.

librosa/core/constantq.py
bmcfee: agreed.

librosa/core/constantq.py
bmcfee: I agree that it's a little contorted, but I think that it's an orthogonal issue to those addressed in this PR.  I'd vote for punting a refactor of this to a later date.

dpwe: I agree it's orthogonal to this PR.  I thought maybe the call to
early_resample had become more meaningless under 'kaiser_fast', but I
see it was always that way.  I'll file an issue ;-)

  DAn.

librosa/core/constantq.py
dpwe: !fixed

librosa/core/constantq.py
dpwe: !resolved

librosa/core/constantq.py
dpwe: Hmm, if I were doing this for money, I'd insist on seeing the actual edge case because I can't figure it out just by staring at it. !resolved

PR #354, opened 2016-05-13
Tempo frequencies

bmcfee:
This PR adds a new frequency function `tempo_frequencies` which makes it easy to convert between lag (the index on autocorrelation and tempograms) and tempo (in BPM).

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/354)

### DIFF
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -238,17 +238,16 @@ def estimate_tempo(onset_envelope, sr=22050, hop_length=512, start_bpm=120,
     >>> # Compute 2-second windowed autocorrelation
     >>> hop_length = 512
     >>> ac = librosa.autocorrelate(onset_env, 2 * sr // hop_length)
-    >>> # Convert tempo estimate from bpm to frames
-    >>> tempo_frames = (60 * sr / hop_length) / tempo
-    >>> plt.plot(librosa.util.normalize(ac),
-    ...          label='Onset autocorrelation')
-    >>> plt.vlines([tempo_frames], 0, 1,
-    ...            color='r', alpha=0.75, linestyle='--',
+    >>> freqs = librosa.tempo_frequencies(len(ac), sr=sr,
+    ...                                   hop_length=hop_length)
+    >>> # Plot on a BPM axis.  We skip the first (0-lag) bin.
+    >>> plt.semilogx(freqs[1:], librosa.util.normalize(ac)[1:],
+    ...              label='Onset autocorrelation', basex=2)
+    >>> plt.axvline(tempo, 0, 1, color='r', alpha=0.75, linestyle='--',
     ...            label='Tempo: {:.2f} BPM'.format(tempo))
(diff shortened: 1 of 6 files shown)


bmcfee:
@dpwe tiny CR request because I think I found a tiny error in the beat tracker/tempo estimator that I'd like your feedback on.

In porting the tempo estimator to use the new `tempo_frequencies` function, I noticed [this line](https://github.com/librosa/librosa/commit/b533e2a0e6bc3e265ad84432822756d6fccbe8c2#diff-5debdd1232d2437f68f9b209ede89065L273) where the tempo associated with each lag bin `t` in the autocorrelation is computed as `60 * sr / (hop_length * (1 + t))`.  This seems incorrect, since `lag=0` should correspond to 0 period and infinite frequency.  It looks like [tempo2.m](https://github.com/dpwe/alignmidi/blob/master/beat2.m#L408) avoids this by adding `0.1` .. I'm not sure where exactly librosa's implementation diverged from that, but the fact that they agree numerically suggests that it doesn't matter too much.

Changing this has little noticeable impact on the regression tests, since the difference is going to be generally tiny for values we're likely to encounter.  However, it does just so happen to hit the sweet spot of changing the estimated tempo of the example track.  This doesn't worry me too much, as I ran it through another dozen examples from SMC and CAL500 and the results sound good.

dpwe:
Yes, I agree with your analysis.

My only defense for 0.1 (which looks really weird) is that the bpm
vector was only used to provide the basis for the lognormal tempo
prior window.  It seems like in librosa it was promoted to also
providing the actual BPM value return.

BTW, I'd change the comment on
https://github.com/librosa/librosa/blob/b533e2a0e6bc3e265ad84432822756d6fccbe8c2/librosa/beat.py#L275
to read "Weight the autocorrelation by a log-normal distribution."
maybe even adding "to incorporate a prior distribution of perceived
BPMs." - but maybe that's too much interpretation, since in practice
the values are used not specifically to match human perception, but
just to tune to whatever evaluation task is being employed.

  DAn.

bmcfee:
Wow, thanks for the quick response!

> It seems like in librosa it was promoted to also
> providing the actual BPM value return.

It previously wasn't, but I made that switch in this PR because it's equivalent, and it eliminates another instance of magic numbers.



> to read "Weight the autocorrelation by a log-normal distribution."
> maybe even adding "to incorporate a prior distribution of perceived
> BPMs."

I like the former better; the latter has bayesian connotations that could be easily misinterpreted as doing proper MAP inference.  (Or maybe I'm just being overly paranoid..)  I'll make that change and merge.

dpwe:
You're right.  I actually conceived of this weighting entirely in that
Bayesian perspective, but that's too much luggage.

  DAn.

bmcfee:
Reviewed 5 of 6 files at r1, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/354)*

PR #372, opened 2016-06-09
Removed util.SMALL_FLOAT

### LINKED ISSUE #358: Replace SMALL_FLOAT with numpy finfo calls
The SMALL_FLOAT constant is somewhat arbitrary, and probably overly conservative when it comes to different floating point precisions.

Instead, we should replace use of this value with `np.finfo(DTYPE).tiny`, where `DTYPE` is the type of the data in question (eg, `float32` or `float64`, or in general, `X.dtype`).

bmcfee:
And replaced by calling `np.finfo(dtype).tiny`, which is a more precise and less arbitrary way to accomplish what we need from the `SMALL_FLOAT` constant.

fixes #358

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/372)

### DIFF
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -247,8 +247,7 @@ def piptrack(y=None, sr=22050, S=None, n_fft=2048, hop_length=None,
 
     # Suppress divide-by-zeros.
     # Points where shift == 0 will never be selected by localmax anyway
-    tiny = np.finfo(shift.dtype).tiny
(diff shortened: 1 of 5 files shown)


bmcfee:
Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/372)*

Issue #399, opened 2016-08-06
Expand window bandwidth dictionary


bmcfee:
Right now, we only have Hann window in there, but we could add all of the scipy window functions.

Side note: the value currently recorded is half width half max. Would it be clearer to change this to be full width half max, and modify cqt accordingly? (Thanks to @lostanlen for pointing this out)

lostanlen:
Actually, I'm starting to think that for energetic interpretation, the most appropriate would probably be equivalent rectangular bandwdith (called equivalent noise bandwidth in the Harris 1976 paper) instead of bandwidth at half-maximum.

bmcfee:
That seems reasonable enough.  Would that lower the cutoff frequency for CQT compared to the current implementation?

lostanlen:
I'm not sure of what you mean. The ERB is ~5% larger than the FWHM for most windows. So the CQT filters will be slightly more localized in the Fourier domain than the curent implementation.

bmcfee:
The `window_bandwidth` function isn't used in basis construction; only to check that the resulting filters don't get too close to nyquist.  If that bandwidth is larger, then there's a higher chance of triggering the nyquist exception; this means that the effective upper bound on frequencies will be lower than if we use FWHM.  This would break backwards compatibility -- I'm okay with that, but I want to understand exactly how much of a change it would introduce.

My question is: how much difference is there here for Hann?

lostanlen:
OK, thanks for clarifying. The bandwidth is 2.5% larger for Hann. So the cutoff frequency is 2.5% lower. That's about 43 cents of a semitone.

lostanlen:
Regarding asymmetric triangular filters, which are used in mel-spectrograms, we could define them with a parameter r = the ratio of the right half-bandwidth to the left half-bandwidth.
What is nice is that the ERB of such a window is equal to 4/3 independently from r.
In the context of a constant-Q filter bank, r = 2^(1/Q)

bmcfee:
Circling back on this after spending some time this morning implementing the calculation for generic filters.

I think I'm on board with switching over to using ENBW for this.  Aside from the energy interpretation that @lostanlen mentioned above, it has the added benefit of being much simpler to compute.  Computational complexity hasn't mattered much in the past for this, since we only had a single hard-coded value (hann).

[...]

PR #400, opened 2016-08-07
Custom SciPy window, window length, and n_fft in STFT

### LINKED ISSUE #397: Support tuple window parameters in window places (stft, etc)
`scipy.signal.get_window` lets you specify parametric windows via tuples: `('kaiser', 4.0)`.

This would be easy to support.

lostanlen:
* I've defined `librosa.filters.get_window`. Right now it falls back to `scipy.signal.get_window`. We could add a try/catch statement in the future for non-SciPy windows in the future
* Now `win_length` and `n_fft` are defaults of each other. If neither of them is provided, falls back to 2048.
* no longer fails when user-defined `n_fft` is below the length of a
user-defined window (by function or array)
* still fails when user-defined `n_fft` is above that length
* calls `librosa.filters.get_window` when window is a tuple of arguments.
win_length is extracted as `window[1]` (second element in the tuple)

Now these work
```
X = librosa.stft(y, window=('kaiser', 4.0)) # n_fft = 2048

X = librosa.stft(y, window='boxcar') # n_fft = 2048
X = librosa.stft(y, window='boxcar', win_length = 1024) # n_fft = 1024
X = librosa.stft(y, window='boxcar', n_fft = 1024) # win_length = 1024

w = scipy.signal.triang(1024)
X = librosa.stft(y, window=w) # n_fft = 1024
```

Fixes #397 in what regards STFT.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/400)

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -127,23 +127,19 @@ def stft(y, n_fft=None, hop_length=None, win_length=None, window=None,
     >>> plt.tight_layout()
 
     """
+    # By default, the window length (also called frame length) is 2048 samples
+    if win_length is None:
+        win_length = 2048
+
     if window is None:
         # Hann window is the default
-        if win_length is None:
-            win_length = 2048
         fft_window = scipy.signal.hann(win_length, sym=False)
     if six.callable(window):
         # Case where window is a user-defined function of win_length
-        if win_length is None:
-            win_length = 2048
         fft_window = window(win_length)
     elif isinstance(window, tuple):
         # Case where window is a tuple of arguments to be passed to get_window
-        if win_length is None:
-            win_length = window[1]
-        elif win_length != window[1]:
-            raise ParameterError('win_length must be equal to window[1]')
-        fft_window = filters.get_window(window)
+        fft_window = filters.get_window(window, win_length)
     else:
         # Case where window is an array
         fft_window = np.asarray(window)
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -148,11 +148,16 @@ def stft(y, n_fft=None, hop_length=None, win_length=None, window=None,
         fft_window = np.asarray(window)
         win_length = fft_window.size
 
+    # By default, the FFT length is equal to the window length
     if n_fft is None:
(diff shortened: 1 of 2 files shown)


lostanlen:
~1 zillion tests are failing.

Notably, having `n_fft` greater than the length of a user-provided `window` vector no longer raises a `ParameterError`.

```
@raises(librosa.ParameterError)
def test_stft_bad_window():
    y = np.zeros(22050 * 5)
    n_fft = 2048
    window = np.ones(n_fft // 2)
    librosa.stft(y, n_fft=n_fft, window=window)
```

I can

bmcfee:
> Notably, having n_fft greater than the length of a user-provided window vector no longer raises a ParameterError

I think that it should, right?  Or do you think we should pad `window` out to `n_fft` in that case?

lostanlen:
I would expect

```
X = librosa.stft(y, window=scipy.signal.triang(1024), n_fft=2048) 
X = librosa.stft(y, window='triang', win_length=1024, n_fft= 2048)
```

to be equivalent. What are the cons ?

bmcfee:
@lostanlen Yup, agreed.  I'm okay with relaxing the test then.  We still want `n_fft >= len(window)`  though, since clipping the window would cause problems.

lostanlen:
The rest of the errors are caused in CQT tests. They're all `n_fft must be larger than win_length` failures.
It's not clear to me whether it's my STFT code, the CQT, the function yielding n_fft in constant_q.py, or the tests that should be changed.

bmcfee:
It looks like you're also getting a bunch of type errors checking for `basestring`; maybe we can fix those first and then figure out how to adjust the remaining testS?

bmcfee:
I'm looking back over this one, in light of #399.

Could we back out the API change for n_fft/win_length defaults? This seems to break a lot of things, and as far as I can tell, is independent of the `get_window` implementation and support for tuple-windows in STFT.

lostanlen:
Yes, I think we could simplify the strategy here. This PR addresses too many problems at once.
I suggest we close this and work from scratch on a new PR that could solve #399.

bmcfee:
make it so

PR #413, opened 2016-09-30
Expanding window options for stft and friends

bmcfee:
Toward implementing #399.

This PR adds the following:

- `filters.get_window` -- a wrapper for `scipy.signal.get_window` that can handle generic callables, ndarrays, and any additional window functions we want to add (eg asymmetric triangles)
- `stft, istft, ifgram, tempogram` support for tuple windows via the above functionality
- removed `window=None` support from stft &co, since string identifiers are now valid. This breaks API.
- all `get_window` calls from stft-like ops are now explicitly `fftbins=True` (`sym=False`).  This breaks previous behavior, but is the right thing to do.  Symmetric windows can still be computed explicitly and passed as window vectors, so there's no loss in functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/413)

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -23,7 +23,7 @@ __all__ = ['stft', 'istft', 'magphase',
 
 
 @cache(level=20)
-def stft(y, n_fft=2048, hop_length=None, win_length=None, window=None,
+def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
          center=True, dtype=np.complex64):
     """Short-time Fourier transform (STFT)
 
@@ -135,9 +135,6 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window=None,
     if hop_length is None:
         hop_length = int(win_length // 4)
 
-    if window is None:
-        window = 'hann'
-
     fft_window = get_window(window, win_length, fftbins=True)
 
     # Pad the window out to n_fft size
@@ -175,7 +172,7 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window=None,
 
 
 @cache(level=30)
-def istft(stft_matrix, hop_length=None, win_length=None, window=None,
+def istft(stft_matrix, hop_length=None, win_length=None, window='hann',
           center=True, dtype=np.float32):
     """
     Inverse short-time Fourier transform (ISTFT).
@@ -263,10 +260,6 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
     if hop_length is None:
         hop_length = int(win_length // 4)
 
-    if window is None:
-        # Default is an asymmetric Hann window.
-        window = 'hann'
-
     ifft_window = get_window(window, win_length, fftbins=True)
 
     # Pad out to match n_fft
@@ -298,8 +291,8 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
 
 
 def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
-           norm=False, center=True, ref_power=1e-6, clip=True,
-           dtype=np.complex64):
+           window='hann', norm=False, center=True, ref_power=1e-6,
+           clip=True, dtype=np.complex64):
     '''Compute the instantaneous frequency (as a proportion of the sampling rate)
     obtained as the time-derivative of the phase of the complex spectrum as
     described by [1]_.
@@ -329,6 +322,10 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
 
     win_length : int > 0, <= n_fft
         Window length. Defaults to `n_fft`.
+        See `stft` for details.
+
+    window : string, tuple, function, or np.ndarray
+        Window function to use in ifgram calculation.
         See `stft` for details.
 
     norm : bool
@@ -387,8 +384,8 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
         hop_length = int(win_length // 4)
 
     # Construct a padded hann window
-    window = util.pad_center(get_window('hann', win_length, fftbins=True),
-                             n_fft)
+    win = util.pad_center(get_window(window, win_length, fftbins=True),
+                          n_fft)
 
     # Window for discrete differentiation
     freq_angular = np.linspace(0, 2 * np.pi, n_fft, endpoint=False)
@@ -419,7 +416,7 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
     if_gram = freq_angular[:n_fft//2 + 1] + bin_offset
 
     if norm:
-        stft_matrix = stft_matrix * 2.0 / window.sum()
+        stft_matrix = stft_matrix * 2.0 / win.sum()
 
     if clip:
         np.clip(if_gram, 0, np.pi, out=if_gram)
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -13,6 +13,7 @@ from . import time_frequency
 from .. import cache
 from .. import util
 from ..util.exceptions import ParameterError
+from ..filters import get_window, hann_asym
 
 __all__ = ['stft', 'istft', 'magphase',
            'ifgram',
@@ -70,12 +71,6 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window=None,
     -------
     D : np.ndarray [shape=(1 + n_fft/2, t), dtype=dtype]
         STFT matrix
-
-
-    Raises
-    ------
-    ParameterError
-        If `window` is supplied as a vector of length `n_fft`.
 
 
     See Also
@@ -138,24 +133,12 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window=None,
 
     # Set the default hop, if it's not already specified
     if hop_length is None:
-        hop_length = int(win_length / 4)
+        hop_length = int(win_length // 4)
 
     if window is None:
-        # Default is an asymmetric Hann window
-        fft_window = scipy.signal.hann(win_length, sym=False)
+        window = hann_asym
 
-    elif six.callable(window):
-        # User supplied a window function
-        fft_window = window(win_length)
-
-    else:
-        # User supplied a window vector.
-        # Make sure it's an array:
-        fft_window = np.asarray(window)
-
-        # validate length compatibility
-        if fft_window.size != n_fft:
-            raise ParameterError('Size mismatch between n_fft and len(window)')
+    fft_window = get_window(window, win_length)
 
     # Pad the window out to n_fft size
     fft_window = util.pad_center(fft_window, n_fft)
@@ -242,11 +225,6 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
     y : np.ndarray [shape=(n,)]
         time domain signal reconstructed from `stft_matrix`
 
-    Raises
-    ------
-    ParameterError
-        If `window` is supplied as a vector of length `n_fft`
-
     See Also
     --------
     stft : Short-time Fourier Transform
@@ -283,25 +261,13 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
 
     # Set the default hop, if it's not already specified
     if hop_length is None:
-        hop_length = int(win_length / 4)
+        hop_length = int(win_length // 4)
 
     if window is None:
         # Default is an asymmetric Hann window.
-        ifft_window = scipy.signal.hann(win_length, sym=False)
+        window = hann_asym
 
-    elif six.callable(window):
-        # User supplied a windowing function
-        ifft_window = window(win_length)
-
-    else:
-        # User supplied a window vector.
-        # Make it into an array
-        ifft_window = np.asarray(window)
-
-        # Verify that the shape matches
-        if ifft_window.size != n_fft:
-            raise ParameterError('Size mismatch between n_fft and window size')
-
+    ifft_window = get_window(window, win_length)
     # Pad out to match n_fft
     ifft_window = util.pad_center(ifft_window, n_fft)
 
@@ -331,7 +297,8 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
 
 
 def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
-           norm=False, center=True, ref_power=1e-6, clip=True, dtype=np.complex64):
+           norm=False, center=True, ref_power=1e-6, clip=True,
+           dtype=np.complex64):
     '''Compute the instantaneous frequency (as a proportion of the sampling rate)
     obtained as the time-derivative of the phase of the complex spectrum as
     described by [1]_.
@@ -419,7 +386,7 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
         hop_length = int(win_length // 4)
 
     # Construct a padded hann window
-    window = util.pad_center(scipy.signal.hann(win_length, sym=False), n_fft)
+    window = util.pad_center(hann_asym(win_length), n_fft)
 
     # Window for discrete differentiation
     freq_angular = np.linspace(0, 2 * np.pi, n_fft, endpoint=False)
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -13,7 +13,7 @@ from . import time_frequency
 from .. import cache
 from .. import util
 from ..util.exceptions import ParameterError
-from ..filters import get_window, hann_asym
+from ..filters import get_window
 
 __all__ = ['stft', 'istft', 'magphase',
            'ifgram',
@@ -136,9 +136,9 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window=None,
         hop_length = int(win_length // 4)
 
     if window is None:
-        window = hann_asym
+        window = 'hann'
 
-    fft_window = get_window(window, win_length)
+    fft_window = get_window(window, win_length, fftbins=True)
 
     # Pad the window out to n_fft size
     fft_window = util.pad_center(fft_window, n_fft)
@@ -265,9 +265,10 @@ def istft(stft_matrix, hop_length=None, win_length=None, window=None,
 
     if window is None:
         # Default is an asymmetric Hann window.
-        window = hann_asym
+        window = 'hann'
 
-    ifft_window = get_window(window, win_length)
+    ifft_window = get_window(window, win_length, fftbins=True)
+
     # Pad out to match n_fft
     ifft_window = util.pad_center(ifft_window, n_fft)
 
@@ -386,7 +387,8 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
         hop_length = int(win_length // 4)
 
     # Construct a padded hann window
-    window = util.pad_center(hann_asym(win_length), n_fft)
+    window = util.pad_center(get_window('hann', win_length, fftbins=True),
+                             n_fft)
 
     # Window for discrete differentiation
     freq_angular = np.linspace(0, 2 * np.pi, n_fft, endpoint=False)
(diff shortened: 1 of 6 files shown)


bmcfee:
Tagging @lostanlen -- this PR implements tuple-style windows in all stft-like ops, and cleans up the code a little.

In doing so, I noticed that calling stft with something like `window=scipy.signal.hann` would actually not preserve the `sym=` parameter, which I think should always be `False` here.  Using `get_window(..., fftbins=True)` implies that `sym=False` for all window construction calls.

This means that behavior has changed slightly from the 0.4.x series, since
```python
D = librosa.stft(y, window=scipy.signal.hann)
```
now behaves differently.  The old behavior can be recovered explicitly by
```python
win = scipy.signal.hann(2048, sym=True)
D = librosa.stft(y, window=win)
```
if one really wants to.

I think this is all correct, but an extra set of eyes on it would be appreciated.

lostanlen:
Pardon my ignorance, but I don't follow the reflection on the keywords `fftbins=True` and `sym=False`. What is meant by asymmetric Hann window ? One null value at the beginning but not at the end ? Does it bring a subtle half-sample offset to the frequencies, so that there is no zeroth frequency, like in the type-II discrete cosine transform ? Or is it even subtler than that ? It would be good if we linked to some textbook or Wikipedia page, because the SciPy docs are not self-contained.
Aside from that, the code looks good to me.

dpwe:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[librosa/filters.py, line 778 at r1](https://reviewable.io:443/reviews/librosa/librosa/413#-KT50EeyKnVrBd1J8C0n:-KT50EeyKnVrBd1J8C0o:barvfwo) ([raw file](https://github.com/librosa/librosa/blob/e21160bd856315fb857b37094065c9dffec631f8/librosa/filters.py#L778)):*
> ```Python
>         The length of the window
> 
>     fftbins : bool, optional
> ```

This is a terrible parameter name.  I understand we inherit it from scipy.signal.get_window, but still... 

---

*[librosa/filters.py, line 780 at r1](https://reviewable.io:443/reviews/librosa/librosa/413#-KT5137Q5I4PQv11eYM7:-KT5137Q5I4PQv11eYM8:bdkn6mk) ([raw file](https://github.com/librosa/librosa/blob/e21160bd856315fb857b37094065c9dffec631f8/librosa/filters.py#L780)):*
> ```Python
>     fftbins : bool, optional
>         If True (default), create a periodic window for use with FFT
>         If False, create a symmetric window for use with filter design
> ```

I infer that this wording comes from the justification for "sym=True" in, e.g., spicy.signal.hann, but it seems confusing here... As far as I know, the only common case where you'd use a "symmetric" window for filter design is odd-length windows, which basically means you're appending the 0th point from the corresponding N-1 even-length "periodic" window.  Secondly, I don't think anyone is going to use librosa's get_window for windowed filter design.  Isn't the librosa use case exclusively for "spectral analysis" (the contrast case in scipy.signal.hann)?  

I can't think of any circumstance in which librosa users actually want an even-length, symmetric window, and I suspect there aren't cases in which they want an odd-length window.  I would support *not* providing this option, and hard-coding fftbins=True in the call to scipy.signal.get_window.  

As you note, if this is not what people want, they can always calculate their window externally and pass it in.

---

*[librosa/filters.py, line 801 at r1](https://reviewable.io:443/reviews/librosa/librosa/413#-KT52tnaiQUDKo6_-HmA:-KT52tnaiQUDKo6_-HmB:b-39h14k) ([raw file](https://github.com/librosa/librosa/blob/e21160bd856315fb857b37094065c9dffec631f8/librosa/filters.py#L801)):*
> ```Python
> 
>     elif isinstance(window, (str, tuple)):
>         # TODO: if we add more window functions, put that index check here
> ```

which index check?  Don't understand this comment.

---

*[librosa/feature/rhythm.py, line 52 at r2](https://reviewable.io:443/reviews/librosa/librosa/413#-KT53IOpnuS6cTqQ7r_Z:-KT53IOpnuS6cTqQ7r__:bd57v4i) ([raw file](https://github.com/librosa/librosa/blob/e25c7e0b318cb12c3b3b4d0ef24be20742bc757f/librosa/feature/rhythm.py#L52)):*
> ```Python
> 
>     window : string, function, tuple, or np.ndarray [shape=(win_length,)]
>         Window function to apply to onset strength function.
> ```

"Window specification"... "See librosa.xx.get_window..."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

bmcfee:
> What is meant by asymmetric Hann window ? One null value at the beginning but not at the end ? Does it bring a subtle half-sample offset to the frequencies, so that there is no zeroth frequency, like in the type-II discrete cosine transform ?

Digging into `scipy.signal.hann` (the same logic exists in other window functions):

```python
 if M < 1:
        return np.array([])
    if M == 1:
        return np.ones(1, 'd')
    odd = M % 2
    if not sym and not odd:
        M = M + 1
    n = np.arange(0, M)
    w = 0.5 - 0.5 * np.cos(2.0 * np.pi * n / (M - 1))
    if not sym and not odd:
        w = w[:-1]
    return w
```

- It only affects even-length windows (which we use by default)
- `sym=False` pads up to an odd-length window on the end, and then clips the last sample out.
- `sym=True`, if I understand correctly, would be equivalent to sub-sample interpolation on an even window so that it's properly "centered".

According to the documentation, `sym=False` is intended to guarantee a periodic window, which is what we want in stft-like applications.

dpwe:
Haha I sniped the review.

I was also super confused by "fftbins=True".  I'm almost tempted to try to
dig through the scipy.signal discussion threads (assuming they are
accessible) to see who is guilty of this, but I think we can sidestep it in
librosa by simply always using "periodic" window for even length windows.
If there are any librosa use cases that call for odd-length windows, we
could make fftbins=(window_length % 2) == 0 or something.

  DAn.

On Sun, Oct 2, 2016 at 11:42 AM, Vincent Lostanlen <notifications@github.com

bmcfee:
> but I think we can sidestep it in
> librosa by simply always using "periodic" window for even length windows.

+:100: -- I think that's what I'm proposing here.  Inside `scipy.signal.get_window`, the first step is `sym = not fftbins`, which I agree is confusing.



> If there are any librosa use cases that call for odd-length windows, we
> could make fftbins=(window_length % 2) == 0 or something.

I'm not even sure that's necessary?  `sym` logic only kicks in when the window is even length anyway.  I think it all shakes out as you would expect.

The only weird case is if you want symmetric windowing on an even-length window, which we kind of did by mistake previously if the user supplied a window function.  You can force this, if you really want to, but constructing the window yourself as I describe above.

dpwe:
I didn't look at the logic.  Haha, not sym and not odd a/k/a using boolean
identities to maximize the chances of confusing programmers.

not (sym or odd)

Basically, I think we assume win_length is always even, and fftbins=True as
some mysterious invocation that code readers don't have to think about.

  DAn.

bmcfee:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[librosa/filters.py, line 780 at r1](https://reviewable.io:443/reviews/librosa/librosa/413#-KT5137Q5I4PQv11eYM7:-KT58Hoa8fbHAP7OwrjP:b-tmbuhg) ([raw file](https://github.com/librosa/librosa/blob/e21160bd856315fb857b37094065c9dffec631f8/librosa/filters.py#L780)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> I infer that this wording comes from the justification for "sym=True" in, e.g., spicy.signal.hann, but it seems confusing here... As far as I know, the only common case where you'd use a "symmetric" window for filter design is odd-length windows, which basically means you're appending the 0th point from the corresponding N-1 even-length "periodic" window.  Secondly, I don't think anyone is going to use librosa's get_window for windowed filter design.  Isn't the librosa use case exclusively for "spectral analysis" (the contrast case in scipy.signal.hann)?  
> 
> I can't think of any circumstance in which librosa users actually want an even-length, symmetric window, and I suspect there aren't cases in which they want an odd-length window.  I would support *not* providing this option, and hard-coding fftbins=True in the call to scipy.signal.get_window.  
> 
> As you note, if this is not what people want, they can always calculate their window externally and pass it in.
</details>
I agree that there's probably no need for forced symmetric windowing in librosa, but I'd rather not limit ourselves if we don't have to.  Right now, we expose a superset of scipy.signal.get_window's API, and I'd prefer to keep it all transparent if possible so that you can use ours as a drop-in replacement.

99% of librosa users will never need to call this function explicitly though, so we can force `fftbins=True` internally and have everything work as it should.

Documentation could for sure be improved though.

---

*[librosa/filters.py, line 801 at r1](https://reviewable.io:443/reviews/librosa/librosa/413#-KT52tnaiQUDKo6_-HmA:-KT57dWx7bqlWVHMC5ww:b-oer227) ([raw file](https://github.com/librosa/librosa/blob/e21160bd856315fb857b37094065c9dffec631f8/librosa/filters.py#L801)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> which index check?  Don't understand this comment.
</details>
There's a separate thread with @lostanlen about implementing some non-standard window functions (eg asymmetric triangles).

If we want those windows to work with CQT, we'll need to index them in the bandwidth table by string identifiers.  As long as we're already indexing them by string ids, we may as well expose that interface to the caller, so you can say `stft(y, window='asym_triang')` and have everything work.

This TODO is a placeholder for when that functionality is added.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

bmcfee:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[librosa/feature/rhythm.py, line 52 at r2](https://reviewable.io:443/reviews/librosa/librosa/413#-KT53IOpnuS6cTqQ7r_Z:-KT5KyTng_CaPVSYm3LT:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/e25c7e0b318cb12c3b3b4d0ef24be20742bc757f/librosa/feature/rhythm.py#L52)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> "Window specification"... "See librosa.xx.get_window..."
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

bmcfee:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[librosa/filters.py, line 778 at r1](https://reviewable.io:443/reviews/librosa/librosa/413#-KT50EeyKnVrBd1J8C0n:-KT66skjSWvNTQKqV3vj:b-pfuemx) ([raw file](https://github.com/librosa/librosa/blob/e21160bd856315fb857b37094065c9dffec631f8/librosa/filters.py#L778)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> This is a terrible parameter name.  I understand we inherit it from scipy.signal.get_window, but still... 
</details>
Agreed, but I want to maintain API compatibility with scipy.signal's implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

dpwe:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[librosa/filters.py, line 780 at r1](https://reviewable.io:443/reviews/librosa/librosa/413#-KT5137Q5I4PQv11eYM7:-KT70Ymihdytub_UjFWJ:b2lipoa) ([raw file](https://github.com/librosa/librosa/blob/e21160bd856315fb857b37094065c9dffec631f8/librosa/filters.py#L780)):*
> I agree that there's probably no need for forced symmetric windowing in librosa, but I'd rather not limit ourselves if we don't have to.  Right now, we expose a superset of scipy.signal.get_window's API, and I'd prefer to keep it all transparent if possible so that you can use ours as a drop-in replacement.
> 
> 99% of librosa users will never need to call this function explicitly though, so we can force `fftbins=True` internally and have everything work as it should.
> 
> Documentation could for sure be improved though.
</details>
Ah, I see.  Preserving all the scipy.signal.get_window args makes sense; maybe explain that's what it is in the docstring?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413#-:-KT71V67yDv4X5-1I6DJ:bnfp4nl)*

bmcfee:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[librosa/filters.py, line 780 at r1](https://reviewable.io:443/reviews/librosa/librosa/413#-KT5137Q5I4PQv11eYM7:-KT797vxW43bOJLETVGx:bvslpl2) ([raw file](https://github.com/librosa/librosa/blob/e21160bd856315fb857b37094065c9dffec631f8/librosa/filters.py#L780)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> Ah, I see.  Preserving all the scipy.signal.get_window args makes sense; maybe explain that's what it is in the docstring?
</details>
I added a note at the beginning of the docstring to that effect.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

bmcfee:
I think this one's ready to go.  @dpwe ?

dpwe:
still figuring out reviewable, approval on its way

dpwe:
Reviewed 1 of 3 files at r2, 1 of 4 files at r3, 4 of 4 files at r4.

---

*[librosa/filters.py, line 775 at r4](https://reviewable.io:443/reviews/librosa/librosa/413#-KTAA7TUnWAZj5oiZqWI:-KTAA7TUnWAZj5oiZqWJ:b-jozwhy) ([raw file](https://github.com/librosa/librosa/blob/8c0a9efbd9ce147f89630607b0a9f817e2692346/librosa/filters.py#L775)):*
> ```Python
>           (e.g., `('kaiser', 4.0)`)
>         - If numeric, it is treated as the beta parameter of the `'kaiser'`
>           window.
> ```

"(mirroring scipy.signal.get_window)" because that also strikes me as odd.

---

*[librosa/core/spectrum.py, line 398 at r4](https://reviewable.io:443/reviews/librosa/librosa/413#-KT9tJW5J2TfsumIS_BJ:-KT9tJW5J2TfsumIS_BK:bqzoyjo) ([raw file](https://github.com/librosa/librosa/blob/8c0a9efbd9ce147f89630607b0a9f817e2692346/librosa/core/spectrum.py#L398)):*
> ```Python
> 
>     # Construct a padded hann window
>     fft_window = util.pad_center(get_window(window, win_length,
> ```

pad_center makes me think of adding padding in the center of the window (!, but something I've done - a rectangle with cosine tapers at each end).  I would have preferred centered_pad.  Realize this is irrelevant to this PR, just saying.

---

*[tests/test_filters.py, line 336 at r4](https://reviewable.io:443/reviews/librosa/librosa/413#-KTA9kbdD7M1QUsOep-S:-KTA9kbdD7M1QUsOep-T:b-2al636) ([raw file](https://github.com/librosa/librosa/blob/8c0a9efbd9ce147f89630607b0a9f817e2692346/tests/test_filters.py#L336)):*
> ```Python
> 
>     for window in ['hann', u'hann', 4.0, ('kaiser', 4.0)]:
>         yield __test, window
> ```

You could test the remaining two signatures: calling with a list/np.array, and calling with a window function itself.

You could in theory also test the fftbins argument, although I am quite convinced that it is successfully passed on.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

bmcfee:
Review status: 4 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[librosa/filters.py, line 775 at r4](https://reviewable.io:443/reviews/librosa/librosa/413#-KTAA7TUnWAZj5oiZqWI:-KTAB_5hy2NUbM8WsTMS:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/8c0a9efbd9ce147f89630607b0a9f817e2692346/librosa/filters.py#L775)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> "(mirroring scipy.signal.get_window)" because that also strikes me as odd.
</details>
Done.

---

*[librosa/core/spectrum.py, line 398 at r4](https://reviewable.io:443/reviews/librosa/librosa/413#-KT9tJW5J2TfsumIS_BJ:-KTABWfAFAJ33j53hiv0:btlepm8) ([raw file](https://github.com/librosa/librosa/blob/8c0a9efbd9ce147f89630607b0a9f817e2692346/librosa/core/spectrum.py#L398)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> pad_center makes me think of adding padding in the center of the window (!, but something I've done - a rectangle with cosine tapers at each end).  I would have preferred centered_pad.  Realize this is irrelevant to this PR, just saying.
</details>
Agreed, but not changing it (here)

---

*[tests/test_filters.py, line 336 at r4](https://reviewable.io:443/reviews/librosa/librosa/413#-KTA9kbdD7M1QUsOep-S:-KTABUQfKZLEgMEMyUqN:b-7we5lo) ([raw file](https://github.com/librosa/librosa/blob/8c0a9efbd9ce147f89630607b0a9f817e2692346/tests/test_filters.py#L336)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> You could test the remaining two signatures: calling with a list/np.array, and calling with a window function itself.
> 
> You could in theory also test the fftbins argument, although I am quite convinced that it is successfully passed on.
</details>
Added a separate test for this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

bmcfee:
if you're still outgefiguring, click the eyeball in the upper-left of a file tab to indicate that the changes are approved.

---

Review status: 4 of 6 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413#-:-KTAFYkaXSEqZqBNEoLE:b7kav3g)*

dpwe:
reviewable hard to use with the super flakey JetBlue wifi...

dpwe:
Reviewed 2 of 2 files at r5.

---

*[tests/test_filters.py, line 336 at r4](https://reviewable.io:443/reviews/librosa/librosa/413#-KTA9kbdD7M1QUsOep-S:-KTASV85a4OzL7lDJFps:btkl8qx) ([raw file](https://github.com/librosa/librosa/blob/8c0a9efbd9ce147f89630607b0a9f817e2692346/tests/test_filters.py#L336)):*
> Added a separate test for this.
</details>
What about testing passing in a window function?

win = librosa.filters.get_window(scipy.signal.hann, 16)
assert np.allclose(win, scipy.signal.hann(16))

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

bmcfee:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/test_filters.py, line 336 at r4](https://reviewable.io:443/reviews/librosa/librosa/413#-KTA9kbdD7M1QUsOep-S:-KTATcf_k29INFUDxSem:b-q9gddt) ([raw file](https://github.com/librosa/librosa/blob/8c0a9efbd9ce147f89630607b0a9f817e2692346/tests/test_filters.py#L336)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> What about testing passing in a window function?
> 
> win = librosa.filters.get_window(scipy.signal.hann, 16)
> assert np.allclose(win, scipy.signal.hann(16))
</details>
Good call; added.  That path was tested implicitly in stft, but we may as well factor it out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

dpwe:
Reviewed 1 of 1 files at r6.

---

*[tests/test_filters.py, line 336 at r4](https://reviewable.io:443/reviews/librosa/librosa/413#-KTA9kbdD7M1QUsOep-S:-KTAXOduZIQ61WkG2Ugs:bb6ywmx) ([raw file](https://github.com/librosa/librosa/blob/8c0a9efbd9ce147f89630607b0a9f817e2692346/tests/test_filters.py#L336)):*
> Good call; added.  That path was tested implicitly in stft, but we may as well factor it out.
</details>
LGTM

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/413)*

PR #416, opened 2016-10-04
write_wav only normalizes on floating point data

### LINKED ISSUE #415: incorrect normalization in write_wav for int types
The `write_wav` function has normalization turned on by default. However normalization does not check the `dtype` of the input and by default normalizes audio to `[-1, 1]`. The normalized range should match the the range of the data type, e.g. `[-32768, +32767]` for `int16`. The current behavior causes `scipy` to write silent audio when the `dtype` is not `float`.

bmcfee:
Fixes #415 .

For now, this does not deprecate `write_wav`, as the IO question is a broader discussion.

Changes here:

- write_wav only normalizes floating point waveforms
- normalization is disabled by default

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/416)

### DIFF
--- a/librosa/output.py
+++ b/librosa/output.py
@@ -198,7 +198,8 @@ def write_wav(path, y, sr, norm=False):
         sampling rate of `y`
 
     norm : boolean [scalar]
-        enable amplitude normalization
+        enable amplitude normalization.
+        For floating point `y`, scale the data to the range [-1, +1].
 
     Examples
     --------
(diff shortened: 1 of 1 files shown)


bmcfee:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/416)*

Issue #412, opened 2016-09-28
CQT filter normalization revisited


bmcfee:
@ejhumphrey pointed out that the CQT filter bank is not scaled by the length of the filters when `norm=None`.  (When `norm!=None`, this doesn't matter because length scaling washes out.)

This should be easy to fix.

ejhumphrey:
can haz assign? it won't let me do that, not being an official collaborator

ejhumphrey:
for posterity, here's a notebook / gist I put together ... more illustrative than reproducible 😞 

http://nbviewer.jupyter.org/urls/dl.dropbox.com/s/y7bc0oh26s5yt7p/cqt-sweep.ipynb

bmcfee:
BTW, I think you're not going to get bit-level comparison to the dl4mir implementation because librosa's filters are fractional-length, and dl4mir had rounded-length windows (as far as I can tell from a skim of the code).

ejhumphrey:
​good point / generally aware of that fact ... wasn't going to b4b it, the
meat of the work is adding the scalar and updating ​corresponding tests.

bmcfee:
I was just thinking about this again: would it be easier to implement this by adding length-normalization to `util.normalize`?  Then you could just call `librosa.cqt(..., norm='length')` and not have to change the cqt code at all.

ejhumphrey:
+1

bmcfee:
I guess it's not 100% correct since this would use the rounded filter length instead of the fractional length.  Maybe that's not a huge deal?

bmcfee:
One more random thought: what happens if you call cqt with `norm=0`?  This ought to be very close to the length hack I suggested before, but requires no modifications to the code.

bmcfee:
I took a hack at this (`norm='length'`), and it doesn't work because the recursive downsampling uses `scale=True`) to preserve signal energy, which I think is correct here.  Any single-octave filter renormalization will not have @ejhumphrey's desired effect.

I think the only sensible way around this is to reweight the cqt output by dividing out the filter lengths.  I don't think this should be the default behavior.

Do we still want to implement this?

bmcfee:
Fixed by merging #417

PR #417, opened 2016-10-07
CQT length scale normalization

bmcfee:
This PR implements #412, and a couple of bug-fixes.  Summary of contents:

- Added a `scale` boolean option to all CQT methods.  If enabled, CQT bands are normalized by `sqrt(n[i])` where `n[i]` is the length of the `i`th filter.  This is analogous to `norm='ortho'` mode in `np.fft.fft`.
- Early downsampling is less aggressive by one octave.  Previously, it was too aggressive, and the top-end of each octave was close to nyquist, which resulted in attenuation.
- Magnitude is now continuous and approximately equal for full, hybrid, and pseudo CQT.  This fixes some of the unresolved continuity errors noted in #347.
- Expanded the CQT unit tests to include a white noise continuity test.

Using `scale=True` means that white noise input will look like white noise output in CQT space (flat spectrum).  With `scale=False` (current default), white noise in looks like `1/sqrt(f)`.

`scale=True` makes an impulse look like `sqrt(f)` as opposed to constant for `scale=False`.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/417)

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -22,7 +22,7 @@ __all__ = ['cqt', 'hybrid_cqt', 'pseudo_cqt']
 def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         bins_per_octave=12, tuning=None, filter_scale=1,
         norm=1, sparsity=0.01, window='hann',
-        scale=True,
+        scale=False,
         real=util.Deprecated()):
     '''Compute the constant-Q transform of an audio signal.
 
@@ -264,7 +264,7 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 @cache(level=20)
 def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                bins_per_octave=12, tuning=None, filter_scale=1,
-               norm=1, sparsity=0.01, window='hann', scale=True):
+               norm=1, sparsity=0.01, window='hann', scale=False):
     '''Compute the hybrid constant-Q transform of an audio signal.
 
     Here, the hybrid CQT uses the pseudo CQT for higher frequencies where
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -365,6 +365,7 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
     if n_bins_pseudo > 0:
         fmin_pseudo = np.min(freqs[pseudo_filters])
+
         cqt_resp.append(pseudo_cqt(y, sr,
                                    hop_length=hop_length,
                                    fmin=fmin_pseudo,
@@ -486,6 +487,8 @@ def pseudo_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     C = fft_basis.dot(D)
 
     if scale:
+        C /= np.sqrt(n_fft)
+    else:
         lengths = filters.constant_q_lengths(sr, fmin,
                                              n_bins=n_bins,
                                              bins_per_octave=bins_per_octave,
@@ -493,7 +496,7 @@ def pseudo_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                              window=window,
                                              filter_scale=filter_scale)
 
-        C /= np.sqrt(lengths[:, np.newaxis])
+        C *= np.sqrt(lengths[:, np.newaxis] / n_fft)
 
     return C
 
@@ -585,14 +588,12 @@ def __early_downsample(y, sr, hop_length, res_type, n_octaves,
             raise ParameterError('Input signal length={:d} is too short for '
                                  '{:d}-octave CQT'.format(len(y), n_octaves))
 
-        # The additional scaling of sqrt(downsample_factor) here is to
-        # implicitly rescale the filters
-        y = np.sqrt(downsample_factor) * audio.resample(y, sr,
-                                                        sr / downsample_factor,
-                                                        res_type=res_type,
-                                                        scale=True)
-
-        sr /= downsample_factor
+        new_sr = sr / float(downsample_factor)
+        y = audio.resample(y, sr,
+                           new_sr,
+                           res_type=res_type,
+                           scale=True)
+        sr = new_sr
 
     return y, sr, hop_length
 
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -22,6 +22,7 @@ __all__ = ['cqt', 'hybrid_cqt', 'pseudo_cqt']
 def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         bins_per_octave=12, tuning=None, filter_scale=1,
         norm=1, sparsity=0.01, window='hann',
+        scale=True,
         real=util.Deprecated()):
     '''Compute the constant-Q transform of an audio signal.
 
@@ -74,6 +75,13 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     window : str, tuple, number, or function
         Window specification for the basis filters.
         See `filters.get_window` for details.
+
+    scale : bool
+        If `True`, scale the CQT response by square-root the length of
+        each channel's filter.  This is analogous to `norm='ortho'` in FFT.
+
+        If `False`, do not scale the CQT. This is analogous to
+        `norm=None` in FFT.
 
     real : [DEPRECATED]
         .. warning:: This parameter name deprecated in librosa 0.5.0
@@ -239,13 +247,24 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         # Compute the cqt filter response and append to the stack
         cqt_resp.append(__cqt_response(my_y, n_fft, my_hop, fft_basis))
 
-    return __trim_stack(cqt_resp, n_bins)
+    C = __trim_stack(cqt_resp, n_bins)
+
+    if scale:
+        lengths = filters.constant_q_lengths(sr, fmin,
+                                             n_bins=n_bins,
+                                             bins_per_octave=bins_per_octave,
+                                             tuning=tuning,
+                                             window=window,
+                                             filter_scale=filter_scale)
+        C /= np.sqrt(lengths[:, np.newaxis])
+
+    return C
 
 
 @cache(level=20)
 def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                bins_per_octave=12, tuning=None, filter_scale=1,
-               norm=1, sparsity=0.01, window='hann'):
+               norm=1, sparsity=0.01, window='hann', scale=True):
     '''Compute the hybrid constant-Q transform of an audio signal.
 
     Here, the hybrid CQT uses the pseudo CQT for higher frequencies where
@@ -355,7 +374,8 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                    filter_scale=filter_scale,
                                    norm=norm,
                                    sparsity=sparsity,
-                                   window=window))
+                                   window=window,
+                                   scale=scale))
 
     if n_bins_full > 0:
         cqt_resp.append(np.abs(cqt(y, sr,
@@ -367,7 +387,8 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                    filter_scale=filter_scale,
                                    norm=norm,
                                    sparsity=sparsity,
-                                   window=window)))
+                                   window=window,
+                                   scale=scale)))
 
     return __trim_stack(cqt_resp, n_bins)
 
@@ -375,7 +396,7 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 @cache(level=20)
 def pseudo_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                bins_per_octave=12, tuning=None, filter_scale=1,
-               norm=1, sparsity=0.01, window='hann'):
+               norm=1, sparsity=0.01, window='hann', scale=False):
     '''Compute the pseudo constant-Q transform of an audio signal.
 
     This uses a single fft size that is the smallest power of 2 that is greater
@@ -462,7 +483,19 @@ def pseudo_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     D = np.abs(stft(y, n_fft=n_fft, hop_length=hop_length))
 
     # Project onto the pseudo-cqt basis
-    return fft_basis.dot(D)
+    C = fft_basis.dot(D)
+
+    if scale:
+        lengths = filters.constant_q_lengths(sr, fmin,
+                                             n_bins=n_bins,
+                                             bins_per_octave=bins_per_octave,
+                                             tuning=tuning,
+                                             window=window,
+                                             filter_scale=filter_scale)
+
+        C /= np.sqrt(lengths[:, np.newaxis])
+
+    return C
 
 
 @cache(level=10)
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -180,7 +180,7 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     y, sr, hop_length = __early_downsample(y, sr, hop_length,
                                            res_type,
                                            n_octaves,
-                                           nyquist, filter_cutoff)
+                                           nyquist, filter_cutoff, scale)
 
     cqt_resp = []
 
@@ -573,7 +573,7 @@ def __early_downsample_count(nyquist, filter_cutoff, hop_length, n_octaves):
 
 
 def __early_downsample(y, sr, hop_length, res_type, n_octaves,
-                       nyquist, filter_cutoff):
+                       nyquist, filter_cutoff, scale):
     '''Perform early downsampling on an audio signal, if it applies.'''
 
     downsample_count = __early_downsample_count(nyquist, filter_cutoff,
@@ -589,10 +589,15 @@ def __early_downsample(y, sr, hop_length, res_type, n_octaves,
                                  '{:d}-octave CQT'.format(len(y), n_octaves))
 
         new_sr = sr / float(downsample_factor)
-        y = audio.resample(y, sr,
-                           new_sr,
+        y = audio.resample(y, sr, new_sr,
                            res_type=res_type,
                            scale=True)
+
+        # If we're not going to length-scale after CQT, we
+        # need to compensate for the downsampling factor here
+        if not scale:
+            y *= np.sqrt(downsample_factor)
+
         sr = new_sr
 
     return y, sr, hop_length
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -563,8 +563,8 @@ def __cqt_response(y, n_fft, hop_length, fft_basis):
 def __early_downsample_count(nyquist, filter_cutoff, hop_length, n_octaves):
     '''Compute the number of early downsampling operations'''
 
-    downsample_count1 = int(np.ceil(np.log2(audio.BW_FASTEST * nyquist /
-                                            filter_cutoff)) - 1)
+    downsample_count1 = max(0, int(np.ceil(np.log2(audio.BW_FASTEST * nyquist /
+                                                   filter_cutoff)) - 1) - 1)
 
     num_twos = __num_two_factors(hop_length)
     downsample_count2 = max(0, num_twos - n_octaves + 1)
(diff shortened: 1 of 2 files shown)


bmcfee:
The pseudo-cqt with `scale=False` gives a flat response to white noise because it's using a fixed window length for all frequencies.

This causes the previously observed discontinuity in frequency response magnitude of hybrid cqt at the full-pseudo boundary; the compensating factor ought to be something like `n[i] / 2**(np.ceil(np.log2(n[i])))`.

ejhumphrey:
*[librosa/core/constantq.py, line 79 at r1](https://reviewable.io:443/reviews/librosa/librosa/417#-KTVh8UZu58k4R1niJAO:-KTVh8UZu58k4R1niJAP:b-aepka7) ([raw file](https://github.com/librosa/librosa/blob/21e7f8693a6d48756a8c4e1f238c935d9215c5f8/librosa/core/constantq.py#L79)):*
> ```Python
>         See `filters.get_window` for details.
> 
>     scale : bool
> ```

how does this affect the choice of `norm=*` in this function? are these independent parameters?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

ejhumphrey:
`scale=False` gives a flat response, so this changes the default behavior?

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417#-:-KTVgitMNPoMdrxo5Xnt:b4pvhp6)*

ejhumphrey:
*[a discussion](https://reviewable.io:443/reviews/librosa/librosa/417#-KTVhkIXYni4BuLBqg-s:-KTVhkIXYni4BuLBqg-t:b9mi1au) (no related file):*
thoughts on a test that pushes white noise through the CQT with `scale=True/False`? worthwhile or overkill?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

bmcfee:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/librosa/librosa/417#-KTVhkIXYni4BuLBqg-s:-KTVisBFSWlQwHeTuipq:bb6mt9b) (no related file):*
<details><summary><i>Previously, ejhumphrey (Eric J. Humphrey) wrote…</i></summary>
> thoughts on a test that pushes white noise through the CQT with `scale=True/False`? worthwhile or overkill?
</details>
:+1:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

bmcfee:
`scale=False` is the default; behavior should not change.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417#-:-KTVitTrlpiifcObveEr:ba949uk)*

bmcfee:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[librosa/core/constantq.py, line 79 at r1](https://reviewable.io:443/reviews/librosa/librosa/417#-KTVh8UZu58k4R1niJAO:-KTViz6alpeLTmUWVYRL:b-vejd44) ([raw file](https://github.com/librosa/librosa/blob/21e7f8693a6d48756a8c4e1f238c935d9215c5f8/librosa/core/constantq.py#L79)):*
<details><summary><i>Previously, ejhumphrey (Eric J. Humphrey) wrote…</i></summary>
> how does this affect the choice of `norm=*` in this function? are these independent parameters?
</details>
They are independent, yes.

They have to be because the length normalization is different for each octave, but the filter construction is constant across octaves.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

ejhumphrey:
*[librosa/core/constantq.py, line 79 at r1](https://reviewable.io:443/reviews/librosa/librosa/417#-KTVh8UZu58k4R1niJAO:-KTVkdK4aIsTpoabf8KK:by6vkdc) ([raw file](https://github.com/librosa/librosa/blob/21e7f8693a6d48756a8c4e1f238c935d9215c5f8/librosa/core/constantq.py#L79)):*
> They are independent, yes.
> 
> They have to be because the length normalization is different for each octave, but the filter construction is constant across octaves.
</details>
so `scale` implies length-scaling only... and the configuration that I would want is `norm=None, scale=True`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

bmcfee:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[librosa/core/constantq.py, line 79 at r1](https://reviewable.io:443/reviews/librosa/librosa/417#-KTVh8UZu58k4R1niJAO:-KTVlx_6_w-G-PEs71KC:b-aj82cd) ([raw file](https://github.com/librosa/librosa/blob/21e7f8693a6d48756a8c4e1f238c935d9215c5f8/librosa/core/constantq.py#L79)):*
<details><summary><i>Previously, ejhumphrey (Eric J. Humphrey) wrote…</i></summary>
> so `scale` implies length-scaling only... and the configuration that I would want is `norm=None, scale=True`?
</details>
No, I think you want `norm=1, scale=True`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

bmcfee:
If we disable `scale` in `hybrid_cqt`'s call to `pseudo_cqt`, then things look nicely flat, except for the huge scaling discrepancy:

image

bmcfee:
Updates: I made a few changes to the re-normalization, but it's now pretty consistently behaving as follows:

image

Where the red curve (hybrid, `scale=False`) has a jump discontinuity and scale error by a factor of `sqrt(log(downsample_factor))`.  Still trying to find a clean way to resolve that.

bmcfee:
Scale discrepancies are now fixed:

image

In the white noise example, the CQT, hybrid cqt, and pseudo-cqt are all basically equal.  The problem was that the early downsampling needs to know whether post-cqt scaling will be applied so that it can correctly compensate.

The test with an impulse is less clean because of the bigger difference between pseudo and full behavior.

I added the `bug` tag to this issue because I think the scale discrepancy we noticed previously in #347 was actually just a bug, not a feature.

bmcfee:
While we're at it: should we make `scale=True` the default?  

Opinions from @dpwe @lostanlen @ebattenberg @ejhumphrey ?

(If you're just joining us, this means that white noise in time looks white in CQT, and breaks backwards compatibility.)

bmcfee:
2e993f3 reduces the early downsampling nyquist check by one.  I was seeing a lot of scalloping at the top end of the octaves, indicating that we were too close to critical sampling.  This is no longer an issue.

ejhumphrey:
regarding the default value of `scale`, I feel defaults should strive to match intuition and take priority over backwards compatibility, and so for me the question is more "what would one expect to get back from a standard CQT?"

that I raised the issue, I would expect a white noise signal to be flat in CQT space (like white noise in an image). It seems like the thing one would want more often, and I (perhaps wrongly) assume filter normalization isn't on the radar of a DSP non-expert.

bmcfee:
> "what would one expect to get back from a standard CQT?"

If we treat [Brown'91](https://www.researchgate.net/profile/Christian_Schoerkhuber/publication/228523955_Constant-Q_transform_toolbox_for_music_processing/links/55126aa60cf20bfdad513a3f.pdf) (Eq. 5) as the standard, then it should be length-normalized: divide by `N[k]` for the `k`th basis.

> I would expect a white noise signal to be flat in CQT space

Dividing by `N[k]` actually doesn't get you a white spectrum on white noise if you use `scipy.fftpack.fft` (or API-compatible friends like pyfftw).  You need a `sqrt(N[k])` to achieve this, as demonstrated by the -1/2 slope on the log-log plots above.  AFAIK, other implementations don't do this -- but maybe Ron's / marlib's do?



> I feel defaults should strive to match intuition and take priority over backwards compatibility

I agree with that.  Since we're breaking backwards compatibility anyway to fix the hybrid_cqt bug (and complex output before that), I'm open to changing the default for scaling as well.  But I want to hash out the implications of it before committing to anything.  For example, it will also change how the downstream chroma features behave -- probably for the better?

dpwe:
>
>
> I would expect a white noise signal to be flat in CQT space
>
> Dividing by N[k] actually doesn't get you a white spectrum on white noise
> if you use scipy.fftpack.fft (or API-compatible friends like pyfftw). You
> need a sqrt(N[k]) to achieve this, as demonstrated by the -1/2 slope on
> the log-log plots above. AFAIK, other implementations don't do this -- but
> maybe Ron's / marlib's do?
>
To be clear, the FFT implementation isn't relevant here, right?  The only
possible major difference between FFT implementations would be global
scaling (sometimes 1/N in one direction, sometimes 1/sqrt(N) in both
directions), which would show up as step discontinuities each octave if it
was "wrong".  It wouldn't change the CQT output from flat to sloped or vice
versa.


> I feel defaults should strive to match intuition and take priority over
> backwards compatibility
>
> I agree with that. Since we're breaking backwards compatibility anyway to
> fix the hybrid_cqt bug (and complex output before that), I'm open to
> changing the default for scaling as well. But I want to hash out the
> implications of it before committing to anything. For example, it will also
> change how the downstream chroma features behave -- probably for the better?
>

I agree with privileging intuition, but I'm not sure flat CQT is the right
intuition.  Compare these two impulse response images from the earlier
discussion of  CQT <https://github.com/librosa/librosa/issues/341>scaling
(the first one has octave contours, but imagine it smoothly decaying within
each octave block):

[image: Inline image 1][image: Inline image 3]


When you plot it to show the time spread too, having the amplitude decay as
1/N[k]^p looks better, otherwise it seems like the low frequency channels
are indicating way more energy because the high amplitude lasts so much
longer.  Integrating along time in the 1/N[k]^p version seems like each CQ
band delivers the same amount of total energy.

  DAn.

bmcfee:
> he only
> possible major difference between FFT implementations would be global
> scaling (sometimes 1/N in one direction, sometimes 1/sqrt(N) in both
> directions), which would show up as step discontinuities each octave if it
> was "wrong".

Correct.  The scipy implementation uses a 1/N scaling on the inverse transform.  We already compensate for this when computing each octave's response [here](https://github.com/librosa/librosa/blob/master/librosa/core/constantq.py#L492-L493).



> but I'm not sure flat CQT is the right
> intuition.

This was only intended to refer to white noise input. I think @ejhumphrey's intuition (correct me if I'm wrong) is that white input should have flat spectra at all frames.  As you point out, an impulse should only look flat when time is integrated out, but at should curve upward within a frame.

For an impulse, this is what we currently get:

image

Integrating squared energy over time:

image

so the scaled version gets a flat response, like you say, up to the last octave.

The linear slope in the top octave seems a bit suspicious -- I suspect what's happening is that no downsampling is happening here, and the lower octaves are relying on downsampling to implement part of the scaling.  I'll investigate.

dpwe:
The impulse responses look lovely, apart from the top octave as you note.

Can you show us the same plots for white noise input?  Or did I miss them
upthread?

Thanks,

  DAn.

bmcfee:
> Can you show us the same plots for white noise input?  Or did I miss them
> upthread?

Those are, more or less, the up-thread plots.  But here's exactly the same code on white input instead of impulse:

image


image

dpwe:
So that's great as it stands IMO.

(My) intuition wants flat responses for both impulses and white noise, but
CQT doesn't work that way (because of the increasing integration window for
lower frequencies).  So we have two settings - one that gives flat
(time-integrated energy for) impulse input, and one that gives flat
response to white noise.

Thanks,

  DAn.

bmcfee:
Great.

Any thoughts about what's going on in that top octave?  The ramp on the impulse response seems a little weird to me, but I think it's unavoidable given that it's going to have a flat response in a fixed-window FFT (ie, the one used to compute the top octave), and we're re-scaling it to simulate having variable window lengths.  Since the filters have constant energy but variable length, the peaks will decay with the length, and that's what we're seeing there.

At the lower octaves, this washes out because we're integrating over wider windows and the downsampling smears out the impulse.  (I think?)

Anyway, I think using white noise is a saner test than an impulse for the scaling issue, so I'm inclined to leave it as is.

If folks agree on that, this just leaves the question of what the default behavior should be.

ejhumphrey:
so, another signal test that came to me as I was dozing off last night is a frequency modulated sine-sweep, i.e. chirp. The signal would be expressed as `y(t) = α*sin(θ(t) + φ)`. In this instance, with `α` being a constant, I would expect (or perhaps want) the max-amplitude of the chirp in a time-frequency representation to be constant also, i.e. a flat response, regardless of whether the analysis filterbank is linear (DFT) or log-spaced (CQT).

is day-off-Eric being dumb, or...? perhaps it's just that I like to think of the CQT as a pitch-linear complement to the DFT, which would guide my intuition for defaults.

bmcfee:
> The signal would be expressed as y(t) = α*sin(θ(t) + φ). In this instance, with α being a constant, I would expect (or perhaps want) the max-amplitude of the chirp in a time-frequency representation to be constant also, i.e. a flat response, regardless of whether the analysis filterbank is linear (DFT) or log-spaced (CQT).

I'm not sure?  It's going to depend on the rate of the sweep.  Lower frequencies will resonate longer than high frequencies with a linear sweep, so it should go down like 1/rate (or something).

bmcfee:
Here's MSE of a linear-speed sine sweep from C1 (~ `2**5`) to C8 (~ `2**12`):
```python
sr = 68000
duration = 5
freqs = librosa.note_to_hz(['C1', 'C8']) / sr
y = np.sin(np.cumsum(2 * np.pi * np.linspace(freqs[0],
                                             freqs[-1],
                                             num=duration*sr)))
```
image


which, does fit @ejhumphrey's intuition about flat response, modulo setup on the left.  It falls off on the right because the CQT goes one octave higher than the sweep.

dpwe:
Any thoughts about what's going on in that top octave? The ramp on the
> impulse response seems a little weird to me, but I think it's unavoidable
> given that it's going to have a flat response in a fixed-window FFT (ie,
> the one used to compute the top octave), and we're re-scaling it to
> simulate having variable window lengths. Since the filters have constant
> energy but variable length, the peaks will decay with the length, and
> that's what we're seeing there.
>
>
I think it's because of the very privileged alignment of the impulse in the
top octave, where it is essentially a single point for all bins.  If you
delayed it by half a sample (i.e., a sinc function, a half-sample delay) it
might look OK.

What is the cutoff frequency of the anti-alias filter before the
downsampling?  Maybe if you LPF the input at 2x that before starting, it
would blur it enough to make it behave the same.

  DAn.

bmcfee:
> I think it's because of the very privileged alignment of the impulse in the
> top octave, where it is essentially a single point for all bins.  If you
> delayed it by half a sample (i.e., a sinc function, a half-sample delay) it
> might look OK.

I gave that a shot: impulse at an odd sample position, then downsampled by 2 with resampy:

image

looks pretty much the same.  But I agree without about sample position; if the impulse happens in between frames (frame center + 1/2 hop_length), we get something like:

image

This looks right to me: once the signal has been downsampled enough, the spacing of the impulse stops mattering.  Above that threshold, higher frequencies will decay more quickly due to the relative position of the impulse compared to the filter's peaks.

I think I'm convinced that this is all behaving as it should.

bmcfee:
So, should I take the responses here as indicating that `scale=True` should be default?  If so, I'll have to make some adjustments to the tests before it's ready for CR.

ejhumphrey:
that's been my position, but I am only one vote.

ejhumphrey:
*[librosa/core/constantq.py, line 79 at r8](https://reviewable.io:443/reviews/librosa/librosa/417#-KTo25chNYd_bQ1a7fWw:-KTo25chNYd_bQ1a7fWx:b-nubsx) ([raw file](https://github.com/librosa/librosa/blob/7af80cb8cd034640fde66112a62dacbf0240ee78/librosa/core/constantq.py#L79)):*
> ```Python
>         See `filters.get_window` for details.
> 
>     scale : bool
> ```

is it worth mentioning anything about the magnitude response's "flatness" here? or is that too colloquial for respective tastes?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

bmcfee:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[librosa/core/constantq.py, line 79 at r8](https://reviewable.io:443/reviews/librosa/librosa/417#-KTo25chNYd_bQ1a7fWw:-KTo2pI_T6b7Mv8Z_pak:b-kztorb) ([raw file](https://github.com/librosa/librosa/blob/7af80cb8cd034640fde66112a62dacbf0240ee78/librosa/core/constantq.py#L79)):*
<details><summary><i>Previously, ejhumphrey (Eric J. Humphrey) wrote…</i></summary>
> is it worth mentioning anything about the magnitude response's "flatness" here? or is that too colloquial for respective tastes?
</details>
I think that's too specific to be useful in the docstring.  It might make a good example though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

ejhumphrey:
*[librosa/core/constantq.py, line 79 at r8](https://reviewable.io:443/reviews/librosa/librosa/417#-KTo25chNYd_bQ1a7fWw:-KTo8ZbfSgLfmTXS8M35:b32v57) ([raw file](https://github.com/librosa/librosa/blob/7af80cb8cd034640fde66112a62dacbf0240ee78/librosa/core/constantq.py#L79)):*
> I think that's too specific to be useful in the docstring.  It might make a good example though.
</details>
+1

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

ejhumphrey:
*[librosa/core/constantq.py, line 489 at r8](https://reviewable.io:443/reviews/librosa/librosa/417#-KTo8dNy1RPD-jpkp0H1:-KTo8dNy1RPD-jpkp0H2:b9imuuv) ([raw file](https://github.com/librosa/librosa/blob/7af80cb8cd034640fde66112a62dacbf0240ee78/librosa/core/constantq.py#L489)):*
> ```Python
>     C = fft_basis.dot(D)
> 
>     if scale:
> ```

this is the same block of code as above, right? function worthy?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

ejhumphrey:
*[tests/test_constantq.py, line 244 at r8](https://reviewable.io:443/reviews/librosa/librosa/417#-KTo9aRaQZ_Unv2_ORtd:-KTo9aRbdrRgloJ6oZDe:b-b0qjxx) ([raw file](https://github.com/librosa/librosa/blob/7af80cb8cd034640fde66112a62dacbf0240ee78/tests/test_constantq.py#L244)):*
> ```Python
> 
> 
> def test_cqt_white_noise():
> ```

is the sine-sweep test worth folding in here? `|question| >> |suggestion|`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

bmcfee:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[librosa/core/constantq.py, line 489 at r8](https://reviewable.io:443/reviews/librosa/librosa/417#-KTo8dNy1RPD-jpkp0H1:-KToCtPIzkVHjqjhZMur:bmmwihf) ([raw file](https://github.com/librosa/librosa/blob/7af80cb8cd034640fde66112a62dacbf0240ee78/librosa/core/constantq.py#L489)):*
<details><summary><i>Previously, ejhumphrey (Eric J. Humphrey) wrote…</i></summary>
> this is the same block of code as above, right? function worthy?
</details>
It's a small repetition, but I don't think it's worth factoring out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

bmcfee:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/test_constantq.py, line 244 at r8](https://reviewable.io:443/reviews/librosa/librosa/417#-KTo9aRaQZ_Unv2_ORtd:-KToCzdMwxfR5N49FLoX:b1zlyoi) ([raw file](https://github.com/librosa/librosa/blob/7af80cb8cd034640fde66112a62dacbf0240ee78/tests/test_constantq.py#L244)):*
<details><summary><i>Previously, ejhumphrey (Eric J. Humphrey) wrote…</i></summary>
> is the sine-sweep test worth folding in here? `|question| >> |suggestion|`
</details>
I think the white noise test is sufficient.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

ejhumphrey:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417#-:-KToGn4Sj9lS9l8_lJWQ:bnfp4nl)*

ejhumphrey:
Reviewed 1 of 1 files at r1, 2 of 2 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/417)*

ejhumphrey:
for the record, I still !❤️  reviewable 😞

dpwe:
I prefer scale=True (white noise, sine sweep is flat, impulse has lower
amplitude as time support grows).

  DAn.

bmcfee:
Ok!  Thanks @dpwe and @ejhumphrey -- I think this one's ready to go then.

Issue #360, opened 2016-05-16
Trim and split


carlthome:
Would it make sense to add a way of trimming silence when loading audio? Currently I'm doing this silliness, which works alright aside from being performance intensive, but the onset detector surely isn't built for detecting offsets like this, I'd assume.

``` Python
path = ... # Path to an audio file with a bunch of leading and trailing silence.
y, sr = librosa.load(path, sr=None)
onsets = librosa.frames_to_time(librosa.onset.onset_detect(y, sr), sr)
offsets = librosa.frames_to_time(librosa.onset.onset_detect(y[::-1], sr), sr)  # Suboptimal!
leading_silence = onsets[0]
trailing_silence = offsets[0]
total_duration = librosa.get_duration(y, sr)
y, _ = librosa.load(path, sr, offset=leading_silence, duration=total_duration-leading_silence-trailing_silence)
```

Also, perhaps if loading with a set duration longer than the actual file, then there should be the option of padding the sequence to the set length, or at least provide a user warning that less than `duration` was loaded.

bmcfee:
Probably a better way to do this is to call `librosa.feature.rmse`, threshold the result, and retain the first and last non-zero indices.

I don't think we should add this to `load`, which is already a little too complicated IMO.  However, it might be handy to include a `trim` function which simplifies this process.  Anyone else have thoughts?

carlthome:
A `trim` function would be very handy. :+1:

dpwe:
We could have trim accept a dB threshold input, and a flag to choose
between interpreting that relative to the largest energy in the file, or
full-scale.  Maybe default to 60 dB below full-scale (rms sample value < 32
in 16 bit ints).  It accepts n_fft (in lieu of window_length) and
hop_length like the underlying rmse.

(I just noticed that rmse uses spectrogram under the hood.  That's quite
extravagant compared to simply summing squared sample values.)

  DAn.

carlthome:
It would also be useful to have an argument for `trim` that trims _in between_ peaks as well. Trivial extension, and useful in a few machine learning applications when data is abundant, e.g. a signature like this `trim(threshold, mode='beginning'|'end'|'both'|'all')`

bmcfee:
> We could have trim accept a dB threshold input, and a flag to choose
> between interpreting that relative to the largest energy in the file, or
> full-scale.  Maybe default to 60 dB below full-scale (rms sample value < 32
[...]

PR #428, opened 2016-11-01
Trim and split

bmcfee:
Implements #360

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/428)

### DIFF
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -33,12 +33,13 @@ import numpy as np
 
 from . import core
 from . import decompose
+from . import feature
 from . import util
 from .util.exceptions import ParameterError
 
 __all__ = ['hpss', 'harmonic', 'percussive',
            'time_stretch', 'pitch_shift',
-           'remix']
+           'remix', 'trim']
 
 
 def hpss(y, **kwargs):
@@ -77,7 +78,7 @@ def hpss(y, **kwargs):
     >>> y, sr = librosa.load(librosa.util.example_audio_file())
     >>> y_harmonic, y_percussive = librosa.effects.hpss(y)
 
-    >>> # Get a more isolated percussive component by widening its margin 
+    >>> # Get a more isolated percussive component by widening its margin
     >>> y_harmonic, y_percussive = librosa.effects.hpss(y, margin=(1.0,5.0))
 
     '''
@@ -93,6 +94,7 @@ def hpss(y, **kwargs):
     y_perc = util.fix_length(core.istft(stft_perc, dtype=y.dtype), len(y))
 
     return y_harm, y_perc
+
 
 def harmonic(y, **kwargs):
     '''Extract harmonic elements from an audio time-series.
@@ -117,13 +119,12 @@ def harmonic(y, **kwargs):
 
     Examples
     --------
-    >>> # Extract harmonic component 
+    >>> # Extract harmonic component
     >>> y, sr = librosa.load(librosa.util.example_audio_file())
     >>> y_harmonic = librosa.effects.harmonic(y)
 
     >>> # Use a margin > 1.0 for greater harmonic separation
     >>> y_harmonic = librosa.effects.harmonic(y, margin=3.0)
-
 
     '''
 
@@ -137,6 +138,7 @@ def harmonic(y, **kwargs):
     y_harm = util.fix_length(core.istft(stft_harm, dtype=y.dtype), len(y))
 
     return y_harm
+
 
 def percussive(y, **kwargs):
     '''Extract percussive elements from an audio time-series.
@@ -160,14 +162,13 @@ def percussive(y, **kwargs):
     librosa.decompose.hpss : HPSS for spectrograms
 
     Examples
-    --------    
-    >>> # Extract percussive component 
+    --------
+    >>> # Extract percussive component
     >>> y, sr = librosa.load(librosa.util.example_audio_file())
     >>> y_percussive = librosa.effects.percussive(y)
 
     >>> # Use a margin > 1.0 for greater percussive separation
     >>> y_percussive = librosa.effects.percussive(y, margin=3.0)
-
 
     '''
 
@@ -374,3 +375,79 @@ def remix(y, intervals, align_zeros=True):
         y_out.append(y[clip])
 
     return np.concatenate(y_out, axis=-1)
+
+
+def trim(y, top_db=60, ref_power=np.max, n_fft=2048, hop_length=512,
+         index=False):
+    '''Trim leading and trailing silence from an audio signal.
+
+    Parameters
+    ----------
+    y : np.ndarray, shape=(n,) or (2,n)
+        Audio signal, can be mono or stereo
+
+    top_db : number > 0
+        The threshold (in decibels) below reference to consider as
+        silence
+
+    ref_power : number or callable
+        The reference power.  By default, it uses `np.max` and compares
+        to the peak power in the signal.
+
+    n_fft : int > 0
+        The number of samples per analysis frame
+
+    hop_length : int > 0
+        The number of samples between analysis frames
+
+    index : bool
+        If `True`, return the start and end of the non-silent
+        region of `y` along with the trimmed signal.
+
+        If `False`, only return the trimmed signal.
+
+    Returns
+    -------
+    y_trimmed : np.ndarray, shape=(m,) or (2, m)
+        The trimmed signal
+
+    index : slice, optional
+        If `index=True` is provided, then this contains
+        the slice of `y` corresponding to the non-silent region:
+        `y_trimmed = y[index]`.
+
+    Examples
+    --------
+    >>> # Load some audio
+    >>> y, sr = librosa.load(librosa.util.example_audio_file())
+    >>> # Trim the beginning and ending silence
+    >>> yt = librosa.effects.trim(y)
+    >>> # Print the durations
+    >>> print(librosa.get_duration(y), librosa.get_duration(yt))
+    61.45886621315193 60.58086167800454
+    '''
+
+    # Convert to mono
+    y_mono = core.to_mono(y)
+
+    # Compute the MSE for the signal
+    mse = feature.rmse(y=y_mono, n_fft=n_fft, hop_length=hop_length)**2
+
+    # Compute the log power indicator and non-zero positions
+    logp = core.logamplitude(mse, ref_power=ref_power, top_db=None) > - top_db
+    nonzero = np.flatnonzero(logp)
+
+    # Compute the start and end positions
+    # End position goes one frame past the last non-zero
+    start = int(core.frames_to_samples(nonzero[0], hop_length))
+    end = min(len(y_mono),
+              int(core.frames_to_samples(nonzero[-1] + 1, hop_length)))
+
+    # Build the mono/stereo index
+    full_index = [slice(None)] * y.ndim
+    full_index[-1] = slice(start, end)
+
+    if index:
+        return y[full_index], full_index[-1]
+    else:
+        return y[full_index]
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -27,6 +27,8 @@ Miscellaneous
     :toctree: generated/
 
     remix
+    trim
+    split
 """
 
 import numpy as np
@@ -39,7 +41,7 @@ from .util.exceptions import ParameterError
 
 __all__ = ['hpss', 'harmonic', 'percussive',
            'time_stretch', 'pitch_shift',
-           'remix', 'trim']
+           'remix', 'trim', 'split']
 
 
 def hpss(y, **kwargs):
@@ -377,8 +379,49 @@ def remix(y, intervals, align_zeros=True):
     return np.concatenate(y_out, axis=-1)
 
 
-def trim(y, top_db=60, ref_power=np.max, n_fft=2048, hop_length=512,
-         index=False):
+def _signal_to_frame_nonsilent(y, frame_length=2048, hop_length=512, top_db=60,
+                               ref_power=np.max):
+    '''Frame-wise non-silent indicator for audio input.
+
+    This is a helper function for `trim` and `split`.
+
+    Parameters
+    ----------
+    y : np.ndarray, shape=(n,) or (2,n)
+        Audio signal, mono or stereo
+
+    frame_length : int > 0
+        The number of samples per frame
+
+    hop_length : int > 0
+        The number of samples between frames
+
+    top_db : number > 0
+        The threshold (in decibels) below reference to consider as
+        silence
+
+    ref_power : callable or float
+        The reference power
+
+    Returns
+    -------
+    non_silent : np.ndarray, shape=(m,), dtype=bool
+        Indicator of non-silent frames
+    '''
+    # Convert to mono
+    y_mono = core.to_mono(y)
+
+    # Compute the MSE for the signal
+    mse = feature.rmse(y=y_mono,
+                       frame_length=frame_length,
+                       hop_length=hop_length)**2
+
+    return (core.logamplitude(mse.squeeze(),
+                              ref_power=ref_power,
+                              top_db=None) > - top_db)
+
+
+def trim(y, top_db=60, ref_power=np.max, frame_length=2048, hop_length=512):
     '''Trim leading and trailing silence from an audio signal.
 
     Parameters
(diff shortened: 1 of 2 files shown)


bmcfee:
I think this one's ready for CR, if @carlthome or @dpwe wants to take a look.

Here's a quick [demo notebook](https://gist.github.com/14d652560d71f6f4d30e142459515b2c) for split:

```python
hop_length = 512
intervals = librosa.effects.split(y, hop_length=hop_length)
interval_times = librosa.samples_to_time(intervals, sr=sr)
plt.figure()
librosa.display.waveplot(y, sr=sr)
for ival in interval_times:
    plt.axvspan(ival[0], ival[1], ymin=-1, ymax=1, alpha=0.25, color='r')
```
image

Trim works pretty much the same way, though the API is a little different.  `trim` returns the trimmed audio directly, and optionally the start/end samples, since I figure this is the most common use case.

`split` returns the array of intervals, but does not actually do any audio slicing.  These would in principle be cheap to do (ie, not require copies), but it would make for an awkward return type (list of `ndarray`).

Does this all seem reasonable?

carlthome:
Sweet! :+1: 

Two nitpicks, nothing important really. See my comments above.

dpwe:
I'm also working on a review

dpwe:
Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions.

---

*[librosa/effects.py, line 407 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_BinnlW0ZwTeqvOnN:-KV_BinnlW0ZwTeqvOnO:b-2p5viq) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L407)):*
> ```Python
>     index : bool
>         If `True`, return the start and end of the non-silent
>         region of `y` along with the trimmed signal.
> ```

It's not completely clear to me from this comment which order the returns are returned in.  Why not always return both, and let the caller discard the second return if they don't want it?  Seems a small price to pay compared to functions whose output shape depends on boolean arguments.

---

*[librosa/effects.py, line 494 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_CLRziFrnEn6-sqyC:-KV_CLRziFrnEn6-sqyD:bhbg7u9) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L494)):*
> ```Python
> 
>     # Compute the log power indicator and non-zero positions
>     logp = (core.logamplitude(mse.squeeze(),
> ```

Why does it have squeeze here but not in trim?  Assuming they are supposed to be the same, I'd be inclined to define _signal_to_frame_energy(y, n_fft, hop_length, ref_power, top_db).

logp is not a great name for this value; it implies it's the log power, whereas in fact its a boolean comparing the power to a threshold.  I'd prefer mse_above_threshold, but that's wordy.

---

*[librosa/effects.py, line 501 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_DNo0kw5WoqM0UCwE:-KV_DNo1wnNe-EO-n__C:b-rdkrra) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L501)):*
> ```Python
>     # https://stackoverflow.com/questions/2619413/efficiently-finding-the-interval-with-non-zeros-in-scipy-numpy-in-python
>     # Find points where the sign flips
>     edges = np.flatnonzero(np.diff((~logp).astype(int)))
> ```

What's the bitwise-not doing there?  nonzero(diff()) is just testing for x[i] != x[i+1], so flipping the sense doesn't add anything, does it?

---

*[librosa/effects.py, line 508 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_Dqet-2qj5_E6Fe-Y:-KV_Dqet-2qj5_E6Fe-Z:b-az2p42) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L508)):*
> ```Python
>     # If the first frame had high energy, count it
>     if logp[0]:
>         edges.insert(0, [0])
> ```

You can do the index offset and the inclusion of the start and end edges with np.flatnonzero(np.diff(np.hstack(False, mse_above_threshold, False)))

---

*[tests/test_effects.py, line 170 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_F8TYdLEw87aEIam9:-KV_F8TYdLEw87aEIamA:b-ke9eby) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L170)):*
> ```Python
>         # Verify duration
>         duration = librosa.get_duration(yt)
>         assert np.allclose(duration, 3.0, atol=1e-1), duration
> ```

Hard-coded parameter in the helper function?  Really?

---

*[tests/test_effects.py, line 171 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_FGLCO7s6HVIUbyK6:-KV_FGLCO7s6HVIUbyK7:bgr4uxu) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L171)):*
> ```Python
>         duration = librosa.get_duration(yt)
>         assert np.allclose(duration, 3.0, atol=1e-1), duration
> ```

This only tests that the returned parts are not silent.  How about adding:

if index:
  rms = librosa.feature.rmse(librosa.to_mono(y))
  logamp = librosa.logamplitude(rms**2, ref_power=ref_power, top_db=None)
  assert np.all(logamp(index) >= -top_db)
  assert np.all(logamp(0:index.indices(log_amp.shape[-1])[0]) < top_db)
  assert np.all(logamp(index.indices(log_amp.shape[-1])[1]):) < top_db)

---

*[tests/test_effects.py, line 197 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_IAClea8l6w3tPpCs:-KV_IAClea8l6w3tPpCt:b-ws0cgd) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L197)):*
> ```Python
>         int_match = librosa.util.match_intervals(intervals, idx_true)
> 
>         print(idx_true)
> ```

Why isn't idx_true an argument?

---

*[tests/test_effects.py, line 198 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_HmGwCH5RH21aPinh:-KV_HmGwCH5RH21aPini:b-3u2lln) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L198)):*
> ```Python
> 
>         print(idx_true)
>         print(intervals)
> ```

Debugging prints that should be removed?

---

*[tests/test_effects.py, line 202 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_IQy3WCV89u_ga_gL:-KV_IQy3WCV89u_ga_gM:b-udli7e) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L202)):*
> ```Python
>             i_true = idx_true[int_match[i]]
> 
>             assert np.all(np.abs(i_true - intervals[i]) <= 2048), intervals[i]
> ```

replace 2048 with hop_length?

---

*[tests/test_effects.py, line 210 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_HV_4dTzhTIxDQA53:-KV_HV_4dTzhTIxDQA54:b-t7vp13) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L210)):*
> ```Python
>     y[::2] *= -1
> 
>     # Zero out all but two few intervals
> ```

s/few //

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

bmcfee:
Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions.

---

*[librosa/effects.py, line 407 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_BinnlW0ZwTeqvOnN:-KV_KCVmADn-e8ForXIn:b-uw3pz3) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L407)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> It's not completely clear to me from this comment which order the returns are returned in.  Why not always return both, and let the caller discard the second return if they don't want it?  Seems a small price to pay compared to functions whose output shape depends on boolean arguments.
</details>
There's precedent in numpy for this style of return, but I agree that it's not a good practice in general.  I'll modify it to always return the slice.

---

*[librosa/effects.py, line 494 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_CLRziFrnEn6-sqyC:-KV_KNtQBjKjvfoXoLZr:b-4r7d5h) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L494)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> Why does it have squeeze here but not in trim?  Assuming they are supposed to be the same, I'd be inclined to define _signal_to_frame_energy(y, n_fft, hop_length, ref_power, top_db).
> 
> logp is not a great name for this value; it implies it's the log power, whereas in fact its a boolean comparing the power to a threshold.  I'd prefer mse_above_threshold, but that's wordy.
</details>
how about `non_silent`?

Good call on factoring out the energy calculation.

---

*[librosa/effects.py, line 501 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_DNo0kw5WoqM0UCwE:-KV_KZclyC4SJgig2cQ3:bjuys8w) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L501)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> What's the bitwise-not doing there?  nonzero(diff()) is just testing for x[i] != x[i+1], so flipping the sense doesn't add anything, does it?
</details>
Right -- this is propagated (and binarized) from the source that I cribbed above.

---

*[librosa/effects.py, line 508 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_Dqet-2qj5_E6Fe-Y:-KV_Kib8uIVfRjAovO8L:b-bzuiym) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L508)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> You can do the index offset and the inclusion of the start and end edges with np.flatnonzero(np.diff(np.hstack(False, mse_above_threshold, False)))
</details>
hstacking there would trigger an additional memory allocation that we don't really need.  I think the current version is both faster and more readable.

---

*[tests/test_effects.py, line 170 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_F8TYdLEw87aEIam9:-KV_KsxCrHXoMzE_nn8c:b-rcglab) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L170)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> Hard-coded parameter in the helper function?  Really?
</details>
Hah :grin:  yep, bad form.

---

*[tests/test_effects.py, line 197 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_IAClea8l6w3tPpCs:-KV_KycGHwml9kbKFcIm:b-phlwzh) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L197)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> Why isn't idx_true an argument?
</details>
it's within scope.  Passing it as an argument makes the test log more difficult to read.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

carlthome:
FYI, a Reviewable inspired tool has been added to GitHub directly apparently (but I forgot to submit it until now, oops). Works pretty sweet though.

dpwe:
Review status: 0 of 2 files reviewed at latest revision, 12 unresolved discussions.

---

*[librosa/effects.py, line 508 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_Dqet-2qj5_E6Fe-Y:-KV_NG4L9hC5dMpof2Xn:bvyq8xx) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L508)):*
> hstacking there would trigger an additional memory allocation that we don't really need.  I think the current version is both faster and more readable.
</details>
Readability is such a squishy standard.  Once you're familiar with the idiom, padding first is natural and way more compact.  But unless you recognize the issue, it's pretty mysterious, so I agree.  I guess the extra alloc tips the balance in your favor. 

My instinct is that if you can replace 6 lines with 1, it's usually worth it, because coding is a constant battle between functionality and attention span.  But that's a very abstract principle.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

librosa/effects.py
```diff
@@ -374,3 +377,143 @@ def remix(y, intervals, align_zeros=True):
...
+
+    top_db : number > 0
+        The threshold (in decibels) below reference to consider as
+        silence
+
+    ref_power : number or callable
+        The reference power.  By default, it uses `np.max` and compares
+        to the peak power in the signal.
+
+    n_fft : int > 0
```
bmcfee: Agreed; `frame_length` would match `util.frame`, but it propagates `n_fft` because that's how it's called in `rmse`.

Of course, we could (should) rename the parameter in `rmse` as well, since it's no longer accurate.  I'll handle that in a separate PR.

bmcfee:
Review status: 0 of 2 files reviewed at latest revision, 12 unresolved discussions.

---

*[librosa/effects.py, line 494 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KVVPUx-_fCS77xrlLEv-r2-494:-KV_QIluCCdShfDQPpp4:b-j4vz7t) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L494)):*
<details><summary><i>Previously, carlthome (Carl Thomé) wrote…</i></summary>
> Why the squeeze here? Shouldn't that happen during mono conversion?
</details>
no; rmse always returns as 2d for compatibility with the other feature functions.  But this point is now moot as the code in question has been refactored.

---

*[librosa/effects.py, line 501 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_DNo0kw5WoqM0UCwE:-KV_QQVxR4curGD3v3uo:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L501)):*
> Right -- this is propagated (and binarized) from the source that I cribbed above.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

bmcfee:
Review status: 0 of 2 files reviewed at latest revision, 12 unresolved discussions.

---

*[tests/test_effects.py, line 171 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_FGLCO7s6HVIUbyK6:-KV_RT23XW-ud3t9K_MM:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L171)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> This only tests that the returned parts are not silent.  How about adding:
> 
> if index:
>   rms = librosa.feature.rmse(librosa.to_mono(y))
>   logamp = librosa.logamplitude(rms**2, ref_power=ref_power, top_db=None)
>   assert np.all(logamp(index) >= -top_db)
>   assert np.all(logamp(0:index.indices(log_amp.shape[-1])[0]) < top_db)
>   assert np.all(logamp(index.indices(log_amp.shape[-1])[1]):) < top_db)
</details>
Done.

---

*[tests/test_effects.py, line 198 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_HmGwCH5RH21aPinh:-KV_RS9nVJ6qiZe4qpsr:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L198)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> Debugging prints that should be removed?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

bmcfee:
Review status: 0 of 2 files reviewed at latest revision, 12 unresolved discussions.

---

*[librosa/effects.py, line 407 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_BinnlW0ZwTeqvOnN:-KV_Zs4cHGNWcQ5fcloz:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L407)):*
> There's precedent in numpy for this style of return, but I agree that it's not a good practice in general.  I'll modify it to always return the slice.
</details>
Done.

---

*[librosa/effects.py, line 494 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KVVPUx-_fCS77xrlLEv-r2-494:-KV_ZtgXv8qgFoxIlIwT:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L494)):*
> no; rmse always returns as 2d for compatibility with the other feature functions.  But this point is now moot as the code in question has been refactored.
</details>
Done.

---

*[librosa/effects.py, line 494 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_CLRziFrnEn6-sqyC:-KV_Zu4vEqjQUamK4OXG:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/librosa/effects.py#L494)):*
> how about `non_silent`?
> 
> Good call on factoring out the energy calculation.
</details>
Done.

---

*[tests/test_effects.py, line 170 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_F8TYdLEw87aEIam9:-KV_ZxZMajsDWs8rmpVc:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L170)):*
> Hah :grin:  yep, bad form.
</details>
Done.

---

*[tests/test_effects.py, line 202 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_IQy3WCV89u_ga_gL:-KV_Zyg0VQ_cq85R-cC6:byuvj6e) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L202)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> replace 2048 with hop_length?
</details>
I made it frame_length rather than hop, which was a little too sensitive.  As long as we're within +- 1 frame, I think it's fine.

---

*[tests/test_effects.py, line 210 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KV_HV_4dTzhTIxDQA53:-KV__7oo7N2-dmvE5PL2:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/4fcf52e44867894fcab7145f1945da0b527b0781/tests/test_effects.py#L210)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> s/few //
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

bmcfee:
Thanks @dpwe and @carlthome for CR!  I think I've hit all the major points.

I've also opened a new pr #432 to rename the `n_fft` parameter in `rmse`, which I'd like to merge first and then update this one accordingly.

Are there any other points that need attention here?

dpwe:
Reviewed 1 of 2 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[librosa/effects.py, line 453 at r5](https://reviewable.io:443/reviews/librosa/librosa/428#-KVaVMV_BC90nJWiMu-q:-KVaVMV_BC90nJWiMu-r:b-jbmmd3) ([raw file](https://github.com/librosa/librosa/blob/d70742e6bd225995af739b1a90dff5c79f361f47/librosa/effects.py#L453)):*
> ```Python
>     index : np.ndarray, shape=(2,)
>         the slice of `y` corresponding to the non-silent region:
>         `y_trimmed = y[index]`.
> ```

Now it's not a slice, so y_trimmed = y[index[0]:index[1]]

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

bmcfee:
Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[librosa/effects.py, line 504 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KVVPUx-_fCS77xrlLEv-r2-504:-KVasqbBMHHrdOSl3R2o:b-5txwq2) ([raw file](https://github.com/librosa/librosa/blob/506626a840bbad4e837312ab473b716da1811b1c/librosa/effects.py#L504)):*
<details><summary><i>Previously, carlthome (Carl Thomé) wrote…</i></summary>
> Instead of inserting the first frame and last if they had high energy, isn't in better to pad the argument to the diff call with low energy frames instead? Then you could remove a lot of lines. E.g. something like:
> 
> 
> 
> ```python
> 
> zs = np.zeros((logp.shape[0], 1))  # Or transposed, I'm not sure.
> 
> edges = np.flatnonzero(np.diff((~np.concat((zs, logp, zs)).astype(int)))
> 
> ```
> 
> 
> 
> Maybe has to do with performance?
</details>
This was discussed in the reviewable thread with @dpwe -- I'd prefer to leave it as is.

---

*[librosa/effects.py, line 453 at r5](https://reviewable.io:443/reviews/librosa/librosa/428#-KVaVMV_BC90nJWiMu-q:-KVasogYxn-ZkGknNuTJ:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/d70742e6bd225995af739b1a90dff5c79f361f47/librosa/effects.py#L453)):*
<details><summary><i>Previously, dpwe (Dan Ellis) wrote…</i></summary>
> Now it's not a slice, so y_trimmed = y[index[0]:index[1]]
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

bmcfee:
Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[librosa/effects.py, line 399 at r2](https://reviewable.io:443/reviews/librosa/librosa/428#-KVVPUx-_fCS77xrlLEv-r2-399:-KVeN0lNkm5uPMo1_2hn:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/506626a840bbad4e837312ab473b716da1811b1c/librosa/effects.py#L399)):*
> Agreed; `frame_length` would match `util.frame`, but it propagates `n_fft` because that's how it's called in `rmse`.
> 
> 
> 
> Of course, we could (should) rename the parameter in `rmse` as well, since it's no longer accurate.  I'll handle that in a separate PR.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

dpwe:
:lgtm:

---

Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428#-:-KVeQuZkyQ3b5lU2cWxA:bnfp4nl)*

bmcfee:
Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

bmcfee:
I added two more sets of tests to `split` to hit some edge cases that we weren't covering before.  Once it passes, I think we're good to merge.

bmcfee:
Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/428)*

Issue #460, opened 2016-12-05
librosa.core.load kwarg res_type 


Cortexelus:
`librosa.core.load` should take keyword argument `res_type` for `librosa.core.resample`

The default `res_type` is `kaiser_best` but usually I want `kaiser_fast` which loads about 2.7x faster  when resampling from 44100 to 22050, and doesn't harm accuracy of my tasks. For a 3-4 minute song that's 10 seconds of load time saved.

bmcfee:
I see no problem Ruth adding that parameter. Do you want to make a pr?

PR #462, opened 2016-12-06
res_type kwarg added to librosa.core.load()

Cortexelus:
#460

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/462)


bmcfee:
Please add a test function for this.  It can be something dead simple (allclose-test against explicit load-without-resample + resample with a given res_type), but it should be there.

bmcfee:
Merging this for now -- I'll add tests sparately.

PR #471, opened 2016-12-09
extensions to normalization

bmcfee:
Implements #318 : 

- [x] `threshold` parameter to set an explicit normalization threshold
- [x] `fill_value` parameter to determine fill behavior for small-norm entries
- [x] throw an exception if input is not finite
- [x] expand documentation

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/471)

### DIFF
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -536,6 +536,25 @@ def axis_sort(S, axis=-1, index=False, value=None):
 def normalize(S, norm=np.inf, axis=0, threshold=None, fill=None):
     '''Normalize an array along a chosen axis.
 
+    Given a norm (described below) and a target axis, the input
+    array is scaled so that
+
+        `norm(S, axis=axis) == 1`
+
+    For example, `axis=0` normalizes each column of a 2-d array
+    by aggregating over the rows (0-axis).
+    Similarly, `axis=1` normalizes each row of a 2-d array.
+
+    This function also supports thresholding small-norm slices:
+    any slice (i.e., row or column) with norm below a specified
+    `threshold` can be left un-normalized, set to all-zeros, or
+    filled with uniform non-zero values that normalize to 1.
+
+    Note: the semantics of this function differ from
+    `scipy.linalg.norm` in two ways: multi-dimensional arrays
+    are supported, but matrix-norms are not.
+
+
     Parameters
     ----------
     S : np.ndarray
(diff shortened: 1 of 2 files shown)


bmcfee:
@stefan-balke Hey, I finally got around to implementing this.  The threshold setting was pretty easy.

Fill has a few quirks though.  

- API is `fill = {None, False, True}`.  `None` is the default behavior (leave untouched); `False` means fill with zeros, `True` means fill uniformly such that the corresponding slice normalizes to 1.

- Uniform fill is impossible when `norm=0`: any uniform fill with a 0 norm will always have 0-norm = `S.shape[axis]`.  The only way to get 0-norm=1 is to fill exactly one entry and zero the rest, but this feels arbitrary to me.  For now, it falls back to fill with zeros, which violates the semantics `norm(S_out, axis) == 1`.  It probably makes more sense to raise an exception in this case though.

- Strange things can happen when the input contains non-finite values.  If `fill=True`, then all non-finites (and their containing slices) are treated as if the norm is below the threshold.  This is mainly due to convenience of implementation, but I think the logic here is reasonable if you interpret these slices as "bad".  OTOH, if `fill=False`, non-finites are set to `NaN`.

So, two questions:

- Should we require `S` to be finite-everywhere?  I tend to think yes.
- Should we forbid the combination `fill=True, norm=0`?  I also think the answer should be yes.

bmcfee:
> Should we require S to be finite-everywhere? I tend to think yes.
> Should we forbid the combination fill=True, norm=0? I also think the answer should be yes.

I made an executive decision on both of these.

Justification for S being finite: if not, normalization generally means dividing inf/inf, nan/nan, or -inf/inf, all of which are nan.  A case could be made for allowing +-inf when normalizing with l0, but this seems like a rather unimportant edge case to me.

For the second point, since uniform filling is impossible in that case, we shouldn't try.

I think this one's ready for CR.

stefan-balke:
:lgtm:

Is this now working with arbitrary tensor sizes? At least you nan-search-strategy would allow for arbitrary axis. If so, we should have tests for this or state the limit in the docs.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/471#-:-KYcEA3LTSJnGaq0UDT6:b-hioj8e)*

bmcfee:
> Is this now working with arbitrary tensor sizes?

I think it already did?  Quick check:
```python
In [15]: x = np.random.randn(3, 4, 5)

In [16]: np.max(np.abs(librosa.util.normalize(x, axis=0)), axis=0)
Out[16]: 
array([[ 1.,  1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.,  1.]])

In [17]: np.max(np.abs(librosa.util.normalize(x, axis=1)), axis=1)
Out[17]: 
array([[ 1.,  1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.,  1.]])

In [18]: np.max(np.abs(librosa.util.normalize(x, axis=2)), axis=2)
Out[18]: 
array([[ 1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.]])

In [19]: np.max(np.abs(librosa.util.normalize(x, axis=None)))
Out[19]: 1.0
```
You're right that tests and docs could be expanded to describe this.

bmcfee:
Actually, tests already cover the `ndim=3` case.

I think this is done.  Anyone care to CR?

stefan-balke:
I'm fine with the code. LGTM

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/471#-:-KYexMLyYarVM11L0ywg:b-ll8c44)*

bmcfee:
Reviewed 1 of 2 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/471)*

Issue #269, opened 2015-10-18
Refactor and improve tempo estimator


bmcfee:
The current tempo estimator is based on the analysis of the first 90 seconds (by default) of the onset strength envelope.  This seems somewhat brittle, and doesn't scale to time-varying tempo estimation #44 .

It shouldn't be difficult to generalize and apply the same tempo inference algorithm to the local onset autocorrelation/tempogram #247 to get time-varying tempo estimates.  Alternately, it could be applied to the temporal average of the tempogram to get a more robust estimator.

bmcfee:
I took a first cut at this, and it works pretty well.  Example notebook is [here](https://gist.github.com/bmcfee/524b452a92c74f17e004).  I've added a couple of debug plots to illustrate what's going on internally between the old method (estimate_tempo, copied from librosa.beat) and the new method.

The new method is essentially identical, except that x_corr estimate is now done by averaging (or somehow aggregating) over the tempogram, rather than taking a 90sec window out to compute the autocorrelation.  This has the benefit of being robust to padding or fade-in effects.

Since this represents a major change in functionality, I think we should punt implementing it until 0.5. Delaying this one has the added benefit of allowing us some flexibility in the API, since some of the parameters of the current method no longer apply.

bmcfee:
Note: this may also be made a bit more modular by using the harmonic estimator from #297 to do more intelligent aggregation of tempo harmonics in resolving octave errors.

bmcfee:
Update:

I have a time-varying tempo estimator living in a notebook now.  It works reasonably well, and supports both aggregation (single tempo) and dynamic tempo estimation.

I think the way to go about this is to drop it in to `beat` as a new function which is called internally from the beat tracker, and deprecate `estimate_tempo`.

Variable-tempo beat tracking #44 can then be implemented as a separate PR.

bmcfee:
Fixed by merging #479

Issue #44, opened 2013-12-29
variable tempo beat tracking


bmcfee:
Beat tracking DP currently takes in a global tempo estimate.  

This could easily be modified to take an array `tempo` the same shape as the onset function, where `tempo[t]` is the tempo at time `t`.

This would also need a dynamic tempo estimator.

bmcfee:
The bottleneck here was speed -- the DP gets slow when the previous step window needs to be changed at each step.

Numbaficiation #157 could help with this though.

bmcfee:
Closing this one out -- PLP provides a variable-tempo beat tracker.  It's not the same as the DP-based method, which we could generalize, but it would be much slower.

PR #479, opened 2016-12-16
Tempo refactor

bmcfee:
Implements #269 - dynamic tempo estimation

This function replaces `estimate_tempo`, which is deprecated in 0.5.

Static tempo estimation is possible by supplying an `aggregate` function to the estimator, which combines onset autocorrelation frames prior to estimating peaks.

Once this is merged, variable-tempo beat tracking #44 is possible, but maybe still not a great idea.

TODO:

- [x] unit tests
- [x] documentation
- [x] fix a numpy index warning

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/479)

### DIFF
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -7,6 +7,13 @@ Beat and tempo
    :toctree: generated/
 
    beat_track
+   tempo
+
+Deprecated
+----------
+.. autosummary::
+   :toctree: generated/
+
    estimate_tempo
 """
 
@@ -17,9 +24,10 @@ from . import cache
 from . import core
 from . import onset
 from . import util
+from .feature import tempogram
 from .util.exceptions import ParameterError
 
-__all__ = ['beat_track', 'estimate_tempo']
+__all__ = ['beat_track', 'tempo', 'estimate_tempo']
 
 
 def beat_track(y=None, sr=22050, onset_envelope=None, hop_length=512,
@@ -174,10 +182,10 @@ def beat_track(y=None, sr=22050, onset_envelope=None, hop_length=512,
 
     # Estimate BPM if one was not provided
     if bpm is None:
-        bpm = estimate_tempo(onset_envelope,
-                             sr=sr,
-                             hop_length=hop_length,
-                             start_bpm=start_bpm)
+        bpm = tempo(onset_envelope=onset_envelope,
+                    sr=sr,
+                    hop_length=hop_length,
+                    start_bpm=start_bpm)[0]
 
     # Then, run the tracker
     beats = __beat_tracker(onset_envelope,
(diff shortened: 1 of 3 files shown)


bmcfee:
Getting some low-frequency modulation artifacts in the time-varying estimator, probably due to windowing the OENV.  This leads to periodic octave errors in some test cases.

bmcfee:
Reviewed 1 of 3 files at r2.
Review status: 1 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/479)*

bmcfee:
@dpwe any chance you could take a look at this one some time?  TLDR is that it generalizes the previous tempo estimator to span the full track (either dynamic or in aggregate), rather than taking a 90-second chunk.

Otherwise, the logic is basically unchanged, but I made it a new function because the API is sufficiently different from `estimate_tempo`.

The one question-mark left over is dealing with modulation of the onset autocorrelation due to windowing in `tempogram`.  This only really matters for dynamic tempo estimation, which we don't currently use anywhere, but I'd like to get it right before merging.

dpwe:
I'll take a look.

  DAn.

dpwe:
From the ipynb:

    # Choose the best peak out of .33, .5, 2, 3 * start_period
    candidates = np.argmax(x_peaks * x_corr) * np.asarray([1./3, 0.5,
1.0, 2, 3])
    candidates = candidates[candidates < ac_window].astype(int)
    best_period = np.argmax(x_corr[candidates])

    return 60.0 * fft_res / candidates[best_period]



What happens here?  We get the index of the greatest peak in the
weighted x_corr, and we scale it by the plausible multiples.  Those
that are within the tolerable range are kept and converted to ints.
Then we choose from among that list the one with the largest x_corr
value.


When is this not the 1.0 scaling?  When there's a large value in the
x_corr function at one of these super/sub multiples, but it's not a
local max because it's on the shoulder of a competing peak?  Does this
actually happen?  Wouldn't that competing peak have won the original
argmax search?  Unless it was outside the range of the search?


I remember putting something like this into my 2007 Matlab code
because it improved performance on the MIREX Moelants/McKinney test
data, but if it was exactly this, I'm surprised it ever helped.


  DAn.

bmcfee:
Thanks for the CR.

And yeah -- that harmonic check was ported over from your original implementation.  I agree that it isn't/shouldn't be necessary, and it's much cleaner to use the prior to weight down the extremes (which are all that could be selected here anyway).

I cut it out, and the unit tests still pass, so I'm happy to go with the simplified version.

---

Review status: 1 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/479#-:-K_1PycBjZc_1w6oHxou:b-lo6d2h)*

bmcfee:
I think this one's good to go.  Any objections to merging?

---

Review status: 1 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/479#-:-K__DBmx1ZZG-e_-vpy0:bjzj51n)*

bmcfee:
Reviewed 1 of 3 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/479)*

Issue #122, opened 2014-11-06
Multitrack support


bmcfee:
The current loader has an option to convert stereo to mono, assuming interleaving of two channels.  This could be generalized to support general mixing of multi-tracks by using the `frame` function.

romino666:
@piem I could extract a stereo track in multitrack using librosa, but I would like the multitrack to be unfiltered.

lostanlen:
@romino666 if by unfiltered you mean non-resampled, consider passing `sr=None`

https://librosa.org/blog/2019/07/17/resample-on-load

romino666:
@lostanlen thanks, unfiltered and surely uncut

romino666:
For Example, if i can import a Michael Jackson song, like Dirty Diana, i can try to convert stereo to mono using `Frame` function to make complete "non-resampled" multitracks with 48 tracks with this solution `sr=None`

romino666:
@dpwe i can convert Dirty Diana into non-resampled multitracks next week.

romino666:
Microsoft Visual Studio Code is free ?

PR #523, opened 2017-03-08
librosa/core/audio.py: fix for files with more than 2 channels

piem:
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/523)

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -147,7 +147,7 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
         y = np.concatenate(y)
 
         if n_channels > 1:
-            y = y.reshape((-1, 2)).T
+            y = y.reshape((-1, n_channels)).T
             if mono:
                 y = to_mono(y)
 


bmcfee:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/523)*

bmcfee:
Thanks for this one.  Quick question: do you have any example files that could be used to test this?  Last time we looked at this general issue was #122 , and decided it wasn't worth pursuing.  OTOH, there's no reason not to be general here, but I'd like for there to be a test in place to demonstrate that the sample ordering is interpreted properly.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/523#-:-KepSLSGjYtcNt5JjTce:bpl5c4c)*

piem:
I would generate it, maybe upmixing a stereo file on three or more channels. I use this one to test for perfect looping:

 `$ sox -r 44100 -c 6 -b 16 -n 44100Hz_100f_sine441_5.1.wav synth 100s sine 441 vol 0.9`

Note that I didn't check if all audioread backends support multichannel.

bmcfee:
> Note that I didn't check if all audioread backends support multichannel.

The ffmpeg decoder seems to handle it, at least.  The audioread API definitely supports multi-channel.

Would you mind adding that synthetic example as a fixture and a small unit test that verifies shape and contents?  The io stuff will all eventually be deprecated anyway, but we may as well do things right for now.

piem:
hi @bmcfee, 

Sure, I'll try to find time to do that when I find some more time.

I tested and with my version of ffmpeg (3.2.4-1~bpo8+1 from debian jessie backports) and the latest audioread HEAD, i still get one channels on a 3 channel file:

```
$ python decode.py ~/drum_3chans.wav
Input file: 1 channels at 44100 Hz; 2.9 seconds.
Backend: ffdec
$ sndfile-info ~/drum_3chans.wav | egrep -i "^(Channels|Duration)"
Channels    : 3
Duration    : 00:00:02.953
```

The problem comes from audioread/ffdec.py, which parses the output of `ffmpeg -i filepath`, [looking for an integer] using a regex. But the channel information gets displayed as `2.1`, unlike with `ffprobe`.

```
$ ffmpeg -i ~/drum_3chans.wav 2>&1 | grep 44100
    Stream #0:0: Audio: pcm_s16le ([1][0][0][0] / 0x0001), 44100 Hz, 2.1, s16, 2116 kb/s
$ ffprobe -i ~/drum_3chans.wav 2>&1 | grep 44100
    Stream #0:0: Audio: pcm_s16le ([1][0][0][0] / 0x0001), 44100 Hz, 3 channels, s16, 2116 kb/s
```

bmcfee:
@piem thanks for digging into it.  I guess the test case will have to wait until the next audioread release, but otherwise this looks good.

PR #402, opened 2016-08-15
Removed auto-import of display submodule

bmcfee:
By popular demand #258 #281 #343 , the display submodule is no longer imported automatically.

I've also relaxed the installation requirements so that matplotlib is an optional dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/402)

Issue #343, opened 2016-04-30
do not always import matplotlib


howthebodyworks:
librosa always imports `matplotlib`, even if you are doing pure analysis and have no interest in graphing.

This is inconvenient for batch soundfile analysis, as matplotlib has a large memory footprint, and an even larger file handle footprint. Moreover, if you run matplotlib in certain environments (I'm using tornado to analyse soundfiles on request) it seems to leak file handles.

Whilst the file handle leak is not, AFAICT, librosa's problem per se, I've run into this issue in enough circumstances to feel that matplotlib should be considered something to be imported only when strictly required, and the memory problem alone is annoying when processing thousands of files in a multiprocess batch.

At first glance it seems that avoiding this could be as simple as only importing `librosa.display` on demand as opposed to in the root `__init__.py`.

Would it be worth me filing a pull request to do that?

howthebodyworks:
related: #338

howthebodyworks:
And while I'm here, let me mention also briefly my opinion that librosa is awesome.

bmcfee:
This is redundant with #281 #314 #258 .

For the next release, we're not going to drop the import because it would break API.  I may revisit this for 0.5.0 since enough people have complained about it.

In the meantime, if it's really causing trouble, you could always suppress the import by mocking the module before loading librosa.  For example:

```python
import mock
import sys
sys.modules.update((mod_name, mock.Mock()) for mod_name in ['matplotlib', 'matplotlib.pyplot'])

import librosa
...
```

TimSchmeier:
FYI: Instead of using the above hack this solved a number of librosa/matplotlib import issues for me

```
import matplotlib
matplotlib.use('Agg')
```

carlthome:
I think it's bad practice for a package to import as needed. It adds unnecessary complexity and unexpected behavior. A package's memory requirements should not fluctuate depending on what functions you import. I expect a bunch of future issues where people wonder why they get OutOfMemory exceptions just because they called librosa.display.waveplot or specshow. If there's a strong use case for not having visualizations builtin then librosa should be split into two completely separate packages. One for analysis and one for visualizations, sharing a mutual interface.

bmcfee:
> I think it's bad practice for a package to import as needed.

Agreed.  Dynamic submodule import is different though, and is standard practice in python (see, eg, sklearn).



[...]

PR #540, opened 2017-04-01
add display

wesky93:
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/540)


bmcfee:
Removing the automatic import of `display` was a deliberate choice, as discussed in #402 and #343.

wesky93:
oh I got it!! I doesn't think like that problem...
thank you!

PR #550, opened 2017-04-09
Unnorm mel

dpwe:
Add a norm= option to filters.mel to allow the mel filters to have max value = 1 for all bands (norm=np.inf or None) instead of the default unit-sum (norm=1).

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/550)


dpwe:
OK, I hadn't sync'd my fork.  Let me try this again.

Issue #539, opened 2017-03-31
unit test for chroma_cens is failing except on py3.4


bmcfee:
[Tagging @stefan-balke ]

I haven't investigated thoroughly, but I suspect that something is going wrong in the underlying BLAS system that can differ on travis by python version.  Either we need to relax the test criteria, or work out if there's a numerical instability that can be resolved within the cens calculation.

stefan-balke:
will check.

bmcfee:
I just tested this out on my local conda environments:

- 2.7
  - numpy 0.11.3, scipy 0.18.1
  - test_cens passes
- 3.4
  - numpy 0.11.3, scipy 0.18.1
  - test_cens passes
- 3.5
  - numpy 0.12.1, scipy 0.19.0
  - test_cens fails

The culprit appears to be scipy: regressing my 3.5 environment back to scipy 0.18.1 fixes the problem.

I added a debug print to the test assertion, and it reads as follows:
```
======================================================================
FAIL: test_features.test_cens
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/bmcfee/miniconda/envs/py35/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/bmcfee/git/librosa/tests/test_features.py", line 644, in test_cens
    atol=1e-15), np.abs(ct_chroma_cens['f_CENS'] - lr_chroma_cens)
AssertionError: [[ 0.          0.          0.          0.          0.          0.          0.
   0.          0.          0.          0.          0.          0.          0.
   0.          0.          0.          0.          0.          0.          0.
   0.          0.          0.        ]
 [ 0.          0.          0.          0.          0.          0.          0.
   0.          0.          0.          0.          0.          0.          0.
   0.          0.          0.          0.          0.          0.          0.
   0.          0.          0.        ]
 [ 0.          0.          0.          0.          0.00510423  0.05317791
   0.11701583  0.01799788  0.05074901  0.10170009  0.07319766  0.0084889
   0.04390135  0.05750489  0.03900799  0.02897911  0.10282193  0.05752425
   0.00286278  0.00159912  0.03659663  0.03217601  0.01264462  0.        ]
 [ 0.          0.          0.          0.          0.00127606  0.0074886
   0.00592027  0.00840787  0.00911527  0.          0.          0.00101242
   0.01000203  0.01432603  0.00554525  0.01845597  0.00380086  0.          0.
[...]

PR #551, opened 2017-04-09
Mel norm

dpwe:
Adds a norm=... option to filters.mel, the mel-weight-matrix generation routine.

The default behavior (norm=1) continues to be "sum normalization" (as in Slaney's mfcc implementation), where the broader, high-frequency mel bands have lower peak amplitudes.

But now norm=None or norm=np.inf makes each mel weight row aim for a max value of 1.0 (although the actual values in the row may not reach 1, depending on the alignment of the underlying triangle and the fft bin samples.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/551)


dpwe:
My added tests simply add three new golden-value mat files in the tests/data directory, with an extra field to specify the norm parameter.  

A weakness of this is that no-one can easily tell what is being tested.

Perhaps, instead, I should have used one of the existing golden test files (or all of them?) and called mel with all 4 norm options (not specified, and ={None, 1, np.inf}), then applied simple sanity tests in each case?

The tight norm=None sanity check (that we expect the peak value to be exactly 1) doesn't quite work, because that will only happen if the Mel center frequency (the peak of the triangle) falls exactly on an FFT bin.  But I can probably come up with something that checks it isn't the same as norm=1.

bmcfee:
Note: test builds are going to continue failing until #539 gets fixed.  Let me know when it's ready for CR and I'll inspect the relevant tests manually.

dpwe:
I think this is good to go, depending on what you think of the tests.  

To help, I've put the notebook I used to generate the new test data files at
https://github.com/dpwe/librosa-mel-test-gen

bmcfee:
Thanks for doing this!  Everything looks good. I have a couple of small requests below, but nothing major.

---

Reviewed 3 of 5 files at r1.
Review status: 3 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[librosa/filters.py, line 256 at r1](https://reviewable.io:443/reviews/librosa/librosa/551#-KhN2bFWcbrCgALL_cYT:-KhN2bFWcbrCgALL_cYU:b9ghowg) ([raw file](https://github.com/librosa/librosa/blob/93a99cd89da1f54ad70ba69796c5ddd30ba5eb53/librosa/filters.py#L256)):*
> ```Python
>         # Slaney-style mel is scaled to be approx constant energy per channel
>         enorm = 2.0 / (mel_f[2:n_mels+2] - mel_f[:n_mels])
>         weights *= enorm[:, np.newaxis]
> ```

I suppose this isn't going to be numerically equivalent to `util.normalize(weights, axis=1, norm=1)` because it's computing the exact area rather than the sampled approximate.  Is that why we don't use normalize here?

For max norm, it's not clear to me that the peak is always going to be 1 (so that max-norm is a no-op), especially when we get near the edges (0, nyq).

---

*[tests/test_filters.py, line 135 at r1](https://reviewable.io:443/reviews/librosa/librosa/551#-KhN39KaadhUzeWfF-Hf:-KhN39KaadhUzeWfF-Hg:bnp2vqn) ([raw file](https://github.com/librosa/librosa/blob/93a99cd89da1f54ad70ba69796c5ddd30ba5eb53/tests/test_filters.py#L135)):*
> ```Python
> 
>     for infile in files('data/feature-melfbnorm-*.mat'):
>         yield (__test_with_norm, infile)
> ```

For completeness, it would be nice to have the notebook that generated the new fixtures included in the repo, preferably with a short header that explains what it does and how to use it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/551#-:-KhN3MbV7rl8ytLA-0tA:b-n4ak4l)*

dpwe:
Review status: 3 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[librosa/filters.py, line 256 at r1](https://reviewable.io:443/reviews/librosa/librosa/551#-KhN2bFWcbrCgALL_cYT:-KhSjfvv7wIluaAbm0FV:ba9h385) ([raw file](https://github.com/librosa/librosa/blob/93a99cd89da1f54ad70ba69796c5ddd30ba5eb53/librosa/filters.py#L256)):*

I suppose this isn't going to be numerically equivalent to `util.normalize(weights, axis=1, norm=1)` because it's computing the exact area rather than the sampled approximate.  Is that why we don't use normalize here?

For max norm, it's not clear to me that the peak is always going to be 1 (so that max-norm is a no-op), especially when we get near the edges (0, nyq).
</blockquote></details>

Yes, that's right.  The underlying shape on a "continuous frequency" axis is a triangle going from zero at mel_f[-1] to one at mel_f[0], then back to zero at mel_f[+1].  But, especially for low bands, the spacing of mel_f is small compared to the FFT bins, and so normalizing on the sampled points rather than the underlying shape would give quite different results.

I don't know how the algorithm came out looking like this (which I agree is less than completely transparent), but the i'th row of "ramps" is a vector of mel_f[i] - <all the FFT bin center frequencies>, so a linear sequence with a negative slope that crosses zero where the FFT bin is closest to the target mel frequency (although in general there will not be an actual zero in the vector).  `lower` and `upper` scale the rows centered on mel_f[i-1] and mel_f[i+1] so that the value at mel_f[i] would be exactly 1.0 if it existed.  Then we intersect these two sampled linear slopes and a floor at zero to create the (sampled) triangle.  

The construction really does produce an underlying triangle with a max at 1.0, but perhaps it should be easier to read?  I think 'ramps' is something you introduced to cleverly halve the number of times we calculate the difference between each mel_f and fftfreqs.  But it could be rewritten to be more transparent, as in the original matlab:

```matlab
  % lower and upper slopes for all bins
  loslope = (fftfrqs - fs(1))/(fs(2) - fs(1));
  hislope = (fs(3) - fftfrqs)/(fs(3) - fs(2));
  % .. then intersect them with each other and zero
  wts(i,1+[0:(nfft/2)]) = max(0,min(loslope, hislope));
```

Not sure how we weight efficiency/succinctness with transparency of the code here.

I think the edges are not a big deal.  We calculate n_mel + 2 frequencies, for n_mel center frequencies, then one extra at each end.  Each triangle is defined between three successive mel points, so each row of the matrix is a complete triangle.

---

*[tests/test_filters.py, line 135 at r1](https://reviewable.io:443/reviews/librosa/librosa/551#-KhN39KaadhUzeWfF-Hf:-KhSqDCFKYscLkVUhXup:bgvs52y) ([raw file](https://github.com/librosa/librosa/blob/93a99cd89da1f54ad70ba69796c5ddd30ba5eb53/tests/test_filters.py#L135)):*

For completeness, it would be nice to have the notebook that generated the new fixtures included in the repo, preferably with a short header that explains what it does and how to use it.
</blockquote></details>

will do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/551)*

bmcfee:
Reviewed 2 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/551)*

dpwe:
I added https://github.com/dpwe/librosa/blob/mel-norm/tests/make_mel_norm_test_data.ipynb

---

Review status: 5 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/551#-:-KhTKHiN8rUU-9VyDCTc:bg3t27g)*

bmcfee:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/551#-:-KhTTsWXOHJ0LSJDvrGq:bnfp4nl)*

bmcfee:
Thanks @dpwe ! Merged.

PR #553, opened 2017-04-13
fixed #552, error in spectral bandwidth shape

### LINKED ISSUE #552: Problem with spectral_bandwidth having only one frame
Hello folks,

I try to use the librosa library for some library I want to implement, which essentially extracts some features from audio files. These files can be very short (including only one frame when transforming it with say stft)

So I stumpled about some unexpected behaviour when using the spectral_bandwith feature.

Let S be some spectrogram (S[bins][frames])
then the following calls will produce (for me) unexpected output:

`res_expected_1 = spectral_bandwidth(sr=sr, S=S)`
`res_expected_2 = spectral_bandwidth(sr=sr, S=S[:,0:2])`
`res_unexpected = spectral_bandwidth(sr=sr, S=S[:,0:1])`

Expected results return for each frame a number. 
So for _res_expected_2_ exactly 2 
and for _res_expected_1_ exactly _len(S[0])_
while res_unexpected produces _bin_num_    results. Which I don't think is what it is intented to do.

in addition I don't know why the results are wrapped in an array like [[frames]] instead of just [frames]

I hope someone can help me out a bit;
Thanks for your time meanwhile

bmcfee:
This is a bugfix for #552, where a single-column input spectrum produces output of the wrong shape.

The fix is simple: replace `squeeze(centroid)` with `centroid[0]`.  I've also added a new test to cover this case.

This should be merge-ready as soon as tests pass.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/553)

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -244,9 +244,9 @@ def spectral_bandwidth(y=None, sr=22050, S=None, n_fft=2048, hop_length=512,
         freq = fft_frequencies(sr=sr, n_fft=n_fft)
 
     if freq.ndim == 1:
-        deviation = np.abs(np.subtract.outer(freq, np.squeeze(centroid)))
+        deviation = np.abs(np.subtract.outer(freq, centroid[0]))
     else:
-        deviation = np.abs(freq - np.squeeze(centroid))
+        deviation = np.abs(freq - centroid[0])
 
     # Column-normalize S
     if norm:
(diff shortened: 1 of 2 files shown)


bmcfee:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/553)*

PR #559, opened 2017-04-14
fixed #529, explicit length for istft

### LINKED ISSUE #529: Explicit length for inverse transform outputs
The inverse transforms (currently `istft`, but in the future `icqt` #165 and maybe `ifmt` #231 )  produce time series according to whole number multiples of the frame length, which makes sense.

However, the forward transform can lose a partial frame (or pad out) so that the inverse length does not match the input signal length.  This can be fixed by the following sequence of steps:

```python
>>> D = librosa.stft(y)
>>> y2 = librosa.istft(D)
>>> y2 = librosa.util.fix_length(y2, len(y))
```

but it's A) clumsy, and B) prone to cause errors if the user forgets to do it.

A simple change to the API could short-circuit this for convenience, eg:
```python
>>> y2 = librosa.istft(D, n=len(y))
```

bmcfee:
#### Reference Issue

Fixes #529 

Users can now specify the target signal length for inverse STFT.

If `length` is shorter than the expected signal length, it is trimmed.

If `length` is longer than the expected signal length, it is zero-padded.

#### Any other comments?

Note that with centering, the forward transform first pads the signal on both ends, and the inverse usually trims off the ending bit corresponding to the padding.  However, if the user calls with `length` greater than the expected signal length, we retain the padded fragment, which when combined with frame windowing, results in a smooth taper to zero before padding kicks in.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/559)

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -179,7 +179,7 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
 
 @cache(level=30)
 def istft(stft_matrix, hop_length=None, win_length=None, window='hann',
-          center=True, dtype=np.float32):
+          center=True, dtype=np.float32, length=None):
     """
     Inverse short-time Fourier transform (ISTFT).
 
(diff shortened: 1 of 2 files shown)


bmcfee:
@dpwe I tagged you on this one just to get some input on how to handle center padding when trimming/padding to a target length.  I'm pretty confident the code is correct, so this doesn't need a full-blown CR.  (Also cool if you don't have time for this.)

bmcfee:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/559)*

PR #563, opened 2017-04-24
expose padding mode to user in stft/ifgram/cqt

### LINKED ISSUE #531: Allow user-specified padding mode in STFT
When STFT's are centered (by default), the signal is first padded and then analyzed.  The padding mode is hard-coded to reflection, but this is probably something that should be controlled by the user.

bmcfee:
#### Reference Issue
Fixes #531 

When STFT is center-aligned, we previously padded by reflection.

This PR allows the user to change the padding mode to anything supported by `np.pad`.  

#### Any other comments?

Note that the fancier modes (eg `linear_ramp`) require additional arguments which we don't expose.  Users desiring fancy padding should pad explicitly before calling `stft`.

The `pad_mode` parameter is also added to `ifgram` *EDIT* and `cqt`, `hybrid_cqt`, `pseudo_cqt`.

No additional action is necessary in `istft`.

~~~Question: should we expose this in `cqt` as well?~~~

<!-- Reviewable:start -->
---
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/librosa/librosa/563)
<!-- Reviewable:end -->

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -25,7 +25,8 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         bins_per_octave=12, tuning=0.0, filter_scale=1,
         norm=1, sparsity=0.01, window='hann',
         scale=True,
-        real=util.Deprecated()):
+        real=util.Deprecated(),
+        pad_mode='reflect'):
     '''Compute the constant-Q transform of an audio signal.
 
     This implementation is based on the recursive sub-sampling method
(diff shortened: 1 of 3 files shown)


bmcfee:
I think this one's good to go as well; eyeballs would be nice but not necessary.

bmcfee:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/563)*

PR #564, opened 2017-04-26
rasterized specshow by default

### LINKED ISSUE #562: rasterize specshow
#### Description
In 0.5, we switched `specshow` from using `imshow` under the hood to `pcolormesh`.  While this gives us a bit more flexibility for pixel layout, it does result in heavyweight figures.

I find myself manually [rasterizing](http://www.astrobetter.com/blog/2014/01/17/slim-down-your-bloated-graphics/) `specshow` plots often, to the point where I'm wondering if it should be baked in as a default option, enabled by default, but easy to override by `librosa.display.specshow(..., rasterized=False)`.

What do folks think?

bmcfee:
#### Reference Issue
Fixes #562 

`display.specshow` now rasterizes color meshes by default.  This reduces the memory footprint of plots, especially when exported to vector graphics formats (eg, pdf or svg).

#### Any other comments?

There should be no noticeable side effects to the user.  Users that want to override this behavior can call `specshow` with `rasterized=False`.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/564)

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -545,6 +545,11 @@ def specshow(data, x_coords=None, y_coords=None,
     kwargs : additional keyword arguments
         Arguments passed through to `matplotlib.pyplot.pcolormesh`.
 
+        By default, the following options are set:
+
+            - `rasterized=True`
+            - `shading='flat'`
+            - `edgecolors='None'`
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


bmcfee:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/564)*

PR #565, opened 2017-04-28
support negative delay in stack_memory. fixes #520

### LINKED ISSUE #520: Reverse stack_memory
It would be useful, and easy, to allow `feature.stack_memory` to pull from the future as well as the past.

This could be achieved either by allowing `n_steps` to be any non-zero integer (in which case, negative comes from the future), or by having a separate flag `reverse`.

bmcfee:
#### Reference Issue
Fixes #520 

This PR drops the requirement of `delay > 0` in `feature.stack_memory`.  A negative delay now corresponds to embedding from the future.  Delay of 0 still causes an exception.

#### Any other comments?

This one should be good to go.

At some point in the future, we should generalize this to support arbitrary dimensionality / axis targets.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/565)

### DIFF
--- a/librosa/feature/utils.py
+++ b/librosa/feature/utils.py
@@ -142,8 +142,12 @@ def stack_memory(data, n_steps=2, delay=1, **kwargs):
     n_steps : int > 0 [scalar]
         embedding dimension, the number of steps back in time to stack
 
-    delay : int > 0 [scalar]
-        the number of columns to step
+    delay : int != 0 [scalar]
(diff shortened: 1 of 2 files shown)


bmcfee:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/565)*

Issue #586, opened 2017-06-09
CQT Display Offset


stefan-balke:
#### Description

* Compute CQT.
* Display with `librosa.display.specshow()` and `y_axis='cqt_note'`
* Note axis has offset by one semitone.

#### Steps/Code to Reproduce

```Python
import librosa

cur_audio, fs = librosa.load('Systematic_Scale-C-Major_Meinard_middle.wav')
C = librosa.cqt(cur_audio, sr=fs)
librosa.display.specshow(librosa.amplitude_to_db(C, ref=np.max),
                         sr=fs, x_axis='time', y_axis='cqt_note')```
```

#### Expected Results

For my audio example, first note should be a C4.

#### Actual Results

A C4#
![screen shot 2017-06-09 at 15 06 41](https://user-images.githubusercontent.com/7326732/26976758-869bf1a4-4d25-11e7-9ca9-19858575c3c1.png)

I think the problem lies here:
https://github.com/librosa/librosa/blob/master/librosa/display.py#L855

Instead of
`core.cqt_frequencies(n+1, fmin=fmin, bins_per_octave=bins_per_octave)`

we need

`core.cqt_frequencies(n, fmin=fmin, bins_per_octave=bins_per_octave)`

The audio file can be obtained from here: https://github.com/stefan-balke/mpa-exc/blob/master/data/Systematic_Scale-C-Major_Meinard_middle.wav

bmcfee:
I'm not able to reproduce this directly.  Would you mind posting a complete example to generate the figure you posted?  (It looks like you've zoomed in on a region.)  Also, could you post the `librosa.show_versions()` output?

When I try to replicate this locally, everything appears to work properly.  In nbagg mode, the mouseover on the bottom row reports `C1` -- when I zoom in, it shows `C1+16`, which is just an artifact of the ticker placement in absolute coordinates.
```python
y, sr = librosa.load(librosa.util.example_audio_file(), offset=10, duration=2)

C = librosa.cqt(y=y, sr=sr)

plt.figure()
librosa.display.specshow(librosa.amplitude_to_db(C, ref=np.max),
                        sr=sr, x_axis='time', y_axis='cqt_note')
plt.xlim([0, 2])
plt.ylim([30, 60])
```
![image](https://user-images.githubusercontent.com/1190540/26977749-faf2b940-4cf6-11e7-829f-225d4f621c35.png)

stefan-balke:
Here you go.

[...]

PR #597, opened 2017-07-07
[CR] centered frequency bins in specshow

bmcfee:
#### Reference Issue

#586

This PR modifies the mesh coordinates in `display.specshow` so that CQT bins are centered at their filter frequencies, rather than edge-aligned.

This PR does not (yet) fix the other issue described in #586, where the tick locations are not centered on the bins.  I made an initial attempt at this, but it didn't quite work due to some quirks in how MPL's loglocator class works.  We might have to implement a custom tick locator class for this.

#### Any other comments?

I also fixed FFT display so that it properly returns `n+1` coordinates like all the other modes.  FFT coords are treated a bit differently though, since it doesn't quite make sense to center the DC component (`f=0`), so FFT bins remain left-aligned.  Anyone @stefan-balke @dpwe have ideas for how to handle this case?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/597)


bmcfee:
Thinking a bit more about this: do we want to set an explicit policy for how bin locations in specshow should be expected to work?  As mentioned above, the positioning for FFT and CQT are currently pretty different: FFT bins are positioned between `f[i]` and `f[i+1]`, where `f[i]` is the `i`th FFT frequency.  Alternatively, CQT bins are now positioned such that the (vertical) center of bin `i` lies at `c[i]` (the `i`th CQT basis frequency).

Overall, bin-centering makes more sense to me, since it behaves better with overlaid plots in natural coordinate space.  However, applying bin-centering to FFT would require either the vertical axis to go negative (so that the DC bin is centered at 0) or to truncate the DC bin to half-height.  I'm okay with the latter, as long as it's done consistently at the top-end for the nyquist bin.  Of course, having a couple of half-height bins also seems strange, but maybe it's worth it, and people won't notice anyway.

In general, the other axis scalings (tempo, chroma, mel) all seem to call for centered bins, since they're either on a log axis, or the responses of filters with a particular center frequency.

dpwe:
More to the point, drawing the d.c. bin as covering 0 to f_s / n_fft is
just wrong; it covers -f_s/(2n_fft) to +f_s/(2n_fft).

Matlab does indeed use a freq axis that goes negative.

Extending beyond 0..f_s/2 is consistent with having n_fft/2 + 1 bins.

But I'm ok with half-height bins instead.

  DAn.

bmcfee:
Note: the current failures on python 3.4 + numba are due to numba backdating numpy to 1.10 in the 3.4 packages.

bmcfee:
CQT and FFT coordinates are done and lgtm.

To fix: Mel and Tempo coordinates.

bmcfee:
Mel is done now, but tempo is a little strange.

The tempo frequency logic goes like:

```python
t[0] = inf
t[1] = 60 * sr / (hop_length)
t[2] = 60 * sr / (2 * hop_length)
t[k] = 60 * sr / (k * hop_length)
```
In log space (log axis scaling), we get:
```python
log(t[k]) = log(60 * sr / hop_length) - log(k)
```

bmcfee:
Reviewed 1 of 6 files at r1, 4 of 5 files at r2, 18 of 19 files at r4.
Review status: 23 of 24 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/597)*

bmcfee:
I think this one's good to go. The tempo scaler gets a bit weird at the high end since it needs to interpolate between [inf, t_max], but it otherwise seems to properly center the bins.

As mentioned above, the test failure is just a numba distribution issue, and irrelevant to this PR.  Unless there are objections ( @stefan-balke ? ) I'll merge this one tomorrow.

bmcfee:
Stefan's out on vacation; anyone else care to CR this?

bmcfee:
Reviewed 1 of 19 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/597)*

PR #598, opened 2017-07-17
Accelerated decompose.nn_filter

### LINKED ISSUE #516: Accelerating decompose.nn_filter
The nearest neighbor filter/kernel additive model filter is not as efficient as it could be.  Most of the lag is in fancy-indexing, which is used to support the `axis` parameter.

I hacked up a quick [gist](https://gist.github.com/bmcfee/fe0ce523df8f824a49848ba4d24187d8) that gives about a 7x speedup by using a combination of direct indexing of sparse matrix data arrays, and swapaxes in place of fancy indexing.

bmcfee:
#### Reference Issue
Fixes #516

This PR is a refactor / rewrite of the `nn_filter` operation to avoid using fancy indexing by accessing the sparse matrix index data structure directly.

#### Any other comments?

We get a pretty good speedup over the previous implementation.  Here's a quick benchmark on a 30-second fragment (`nn_filter` is the new version, `librosa.decompose.nn_filter` is the old version):

```python
In [1]: y, sr = librosa.load(librosa.util.example_audio_file(), offset=20, duration=30)
In [2]: S = librosa.amplitude_to_db(librosa.stft(y))
In [3]: R = librosa.segment.recurrence_matrix(S, sparse=True, metric='cosine', mode='affinity')

In [4]: %timeit f1 = nn_filter(S, rec=R, aggregate=np.median)
Out [4]: 1.26 s ± 17.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: %timeit f2 = librosa.decompose.nn_filter(S, rec=R, aggregate=np.median)
Out [5]: 5.42 s ± 186 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit f1 = nn_filter(S, rec=R, aggregate=np.average)
Out [6]: 184 ms ± 11.1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit f2 = librosa.decompose.nn_filter(S, rec=R, aggregate=np.average)
Out [7]: 5.28 s ± 227 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/598)

### DIFF
--- a/librosa/decompose.py
+++ b/librosa/decompose.py
@@ -509,24 +509,52 @@ def nn_filter(S, rec=None, aggregate=None, axis=-1, **kwargs):
                              'rec.shape={} for S.shape={}'.format(rec.shape,
                                                                   S.shape))
 
-    s_out = S.copy()
+    return __nn_filter_helper(rec.data, rec.indices, rec.indptr,
+                              S.swapaxes(0, axis), aggregate).swapaxes(0, axis)
 
-    index = [slice(None)] * s_out.ndim
 
-    for i in range(rec.shape[0]):
-        index[axis] = i
+def __nn_filter_helper(R_data, R_indices, R_ptr, S, aggregate):
+    '''Nearest-neighbor filter helper function.
+
+    This is an internal function, not for use outside of the decompose module.
+
+    It applies the nearest-neighbor filter to S, assuming that the first index
+    corresponds to observations.
+
+    Parameters
+    ----------
+    R_data, R_indices, R_ptr : np.ndarrays
+        The `data`, `indices`, and `indptr` of a scipy.sparse matrix
+
+    S : np.ndarray
+        The observation data to filter
+
+    aggregate : callable
+        The aggregation operator
+
+
+    Returns
+    -------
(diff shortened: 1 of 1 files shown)


bmcfee:
Note: the helper function does not `numba.jit` because it still uses `np.take` (not jittable) as well as helper functions passed by reference (`aggregate`).  The latter could be fixed by doing some scoping tricks to first jit-compile and cache a local function with the hard-coded aggregate op, but the former problem will have to wait until `numba` supports `np.take`.

bmcfee:
Note: tests pass except for the numba/numpy compat issue on 3.4.

bmcfee:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/598)*

bmcfee:
Results are `np.allclose` compatible when using high-precision (`float64` mode), so I'm gonna skip CR on this.

PR #622, opened 2017-09-02
Fix hz_to_midi docstring

carlthome:
`hz_to_midi` returns floats which is nice (like most `ftom` implementations) but the docstring makes it sound like results would be rounded.

Sidenote: maybe A4's frequency should be a parameter?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/622)

### DIFF
--- a/librosa/core/time_frequency.py
+++ b/librosa/core/time_frequency.py
@@ -509,7 +509,7 @@ def midi_to_hz(notes):
 
 
 def hz_to_midi(frequencies):
-    """Get the closest MIDI note number(s) for given frequencies
+    """Get MIDI note number(s) for given frequencies
 
     Examples
     --------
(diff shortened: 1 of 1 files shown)


bmcfee:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/622)*

bmcfee:
LGTM, thanks!

> Sidenote: maybe A4's frequency should be a parameter?

I don't think so.  MIDI has a [tuning standard](https://en.wikipedia.org/wiki/MIDI_tuning_standard), and overriding the reference tuning would produce numbers that are not midi.

bmcfee:
@carlthome we good to merge on this?

carlthome:
Yes! :smile:

Thanks for clarifying that MIDI notes have a fixed reference pitch. Did not know that.

bmcfee:
Thanks!

PR #632, opened 2017-10-04
util.frame() validation enhancements

### LINKED ISSUE #631: frame() validation is overkill / incorrectly placed
#### Description
`util.frame()` currently calls `valid_audio()` as part of its input checking.  This is probably not correct, since we may want to frame other signals (where nan/inf are okay).

Additionally, the call to `valid_audio()` happens after a check that already assumes an ndarray input.

All we really need in `frame()` is for the input to be:

- `ndarray`
- `ndim==1`
- shorter than the frame length
- contiguous
- `hop_length >= 1`

We're currently relying on `valid_audio` to check the first two properties, but these should be explicitly checked in `frame()` instead.

The call to `valid_audio` is also frequently redundant, since almost everywhere we call `frame` is after a validation check anyway.

bmcfee:
#### Reference Issue
Fixes #631 

This PR makes validation requirements for `util.frame()` more explicit, and relaxes them where appropriate.  Tests have been expanded to cover the various failure modes.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/632)

### DIFF
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -58,10 +58,12 @@ def frame(y, frame_length=2048, hop_length=512):
     Raises
     ------
     ParameterError
-        If `y` is not contiguous in memory, framing is invalid.
-        See `np.ascontiguous()` for details.
+        If `y` is not contiguous in memory, not an `np.ndarray`, or
+        not one-dimensional.  See `np.ascontiguous()` for details.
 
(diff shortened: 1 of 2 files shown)


bmcfee:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/632)*

PR #633, opened 2017-10-04
implemented multiband tempogram. fixes #518

### LINKED ISSUE #518: support subband onsets in tempogram
`feature.tempogram` currently supports either waveform or onset envelope input.  If an onset envelope is provided, it must be a 1d array.  This makes it somewhat cumbersome to work with sub-band onset envelopes as produced by [`onset.onset_strength_multi`](https://librosa.github.io/librosa/generated/librosa.onset.onset_strength_multi.html).

In principle, it shouldn't be difficult to extend `tempogram` to support two-dimensional envelopes of shape `(m, n)` for `m` channels and `n` frames.  It would then produce a tensor of shape `(m, win_length, n)`, ie, a tempogram for each input channel.

Note: This might require a slight rewrite, due to internal use of `util.frame`, which requires input to be memory-contiguous.

Maybe @mcartwright can chime in on other API changes that might make downstream rhythm analysis easier?

bmcfee:
#### Reference Issue
Fixes #518 

This PR adds multi-channel tempogram support, eg for analyzing the output of `onset_strength_multi`.  The implementation is naive in that it simply iterates over each band of the input onset envelope and combines the results.  Probably there's some redundant work being done here, but this works for now.

#### Any other comments?

Testing is regression against independent sub-band calculation.

This also works nicely with the fast Mellin transform by setting `axis=1`, which should make it easier to reproduce Prockup'15's methods.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/633)

### DIFF
--- a/librosa/feature/rhythm.py
+++ b/librosa/feature/rhythm.py
@@ -3,8 +3,6 @@
 '''Rhythmic feature extraction'''
 
 import numpy as np
-import scipy.signal
-import six
 
 from .. import util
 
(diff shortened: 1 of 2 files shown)


bmcfee:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/633)*

Issue #647, opened 2017-12-12
Spectral flatness


keunwoochoi:
I just started to wonder why don't we have spectral flatness in `librosa.feature`. Is there any historical reason? I can PR if it'd be useful.

bmcfee:
Because nobody implemented it :grin: 

But seriously, a PR would be great.

keunwoochoi:
Cool, few things to confirm. 
* It's based on a power spectrum, but the API would still wants `S=None` to be spectrogram magnitude for consistency with other `spectral_` siblings?

* Which way would you prefer, either `scipy.stats.mstats.gmean(S, axis=1)` or `np.exp(np.mean(np.log(X), axis=1))`?

..anything else?

bmcfee:
> It's based on a power spectrum, but the API would still wants S=None to be spectrogram magnitude for consistency with other spectral_ siblings?

Yeah, it should match the styling of the other `spectral_` functions.



> Which way would you prefer, either scipy.stats.mstats.gmean(S, axis=1) or np.exp(np.mean(np.log(X), axis=1))?

Definitely not `log(S)` for stability reasons.  I assume (but have not checked) that scipy's `gmean` is numerically stable.  If not, we might want to just code it directly with a noise floor parameter to prevent underflow.

keunwoochoi:
Seems like it's not?
```python
def gmean(a, axis=0):
    a, axis = _chk_asarray(a, axis)
    log_a = ma.log(a)
    return ma.exp(log_a.mean(axis=axis))
```

`amin=1e-10` as done in `power_to_db` seems the consistent choice. To summarise,

```python
def spectral_flatness(y=None, S=None, n_fft=2048, hop_length=512, amin=1e-5): # seems like we don't need sr and freq
    if S is None: compute S using y, n_fft, hop_length
    # few sanity check for S
    # sanity check for amin
    gmean = np.exp(np.mean(np.log(np.maximum(amin, S ** 2)), keepdims=True))
    amean = np.mean(np.maximum(amin, S ** 2), keepdims=True)
    return  gmean / amean

```

keunwoochoi:
Ok writing test code doesn't seem very easy, but will try. But what should I do to put the images in the documentation? (Is it automatic?!)

bmcfee:
> Ok writing test code doesn't seem very easy, but will try.

Do you have a reference implementation to compare against?  Most of the tests, aside from error checking, are just regression tests against old matlab code.



[...]

PR #648, opened 2017-12-17
add spectral_flatness, its docs, and a simple test

keunwoochoi:
#### Reference Issue
#647 

Add a function to compute the spectral flatness.

#### Any other comments?

Decided to PR, knowing the test might be not enough, just to proceed :-) [Essentia](http://essentia.upf.edu/documentation/reference/std_Flatness.html) has one, and there are some other sources.
* Is the current `def test_spectral_flatness_errors()` too simple? I assumed there's no need to test on `y, S, n_fft, hop_length`, which only leaves `amin`. 
* To make test great again, I guess I should choose a proper reference, add a data file in tests/data, and write a test code that loads the file and compare. Essential seems a legit choice, let me know if you have any other idea.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/648)


bmcfee:
Thanks for kicking this one off!  A few comments:



> Is the current def test_spectral_flatness_errors() too simple?

It looks like there's no coverage on the function itself here, so yes.  Check out the other `spectral_*` tests to see what we usually do for hitting corner cases and regressions.


> I assumed there's no need to test on y, S, n_fft, hop_length, which only leaves amin.
> I guess I should choose a proper reference, add a data file in tests/data, and write a test code that loads the file and compare.

Yeah.  If you use essentia, you'll need to be careful to isolate the stage of processing being tested so that we're not chasing down differences in stft implementations.  That might be tricky to do.  OTOH, this feature is simple enough that we might not need a ref/regression test, and we can get away with simple unit tests on synthetic signals with know outputs.

Beyond that, I'll stick a couple of notes in CR.

bmcfee:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[librosa/feature/spectral.py, line 503 at r1](https://reviewable.io:443/reviews/librosa/librosa/648#-L0e77xiOieU6JfJ-ReS:-L0e77xiOieU6JfJ-ReT:bhgt9nf) ([raw file](https://github.com/librosa/librosa/blob/4325a8d23415672a72e58d7400c19d8764402a32/librosa/feature/spectral.py#L503)):*
> ```Python
> 
> 
> def spectral_flatness(y=None, S=None, n_fft=2048, hop_length=512, amin=1e-5):
> ```

For consistency with `power_to_db`, `amin` should be `1e-10` by default here since you're squaring `S` before thresholding.

---

*[librosa/feature/spectral.py, line 552 at r1](https://reviewable.io:443/reviews/librosa/librosa/648#-L0e6_P31xStjCCQBsVe:-L0e6_P31xStjCCQBsVf:b5ykf8i) ([raw file](https://github.com/librosa/librosa/blob/4325a8d23415672a72e58d7400c19d8764402a32/librosa/feature/spectral.py#L552)):*
> ```Python
>         raise ParameterError('amin must be strictly positive')
> 
>     S, n_fft = _spectrogram(y=y, S=S, n_fft=n_fft, hop_length=hop_length)
> ```

Maybe worth explicitly stating the `power` parameter here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/648)*

keunwoochoi:
* amin to be 1e-10: you're right.
* to be precise, you meant
i) 
```python
S, n_fft = _spectrogram(y=y, S=S, n_fft=n_fft, hop_length=hop_length,
                            power=2.)
# and later..
    gmean = np.exp(np.mean(np.log(np.maximum(amin, S)),
                           axis=0, keepdims=True))
    amean = np.mean(np.maximum(amin, S), axis=0, keepdims=True)

```
? This would make the code cleaner, but one thing I just noted is that `_spectrogram` doesn't change `S` to be `S**2` if `S is not None`, so it doesn't work. 

So now I guess what you meant should be..

ii)
```python
S, n_fft = _spectrogram(y=y, S=S, n_fft=n_fft, hop_length=hop_length,
                            power=1.)
```
and keep the other code as it is.

bmcfee:
> (Sidenote: maybe it's a good idea to specify power=1 (or 2) for other lines in another PR.)

Alternately, you could have it call _spec with `power=2` by default, but allow the user to directly override the spectrogram without forcing a square on it.  I think melspectrogram works this way (thanks to your previous PR).

keunwoochoi:
> but allow the user to directly override the spectrogram without forcing a square on it.

I'm confused a little bit, is it about `spectral_flatness`? In melspectrogram, if user wants it to be`power=2.0` while they pass a valid `S`, `S` should be power spectrogram (as far as I understand what I did!). To set `power=2.0` by default in `spectral_flatness` the input `S` should be also power spectrogram, which I think is ok (so that we don't square twice) but the API is not consistent with its families.. or am I getting something wrong?

bmcfee:
> I'm confused a little bit, is it about spectral_flatness? In melspectrogram, if a user wants it to bepower=2.0 while they pass a valid S, S should be power spectrogram (as far as I understand what I did!). To set power=2.0 by default in spectral_flatness the input S should be also power spectrogram, which I think is ok (so that we don't square twice) but the API is not consistent with its families.. or am I getting something wrong?

Yeah, I'm sorry that this bit of internals isn't well documented.  The semantics of `_spectrogram` are as follows:

- If provided with a signal, generate a magnitude spectrogram raised to the given power.
- If provided with a spectrogram, return it unmodified.

So if you want a power spectrogram by default, call with `power=2`.  Better yet, make `power` a parameter to `spectral_flatness` and pass it all through as in `melspectrogram`.

bmcfee:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[librosa/feature/spectral.py, line 552 at r1](https://reviewable.io:443/reviews/librosa/librosa/648#-L0e6_P31xStjCCQBsVe:-L1If6zEpgTaqLguwyYx:b-kodx10) ([raw file](https://github.com/librosa/librosa/blob/4325a8d23415672a72e58d7400c19d8764402a32/librosa/feature/spectral.py#L552)):*

Maybe worth explicitly stating the `power` parameter here?
</blockquote></details>

I think the consistent way to do this, in reference to `melspectrogram`, is to make `power=2` a default parameter to `spectral_flatness`, and then call `_spectrogram(..., power=power)`.

---

*[tests/test_features.py, line 378 at r2](https://reviewable.io:443/reviews/librosa/librosa/648#-L1IfOMwQo0KuvYjF887:-L1IfOMwQo0KuvYjF888:bfx2qcx) ([raw file](https://github.com/librosa/librosa/blob/28d0a712c9e7cf051be52fb41f12a8f7f274e836/tests/test_features.py#L378)):*
> ```Python
>     # a known case
>     y, _ = librosa.load(__EXAMPLE_FILE)
>     flatness_ref = np.load("data/feature-spectral-flatness.npy")
> ```

please document where this data file came from

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/648)*

keunwoochoi:
1. 
> Better yet, make power a parameter to `spectral_flatness` and pass it all through as in `melspectrogram`

But by doing it we're assuming there are cases where the `power!=2` even if the definition of spectral flatness is based on a power spectrum, which doesn't seem a big deal. To summary again, my understanding is..

- The api has `power=2` by default.
- The api asks amplitude spectrogram
- When the function receives `S`, it will do `S=S**power`.
- When the function receives `y`, it will compute a power spectrogram. 
- Most of the users wouldn't care to change `power` and will input `y` or amplitude spectrogram `S` (because that's how spectral siblings work).
- Some of the users might wanna pass perhaps `power=2` e.g., when they already computed power spectrograms, and those would (should) be probably careful enough. 

Seems alright for me.

2. 
> please document where this data file came from

It comes from the very current implementation, assuming it's correct so that it could be used for further changes. If it's fine, how should I explain it? One way would be to mention the provisional version that this function would be included for the first time. I guess it's.. 0.6?

bmcfee:
> But by doing it we're assuming there are cases where the power!=2 even if the definition of spectral flatness is based on a power spectrum, which doesn't seem a big deal.

Yup, that's correct. It makes me a bit nervous to have a magic number (2) kicking around multiple times in the the function, since one might want to bypass it.  (It's also calculated multiple times, which is inefficient.)

Maybe a simpler approach would be to keep the `power=2` parameter, but set `power=1` in the `_spectrogram` call, and take the gmean/amean of `S**power` instead.  This way, if a user has a pre-computed power spectrogram, they cal call `spectral_flatness(S=powspec, power=1)` and get the right answer.  How's that sound?



> It comes from the very current implementation, assuming it's correct so that it could be used for further changes. If it's fine, how should I explain it?

I don't think we should do it this way -- regression tests in librosa are generally included for measuring compatibility with prior implementations.  If you want a correctness test, it would be better to include a small, self-contained example as inputs and outputs directly in the source code.  These can be very small input spectrograms (like, 3x5 or something) where the results can be verified easily by hand.

bmcfee:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[librosa/feature/spectral.py, line 505 at r2](https://reviewable.io:443/reviews/librosa/librosa/648#-L1Y0RgYEAsTTtElzOLO:-L1Y0RgYEAsTTtElzOLP:b7a7haw) ([raw file](https://github.com/librosa/librosa/blob/28d0a712c9e7cf051be52fb41f12a8f7f274e836/librosa/feature/spectral.py#L505)):*
> ```Python
> def spectral_flatness(y=None, S=None, n_fft=2048, hop_length=512, amin=1e-10):
>     '''Compute spectral flatness
> ```

Please describe what spectral flatness is, and if possible, provide a reference

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/648)*

keunwoochoi:
> Maybe a simpler approach 

Done!

> If you want a correctness test, it would be better to include a small, self-contained example

Done!

> test 

Done!

> Documentation

Done!

Anything else?

bmcfee:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[librosa/feature/spectral.py, line 587 at r3](https://reviewable.io:443/reviews/librosa/librosa/648#-L1rbcFQF1NNWMlrMe7C:-L1rbcFQF1NNWMlrMe7D:b-s45c9g) ([raw file](https://github.com/librosa/librosa/blob/fd96b58a8763ae9963b4d0a92977f1617756a807/librosa/feature/spectral.py#L587)):*
> ```Python
>     gmean = np.exp(np.mean(np.log(np.maximum(amin, S ** power)),
>                            axis=0, keepdims=True))
>     amean = np.mean(np.maximum(amin, S ** power), axis=0, keepdims=True)
> ```

One last request: `np.maximum(amin, S**power)` is computed twice here, but it could be done more efficiently by caching the result.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/648)*

bmcfee:
Reviewed 1 of 3 files at r3.
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/648)*

keunwoochoi:
> by caching the result.

Understood. But I couldn't find how to do it if there's a generic way to do so. My guess is.. `from .. import cache` on the top and decorate the function with `@cache(level=20)`?

bmcfee:
> Understood. But I couldn't find how to do it if there's a generic way to do so.

Oh i just mean store it: `S_thresh = np.maximum(...)`

keunwoochoi:
Ohhh. Ok, done!

bmcfee:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/648)*

bmcfee:
Thanks!  I'll merge this once the light turns green.

Issue #655, opened 2018-01-10
FutureWarning when calling librosa.display.specshow()


aleksandrevic:
#### Description
When trying to calculate and then plot the mfccs of a wav file, an error message appears and no mfcc plot 

#### Steps/Code to Reproduce

y, sr = librosa.load('~/PycharmProjects/MA/Roland-JV-2080-PizzicatoStr-C5.wav/')

hop_length = 512

mfccs = librosa.feature.mfcc(y=y, sr=sr, hop_length=hop_length, n_mfcc=13)

plt.figure(figsize=(10, 4))
librosa.display.specshow(mfccs, x_axis='time')
plt.colorbar()
plt.title('MFCC')
plt.tight_layout()

#### Expected Results
Plot of MFCCs

#### Actual Results
Error message and no plot:

/Users/aleksandar/PycharmProjects/MA/venv/lib/python3.6/site-packages/librosa/display.py:656: FutureWarning: Conversion of the second argument of issubdtype from `complex` to `np.complexfloating` is deprecated. In future, it will be treated as `np.complex128 == np.dtype(complex).type`. if np.issubdtype(data.dtype, np.complex):

Checking the variables, the mfccs have been calculated properly.

#### Versions
Darwin-17.3.0-x86_64-i386-64bit
Python 3.6.4 (v3.6.4:d48ecebad5, Dec 18 2017, 21:07:28) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]
NumPy 1.14.0
Python 3.6.4 (v3.6.4:d48ecebad5, Dec 18 2017, 21:07:28) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.
SciPy 1.0.0
librosa 0.5.1
PyDev console: using IPython 6.2.1

bmcfee:
That's strange -- a warning should not interrupt any functionality, but thanks for reporting it.

Perhaps the plot is being rendered properly, but you need to say `plt.show()` to draw the figure?  I haven't used pydev console, so I'm not sure how it interacts with matplotlib.

aleksandrevic:
@bmcfee yeah that was missing, I just copy pasted... thank you... omg it's so embarrassing haha, that's why I tried it in the google group first... Sorry I am a beginner :-)

But is there any chance to not get the warning?

bmcfee:
[...]

PR #656, opened 2018-01-13
Fix FutureWarning from NumPy

carlthome:
Fix NumPy warnings:

>FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.

>FutureWarning: Conversion of the second argument of issubdtype from `complex` to `np.complexfloating` is deprecated. In future, it will be treated as `np.complex128 == np.dtype(complex).type`.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/656)

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -653,7 +653,7 @@ def specshow(data, x_coords=None, y_coords=None,
     >>> plt.tight_layout()
     '''
 
-    if np.issubdtype(data.dtype, np.complex):
+    if np.issubdtype(data.dtype, np.complexfloating):
         warnings.warn('Trying to display complex-valued input. '
                       'Showing magnitude instead.')
         data = np.abs(data)
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -287,7 +287,7 @@ def pitch_shift(y, sr, n_steps, bins_per_octave=12):
     ...                                          bins_per_octave=24)
     '''
 
-    if bins_per_octave < 1 or not np.issubdtype(type(bins_per_octave), np.int):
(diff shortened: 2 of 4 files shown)


bmcfee:
Thanks for fixing this!  Issue ref: #655 

Are there any other points in the code that do similar comparisons that need to be updated?

carlthome:
>Are there any other points in the code that do similar comparisons that need to be updated?

Oops, yup. Done!

bmcfee:
Reviewed 4 of 4 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/656)*

bmcfee:
Cool, thanks for updating.  Last question: does this change our minimum numpy version requirement?

carlthome:
Not sure. I'm trying to grok [when np.complexfloating was introduced](https://github.com/numpy/numpy/blame/10ccfe747a68d974ff16a5c0765054b816d5486f/numpy/core/numerictypes.py#L66) but can't really find it (not familiar with NumPy's codebase).

Maybe we should have an additional Travis run in the build matrix that only installs the minimum package versions listed in setup.py? I could do a separate pull request with that change to .travis.yml.

bmcfee:
Well, git blame says that line of the docs was added 8 years ago, so I think we're safe.  It does show up in the [numpy 1.6 docs](https://docs.scipy.org/doc/numpy-1.6.0/reference/arrays.dtypes.html?highlight=complexfloating).

I think we're ready to merge?

carlthome:
Yes! Ready.

bmcfee:
Thanks!

PR #663, opened 2018-02-05
fixed #657 by reimplementing the delta function. Deprecates trim=

### LINKED ISSUE #657: 2nd order delta is incorrect
#### Description
2nd order delta results is one unit right shifted comparing with expected results.

#### Steps/Code to Reproduce
```
In [1]: import librosa

In [2]: librosa.feature.delta([-2, -1, 0, 1, 2], width=3, order=2)
Out[2]: array([ 0.25,  0.5 ,  0.25,  0.  , -0.25])
```
#### Expected Results
```
array([ 0.5, 0.25, 0, -0.25, -0.5])
```

#### Actual Results
```
array([ 0.25,  0.5 ,  0.25,  0.  , -0.25])
```

#### Versions
```
Linux-4.4.0-31-generic-x86_64-with-debian-stretch-sid
Python 3.5.4 (default, Dec  5 2017, 08:24:28) 
[GCC 5.4.0 20160609]
NumPy 1.13.3
SciPy 1.0.0
librosa 0.5.1
```

bmcfee:
#### Reference Issue
Fixes #657 

There was a bug in `feature.delta` that caused drift at higher order features.  This was caused by implementing smoothed differencing by iteratively calling `scipy.signal.lfilter`.

~~~The function has been rewritten to use `scipy.ndimage.convolve1d`, so that results stay centered.~~~
This function has been rewritten to use Savitsky-Golay filtering via `scipy.signal.savgol_filter`.  This gives a more general implementation of the linear smoothing previously used in `delta`, and reduces the influence of boundary effects, producing more accurate high-order difference estimates.

Compared to the code example in #657:

```python
In [1]: import librosa

In [2]: librosa.feature.delta([-2, -1, 0, 1, 2], width=3, order=2)
Out[2]: array([ 0.25,  0.5 ,  0.25,  0.  , -0.25])
```
##### Expected Results
```array([ 0.5, 0.25, 0, -0.25, -0.5])```
##### Actual Results
```array([ 0.25,  0.5 ,  0.25,  0.  , -0.25])```

we now get for first-order (`mode='nearest'`):
```python
array([ 0.5,  1. ,  1. ,  1. ,  0.5])
```
and second-order (*updated* **updated again**):
```python
array([  1.00000000e+00,   8.88178420e-16,   4.44089210e-16,
         0.00000000e+00,  -1.00000000e+00])
```
With `mode='interp'`, first-order:
```python
array([ 1.,  1.,  1.,  1.,  1.])
```
and second-order:
```python
array([ -2.03182142e-15,   8.88178420e-16,   4.44089210e-16,
         0.00000000e+00,  -5.89125247e-16])
```
#### Any other comments?

Because ~~~`convolve1d` pads implicitly~~~ `savgol_filter` pads explicitly, padding and trimming is no longer required in the `delta` API.  As such, the `trim=` parameter has been deprecated, and will be removed in 0.7.  Calls to `librosa.delta` now preserve shape and dtype.  This is a breaking change if people relied on `trim=False` behavior previously, but I expect that this feature was not used (often) outside of our unit tests.

By default, it now uses `interp` padding.  As noted above, this reduces the influence of boundary effects.  The old behavior can be recovered by setting `mode='nearest'`.

The requirements on `width` have been tightened, and now depend on the length of the signal.

The API now exposes `kwargs` parameters to `savgol_filter`, in case a user wants to change the sampling points or constant-padding behavior.

The minimum scipy version has been bumped up to 0.14.

<!-- Reviewable:start -->
---
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/librosa/librosa/663)
<!-- Reviewable:end -->


bmcfee:
@chenjiasheng would you mind trying this PR out to see if it fixes your problem?

bmcfee:
Follow-up note: I think I'm in favor of the accelerated version where `delta(delta(x)) != delta(x, order=2)`, for a few reasons:

1. If anyone wants to call `delta(delta(x))`, they're still free to do so.  The delta op is cheap enough that this isn't a huge deal.
2. It's faster and more memory-efficient to touch the data only once.
3. Convolving once by the second-order operator requires only a single pad operation, which explicitly approximates the input signal.  Convolving twice by first-order operators requires two pad operations, the first of which approximates the input signal and the second approximates its first delta.  Given the choice between the two, I think it makes more sense to only approximate the input signal.

bmcfee:
Summarizing an out-of-band conversation with @lostanlen:

Much of `delta`'s functionality is now implemented in [`scipy.signal.savgol_filter`](https://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.signal.savgol_filter.html).  This function was added in 0.14 (2016), and was not available when we added `delta`. It might make sense to rewrite `delta` to be a thin wrapper on top of `savgol_filter`:

- This would let us use its `interp` mode for padding, which does clever things at the boundaries.  It also exposes the padding mode parameters, if people want the older behavior.
- It would give us automatic support for higher order polynomial smoothing filters.  Otherwise, keeping `polyorder=1` preserves the current behavior.
- It would require upping the minimum scipy version (currently 0.13).  I'm okay with this, as scipy's now at 1.0 anyway.

bmcfee:
Some caveats to the above:

- `savgol_filter` requires `width <= x.shape[axis]` when `mode='interp'`.  This is reasonable, I think.
- `savgol_filter` requires (undocumented) that `polycoef <= deriv (order)`, so that you can't use a linear smoothing to compute a second derivative.  This is also, I think, reasonable, and was implicitly handled by iterating the operator.  If we set `polycoef = deriv (order)`, everything matches up well with the previous implementation.

For comparison, here's a plot of first and second differences of a sin using the current (fixed) `delta` from this branch, and the savitsky-golay filter with `mode='interp'`:
image
Note the lack of jump discontinuities at the boundaries using SG.

bmcfee:
Reviewed 3 of 3 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/663)*

lostanlen:
`ParameterError('width must be an odd integer >= 3')`
is likely not going to help the user in case
`(mode=='interp' and width > data.shape[axis])`

bmcfee:
Good catch, thanks.

Issue #662, opened 2018-02-05
db_to_amplitude(amplitude_to_db(S)) doesn't preserve peak amplitude?


carlthome:
#### Description
I'd expect roughly the same peak amplitude when decibel scaling a spectrogram, and then going back to linear amplitudes before the inverse spectral transform. Am I missing something obvious?

#### Steps/Code to Reproduce
```python
import librosa as lr

y = lr.load(lr.util.example_audio_file(), None)[0]
S = lr.amplitude_to_db(lr.stft(y))
y2 = lr.istft(lr.db_to_amplitude(S))
print(y.max(), y2.max())
```

#### Expected Results
> 0.64982605 0.64982605

#### Actual Results
> 0.64982605 0.16012064

#### Versions
Linux-4.4.0-92-generic-x86_64-with-debian-stretch-sid
Python 3.6.3 |Anaconda, Inc.| (default, Nov 20 2017, 20:41:42) 
[GCC 7.2.0]
NumPy 1.14.0
SciPy 1.0.0
librosa 0.5.1

bmcfee:
Strange.  We have integration tests for `amplitude_to_db` and `db_to_amplitude`, and a quick check verifies that they work as expected:
```python
In [5]: D = lr.stft(y)

In [6]: S = lr.amplitude_to_db(D, top_db=None)

In [7]: Dinv = lr.db_to_amplitude(S)

In [10]: np.max(np.abs(D))
Out[10]: 215.17365

In [11]: np.max(np.abs(Dinv))
Out[11]: 215.17369
```

Whatever is going on must be down to `i/stft`, probably a windowing artifact?

bmcfee:
Aha, it's down to phase loss.  You're reconstructing from just the magnitudes, so you're missing out on some constructive interference:

```python
>>> y, sr = librosa.load(librosa.util.example_audio_file())
>>> D = librosa.stft(y)
>>> Smag, p = librosa.magphase(D)
>>> S = librosa.amplitude_to_db(Smag, top_db=None)
>>> Sinv = librosa.db_to_amplitude(S)
>>> y2 = librosa.istft(Sinv * p)
>>> y3 = librosa.istft(D)
>>> print(y.max(), y2.max(), y3.max())
0.650596 0.650596 0.650596
```

carlthome:
Ah, thanks! 🐤  🐨 

Maybe we could have a warning in `amplitude_to_db` when the input is dtype complex about that phase will be discarded.

[...]

PR #671, opened 2018-02-12
Add warnings when phase is lost

carlthome:
Fixes my own confusion in https://github.com/librosa/librosa/issues/662

I don't think there's an easy way of not duplicating the warning in both `power_to_db` and `amplitude_to_db` but would gladly fix that if possible.

I'm also unsure if there's a non-complex time-frequency representation of amplitudes that are not magnitudes? I guess it would be something like `np.sign(S)*np.abs(S)` but that feels somewhat exotic.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/671)

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -850,6 +850,8 @@ def power_to_db(S, ref=1.0, amin=1e-10, top_db=80.0):
 
     """
 
+    S = np.array(S)
+
     if amin <= 0:
         raise ParameterError('amin must be strictly positive')
 
@@ -943,6 +945,9 @@ def amplitude_to_db(S, ref=1.0, amin=1e-5, top_db=80.0):
     -----
     This function caches at level 30.
     '''
+
+    S = np.array(S)
+
     if np.issubdtype(S.dtype, np.complexfloating):
         warnings.warn('Input was complex. Phase information will be discarded.'
                       'Only input magnitudes to avoid this warning (e.g. `np.abs(S)`).')
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -537,7 +537,8 @@ def magphase(D, power=1):
 
     """
 
-    mag = np.pow(np.abs(D), power)
+    mag = np.abs(D)
+    mag **= power
     phase = np.exp(1.j * np.angle(D))
 
     return mag, phase
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -481,7 +481,7 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
     return if_gram, stft_matrix
 
 
-def magphase(D):
+def magphase(D, power=1):
     """Separate a complex-valued spectrogram D into its magnitude (S)
     and phase (P) components, so that `D = S * P`.
 
@@ -490,12 +490,15 @@ def magphase(D):
     ----------
     D       : np.ndarray [shape=(d, t), dtype=complex]
         complex-valued spectrogram
+    power : float > 0
+        Exponent for the magnitude spectrogram,
+        e.g., 1 for energy, 2 for power, etc.
 
 
     Returns
     -------
     D_mag   : np.ndarray [shape=(d, t), dtype=real]
-        magnitude of `D`
+        magnitude of `D`, raised to `power`
     D_phase : np.ndarray [shape=(d, t), dtype=complex]
         `exp(1.j * phi)` where `phi` is the phase of `D`
 
@@ -534,7 +537,7 @@ def magphase(D):
 
     """
 
-    mag = np.abs(D)
+    mag = np.pow(np.abs(D), power)
     phase = np.exp(1.j * np.angle(D))
 
     return mag, phase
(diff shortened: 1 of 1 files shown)


librosa/core/spectrum.py
```diff
@@ -946,8 +955,8 @@ def amplitude_to_db(S, ref=1.0, amin=1e-5, top_db=80.0):
     else:
         ref_value = np.abs(ref)
 
-    magnitude **= 2
-    return power_to_db(magnitude, ref=ref_value**2, amin=amin**2,
+    power = magnitude**2
```
carlthome: I get the mutation was to avoid a copy, _but_ for newbs like me it's a lot clearer with the name `power` because when I was new to LibROSA's code base I conflated magnitudes and power, _a lot_.

librosa/core/spectrum.py
```diff
@@ -849,9 +850,15 @@ def power_to_db(S, ref=1.0, amin=1e-10, top_db=80.0):
 
     """
 
+    S = np.array(S)
```
Nodd: Maybe use `np.asarray` to avoid unnecessary copies ? (same below)
`np.array` copies by default, see for example: https://stackoverflow.com/questions/14415741/numpy-array-vs-asarray

carlthome: Thanks for the tip!

bmcfee:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[librosa/core/spectrum.py, line 958 at r1](https://reviewable.io:443/reviews/librosa/librosa/671#-L58NVRUxuOtZJxNv_Sf-r1-958:-L59aJ_sORPrycLaFzy9:bqcth0w) ([raw file](https://github.com/librosa/librosa/blob/4b39d997b4f8cd174f8eee4fcd0e60da8128940a/librosa/core/spectrum.py#L958)):*
<details><summary><i>Previously, carlthome (Carl Thomé) wrote…</i></summary><blockquote>

I get that the mutation was to avoid a copy, _but_ for newbs like me it's a lot clearer with the name `power` because when I was new to LibROSA's codebase I conflated magnitudes and power, _a lot_.
</blockquote></details>

Yes, this was done to avoid a copy.  Maybe it would be better to rename the variable to something less confusing, rather than invoke an unnecessary copy?

---

*[librosa/core/spectrum.py, line 860 at r3](https://reviewable.io:443/reviews/librosa/librosa/671#-L59_RUr8uJRU99mSsx5:-L59_RUr8uJRU99mSsx6:botda3q) ([raw file](https://github.com/librosa/librosa/blob/fddfffc20c7df4d8f1cbed6e7166d50bd25e664b/librosa/core/spectrum.py#L860)):*
> ```Python
>     if np.issubdtype(S.dtype, np.complexfloating):
>         warnings.warn('Input was complex. Phase information will be discarded.'
>                       'Only input magnitudes to avoid this warning (e.g. `np.abs(S)`).')
> ```

I think this warning (and others) could be more useful if it was made more explicit.  A novice user might not know how to make sense of this.  Instead, I suggest something like: `power_to_db was called on complex input, and phase information will be discarded. To suppress this warning, call power_to_db(np.abs(S), ...) instead.`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/671)*

bmcfee:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/671)*

carlthome:

---

*[librosa/core/spectrum.py, line 853 at r2](https://reviewable.io:443/reviews/librosa/librosa/671#-L58NVRUxuOtZJxNv_Sf-r2-853:-L59qProqpls9QFUoWJC:b-896fix) ([raw file](https://github.com/librosa/librosa/blob/50fcd3aa3438a7231028e98915268de6cc50f515/librosa/core/spectrum.py#L853)):*
<details><summary><i>Previously, carlthome (Carl Thomé) wrote…</i></summary><blockquote>

Thanks for the tip!
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/671)*

carlthome:
Please let me know if I'm overthinking this but I think it's nice to use magphase if it's already there, but I also get that it might feel more familiar to just type `np.abs` for many people.

bmcfee:
> Please let me know if I'm overthinking this but I think it's nice to use magphase if it's already there, but I also get that it might feel more familiar to just type np.abs for many people.

I'm fine with either, though `abs` will be more familiar, shorter, and more efficient.  (The latter because it doesn't have to allocate space to store the phase array.)

carlthome:
> I'm fine with either, though abs will be more familiar, shorter, and more efficient. (The latter because it doesn't have to allocate space to store the phase array.)

Yeah, hopefully users that already know the difference between amplitude, magnitude and power will just keep doing np.abs.

For newcomers it can be nice to get some directions toward the docstring of magphase for some additional insight. I remember being confused by `specshow(logamplitude(np.abs(stft(my_audio_waveform))**2))` a year ago. 🐨

bmcfee:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/671)*

carlthome:
Ready for merge, @bmcfee.

bmcfee:
Thanks!

PR #679, opened 2018-02-21
implemented PCEN. fixes #615. needs tests

### LINKED ISSUE #615: Per-channel energy normalization (PCEN)
#### Description
PCEN [(Wang et al.)](https://arxiv.org/pdf/1607.05666) seems to be a good alternative to log-amplitude scaling for speech recognition systems; see also: [(Battenberg, et al.)](https://arxiv.org/abs/1705.04400).  Since it's fairly easy to implement in the static (non-trainable) form, do folks thing it's worth including in librosa?

Here's a quick reference implementation of eq.1 in the paper, using the default parameters listed in section 2:
```python
def pcen(E, alpha=0.98, delta=2, r=0.5, s=0.025, eps=1e-6):   
    M = scipy.signal.lfilter([s], [1, s - 1], E)
    smooth = (eps + M)**(-alpha)
    return (E * smooth + delta)**r - delta**r
```
It could probably be made more numerically stable.  It could also be vectorized to support channel-dependent parameters, which might make it more useful for things like CQT.

bmcfee:
#### Reference Issue
#615 

This PR adds the PCEN method of [Wang et al.](https://arxiv.org/pdf/1607.05666)

The default parameters are set to correspond (as closely as possible) to those reported in the paper.

The API allows the user to specify the IIR smoothing filter either by setting a time constant (from which the filter coefficient is derived), or explicitly setting the parameter.

#### Any other comments?

This PR also adds a superflux-style extension that allows for vibrato-sensitive automatic gain control.  This is disabled by default, and enabled by setting `max_size` to an integer greater than 1.

It still needs tests.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/679)

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -6,6 +6,7 @@ import warnings
 import numpy as np
 import scipy.fftpack as fft
 import scipy
+import scipy.ndimage
 import scipy.signal
 import scipy.interpolate
 import six
@@ -23,7 +24,7 @@ __all__ = ['stft', 'istft', 'magphase', 'iirt',
            'perceptual_weighting',
            'power_to_db', 'db_to_power',
            'amplitude_to_db', 'db_to_amplitude',
-           'fmt']
+           'fmt', 'pcen']
 
 
 @cache(level=20)
@@ -1279,6 +1280,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
 
     # Truncate and length-normalize
     return result[idx] * np.sqrt(n) / n_fmt
+
+
+@cache(level=30)
+def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
+         time_constant=0.395, eps=1e-6, b=None, max_size=1):
+    '''Per-channel energy normalization (PCEN) [1]_
+
+    This function normalizes a time-frequency representation `S` by
+    performing automatic gain control, followed by non-linear compression:
+
+        P = (S / (eps + M)**gain + bias)**power - bias**power
+
+    where `M` is the low-pass filtered output of `S`:
+
+        M[f, t] = (1 - b) M[f, t - 1] + b * S[f, t]
+
+    and if `b` is not provided, it is calculated as:
+
+        b = 1 - exp(-hop_length / (sr * time_constant))
+
+    This normalization is designed to suppress background noise, and
+    emphasize foreground signals, and can be used as an alternative to
+    decibel scaling (`amplitude_to_db`).
+
+    This implementation also supports frequency-bin smoothing by specifying
+    `max_size > 1`.  If this option is used, the filtered spectrogram `M` is
+    computed as
+
+        M[f, t] = (1 - b) M[f, t - 1] + b * R[f, t]
+        
+    where `R` has been max-filtered along the frequency axis, similar to
+    the Superflux algorithm implemented in `onset.onset_strength`:
+
+        R[f, t] = max(R[f - max_size//2: f + max_size//2, t])
+
+    This can be used to perform automatic gain control on signals that cross
+    or span multiple frequency bans, which may be desirable for spectrograms
+    with high frequency resolution.
+
+    .. [1] Wang, Y., Getreuer, P., Hughes, T., Lyon, R. F., & Saurous, R. A.
+       (2017, March). Trainable frontend for robust and far-field keyword spotting.
+       In Acoustics, Speech and Signal Processing (ICASSP), 2017
+       IEEE International Conference on (pp. 5670-5674). IEEE.
+
+    Parameters
+    ----------
+    S : np.ndarray (non-negative) [shape=(n, m)]
+        input spectrogram
+
+    sr : number > 0 [scalar]
+        Sampling rate of audio
+
+    hop_length : int > 0 [scalar]
+        Hop length of `S`
+
+    gain : number > 0 [scalar]
+        Gain factor.  Typical values should be slightly less than 1.
+
+    bias : number >= 0 [scalar]
+        The bias point of the non-linear compression
+
+    power : number > 0 [scalar]
+        The compression factor.  Typical values should be between 0 and 1.
+
+    time_constant : number > 0 [scalar]
+        The time constant for IIR filtering, measured in seconds.
+
+    eps : number > 0 [scalar]
+        A small constant used to ensure numerical stability of the filter.
+
+    b : number in [0, 1]  [scalar]
+        The filter coefficient for the low-pass filter.
+        If not provided, it will be inferred from `time_constant`.
+
+    max_size : int > 0 [scalar]
+        The size of the frequency axis max filter.
+        If left as `1`, no filtering is performed.
+
+    Returns
+    -------
+    P : np.ndarray, non-negative [shape=(n, m)]
+        The per-channel energy normalized version of `S`.
+
+    See Also
+    --------
+    amplitude_to_db
+    librosa.onset.onset_strength
+
+    Examples
+    --------
+
+    Compare PCEN to log amplitude (dB) scaling on Mel spectra
+
+    >>> import matplotlib.pyplot as plt
+    >>> y, sr = librosa.load(librosa.util.example_audio_file(),
+    ...                      offset=10, duration=10)
+
+    >>> # We'll use power=1 to get a magnitude spectrum,
+    >>> # instead of a power spectrum
+    >>> S = librosa.feature.melspectrogram(y, sr=sr, power=1)
+    >>> log_S = librosa.amplitude_to_db(S, ref=np.max)
+    >>> pcen_S = librosa.pcen(S)
+    >>> plt.figure()
+    >>> plt.subplot(2,1,1)
+    >>> librosa.display.specshow(log_S, x_axis='time', y_axis='mel')
+    >>> plt.title('log amplitude (dB)')
+    >>> plt.colorbar()
+    >>> plt.subplot(2,1,2)
+    >>> librosa.display.specshow(pcen_S, x_axis='time', y_axis='mel')
+    >>> plt.title('Per-channel energy normalization')
+    >>> plt.colorbar()
+    >>> plt.tight_layout()
+
+    Compare PCEN with and without max-filtering
+
+    >>> pcen_max = librosa.pcen(S, max_size=5)
+    >>> plt.figure()
+    >>> plt.subplot(2,1,1)
+    >>> librosa.display.specshow(pcen_S, x_axis='time', y_axis='mel')
+    >>> plt.title('Per-channel energy normalization (no max-filter)')
+    >>> plt.colorbar()
+    >>> plt.subplot(2,1,2)
+    >>> librosa.display.specshow(pcen_max, x_axis='time', y_axis='mel')
+    >>> plt.title('Per-channel energy normalization (max_size=5)')
+    >>> plt.colorbar()
+    >>> plt.tight_layout()
+
+    '''
+    if np.issubdtype(S.dtype, np.complexfloating):
+        warnings.warn('pcen was called on complex input so phase '
+                      'information will be discarded. To suppress this warning, '
+                      'call pcen(magphase(D)[0]) instead.')
+
+        S = np.abs(S)
+
+    if max_size == 1:
+        ref_spec = S
+    else:
+        ref_spec = scipy.ndimage.maximum_filter1d(S, max_size, axis=0)
+
+    if b is None:
+        b = 1 - np.exp(- float(hop_length) / (time_constant * sr))
+
+    smooth = (eps + scipy.signal.lfilter([b], [1, b - 1], ref_spec))**gain
+
+    return (S / smooth + bias)**power - bias**power
 
 
 def _spectrogram(y=None, S=None, n_fft=2048, hop_length=512, power=1):
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1439,8 +1439,7 @@ def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
         warnings.warn('pcen was called on complex input so phase '
                       'information will be discarded. To suppress this warning, '
                       'call pcen(magphase(D)[0]) instead.')
-
-    S = np.abs(S)
+        S = np.abs(S)
 
     if max_size == 1:
         ref_spec = S
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1426,7 +1426,7 @@ def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
     if time_constant <= 0:
         raise ParameterError('time_constant={} must be strictly positive'.format(time_constant))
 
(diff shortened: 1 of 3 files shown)


librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+
+    if max_size == 1:
+        ref_spec = S
+    else:
+        ref_spec = scipy.ndimage.maximum_filter1d(S, max_size, axis=0)
+
+    if b is None:
+        b = 1 - np.exp(- float(hop_length) / (time_constant * sr))
+
+    smooth = (eps + scipy.signal.lfilter([b], [1, b - 1], ref_spec))**gain
```
lostanlen: Off band I explained why I liked
`smooth = np.exp(gain * (np.log(eps) + np.log1p(scipy.signal.lfilter([b], [1, b-1], ref_spec/eps)))`
better, because it's more numerically accurate for background time-frequency bins.
But it might be much, much slower (two special functions instead of one). It's a tradeoff I guess. I can't tell how important this is.

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+
+    return (S / smooth + bias)**power - bias**power
```
lostanlen: Same remark here:
`exp(power *(log(bias) + log1p(S/(bias*smooth)))) - bias**power`

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+    >>> plt.colorbar()
+    >>> plt.subplot(2,1,2)
+    >>> librosa.display.specshow(pcen_max, x_axis='time', y_axis='mel')
+    >>> plt.title('Per-channel energy normalization (max_size=5)')
+    >>> plt.colorbar()
+    >>> plt.tight_layout()
+
+    '''
+    if np.issubdtype(S.dtype, np.complexfloating):
+        warnings.warn('pcen was called on complex input so phase '
```
lostanlen: I tend to advocate for errors rather than warnings, but that's just me.

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+
+    and if `b` is not provided, it is calculated as:
+
+        b = 1 - exp(-hop_length / (sr * time_constant))
+
+    This normalization is designed to suppress background noise, and
+    emphasize foreground signals, and can be used as an alternative to
+    decibel scaling (`amplitude_to_db`).
+
+    This implementation also supports frequency-bin smoothing by specifying
```
lostanlen: Adjectived nouns are hard. "Frequency-bin smoothing" -> "smoothing across frequency bins"

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+    `max_size > 1`.  If this option is used, the filtered spectrogram `M` is
+    computed as
+
+        M[f, t] = (1 - b) M[f, t - 1] + b * R[f, t]
+        
+    where `R` has been max-filtered along the frequency axis, similar to
+    the Superflux algorithm implemented in `onset.onset_strength`:
```
lostanlen: SuperFlux (CamelCase)

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
 
+@cache(level=30)
+def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
+         time_constant=0.395, eps=1e-6, b=None, max_size=1):
+    '''Per-channel energy normalization (PCEN) [1]_
+
+    This function normalizes a time-frequency representation `S` by
+    performing automatic gain control, followed by non-linear compression:
+
+        P = (S / (eps + M)**gain + bias)**power - bias**power
```
lostanlen: `P[f, t] = (S[f, t] / (eps + M[f, t])**gain + bias)**power - bias**power`

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+
+    where `M` is the low-pass filtered output of `S`:
+
+        M[f, t] = (1 - b) M[f, t - 1] + b * S[f, t]
```
lostanlen: `M[f, t] = (1 - b) * M[f, t - 1] + b * S[f, t] `

Missing asterisk

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
     return result[idx] * np.sqrt(n) / n_fmt
 
 
+@cache(level=30)
+def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
+         time_constant=0.395, eps=1e-6, b=None, max_size=1):
+    '''Per-channel energy normalization (PCEN) [1]_
+
+    This function normalizes a time-frequency representation `S` by
+    performing automatic gain control, followed by non-linear compression:
```
lostanlen: non-linear -> nonlinear (Merriam Webster / Harper & Collins)

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+
+        P = (S / (eps + M)**gain + bias)**power - bias**power
+
+    where `M` is the low-pass filtered output of `S`:
+
+        M[f, t] = (1 - b) M[f, t - 1] + b * S[f, t]
+
+    and if `b` is not provided, it is calculated as:
```
lostanlen: "and if" -> "If"
let's break it into 2 sentences

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+    Parameters
+    ----------
+    S : np.ndarray (non-negative) [shape=(n, m)]
+        input spectrogram
+
+    sr : number > 0 [scalar]
+        Sampling rate of audio
+
+    hop_length : int > 0 [scalar]
+        Hop length of `S`
```
lostanlen: , expressed in audio samples.

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+
+    gain : number > 0 [scalar]
+        Gain factor.  Typical values should be slightly less than 1.
+
+    bias : number >= 0 [scalar]
+        The bias point of the non-linear compression
+
+    power : number > 0 [scalar]
+        The compression factor.  Typical values should be between 0 and 1.
```
lostanlen: It's not a factor, it's an exponent.
Maybe say that the lower the `power`, the stronger the compression?

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
+         time_constant=0.395, eps=1e-6, b=None, max_size=1):
+    '''Per-channel energy normalization (PCEN) [1]_
+
+    This function normalizes a time-frequency representation `S` by
+    performing automatic gain control, followed by non-linear compression:
+
+        P = (S / (eps + M)**gain + bias)**power - bias**power
+
+    where `M` is the low-pass filtered output of `S`:
```
lostanlen: "low-pass filtered output" 
Is "low-pass" an adverbed noun here? 

How about
"where every frequency channel in `M` is obtained from `S` by IIR low-pass filtering through time"?

(I don't love the term "frequency channel" otutside of radiocommunications but it matches the C in PCEN ... )

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+
+        R[f, t] = max(R[f - max_size//2: f + max_size//2, t])
+
+    This can be used to perform automatic gain control on signals that cross
+    or span multiple frequency bans, which may be desirable for spectrograms
+    with high frequency resolution.
+
+    .. [1] Wang, Y., Getreuer, P., Hughes, T., Lyon, R. F., & Saurous, R. A.
+       (2017, March). Trainable frontend for robust and far-field keyword spotting.
+       In Acoustics, Speech and Signal Processing (ICASSP), 2017
```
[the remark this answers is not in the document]
bmcfee: where?  I don't see it.  (Bib is copypasta from google scholar apa style.)

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,152 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+        The compression factor.  Typical values should be between 0 and 1.
+
+    time_constant : number > 0 [scalar]
+        The time constant for IIR filtering, measured in seconds.
+
+    eps : number > 0 [scalar]
+        A small constant used to ensure numerical stability of the filter.
+
+    b : number in [0, 1]  [scalar]
+        The filter coefficient for the low-pass filter.
```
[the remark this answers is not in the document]
bmcfee: If `b` is provided, the time constant parameter is ignored.  Or do you just mean the implicit time constant of the filter?

librosa/core/spectrum.py
bmcfee: This is a warning for consistency with the _to_db functions.  It's common to skip the `abs` step when it's obvious that phase should be discarded -- I think pcen is such a case.

librosa/core/spectrum.py
bmcfee: I think it's clear from context that these are element-wise operations?

librosa/core/spectrum.py
bmcfee: I'll look into it once the rest of the dust has settled, but for now, I'm inclined to keep it as is for simplicity.

bmcfee:
Thanks @lostanlen for the doc review!

Dangling to-dos:

- [ ] input checking
- [ ] unit tests
- [ ] stability test for smoothing operation

librosa/core/spectrum.py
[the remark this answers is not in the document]
bmcfee: Oh, you mean in the ICASSP title.  This style matches [the website](https://2018.ieeeicassp.org/), so I'm okay with leaving it as is.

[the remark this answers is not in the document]
lostanlen: You're right, upon verification there is inconsistency even within IEEE ... ¯\_(ツ)_/¯ 
https://signalprocessingsociety.org/blog/icassp-2018-2018-ieee-international-conference-acoustics-speech-and-signal-processing

librosa/core/spectrum.py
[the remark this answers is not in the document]
lostanlen: Yes, the implicit time constant of the filter.
Defining b as "the filter coefficient" does not help the reader who is fiddling with parameters assuming they skip the recurrence equation.

librosa/core/spectrum.py
lostanlen: yes but laying `S[f,t]` and `M[f,t]` in full highlights which variables are arrays and which are scalars.

bmcfee:
Unit tests cover parameter validation, and a few simple edge cases:

- disabling agc and bias should simplify to exponent compression
- disabling iir filter (`b=1`) and `bias=0` should produce all ones
- zeros should map to zeros
- shape and finitude are checked

Since we don't have a reference implementation for this, I'm not sure what else makes sense to test.

bmcfee:
I just tried out the stability comparison on @lostanlen's bird/car example.  The residual magnitudes for `eps=1e-6` are at most 4e-15 (with double-precision inputs).  For `eps=1e-20`, they're at most 8e-15.  In both cases, the residual difference concentrates on the louder parts of the signal, so I don't think the difference is significant:

image

As for timing, the log1p version is actually a bit faster:
```47.8 ms ± 800 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)```
vs
```69 ms ± 1.13 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)```
I'm not sure what entirely is going on there, but it does appear to be consistent and scale with the data.  On a longer track (4:44) with default stft parameters, the timing goes to ~1.0s for the current implementation and 740ms for the log calculation.  This seems like a speedup worth taking.  Thanks @lostanlen for the suggestion!

**Comment by bmcfee** [comment]
Reviewed 1 of 1 files at r6.
Review status: 1 of 3 files reviewed at latest revision, 20 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/679)*

File: librosa/core/spectrum.py [hunk 15]
```diff
@@ -1281,6 +1282,178 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+    >>> plt.colorbar()
+    >>> plt.tight_layout()
+
+    '''
+
+    if power <= 0:
+        raise ParameterError('power={} must be strictly positive'.format(power))
+
+    if gain < 0:
+        raise ParameterError('gain={} must be non-negative'.format(gain))
```
lostanlen: no hyphen needed in nonnegative (Merriam-Webster + Collins)

same remark a few lines below

librosa/core/spectrum.py
```diff
@@ -1281,6 +1282,178 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
...
+
+    if not 0 <= b <= 1:
+        raise ParameterError('b={} must be between 0 and 1'.format(b))
+
+    if np.issubdtype(S.dtype, np.complexfloating):
+        warnings.warn('pcen was called on complex input so phase '
+                      'information will be discarded. To suppress this warning, '
+                      'call pcen(magphase(D)[0]) instead.')
+
+    S = np.abs(S)
```
lostanlen: Shouldn't this be indented into the if statement? If `S` is real and above `-eps`, there is no need for `abs`. If `S` is real and has values below `-eps`, `S` is certainly not a magnitude spectrogram and PCEN should not be used, yet the code will run without error. This might introduce subtle bugs in the pipeline.

Issue #573, opened 2017-05-08
inconsistency with python_speech_features


chananshgong:
I compared the mfcc of [librosa](https://github.com/librosa/librosa/) with [python_speech_analysis](https://github.com/jameslyons/python_speech_features) package and got **totally different results**. 

Which one is correct?
librosa list of first frame coefficients:

> [-395.07433842032867, -7.1149347948192963e-14, 3.5772469223901538e-14, -1.7476140989485184e-14, 3.1665300829452658e-14, -4.4214136625668904e-14, 6.7157035631648599e-14, 1.5013974158050108e-14, 2.9512326634271699e-14, 7.2275398398734558e-14, -1.5043753316598812e-13, -2.2358383003147776e-14, 1.6209256159527285e-13]

python_speech_analysis list of first frame coefficients:

> [-169.91598446684722, 1.3219891974654943, 0.22216979881740945, -0.7368248288464827, 0.26268194306407788, 1.8470757480486224, 3.2670900572694435, 2.3726120692753563, 1.4983949546889608, 0.67862219561000914, -0.44705590991616034, 0.39184067109778226, -0.48048214059101707]

```
import librosa
import python_speech_features
from scipy.signal.windows import hann

n_mfcc = 13
n_mels = 40
n_fft = 512 # in librosa, win_length is assumed to be equal to n_fft implicitly
hop_length = 160
fmin = 0
fmax = None
y, sr = librosa.load(librosa.util.example_audio_file())
sr = 16000  # fake sample rate just to make the point

# librosa
mfcc_librosa = librosa.feature.mfcc(y=y, sr=sr, n_fft=n_fft,
                                    n_mfcc=n_mfcc, n_mels=n_mels,
                                    hop_length=hop_length,
                                    fmin=fmin, fmax=fmax)

# python_speech_features
# no preemph nor ceplifter in librosa, so setting to zero
# librosa default stft window is hann
mfcc_speech = python_speech_features.mfcc(signal=y, samplerate=sr, winlen=n_fft / sr, winstep=hop_length / sr,
                                          numcep=n_mfcc, nfilt=n_mels, nfft=n_fft, lowfreq=fmin, highfreq=fmax,
                                          preemph=0, ceplifter=0, appendEnergy=False, winfunc=hann)


print(list(mfcc_librosa[:, 0]))
print(list(mfcc_speech[0, :]))
```

bmcfee:
[...]

Issue #437, opened 2016-11-12
DCT type question


georgid:
Why did you bother to implement the DCT, and not use the scipy's [DCT](https://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.fftpack.dct.html)  
As far as I see it is used only in the MFCC. Could the reason be that for MFCC a particular detail is not  in scikits implementation?

Also I am wondering, why do you use DCT type III by default and not have the option of DCT type II? DCT type III is used by htk, MFCC,  whereas DCT type II is used in Slaney's MFCC.

bmcfee:
> Why did you bother to implement the DCT, and not use the scipy's DCT

This was inherited from the matlab implementation we were basing the original implementation on.  For backwards compatibility, it was simpler to translate the code directly.

> Also I am wondering, why do you use DCT type III by default and not have the option of DCT type II? DCT type III is used by htk, MFCC, whereas DCT type II is used in Slaney's MFCC.


Sounds like a good idea.  Pull requests are always welcome!

bmcfee:
Circling back on this one: librosa's dct does not match scipy's. For reference: librosa's dct was based on this implementation: https://labrosa.ee.columbia.edu/matlab/rastamat/spec2cep.m in type-3 mode, and these two implementations do match (down to around 6 decimal places).

scipy's dct behaves differently, but it doesn't expose an interface to the basis, so it's not trivial to work out exactly how it's different.

It would be fairly easy to add functionality in `mfcc` to call out to `scipy.fftpack.dct(...)` with pass-through parameters, instead of using the existing one, but making this the default would be a breaking change.  It still seems like a reasonable feature to add though.

bmcfee:
After digging into this a bit more over on #682 -- it seems like the documentation is actually incorrect.  Librosa's dct basis matches scipy to numerical tolerance as type-II with `norm='ortho'`, not type-III as advertised.  I'll clarify this in the PR.

georgid:
Good to know, thank you!

PR #682, opened 2018-02-22
Expanded MFCC options

bmcfee:
#### Reference Issue
#437 
#573 

This expands the mfcc interface to allow the user to use scipy's dct implementation.  This may better approximate alternative mfcc implementations, such as Slaney's auditory toolbox (uses type-II dct) or HTK (type-III + liftering).

#### Any other comments?

Maybe we should roll liftering into this PR as well?  It probably wouldn't hurt.

As it currently stands, there are no regression tests for mfcc.  This is largely due to the integrated implementation in rastamat being hard to factor out for component-wise testing.  I'm curious if anyone has thoughts on how sanely test this though.

~~~Side note: the rastamat (and by inheritance, librosa) type-III dct does not match with scipy's, with or without normalization.  The differences are pretty stark, and I'm not 100% sure I understand why this should be the case.  That said, I think I prefer librosa's dct because for a constant signal (say, all ones), it provides an MFCC output that closely resembles an impulse; scipy's dct does not.~~ Disregard -- I was using `fftpack.dct` incorrectly.  It does turn out that the librosa docs are wrong though: the dct basis used is Type-II (normalized), not Type-III.

<!-- Reviewable:start -->
---
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/librosa/librosa/682)
<!-- Reviewable:end -->

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -1305,7 +1305,7 @@ def tonnetz(y=None, sr=22050, chroma=None):
 
 
 # -- Mel spectrogram and MFCCs -- #
-def mfcc(y=None, sr=22050, S=None, n_mfcc=20, dct_type=None, norm=None, **kwargs):
+def mfcc(y=None, sr=22050, S=None, n_mfcc=20, dct_type=2, norm='ortho', **kwargs):
     """Mel-frequency cepstral coefficients (MFCCs)
 
     Parameters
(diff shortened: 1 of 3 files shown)


bmcfee:
I added a couple of simple tests for matching the energy envelope (mfcc 0) and general shape checks.

Docs have been updated.

`filters.dct` has been deprecated in favor of a direct call to `scipy.fftpack.dct`.  It's a bit less efficient, since it constructs the full basis and then truncates, but in the long run it means less code for us to maintain.

bmcfee:
Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/682)*

bmcfee:
Reviewed 1 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/682)*

librosa/feature/spectral.py
```diff
@@ -1370,13 +1382,26 @@ def mfcc(y=None, sr=22050, S=None, n_mfcc=20, **kwargs):
...
+    >>> plt.title('HTK-style (dct_type=3)')
+    >>> plt.colorbar()
+    >>> plt.tight_layout()
     """
 
     if S is None:
         S = power_to_db(melspectrogram(y=y, sr=sr, **kwargs))
 
-    return np.dot(filters.dct(n_mfcc, S.shape[0]), S)
+    return scipy.fftpack.dct(S, axis=0, type=dct_type, norm=norm)[:n_mfcc]
```
Marcel-G, line 1404: norm is also a valid arg for `melspectrogram`

bmcfee: This PR was merged over a year ago.  If this is still an issue, please open a new report.  Thanks!

Issue #683, opened 2018-02-26
Specify axis instance in plotting functions


choldgraf:
I noticed that there's no option to say "plot this spectrogram on this specific axis object". The code just uses `plt.gca`. Would you be friendly to `specshow(..., ax=myaxis)`. And under the hood it'd do `if ax is None: axis = plt.gca`?

bmcfee:
Good catch -- I've implemented that exact thing in related packages (jams, mir_eval), but somehow never did it for librosa.  And yeah, a pr for this would be great!

bmcfee:
Closed by merging #687

PR #687, opened 2018-03-07
Specifying axis instances in plotting functions

pfriesch:
Adding https://github.com/librosa/librosa/issues/683

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/687)


bmcfee:
Thanks for implementing this!  It looks good, but I have two questions:

1. Is there a convention for the parameter name for matplotlib axes?  I see `ax=` often (eg, seaborn and mir_eval both use this), but I don't know if it's standard.
2. Could you add a simple test for this?  Something like making two subplots, and drawing to them out of order using the `axes=` parameter ought to do the trick.  We already have plenty of tests implicitly drawing to gca, so I think we don't have to worry about that case.

isn0gud:
I don't know about any convention, but `ax=` sounds reasonable. I am gonna change it then.

I actually ran the nose tests and it seemed like most of them were failing, so I assumed the project gave up on tests. But yeah! I will surely write some tests. I am just gonna convert the existing tests to use an external axes instance then.

bmcfee:
> I actually ran the nose tests and it seemed like most of them were failing, so I assumed the project gave up on tests.

:fearful: definitely not!  The tests should not be failing, though you might have to install a specific version of matplotlib to get them to run.  (We do some ugly things to yank the image comparison test code out of mpl, and it doesn't work for every mpl version.)


> I am just gonna convert the existing tests to use an external axes instance then.

I would prefer it if the existing tests are left as is, and we just have two new tests for the axis override case: one for waveplot and one for specshow.

isn0gud:
>I would prefer it if the existing tests are left as is, and we just have two new tests for the axis override case: one for waveplot and one for specshow.

Sorry that was what I meant.

isn0gud:
I changed the name and added a test. However, the test feels a bit hacky. Maybe you have a pointer how it would to do it better? A different decorator would probably make sense?!

bmcfee:
@friesch Thanks for adding the tests.  I think I see what you mean about the decorator -- as I understand it, the regression test decorators are meant to work with a single figure, so adding multiple figures (to get axes) seems a bit circuitous.

Maybe it would be simpler to do something like the following:

```python
ax_left = plt.subplot(1,2,1)
ax_right = plt.subplot(1,2,2)
# ax2 is now the current axes
librosa.display.specshow(..., cmap='gray')  # draw on gca, implicitly ax_right
librosa.display.specshow(..., cmap='magma', ax=ax_left)  # draw on the left subplot != gca()
```
Using different colormaps in each subplot should make it really easy to detect if it's behaving correctly: left plot should be magma, right plot should be grayscale.

You could do something similar for waveplot, using different colors instead of colormaps.

friesch:
I am not exactly sure what function `plt.sci(out)` has in `specshow`, but I guess it should only be called if the axes is not given.

librosa/display.py
```diff
@@ -356,6 +356,9 @@ def waveplot(y, sr=22050, max_points=5e4, x_axis='time', offset=0.0, max_sr=1000
     x_axis : str {'time', 'off', 'none'} or None
         If 'time', the x-axis is given time tick-marks.
 
+    axes : matplotlib.axes.Axes or None
```
choldgraf: this should be `ax`, no?

librosa/display.py
```diff
@@ -435,7 +438,7 @@ def waveplot(y, sr=22050, max_points=5e4, x_axis='time', offset=0.0, max_sr=1000
         y_top = y
         y_bottom = -y
 
-    axes = plt.gca()
+    axes = ax if ax is not None else plt.gca()
```
choldgraf: here you may wanna check to make sure users actually give an axis object (if not None), because often people will get confused when they use `plt.subplots` w/ more than one axis, it'll give then a numpy array and then the plotting calls will have stranger error messages.

(the way I usually do this is to have a hidden `_check_ax` function to make this check throughout, see https://github.com/librosa/librosa/pull/689/files#diff-1ad5980f89cf6076183eced46362487aR901 for example)

librosa/display.py
```diff
@@ -542,6 +546,9 @@ def specshow(data, x_coords=None, y_coords=None,
     bins_per_octave : int > 0 [scalar]
         Number of bins per octave.  Used for CQT frequency scale.
 
+    axes : matplotlib.axes.Axes or None
```
choldgraf: also ax

librosa/display.py
```diff
@@ -887,4 +895,4 @@ def __coord_n(n, **_kwargs):
 
 def __coord_time(n, sr=22050, hop_length=512, **_kwargs):
     '''Get time coordinates from frames'''
-    return core.frames_to_time(np.arange(n+1), sr=sr, hop_length=hop_length)
+    return core.frames_to_time(np.arange(n+1), sr=sr, hop_length=hop_length)
```
choldgraf: could you add in an empty new line at the end?

choldgraf:
a couple small suggestions, but this looks quite nice! thanks @friesch :-)

friesch:
Done :)

choldgraf:
woot! I'm +1 on merge...@bmcfee will need to hit the green button tho :-)

bmcfee:
Reviewed 4 of 5 files at r3.
Review status: 4 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[librosa/display.py, line 687 at r4](https://reviewable.io:443/reviews/librosa/librosa/687#-L83U3hfQuj_iGSelbny:-L83U3hg5A3bFENn0cUx:baja141) ([raw file](https://github.com/librosa/librosa/blob/701f24b0d3ae93446f4f32a9c92cc7f788b3b1ee/librosa/display.py#L687)):*
> ```Python
>     out = axes.pcolormesh(x_coords, y_coords, data, **kwargs)
>     if ax is None:
>         plt.sci(out)
> ```

Having dug through matplotlib a bit, I think we might actually want to keep this in here regardless of whether ax was provided.

What `plt.sci()` does is set the current image in the pyplot stateful api.  I think the convention is that this should always be the most recently drawn image, but I'm not 100% on that and have not found any definition docs to that effect.  We might have to ping the matplotlib developers for advice here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/687)*

bmcfee:
Review status: 4 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[librosa/display.py, line 687 at r4](https://reviewable.io:443/reviews/librosa/librosa/687#-L83U3hfQuj_iGSelbny:-L83xC7GuBaAGOTdrQmK:b-2zkuqu) ([raw file](https://github.com/librosa/librosa/blob/701f24b0d3ae93446f4f32a9c92cc7f788b3b1ee/librosa/display.py#L687)):*

Having dug through matplotlib a bit, I think we might actually want to keep this in here regardless of whether ax was provided.

What `plt.sci()` does is set the current image in the pyplot stateful api.  I think the convention is that this should always be the most recently drawn image, but I'm not 100% on that and have not found any definition docs to that effect.  We might have to ping the matplotlib developers for advice here.
</blockquote></details>

Ok, I checked in with the MPL devs, and it's a rather sticky situation.  The recommendation is generally that a function like this should be setting `sci` regardless of pyplot or object interface mode.

Where it gets sticky is that, as a library, we shouldn't be relying on the pyplot interface anyway, since importing it has side effects (choosing a backend).  This has caused us some headaches in the past, which lead to the optional import of the `display` module.

I think the best solution, for now, is to leave the `sci` call in, and we'll handle factoring out pyplot dependencies as a separate issue down the road.  There are only a couple of places where this comes up, but it will be better to handle them all at once.

@choldgraf @pfriesch does that sound good to you?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librosa/librosa/687)*

pfriesch:
Getting a `ValueError: Argument must be an image, collection, or ContourSet in this Axes` then though. I guess it is the case because the given axes is not the active one.

bmcfee:
Ah, that makes total sense.

In that case, leave it as is (only sci if ax=gca).  At that point, I think we're good to merge unless there's anything else?

Thanks again for putting the work into this!

bmcfee:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/librosa/librosa/687)*

choldgraf:
+1 to merge

bmcfee:
FTR: I just made a new issue for factoring out pyplot dependencies.  If any of you feel inclined to take that on as well, go right ahead :grin: but otherwise I'm happy to do it in the coming weeks.  (I have a good idea of how it should work now, but it does get into some rather grungy details of probing `sys.modules` and redirecting getters for, eg, colormap and current axes.)

bmcfee:
Thanks again @pfriesch and @choldgraf !

choldgraf:
wooo!
--

Issue #157, opened 2015-01-24
numbafy


bmcfee:
Wherever it makes sense to accelerate loop code, we should jit-compile.

Specifically,
- [x] util.match_events
- [x] util.match_intervals
- [ ] util.peak_pick
- [ ] beat.__beat_track_dp
- [ ] core.constantq.cqt internals
  - [ ] trim_stack
  - [x] num_two_factors
- [ ] core.spectrum.stft
- [ ] core.spectrum.istft
- [ ] decompose.nn_filter
- [ ] feature.util.sync
- [x] feature.util.stack_memory
- [x] segment.lag_to_recurrence, segment.recurrence_to_lag
- [ ] effects.remix
- [ ] filters.mel_frequencies
- [ ] filters.chroma

bmcfee:
I took a quick cut at cythonizing stft today.

A naive, frame-oriented cython implementation gets almost as fast as our current batched (pure-python) implementation.

bmcfee:
Updating: now that #323 is merged, we have optional numba jit decoration.  This will be considerably easier to deploy than cythonization.

bmcfee:
Tagging @ejhumphrey if he wants to take a look into this

bmcfee:
Maybe as a subtask, it would help to implement a profiling test suite so we know where the hot spots are.

stefan-balke:
Like that one? https://pypi.python.org/pypi/pytest-profiling

That would be the output for `py.test test_core.py --profile`

```
Profiling (from prof/combined.prof):
Sat Oct  8 17:08:24 2016    prof/combined.prof

         4845410 function calls (4844615 primitive calls) in 127.496 seconds

   Ordered by: cumulative time
   List reduced from 272 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2120    0.097    0.000  127.498    0.060 {built-in method builtins.exec}
     2120    0.025    0.000  127.395    0.060 pluggy.py:586(execute)
     2120    0.208    0.000  127.367    0.060 python.py:277(pytest_pyfunc_call)
     1228    0.092    0.000  101.133    0.082 test_core.py:761(__test)
     2681    0.316    0.000  100.665    0.038 spectrum.py:770(fmt)
     2465    0.338    0.000   98.920    0.040 interpolate.py:408(__init__)
     2458    1.273    0.001   98.318    0.040 interpolate.py:2312(splmake)
[...]

PR #688, opened 2018-03-14
Match events

danagilliann:
#### Reference Issue
Fixing a bit of #157 

This Numbafies the match_events function. The function uses Numba and the Numpy functions used in the function are supported by Numba.

#### Any other comments?
It uses some Numpy functions to be used as an alternative for a dictionary


bmcfee:
Thanks for getting this going!

It looks like the tests are failing now.  You can see the log [here](https://ci.appveyor.com/project/bmcfee/librosa/build/1.0.96/job/4m92iaqng7a38daj) to help track down what's going wrong.  I'll hold off on starting the code review until the tests are passing, but please let me know if you run into trouble debugging.

lostanlen:
What is the different between this and `mir_eval.util.match_events` ?

bmcfee:
> What is the different between this and mir_eval.util.match_events ?

mir_eval's event matching Hopcroft-Karp (maximal bipartite matching), and is meant for assigning at most one match per event.  It is intended for use in evaluation, and supports a window parameter to specify what can or cannot be counted as a match.

librosa's event matching is much simpler proximity search, and does not prevent multiple assignment.  It is primarily intended for sample alignment problems, such as rounding a frame boundary to the nearest zero-crossing, or matching a segment boundary to the nearest beat boundary.

So, tldr, the functions have nothing to do with each other, but happen to share a name.

danagilliann:
@bmcfee how should I go about the code coverage decreasing? Are there any additional tests I need to write?

bmcfee:
That's weird -- looks like appveyor broke.  I'll try a fresh build.

danagilliann:
@bmcfee TRAVIS IS WORKING!!!! 🎉

bmcfee:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[librosa/util/utils.py, line 907 at r4](https://reviewable.io/reviews/librosa/librosa/688#-LArneXq5KeCs8v7DyeV:-LArneXq5KeCs8v7DyeW:bgzx25j) ([raw file](https://github.com/librosa/librosa/blob/d482ea4b82cf3c592543f3f9c7f768720107c01c/librosa/util/utils.py#L907)):*
> ```Python
>     # If we can't match left or right, then only strict equivalence
>     # counts as a match.
>     if not (left or right) and not set(events_from).issubset(set(events_to)):
> ```

Since this bit is no longer under numba-jit, can we revert it to use `np.in1d`?  Casting to `set` twice will invoke more memory copies than are necessary.

It also might be more efficient to check `len(np.setdiff1d(events_from, events_to)) == 0` than doing `all(np.in1d(...)`.

---

*[librosa/util/utils.py, line 925 at r4](https://reviewable.io/reviews/librosa/librosa/688#-LAroR6L4OiWkHZk5PJs:-LAroR6L4OiWkHZk5PJt:b-h4mlta) ([raw file](https://github.com/librosa/librosa/blob/d482ea4b82cf3c592543f3f9c7f768720107c01c/librosa/util/utils.py#L925)):*
> ```Python
> 
>     # array of matched items
>     output = np.empty_like(events_from, int)
> ```

Second arg should have an explicit name, `dtype=np.int`

---

*[librosa/util/utils.py, line 931 at r4](https://reviewable.io/reviews/librosa/librosa/688#-LAroVqZB-q0aI5lqXsZ:-LAroVqZB-q0aI5lqXs_:bj2rlwp) ([raw file](https://github.com/librosa/librosa/blob/d482ea4b82cf3c592543f3f9c7f768720107c01c/librosa/util/utils.py#L931)):*
> ```Python
> 
> @jit(nopython=True)
> def logic_match_events(output, events_from, events_to, left=True, right=True):
> ```

Since this is an internal function, its name should start with `__`; maybe `__match_events_helper` or something.

---

*[librosa/util/utils.py, line 943 at r4](https://reviewable.io/reviews/librosa/librosa/688#-LArokf139cIzcCaCkhp:-LArokf139cIzcCaCkhq:buvpg10) ([raw file](https://github.com/librosa/librosa/blob/d482ea4b82cf3c592543f3f9c7f768720107c01c/librosa/util/utils.py#L943)):*
> ```Python
> 
>     # iterate over indices in matching_indices
>     for ind in range(0, len(matching_indices)):
> ```

you don't need the `0` here.

General python style thing: this loop would be clearer if written as:

`for ind, middle_ind in enumerate(matching_indices):`

---

*[librosa/util/utils.py, line 948 at r4](https://reviewable.io/reviews/librosa/librosa/688#-LArq1pREowons6U8NgJ:-LArq1pREowons6U8NgK:b-omynzp) ([raw file](https://github.com/librosa/librosa/blob/d482ea4b82cf3c592543f3f9c7f768720107c01c/librosa/util/utils.py#L948)):*
> ```Python
> 
>         middle_ind = matching_indices[ind]
>         sorted_from_num = sorted_from[ind]
> ```

I think i see where the bounds errors were coming from before: using `-1` as a place-holder for array indices can be problematic, since -1 is a valid array index in python (last position).  Likewise, `len(matching_indices)` is not a valid array index, since it's out of bounds.

Would it be possible to rewrite these two special cases using `None` instead?  That way, the semantics are clear that these are not valid indices.

---

*[librosa/util/utils.py, line 955 at r4](https://reviewable.io/reviews/librosa/librosa/688#-LArp7Pp4fg9uPPNo3Vf:-LArp7Pp4fg9uPPNo3Vg:b-2sf3fv) ([raw file](https://github.com/librosa/librosa/blob/d482ea4b82cf3c592543f3f9c7f768720107c01c/librosa/util/utils.py#L955)):*
> ```Python
> 
>         # Permitted to look to the left
>         if middle_ind > 0 and left:
> ```

general thing: it'll be a little quicker if you check `left and middle_ind > 0` instead of the other way around.

This is because `left` is constant over the function, and `and` is a short-circuit operator: f its first argument is false, the second is never checked.  Likewise for the right check below.

---

*[librosa/util/utils.py, line 968 at r4](https://reviewable.io/reviews/librosa/librosa/688#-LArpb0c3jmYIPDZq9Uo:-LArpb0c3jmYIPDZq9Up:b-r8kgvx) ([raw file](https://github.com/librosa/librosa/blob/d482ea4b82cf3c592543f3f9c7f768720107c01c/librosa/util/utils.py#L968)):*
> ```Python
>                     abs(sorted_to[right_ind] - sorted_from_num) and
>                     abs(sorted_to[left_ind] - sorted_from_num) <
>                     abs(sorted_to[middle_ind] - sorted_from_num))):
> ```

There's a bit of redundant calculation here, and it's difficult to read.  I suggest rewriting these conditions with temporary variables:

`left_diff = abs(sorted_to[left_ind] - sorted_from_num`
`right_diff = abs(sorted_to[right_ind] - sorted_from_num`
`mid_diff = abs(sorted_to[middle_ind] - sorted_from_num`

---


*Comments from [Reviewable](https://reviewable.io/reviews/librosa/librosa/688)*

bmcfee:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[librosa/util/utils.py, line 977 at r5](https://reviewable.io/reviews/librosa/librosa/688#-LBN8SLU9qMo9EUpFozV:-LBN8SLU9qMo9EUpFozW:bau7u0m) ([raw file](https://github.com/librosa/librosa/blob/aec67fa2427bbff0e34f7d0c8b5745600b58732c/librosa/util/utils.py#L977)):*
> ```Python
>             right_diff = abs(sorted_to[right_ind] - sorted_from_num)
> 
>         if ((left and left_flag) and
> ```

If I understand properly, `left_flag` can only be true if `left` is true.  You can then simplify this check (and the one below) to only look at `left_flag`.  No major efficiency change here, just readability.

---


*Comments from [Reviewable](https://reviewable.io/reviews/librosa/librosa/688)*

bmcfee:
Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io/reviews/librosa/librosa/688)*

danagilliann:
Hello! I did some benchmarking, and you can see the results [here](https://gist.github.com/danagilliann/ad65e4877570ce2cd621de5c09e89559).

In general, the numba-fication of the function has sped up the process by a signifcant amount of time. For the benchmarking, I had `beats` as `events_to` and compared that to lists of random numbers with lengths of 1, 10, 100, and 1000 populated based on the range `[0, len(audio_buffer)]`.

For the list of random numbers, here are the results of the speed up
length 1 : 7x
length 10: 7x
length 100: 15x
length 1000: 12x

🎉

bmcfee:
Followup: running this with beats->zero crossings on the one-minute included example yields speedup from 30-70x (depending on the number of beat events). This is huge!

Otherwise, the ci failures are nbd here, and out of scope for this pr. Lgtm!

PR #689, opened 2018-03-14
adding axes kwarg

choldgraf:
This is a first-pass at adding the ability to pre-specify a matplotlib axis object when doing plotting. This way you can do

```
fig, ax = plt.subplots()
lr.myplotfunction(..., axes=ax)
```

LMK how this looks (or if I'm missing any functions this could be added to)

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/689)

### DIFF
--- a/docs/examples/plot_music_sync.py
+++ b/docs/examples/plot_music_sync.py
@@ -97,23 +97,19 @@ plt.colorbar()
 # Alternative Visualization in the Time Domain
 # --------------------------------------------
 #
-# We can also visualize the wariping path directly on our time domain signals.
+# We can also visualize the warping path directly on our time domain signals.
 # Red lines connect corresponding time positions in the input signals.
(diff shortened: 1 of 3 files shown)


bmcfee:
Thanks @choldgraf !  You perhaps missed that there's already a PR for this open? #687 

Maybe you two can team up?

choldgraf:
Whoops! No I did not see that haha. Don't think I can push to their pr, but
happy to do so if I can be of help there.

choldgraf:
Actually that pr looks pretty much ready to go. I'm not sure how I could give input in a way that wouldn't add unnecessary complexity to the pr (it seems like a good enough solution)... I think I'll just close this one but please do ping me if there's something specific I can do there

friesch:
@choldgraf I added your example to the PR. Anything else you want me to add?

Issue #690, opened 2018-03-20
Remove pyplot requirements from display submodule


bmcfee:
#### Description
The `display` submodule is currently written primarily using the `pyplot` interface to matplotlib.

This is generally not a great thing for library code, because it forces an import to `pyplot`, which has side effects (setting a backend, for example).

There are relatively few parts where we need to talk to `pyplot` though, and they can generally be worked around if the user supplies axes arguments as per #687.  We should be able to rewrite things so as to not require `plt` unless forced to.

This will require a little bit of trickery and indirection to get right and preserve the current behavior without imposing unwarranted side effects.  pyplot import is unavoidable when display functions are called without `axes` parameters, but we can be more careful about how this is managed.

bmcfee:
For future reference, here are the parts that depend on `matplotlib.pyplot`:

- `cmap` uses `plt.get_cmap`.  This should be rewritten to use `matplotlib.cm.get_cmap` instead.
- `specshow` uses `plt.sci` (if the user does not supply axes).  This could be wrapped in a helper function that handles dynamic `plt` import, and is only called if the user does not supply axes.
- `__check_axes` uses `plt.gca()` and `plt.Axes` (for a type check).  The former is unavoidable, but could use a dynamic import to delay side effects as long as possible.  The latter should be rewritten to use `matplotlib.axes.Axes`.

PR #700, opened 2018-04-20
add display to __init__.py

SongweiGe:
to fix module 'librosa' has no attribute 'display'

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/700)

### DIFF
--- a/librosa/__init__.py
+++ b/librosa/__init__.py
@@ -12,6 +12,7 @@ from . import cache
 from . import core
 from . import beat
 from . import decompose
+from . import display
 from . import effects
 from . import feature
 from . import filters


stefan-balke:
Hey,

this is on purpose: #343

bmcfee:
Thanks @SongweiGe , but as @stefan-balke mentioned, this is intentional behavior following the thread he linked to.  To summarize: importing matplotlib (actually, matplotlib.pyplot) has side-effects and can cause memory leakage.

There's an open issue #690 to remove the `matplotlib.pyplot` dependency from `librosa.display`, which if done, should avoid any side-effects.  Once that's done -- if it's even possible -- then it might make sense to auto-import `display` again.

SongweiGe:
Got it. Thanks guys!

PR #704, opened 2018-04-30
Isolated matplotlib.pyplot dependencies; fixes #690

### LINKED ISSUE #690: Remove pyplot requirements from display submodule

bmcfee:
#### Reference Issue
Fixes #690 

This PR isolates all dependencies on `matplotlib.pyplot` to avoid import side-effects.

As noted in #690, the issue here is that `librosa.display` should be compatible with matplotlib's stateful (`pyplot`) and objective APIs.  Compatibility with `pyplot` requires accessing the `matplotlib.pyplot` module for managing state (getting current axes, setting current images, etc), but importing this module incurs side-effects.  If the user is only using the object interface, then we shouldn't import pyplot.

Otherwise, there should be no externally visible changes to the API or functionality.

#### Any other comments?

Is it worth revisiting the dynamic import issue for `display` #343 , now that the side effect issue should be neutralized?  On the one hand, dynamic submodule imports are ugly.  On the other hand, it would turn an optional dependency (matplotlib) back into a hard dependency.  It's mainly a trade-off between dependency simplicity and (space/memory) efficiency at runtime.

What do folks think?  I'm inclined to leave it as is (optional), since it's been that way for a couple of years and people should be used to it by now.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/704)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,10 +2,10 @@ build: false
 
 environment:
   matrix:
-    - PYTHON_VERSION: 3.6
-      MINICONDA: C:\Miniconda3
     - PYTHON_VERSION: 2.7
       MINICONDA: C:\Miniconda
+    - PYTHON_VERSION: 3.6
+      MINICONDA: C:\Miniconda3
 
 cache:
     - "%MINICONDA%\\envs -> appveyor.yml"
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -20,7 +20,8 @@ Display
 import warnings
 
 import numpy as np
-import matplotlib.pyplot as plt
+from matplotlib.cm import get_cmap
+from matplotlib.axes import Axes
 from matplotlib.ticker import Formatter, ScalarFormatter
 from matplotlib.ticker import LogLocator, FixedLocator, MaxNLocator
 from matplotlib.ticker import SymmetricalLogLocator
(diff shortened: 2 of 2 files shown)


bmcfee:
@choldgraf you have any thoughts on the above?

choldgraf:
I like it!

bmcfee:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/librosa/librosa/704)*

bmcfee:
Update: I thought more about the `display` import issue, and think it's worth leaving as is (optional).  Even without side effects, mpl is huge, and quite independent of the rest of librosa.  Once CI finishes, I think this is good to go.

Issue #706, opened 2018-05-07
General match_intervals improvements


bmcfee:
#### Description
In auditing `match_intervals` in preparation for numbafication #157, I noticed a couple of undocumented / undefined behaviors that should be made explicit.  I'm tagging these as "bug" because they are definitely not intended behavior, and could silently cause some strange behaviors down the road.

1. What should happen when a query interval has no overlap with any of the targets?  The current logic for this is [here](https://github.com/librosa/librosa/blob/master/librosa/util/utils.py#L837), which would amount to argmax over an all zeros vector.  `np.argmax` breaks ties in favor of the first maximal element, which in this case, would be 0 (the first target interval).  This seems bad and wrong here.

  Some suggested alternative behaviors:

  - Use a null value.  `None` seems reasonable, but it would require returning a list of ints instead of a `ndarray(dtype=int)`.  `NaN` also seems reasonable, but would require `dtype=float`, which would not work with fancy indexing.
  - Throw an exception.  At the very least, we should throw a warning when this happens.
  - Map to the "closest" non-overlapping interval [in the set sense](https://en.wikipedia.org/wiki/Distance#Distances_between_sets_and_between_a_point_and_a_set).
  - Maybe have a user-supplied option to switch between some of these modes?

2. How should we break ties?  The current interval scoring uses the [raw intersection](https://github.com/librosa/librosa/blob/master/librosa/util/utils.py#L835) between intervals, and then selects the first maximal target as the match.  This seems wrong / under-determined when there are overlapping intervals.  I think [Jaccard similarity](https://en.wikipedia.org/wiki/Jaccard_index) is a better selection criterion, being a proper metric and all.  This would break ties in favor of the "tightest" matching interval, which I think is sensible.

### A note on candidate filtering

@danagilliann and I worked out a slightly faster matching algorithm than what we currently have.  It's still `O(nm)` in the worst case, but in the average case, should behave more like `O(m log m + n log m)`.

Here's some simplified pseudo-code of the current method (brute-force quadratic search):
```python
output = []
for query in from_intervals:
   best_score = 0
   best_idx = 0
   for idx, candidate in enumerate(to_intervals):
      score = similarity(query, candidate)
      if score > best_score:
          best_score, best_idx = score, idx
   output.append(best_idx)
```

[...]

PR #714, opened 2018-05-18
Rewrite of match_intervals

bmcfee:
#### Reference Issue
Implements #706 

This PR includes a rewrite of the `match_intervals` function as described in #706.

This was originally intended as an acceleration enhancement, but we discovered some unspecified behavior bugs in the process, and it is now more of a bug-fix PR.

The behavior of `match_intervals` has changed somewhat.  In the old version, it would find the interval with maximum intersection, and if none exists, it would match to 0 (unintended behavior).

The new implementation operates in two modes, `strict=True|False`.  In both cases, overlapping intervals are matched to maximize Jaccard similarity.  When `strict=True`, if a query interval is disjoint from all targets, an exception is now thrown.  When `strict=False`, a disjoint query interval will instead match to the target with boundary closest to the query -- this can be interpreted as tie-breaking when no intersection exists.

`strict=True` is the default.

#### Any other comments?

The new implementation is a bit slower than the old implementation for common input sizes.  However, it is still acceptably fast, and now has well-specified behavior for all inputs.

Because there is now substantial helper code for the matching functions (intervals and events), they have been broken out of `util.utils` and placed in a separate submodule `util.matching`.  This should not have any user-facing effect.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/714)


bmcfee:
Note: this PR adds a hard requirement on `numba >= 0.38` because it uses `searchsorted(..., side=)`.

bmcfee:
While we're at it, I threw in a bunch of `# pragma: no cover` lines on all of our jitted functions.  Hopefully that fixes our broken coverage reports.

bmcfee:
Reviewed 6 of 7 files at r1, 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/librosa/librosa/714)*

bmcfee:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io/reviews/librosa/librosa/714)*

Issue #729, opened 2018-06-22
joblib 0.12 causes TypeError at import


mvollrath:
#### Description
librosa fails to import with joblib 0.12 installed.  Works with joblib 0.11.

#### Steps/Code to Reproduce
```
import librosa
```

#### Expected Results
No `TypeError`

#### Actual Results
```
  File "...", line 19, in <module>
    import librosa
  File "/usr/local/lib/python2.7/dist-packages/librosa/__init__.py", line 12, in <module>
    from . import core
  File "/usr/local/lib/python2.7/dist-packages/librosa/core/__init__.py", line 104, in <module>
    from .time_frequency import *  # pylint: disable=wildcard-import
  File "/usr/local/lib/python2.7/dist-packages/librosa/core/time_frequency.py", line 10, in <module>
    from ..util.exceptions import ParameterError
  File "/usr/local/lib/python2.7/dist-packages/librosa/util/__init__.py", line 67, in <module>
    from .utils import *  # pylint: disable=wildcard-import
  File "/usr/local/lib/python2.7/dist-packages/librosa/util/utils.py", line 111, in <module>
    def valid_audio(y, mono=True):
  File "/usr/local/lib/python2.7/dist-packages/librosa/cache.py", line 49, in wrapper
    if self.cachedir is not None and self.level >= level:
  File "/usr/local/lib/python2.7/dist-packages/joblib/memory.py", line 847, in cachedir
    DeprecationWarning, stacklevel=2)
TypeError: expected string or buffer
```

#### Versions
```
Linux-4.4.0-119-generic-x86_64-with-Ubuntu-16.04-xenial
('Python', '2.7.12 (default, Dec  4 2017, 14:50:18) \n[GCC 5.4.0 20160609]')
('NumPy', '1.14.3')
('SciPy', '1.1.0')
('librosa', '0.5.1')
```

mvollrath:
Looks like joblib's deprecation warning is raising a `TypeError`, not as gentle as they intended..

bmcfee:
Yikes, thanks for catching this!  It looks like the changelog for joblib isn't even up to date with 0.12.  Any idea what the change was?

bmcfee:
Ah, I see in the source.  They renamed a parameter from `cachedir` to `location`.  This is a relatively easy fix if we bump up the minimum version requirement.  I can try to do this over the weekend and push out a quick point release.

Long term, we might want to rethink how the whole caching system works, or try to push the caching level functionality upstream to joblib.

[...]

PR #730, opened 2018-06-22
Don't use the deprecated cachedir property of joblib Memory

mvollrath:
#### Reference Issue
#729 

The `Memory.cachedir` deprecation warning in joblib 0.12 is broken causing a `TypeError` when importing librosa.  Using the `location` attribute gets the same results.

#### Any other comments?
We still use `cachedir` in the constructor, but this is not broken yet.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/730)

### DIFF
--- a/librosa/cache.py
+++ b/librosa/cache.py
@@ -46,7 +46,7 @@ class CacheManager(Memory):
                     func, 'return decorated(%(signature)s)',
                     dict(decorated=dec(func)), __wrapped__=func)
 
-            if self.cachedir is not None and self.level >= level:
+            if self.location is not None and self.level >= level:
                 return decorator_apply(self.cache, function)
 
             else:
--- a/librosa/cache.py
+++ b/librosa/cache.py
@@ -15,8 +15,8 @@ class CacheManager(Memory):
     field, thereby allowing librosa.cache to act as a decorator function.
     '''
 
-    def __init__(self, cachedir, level=10, **kwargs):
-        super(CacheManager, self).__init__(cachedir, **kwargs)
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks for taking the lead on this!  I expect that CI testing will not work right away due to build caches and inherent latency of deploying the new version of joblib through pypi and conda.

bmcfee:
:lgtm:

---

Reviewed 2 of 2 files at r2.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/librosa/librosa/730#-:-LFrwTxR8SdIGr614csO:bnfp4nl)*

bmcfee:
I think this is ready to go.  Thanks for fixing this!

mariusnita:
Btw is there a reason for using imprecise operators for dependency versions? For example, librosa 0.6.0 broke when joblib released 0.12.0. Had it been using `==` with a known good version, this couldn't have happened.

bmcfee:
> Btw is there a reason for using imprecise operators for dependency versions?

We generally don't expect breaking changes to happen without warning.  Usually, a breaking change would be known well in advance and we could plan around it without having tight coupling of releases.



> Had it been using == with a known good version, this couldn't have happened.

This is true, but it requires synchronizing releases to make sure that the entire ecosystem is internally consistent.  Generally I prefer to weakly assume semantic versioning, and hope for the best.  In almost 6 years, this is really the first time this has happened (notwithstanding one other isolated incident).

mariusnita:
Makes sense. Thanks!

PR #734, opened 2018-06-27
More detailed docstring for spectral_rolloff

### LINKED ISSUE #720: Get mean, maximum and minimum frequency
Hi
How can I get mean, maximum and minimum frequency of a .WAV file?

ZiyaoWei:
#### Reference Issue

Fixes #720 

This adds a more detailed explanation for the `spectral_rolloff` function, including what the function does and an example of potential usage. It might be worded too abstrusely.

#### Any other comments?

I said "this won't break the build" as a joke since it's a docs only change :-) But come to think of it maybe I should add a doctest for low rolloff percentage, to illustrate the example usage and risk breaking the build?

Also - not sure what's the convention in the scientific Python ecosystem but I am somewhat surprised by `rolloff_percentage` being [0, 1] instead of [0, 100].

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/734)

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -401,7 +401,13 @@ def spectral_contrast(y=None, sr=22050, S=None, n_fft=2048, hop_length=512,
 
 def spectral_rolloff(y=None, sr=22050, S=None, n_fft=2048, hop_length=512,
                      freq=None, roll_percent=0.85):
-    '''Compute roll-off frequency
+    '''Compute roll-off frequency.
+
+    The roll-off frequency is defined for each frame as the center frequency
+    for a spectrogram bin such that at least roll_percent (0.85 by default)
+    of the energy of the spectrum in this frame is contained in this bin and
+    the bins below. This can be used to, e.g., approximate the maximum (or
+    minimum) frequency by setting roll_percent to a value close to 1 (or 0).
 
     Parameters
     ----------
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -447,9 +447,16 @@ def spectral_rolloff(y=None, sr=22050, S=None, n_fft=2048, hop_length=512,
     From time-series input
 
     >>> y, sr = librosa.load(librosa.util.example_audio_file())
+    >>> # Approximate maximum frequencies with roll_percent=0.85 (default)
     >>> rolloff = librosa.feature.spectral_rolloff(y=y, sr=sr)
     >>> rolloff
     array([[ 8376.416,   968.994, ...,  8925.513,  9108.545]])
+    >>> # Approximate minimum frequencies with roll_percent=0.1
+    >>> rolloff = librosa.feature.spectral_rolloff(y=y, sr=sr, roll_percent=0.1)
+    >>> rolloff
+    array([[ 75.36621094,  64.59960938,  64.59960938, ...,  75.36621094,
+         75.36621094,  64.59960938]])
+
 
     From spectrogram input
 


bmcfee:
This looks great, thanks for doing it!

> But come to think of it maybe I should add a doctest for low rolloff percentage, to illustrate the example usage and risk breaking the build?

We don't actually have doctests as part of CI, since it takes a considerable amount of time to execute all of the examples in the documentation.  However, if you wanted to add an example following the standard formatting, that would be great.  For this, you can bypass the unit tests by adding ``[ci skip]]`` to the end of the commit message.

bmcfee:
Reviewed 1 of 1 files at r1.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/librosa/librosa/734)*

ZiyaoWei:
I'm trying to add an example of doing spectral_rolloff with a 0.1 roll_percent, which is straightforward enough.

However, it turns out I can't replicate most (any?) of the results in the docstrings in spectral.py, and the errors don't seem to be floating point errors (which happens for tests in, e.g., utils.py) - they are off by a large margin. E.g., the docs say

    >>> y, sr = librosa.load(librosa.util.example_audio_file())
    >>> rolloff = librosa.feature.spectral_rolloff(y=y, sr=sr)
    >>> rolloff
    array([[ 8376.416,   968.994, ...,  8925.513,  9108.545]])

but the result I got is:

    >>> y, sr = librosa.load(librosa.util.example_audio_file())
    >>> rolloff = librosa.feature.spectral_rolloff(y=y, sr=sr)
    >>> rolloff
    array([[7019.82421875,  968.99414062,  968.99414062, ..., 4661.93847656,
            4478.90625   , 4435.83984375]])

(note the second elements coincide but that looks more like a happenstance)

I also checked out c651d2be2376f4a41c49106c8ebe1914ee585649 which is when the docstrings were last updated, but I can't repro the results either. I'm assuming it's due to dependency versions:

    >>> import librosa
    /Users/wzy/Research/librosa/librosa/core/audio.py:34: UserWarning: Could not import scikits.samplerate. Falling back to scipy.signal
      'Falling back to scipy.signal')
    >>> y, sr = librosa.load(librosa.util.example_audio_file())
    >>> rolloff = librosa.feature.spectral_rolloff(y=y, sr=sr)
    >>> rolloff
    array([[ 8214.91699219,   968.99414062,   979.76074219, ...,
             5092.60253906,  5383.30078125, 11025.        ]])

Not sure what's the best thing to do here - I can see several options:

1) don't add the new example;
2) only add a new example;
3) 2) + update the examples for spectral_rolloff;
4) 3) + completely update all examples in spectral.py

which are ordered by the trust we put in to my dev environment, or

5) try to figure out how to replicate the docstring behavior first and dig to see if there are any bugs/environment issues that's causing this.

I guess my environment is at least one possible system configuration, so ignoring it won't be right, but 5) might not be doable or become a timesink. I guess I could do some option from 1-4 and file a different issue for 5?

bmcfee:
I suspect the docstring examples are just out of sync with changes elsewhere in the library, eg, reflection-padding in stft, or possibly changes to sample rate conversion.  The value in the second column is too close to be chance IMO -- I bet if more than two values were printed, it would line up fine aside from edge effects.

At this point, I'm not *too* concerned with getting numerical precision in the docstring examples.  It'd be great if we could, but it's not a high priority.  This is why I generally prefer plot examples to python console outputs, since the former are easier to not take so literally.



> 2 + update the examples for spectral_rolloff;

I would opt for either adding the new example and leaving the old ones as is, or confining all updates to just this one method for now.  For sure don't do more than that, as the general problem is much larger in scope than this one docstring.

bmcfee:
LGTM, thanks!  Is there anything else for this one, or can we merge?

bmcfee:
Merged. Thanks again!

PR #750, opened 2018-07-23
Add chirp signal generator function

### LINKED ISSUE #736: Signal generators: tone and chirp
#### Description
It's often useful to be able to synthesize a pure sine tone, or either a linear or exponential sine sweep.  It's not hard to implement these by hand, but it would be nice to have them as primitives in the library, much like we currently have ``librosa.click``.

For signatures, I'm thinking:

```python
tone(frequency, sr, length=None, duration=None)
```
- The first two are obvious (Hz)
- `length` would be in samples, as in `istft`, `click`, or the effects module.
- `duration` would allow the user to specify the duration in seconds rather than samples, though `length` would take priority.  At least one of `lenth` or `duration` must be provided.

```python
chrip(fmin, fmax, sr, length=None, duration=None, linear=False)
```
- The first 5 are as above
- `linear` is a boolean that toggles between linear and exponential sweep.

ZiyaoWei:
#### Reference Issue
Partially fixes #736

This implements the chirp signal generator as specified in #736. I'll either update this with the tone generator or add that function in a separate pull request.

#### Any other comments?
It takes me an embarrassingly long time to get around to do this :-/
Turns out `scipy` already has a `chirp` function, but it does have a different signature and has a couple small caveats (e.g., by default it doesn't start at 0 which would cause a "pop" sound in the generated .wav file, at least on the machine I am using). This function is just a thin wrapper on top of it - let me know whether this makes sense!

I also see a bunch of warnings from `pytest` that generator tests are going to be deprecated - it seems [parameterize](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize) is the replacement - should I use that instead or is `yield` tests still preferred?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/750)

### DIFF
--- a/librosa/core/__init__.py
+++ b/librosa/core/__init__.py
@@ -17,6 +17,7 @@ Audio processing
     zero_crossings
     clicks
     tone
+    chirp
 
 Spectral representations
 ------------------------
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -17,7 +17,7 @@ from .. import util
 from ..util.exceptions import ParameterError
 
 __all__ = ['load', 'to_mono', 'resample', 'get_duration',
-           'autocorrelate', 'zero_crossings', 'clicks', 'tone']
+           'autocorrelate', 'zero_crossings', 'clicks', 'tone', 'chirp']
 
 # Resampling bandwidths as percentage of Nyquist
 BW_BEST = resampy.filters.get_filter('kaiser_best')[2]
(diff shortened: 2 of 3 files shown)


bmcfee:
Thanks for doing this!

I'm currently on vacation, and can't do a review. Anyone else feel like reviewing this one?

Side note: we still use nosetest, not pytest, so don't worry about the warnings for now. There's an issue to convert the test suite away from yield tests to proper fixtures, but it's a pretty big job.

stefan-balke:
Sure.

First question: Do we want it to be a wrapper for `scipy.signal.chirp`?
If so, we should also add the offered quadratic and hyperbolic chirps.

ZiyaoWei:
FWIW, the reason why I did this are:

1) librosa already has a scipy dependency, and the `chirp` function looks almost exactly what's needed and its implementation makes sense (of course);
2) however, it has a different signature and isn't quite as convenient as the one proposed, so maybe it's worth it to write a wrapper around it.

Let me know whether those make sense - I'm fine with redo this or even just throw it away in favor of the scipy function :-)

re: additional choices of chirps - I just followed the spec and (personally) assumed that exponential (logarithmic in the scipy terminology) is the most useful for music processing, linear is somewhat useful while others are corner cases enough that people should just use the scipy function instead - but I am a complete n00b here so definitely correct those if I'm wrong!

ZiyaoWei:
Also I plan to write the pure tone generator using `numpy.sin` along the same vein - hopefully that's a sane choice.

ZiyaoWei:
The CI failure is caused by me overlooking Python 2 integer division behavior - fixing.

stefan-balke:
> Also I plan to write the pure tone generator using numpy.sin along the same line - hopefully that's a sane choice.

Well, yes, you don't need to have your own sine table lookup :)

The question related to using `scipy` or not is: Should this be a util for people in need of a sweep/chirp, then that's okay. Should it be educational? Well, then we should do it from scratch.

bmcfee:
> The question related to using scipy or not is: Should this be a util for people in need of a sweep/chirp, then that's okay. Should it be educational? Well, then we should do it from scratch.

That's a pretty reasonable question.  I just took a look through scipy's chirp, and it's... not easy to unpack.  By contrast, here's a custom hack that I had kicking around a notebook:
```python
def make_signal(sr, duration, fmin='C1', fmax='C8'):
    if fmin is None:
        fmin = 0.01
    else:
        fmin = librosa.note_to_hz(fmin) / sr

    if fmax is None:
        fmax = 0.5
    else:
        fmax = librosa.note_to_hz(fmax) / sr

    return np.sin(np.cumsum(2 * np.pi * np.logspace(np.log10(fmin),
                                                    np.log10(fmax),
                                                    num=int(duration * sr))))
```
which is pretty self-contained, but obviously does less than scipy's.

That said, I'm fine with wrapping scipy's as is done here.  While a simple reference implementation would be nice, I put more weight on reusing existing (well-tested and maintained) code.

ZiyaoWei:
The ssh config is messed up on my machine and will update later.

ZiyaoWei:
Updated the pull request but messed up the commit history (i.e., I folded the new commit into the existing pull request). Changes are:

1) addressed the comments;
2) added a new `phi` argument a la #753;
3) some formatting to avoid really long lines.

bmcfee:
Thanks again @ZiyaoWei, this looks good to me!  Is there anything else to be done here, or shall I merge?

ZiyaoWei:
Thanks for the review @bmcfee - feel free to merge if you think it looks good!

bmcfee:
Merged.  Thanks @ZiyaoWei! :shipit:

PR #753, opened 2018-07-27
Add pure sine tone generator

### LINKED ISSUE #736: Signal generators: tone and chirp

ZiyaoWei:
#### Reference Issue
Partially fixes #736
This implements the pure sine tone signal generator as specified in #736.
#### Any other comments?
None

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/753)

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -17,7 +17,7 @@ from .. import util
 from ..util.exceptions import ParameterError
 
 __all__ = ['load', 'to_mono', 'resample', 'get_duration',
-           'autocorrelate', 'zero_crossings', 'clicks']
+           'autocorrelate', 'zero_crossings', 'clicks', 'tone']
 
 # Resampling bandwidths as percentage of Nyquist
 BW_BEST = resampy.filters.get_filter('kaiser_best')[2]
@@ -332,8 +332,8 @@ def get_duration(y=None, sr=22050, S=None, n_fft=2048, hop_length=512,
     S : np.ndarray [shape=(d, t)] or None
         STFT matrix, or any STFT-derived matrix (e.g., chromagram
         or mel spectrogram).
-        Durations calculated from spectrogram inputs are only accurate 
-        up to the frame resolution. If high precision is required, 
+        Durations calculated from spectrogram inputs are only accurate
+        up to the frame resolution. If high precision is required,
         it is better to use the audio time series directly.
 
     n_fft       : int > 0 [scalar]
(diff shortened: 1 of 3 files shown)


lostanlen:
Hello @ZiyaoWei, thank you for doing this. It would be good if you could add a `phase` keyword argument, defaulting to zero.

lostanlen:
Also, I would recommend that the sine wave uses `np.cos` rather than `np.sin`. That way, the default case of `phase=0` will make the wave symmetric.

ZiyaoWei:
@lostanlen - thanks for the suggestion - I'll definitely add the `phi` parameter (it should be using degrees and not radians I assume?)! For my edification - why is `cos` + 90 degrees preferred to `-sin`?

lostanlen:
Yes, `phi` is a better name for the keyword that what I proposed earlier (`phase`), given that it matches `scipy.signal.chirp`.

I would prefer `cos` over `sin` because `cos(omega t)` is the real part of `exp(i omega t)`. That way, `phi` is equal the Hilbert analytic phase of the tone.

See Eq 4.47 of Mallat's Wavelet Tour: http://links.uwaterloo.ca/amath391w13docs/Mallat3.pdf
and this paper of Flandrin http://perso.ens-lyon.fr/patrick.flandrin/SPIE01_PF.pdf

ZiyaoWei:
@lostanlen - thanks for the explanation! That's quite a bit more technical than I thought :-)

ZiyaoWei:
(Updated - I amended the commit so not sure if GitHub will show it as updated or not)

bmcfee:
This looks great, thanks for doing it so quickly!  And thanks @lostanlen for the input.

bmcfee:
@ZiyaoWei Hey, @lostanlen and I discussed this offline, and think it would be better to use radians rather than degrees for `phi`, both here and in #750.  This disagrees with `scipy.signal`'s API, but elsewhere in librosa, phase is always in radians (or complex exponential) as produced by `np.angle`.

Would you mind making that change in both places?  To keep the definition clean, I'd suggest using `phi=None` as a default, and converting `None` to `-np.pi * 0.5` at runtime; this would keep us from having a long string of numbers in the function signature.  Also, if the convention here is `cos(x + phi)`, then I'd prefer a default `phi = - np.pi * 0.5` so that the signal starts with positive slope (equiv. to sine) rather than negative.  This is still consistent with @lostanlen's argument for `cos(x + phi)`, but still gives us the expected sine-like behavior by default.

Thanks!

ZiyaoWei:
@bmcfee Sure, will make that change today - I will also add the return value shape in the documentation too a la #750 , thanks for reviewing!

librosa/core/audio.py
```diff
@@ -712,3 +712,66 @@ def clicks(times=None, frames=None, sr=22050, hop_length=512,
             click_signal[start:end] += click
 
     return click_signal
+
+
+def tone(frequency, sr=22050, length=None, duration=None, phi=90):
```
bmcfee: (Redundant note here, just for maintaining GH status)

1. phi should be in radians, not degrees
2. default should be `None`
3. `phi=None` should map to `-np.pi/2` (not `+np.pi/2`)

ZiyaoWei:
Didn't quite keep the promise - only got around to it now :-/ Will update `chirp` shortly too.

bmcfee:
Thanks!  I think this one's good to merge.

I suspect this will cause a merge conflict on #750, so maybe wait a minute and rebase master after I merge this?

lostanlen:
Congratulations on your merged PR @ZiyaoWei! I'm sorry that I gave you the runaround with that `phi` kwarg. This looks great now.
We'll make sure to mention the addition of `librosa.tone` as a new feature to v0.6.2 in the release notes, to give it some publicity.

PR #755, opened 2018-08-03
Fixed #754 to avoid futurewarnings in index sequences

### LINKED ISSUE #754: Tuple indexing (numpy 1.15)
#### Description

As of the latest numpy release, we're getting the following messages in the build log:
```
test_beat.test_onset_strength('data/beat-onset-000.mat',) ... /home/travis/build/librosa/librosa/librosa/util/utils.py:1377: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  data_agg[idx_agg] = aggregate(data[idx_in], axis=axis)
```
Mostly coming up whenever `util.sync` is called.

This is relatively easy to fix, and we should do it before the 0.6.2 release.

bmcfee:
#### Reference Issue
Fixes #754 

Numpy 1.15 added `FutureWarning` for using lists as ndarray indexes.  This PR tuple-izes all of the offending cases to avoid the warnings going forward.

#### Any other comments?

I don't anticipate any problems with this change.  As long as the tests pass and the warnings are gone, I think we can merge without CR.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/755)

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -447,7 +447,7 @@ def autocorrelate(y, max_size=None, axis=-1):
     subslice = [slice(None)] * autocorr.ndim
     subslice[axis] = slice(max_size)
 
-    autocorr = autocorr[subslice]
+    autocorr = autocorr[tuple(subslice)]
 
     if not np.iscomplexobj(y):
         autocorr = autocorr.real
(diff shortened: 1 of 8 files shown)


bmcfee:
Note: one of the warning sources is upstream in scipy.fftpack, so we can't get them all right now.

bmcfee:
Log is clean now -- all remaining futurewarnings are upstream.

PR #756, opened 2018-08-05
Conda 4.3 requires noarch/repodata.json and noarch/repodata.json.bz2

vbreeden:
From Conda: 'As of conda 4.3, a valid channel must contain a `noarch/repodata.json` and associated `noarch/repodata.json.bz2` file, even if `noarch/repodata.json` is empty.'

#### Any other comments?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/756)


bmcfee:
Thanks for this.  I'm a little confused though -- shouldn't this PR be made against the [feedstock](https://github.com/librosa/librosa-feedstock) and not the main repo?

vbreeden:
I suspect you are correct. This is my first go at contributing to a community project so there are still things I'm learning. I'll reproduce the error this evening to double check and then create a PR against the feedstock.

vbreeden:
I did some more research into this. It turns out that my condarc was pointing directly to https://github.com/librosa/librosa instead of conda-forge. With the proper channel in place the problem goes away so this pull request can be rejected and closed.

bmcfee:
Ok, thanks for looking into it!

PR #760, opened 2018-08-16
adding specific units to the time axis for display

### LINKED ISSUE #737: Display axis decorators: seconds, ms
#### Description

We currently support two temporal decoration modes: `time` and `lag`.  These modes use seconds as the underlying data representation, and the `TimeFormatter` object adaptively scales the units to an appropriate resolution given the display range.

It would be good to be able to override this, and display in fixed units.  This could be easily done by adding a couple of new modes:

- `seconds`: always use seconds
- `ms`: always use milliseconds
- `lag_seconds`: always use seconds (including negative for lag)
- `lag_ms`: as above, but for milliseconds.

Implementing this is simply a matter of adding some control logic to the `TimeFormatter`, and adjusting the automatic axis labels to `Time (s)` or `Time (ms)` accordingly.

kyungyunlee:
#### Reference Issue

Fixes #737 

Currently, the time axis unit is automatically set to either seconds, minutes or hours, depending on the given time range from the user. 
In this fix, users can specify units to either `'s' (seconds)`, `'ms' (milliseconds)` or default to `None`, in which the default behaves the same as the original implementation. 

#### Any other comments?
I was thinking of adding tests in the `tests` directory, but wasn't quite sure how to go about it. I am guessing I should be adding some images to the `baseline_images` folder? 

Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/760)


bmcfee:
Thanks for getting this going so quickly!

Re: tests, it's a little backwards to get set up. If you take a look at the display test module, you'll see a bunch of image comparison decorator tests. The tests render the figure, and then compare it (by pixel rmse) to a reference image.

To create the reference image for a new test, run the tests once to create the output, check that the output looks as expected, and then copy it over to the baseline images directory.

Once the tests are implemented and working, we can start on code review.

kyungyunlee:
I added 2 tests: 1 parameter checking test for `TimeFormatter` and 1 spectrogram test checking the correctness for displaying x axis in `s` and `ms`. 
Parameter checking for `specshow` already exists, so I didn't add that one. 
Let me know if I need to add more tests :)

bmcfee:
Thanks for adding the tests.  It looks like there is still a bunch of new code that is not covered, see the [coveralls status](https://coveralls.io/builds/18550865) for a line-by-line breakdown for where to add test coverage (lines in red).

As a side note, in `waveplot`, it looks like you've added logic to change the data axes to milliseconds. I thought we had discussed in #737 to keep the data units always in seconds, and only change the display axes?  This is important for getting shared axes across subplots to behave as expected.

kyungyunlee:
> As a side note, in waveplot, it looks like you've added logic to change the data axes to milliseconds. I thought we had discussed in #737 to keep the data units always in seconds, and only change the display axes? This is important for getting shared axes across subplots to behave as expected.

Oh, I thought we were allowing the time unit to be specified for `waveplot` as well, options being `{'time', 's', 'ms', 'off', 'none'}. When I meant to "keep the unit always in seconds", I was referring to keeping seconds as the minimum unit even when the data range is very small (less than 1 seconds). 

Wow I totally was not understanding how the coverage stuff works..now I get it..I will add more tests!

bmcfee:
> Oh, I thought we were allowing the time unit to be specified for waveplot as well, options being `{'time', 's', 'ms', 'off', 'none'}. When I meant to "keep the unit always in seconds", I was referring to keeping seconds as the minimum unit even when the data range is very small (less than 1 seconds).

Ah, I see.  To clarify my point: it's all around simpler if the data units are always kept fixed to seconds, and only the axis decorations change how that's formatted.  This is how specshow works more generally, eg, for plotting frequency axes with hz/note decorations, or non-uniformly sampled time axes (eg beat-synchronous observations).  Once the data units are fixed, it makes it possible to do things like:
```python
>>> ax = plt.subplot(2,1,1)
>>> librosa.display.waveplot(y, sr)
>>> plt.subplot(2,1,2, sharex=ax)
>>> librosa.display.specshow(S, x_axis='time')
>>> plt.xlim([5, 10])
```
and have both subplots stay aligned, even after changing the view (`xlim` to 5-10 seconds).

If the data units are not fixed, then this becomes much more complex to deal with, so I'd prefer to avoid it unless we have a strong reason otherwise.

> Wow I totally was not understanding how the coverage stuff works..now I get it..I will add more tests!

:+1:!

kyungyunlee:
Yeah it totally makes sense. My bad for not considering as far as the shared axis. 

I tried something like this, where the x_axis is specified to `ms` and shares the axis with the `waveplot`. In this case, I guess I would expect the axis to be shown in `ms`.  Instead it displays the time altogether in `seconds` due to the `waveplot` but the spectrogram is actually is plotted between 0 - 4.5ms. 

```
ax = plt.subplot(2, 1, 1)
librosa.display.specshow(np.log(S), x_axis='ms')
plt.subplot(2, 1, 2, sharex=ax)
librosa.display.waveplot(y, sr)
```
test

Simiilar problem when waveplot is plotted first. Now, the waveplot is displayed between 0 - 4500seconds.
```
ax = plt.subplot(2,1,1)
librosa.display.waveplot(y, sr)
plt.subplot(2,1,2, sharex=ax)
librosa.display.specshow(np.log(S), x_axis='ms')
```
test2

Again, I am not sure if it is really necessary to consider all these possibilities (although it does appear to be a possible attempt by users).

bmcfee:
> Again, I am not sure if it is really necessary to consider all these possibilities (although it does appear to be a possible attempt by users).

This definitely smells like a bug to me: probably the axis decorations will have to be the same (since the axes are shared), but the data limits are definitely wrong.  Maybe it would make sense to include these two examples as test cases.

kyungyunlee:
I am struggling to find a way to test this bug...
I was planning to add an assertion to compare the data points (x and y or an array) from the correct time plot with the buggy time plot, but the way to access data points when two plots are sharing axis is hard to find. 
I tried `get_path()` option for waveplot function which returns `PolygonCollection`, but this doesn't take into account the shared axis. 

Any ideas on implementing the test?

bmcfee:
> Any ideas on implementing the test?

Image comparison tests are tricky to write because they're inherently regression tests, so you have to have a correct implementation to start with.

In this case, I suppose you could fake it by rendering the figure without shared axes (giving the correct-looking output), saving the output as the target/baseline image, and then revising the test code to use shared axes.

kyungyunlee:
Added 2 tests, and made a short note about the bugs in `test_display.py`. 
Additional thing that should be fixed is, from looking at matplotlib demos, when sharing axis, there is only one "shared" x label on the bottom plot. 
Here when I tried plotting, I see x labels on both plots. 
(top figure: current output, bottom figure:correct output)
sharex_specshow_ms
sharex_specshow_ms-expected


So, overall 2 issues here.
1. wrong time range for upper plots
2. x_label appearing on both plots when `sharex`  is True. 

Thanks!

bmcfee:
I'm not 100% sure, but I think the x-label replication is expected behavior.

librosa/display.py
```diff
@@ -917,5 +969,10 @@ def __coord_n(n, **_kwargs):
...
 
 def __coord_time(n, sr=22050, hop_length=512, **_kwargs):
-    '''Get time coordinates from frames'''
+    '''Get time coordinates in seconds from frames'''
     return core.frames_to_time(np.arange(n+1), sr=sr, hop_length=hop_length)
+
+
+def __coord_milliseconds(n, sr=22050, hop_length=512, **_kwargs):
+    '''Get time coordinates in milliseconds from frames'''
+    return 1000 * __coord_time(n, sr, hop_length, **_kwargs)
```
bmcfee: I think this is the problem with your mis-aligned plots.

All time-like plots should use `__coord_time` and have positions in seconds; `__coord_milliseconds` should not exist.  All that we need to do is make sure that `__decorate_axis` sets the timeformatter correctly, which it seems to.

kyungyunlee: Ah..I see! 
But, when we allow unit specification to `ms` in specshow(), doesn't this mean that we need unit conversion from seconds to milliseconds? Currently, this is not handled in `__decorate_axis`.

bmcfee: Yes, but that should only be done within the TimeFormatter (and when specifying the axis label).  The coordinate grid should still be in seconds.

kyungyunlee: Oops forgot about not allowing time unit specification in waveplot.
Adding unit conversion in timeformatter works and fixes the bug :)

One last thing though.. the axis label will conflict with the actual plotted unit, since currently the axis labeling is hard coded within `__decorate_axis`, it does not take into account the `sharex` situation. Hmm...

kyungyunlee:
I put time unit related functions inside TimeFormatter and things are working now! Except the axis labeling I mentioned above in the last comment... 
I just added test images as shown above for now (with `Time (ms)`).
I think this might be the last thing to be fixed!

lostanlen:
Hi @kyungyunlee, thanks for the edits! I'm requesting minor changes to the documentation.
The most important one is at the top of the `TimeFormatter` docstring. It is important to document that `unit` defaults to `None`, and that `None` is an admissible kwarg value, not just `s` or `ms`.

I also noticed some hyphenation mistakes in the docstring that were there before you started editing (in words such as "midpoint" and "halfway"), so I'm suggesting to take advantage of this PR to fix them too.

kyungyunlee:
I checked and fixed all the comments. 

> I also noticed some hyphenation mistakes in the docstring 

Regarding mistakes, I also found another spelling error in `test_display.py`, so I fixed it as well. 
It was saying : 
```
def test_waveplot_bad_maxploints(): 
    ....
```
Now changed to `maxpoints`.

bmcfee:
This all looks great, thanks @kyungyunlee (and @lostanlen for the CR)!

The only nitpick that I have is that the new `units` parameter to `TimeFormatter` was placed before an existing parameter.  This will break API compatibility if, for some reason, anyone was using `TimeFormatter` with positional arguments.  In general, when adding new parameters, it's best to put them at the end unless there's a strong argument otherwise.  Aside from that, this looks ready to :shipit: !

librosa/display.py
```diff
@@ -818,17 +852,41 @@ def __decorate_axis(axis, ax_type):
         axis.set_label_text('BPM')
 
     elif ax_type == 'time':
-        axis.set_major_formatter(TimeFormatter(lag=False))
+        axis.set_major_formatter(TimeFormatter(unit=None, lag=False))
```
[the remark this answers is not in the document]
bmcfee: This isn't actually necessary; since it's calling with explicit keywords, order does not matter.  I'd say it's fine as is.

bmcfee:
Thanks @kyungyunlee for the quick fix!  I think this is :shipit: .

librosa/display.py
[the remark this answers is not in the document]
lostanlen: yes, it does not change the behavior of the program, but I figured it might be good to "set the example" right in the code base. but if you say that that's an overcorrection, let's forget about it

bmcfee:
Done and done!  Thanks again @kyungyunlee and @lostanlen !

Issue #391, opened 2016-07-22
Rewrite tests in py.test


bmcfee:
This will be a huge job, but our test suite is currently the stuff of nightmares.

stefan-balke:
Puh, that's gonna be a lot of work. Should it cover #319? Would make sense, wouldn't it?

bmcfee:
Yup, loads of work!

I think it makes more sense to port the tests as they currently are, and then worry about rewriting them ala #319.

stefan-balke:
Quick thought on this:
Can we run `nosetests` and `py.test` in parallel? At least the new tests could be directly written in `py.test` and I think there are some tests where the porting will be very easy.

bmcfee:
pytest has some trouble with yield-tests (see [here](https://docs.pytest.org/en/latest/nose.html)), but it could work.  There's no harm in trying, anyway.

bmcfee:
This issue is about to become more pressing, as matplotlib 3.0 is dropping the (deprecated) image comparison test functionality that we use.

For now, we can pin the test dependencies to matplotlib >=1.5, <3.0, but we'll need to fix it long term.

PR #775, opened 2018-11-01
fix failing test CI's / migrate to pytest

bmcfee:
Subsequent fixing on #754 

This PR fixes some FutureWarnings introduced in numpy 1.15 when fancy-indexing arrays with non-tuple sequences.

This is primarily a problem because Travis builds are failing due to log exhaustion.  It turns out that most of the futurewarnings we're seeing are not due to librosa code, but dependencies on bits of `scipy` that, as of 1.1.0, are not yet future-proof.  We can't fix those from here, but there were a few last bits that hadn't been fixed in the previous pr #755.

#### Any other comments?

How do we want to handle upstream warnings?  I don't want to temporarily disable `FutureWarnings` or log output in general, because those are still extremely valuable.  On the other hand, I'd like to avoid choking the log with warnings generated upstream from our code.  Does anyone know if there's a way to set warnings to suppress warnings generated outside of a particular package?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/775)


bmcfee:
This fixed the warning log choking, but it is now failing because matplotlib 3 removed the nose classes submodule that we use for display regression tests.

bmcfee:
At this point, I think the best path forward on this is to start with the much-dreaded #391 : migrating from nosetests to pytest.

There are a few hurdles here:

1. our tests are almost entirely `yield` tests, which pytest deprecated in 3 and will remove in 4.  This is okay for now, but it will be a major undertaking to fix properly.
2. Many of the exception-catching bits will need to be rewritten to use `pytest` mark rather than `nose` raises.  My current count has 168 uses of `raises` in the test suite: plenty, but doable.
3. We'll need to redo the display module to remove the reliance on the noseclasses submodule.  I think this won't be too big of a deal, as the pytest wrapper looks pretty similar.

bmcfee:
A preliminary cut at porting to pytest is now in place.  It mostly works, with the following caveats:

1. There are numerous failures due to quirks in how pytest implements yield tests.  Tests which pass in nose are failing in pytest, I think because it's not catching the failure marker in certain cases.  Possibly switching it from `noses.raises(...)` to `pytest.mark.xfail(raises=...)` will fix this, but it's not 100% for certain.

2. We're still using yield tests all over the place.  This is bad, and breaks things in the future, but overhauling all of the tests is a massive job.  I rewrote the tests in the display module to use proper fixtures, so we have some basis to move forward.

3. We're still using a bunch of nose internals `eq_`, `raises`, etc.  These can be pulled out with a little bit of work.

bmcfee:
Coverage is dropped right now because I added a few test skips.  These tests are correct in the sense that they work when executed in nose, but something is amiss when run in pytest.  Probably the simplest thing is to refactor them into simpler tests.

bmcfee:
Remaining tests to fix:
- `normalize`
- `match_events_onesided`

These ones have been tricky to refactor because they generate and modify fixtures on the fly, or otherwise have weird scoping behavior.

As a fun side note, I found a few minor variable scope bugs in some of the tests that I did convert to use proper fixtures.  Nothing major that impacts correctness of the tests in these specific cases, but the code was certainly "wrong".  In the long run, implementing #391 ought to pull out more such cases.

bmcfee:
Test coverage is down a little, primarily in the event matching code.  I think this is mainly due to unorthodox use of generator tests, and I'll defer fixing it for real until #783 .  This is otherwise good to go.

PR #780, opened 2018-11-10
change 'seconds' to 'microseconds'

yingshaoxo:
It's different in mathematics.

### DIFF
--- a/librosa/core/time_frequency.py
+++ b/librosa/core/time_frequency.py
@@ -119,7 +119,7 @@ def samples_to_frames(samples, hop_length=512, n_fft=None):
 
 
 def frames_to_time(frames, sr=22050, hop_length=512, n_fft=None):
-    """Converts frame counts to time (seconds).
+    """Converts frame counts to time (microseconds).
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


lostanlen:
I don't understand what this PR does. In signal processing, the sample rate is measured in Hertz, not MegaHertz, so the time values are in seconds, not microseconds.
Could you please explain?

lostanlen:
I need a minimal working example of what you describe before I can reply. What signal are you using? What is your sampling rate? Can you please post your code?

yingshaoxo:
https://github.com/yingshaoxo/auto_everything/blob/master/auto_everything/video/__video.py#L182

It may difficult to understand.

Here is the place I divided by 1000: https://github.com/yingshaoxo/auto_everything/blob/master/auto_everything/video/__video.py#L124





If you feel bad about these codes.

I can write you a simple version of codes instead, thanks.

bmcfee:
@yingshaoxo as @lostanlen says, the librosa code is correct here.  In librosa, sample rates are always interpreted in Hz and times are always interpreted in seconds.

lostanlen:
@yingshaoxo I looked at your code on 
https://github.com/yingshaoxo/auto_everything/blob/master/auto_everything/video/__video.py#L182
you shouldn't be passing `parts` (output of `librosa.effects.split`) to `frames_to_time`.
`parts` contains the amplitude of signal through time, whereas the expected input of `frames_to_time` is an array of time samples, i.e. a range

PR #785, opened 2018-11-14
Remove asserts

### LINKED ISSUE #784: Remove assertions from library
#### Description
@lostanlen noticed recently that there are a few spots in the code with `assert` statements that should really be exceptions.  This is bad practice because it makes downstream error handling difficult and non-diagnostic.  Replacing these assertions with exceptions ought to be easy.

Offending lines:

https://github.com/librosa/librosa/blob/c9c1b6d74a06afec9aad07425b6e59fd39bdb364/librosa/core/audio.py#L367

https://github.com/librosa/librosa/blob/27174ba111faa859f67cbb22f0ec3bc646196da6/librosa/core/spectrum.py#L1265

#### Versions
0.6.2 and below.

bmcfee:
#### Reference Issue
Fixes #784 

This PR replaces runtime assertions by proper exceptions.  No other change to functionality.

#### Any other comments?

The exception in the fast mellin transform code should, I think, never occur in practice.  It's difficult to construct a test case to trigger this, so we'll have a small dip in test coverage for now.

In the future, it might be nice to refactor FMT a bit to make it easier to understand what's going on, but I'll leave that for future work.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/785)

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -357,6 +357,11 @@ def get_duration(y=None, sr=22050, S=None, n_fft=2048, hop_length=512,
     -------
     d : float >= 0
         Duration (in seconds) of the input time series or spectrogram.
+
+    Raises
+    ------
+    ParameterError
+        if none of `y`, `S`, or `filename` are provided.
     """
 
     if filename is not None:
@@ -364,7 +369,8 @@ def get_duration(y=None, sr=22050, S=None, n_fft=2048, hop_length=512,
             return fdesc.duration
 
     if y is None:
-        assert S is not None
+        if S is None:
+            raise ParameterError('At least one of (y, sr), S, or filename must be provided')
 
         n_frames = S.shape[1]
         n_samples = n_fft + hop_length * (n_frames - 1)
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -362,6 +362,14 @@ def get_duration(y=None, sr=22050, S=None, n_fft=2048, hop_length=512,
     ------
     ParameterError
         if none of `y`, `S`, or `filename` are provided.
+
+    Notes
+    -----
+    `get_duration` can be applied to a file (`filename`), a spectrogram (`S`),
+    or audio buffer (`y, sr`).  Only one of these three options should be
+    provided.  If you do provide multiple options (e.g., `filename` and `S`),
+    then `filename` takes precedence over `S`, and `S` takes precedence over
+    `(y, sr)`.
     """
 
     if filename is not None:
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1264,7 +1264,7 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
     # Make sure that all sample points are unique
     # This should never happen!
     if len(np.unique(x_exp)) != len(x_exp):
-        raise RuntimeError('Redundant sample positions in FMT')
+        raise RuntimeError('Redundant sample positions in Mellin transform')
 
     # Resample the signal
     y_res = f_interp(x_exp)
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1262,7 +1262,9 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
         x_exp = np.clip(x_exp, float(t_min) / n, x[-1])
 
     # Make sure that all sample points are unique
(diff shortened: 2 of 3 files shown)


librosa/core/audio.py
```diff
@@ -357,14 +357,20 @@ def get_duration(y=None, sr=22050, S=None, n_fft=2048, hop_length=512,
...
     """
 
     if filename is not None:
         with audioread.audio_open(filename) as fdesc:
             return fdesc.duration
 
     if y is None:
-        assert S is not None
+        if S is None:
+            raise ParameterError('At least one of (y, sr), S, or filename must be provided')
```
lostanlen: I'm a bit circumspect with this `at least one of X, Y, Z` design
where `X` is `(y,sr)`, `Y` is `S` and `Z` is `filename`
why not make these three options mutually exclusive?
at the moment `Z` takes precedence over `Y`, which takes precedence over `X`
this is undocumented

librosa/core/spectrum.py
```diff
@@ -1262,7 +1262,9 @@ def fmt(y, t_min=0.5, n_fmt=None, kind='cubic', beta=0.5, over_sample=1, axis=-1
         x_exp = np.clip(x_exp, float(t_min) / n, x[-1])
 
     # Make sure that all sample points are unique
-    assert len(np.unique(x_exp)) == len(x_exp)
+    # This should never happen!
+    if len(np.unique(x_exp)) != len(x_exp):
+        raise RuntimeError('Redundant sample positions in FMT')
```
lostanlen: spell out "fast Mellin transform"
FMT is too uncommon of a TLA to be used

what does "this should never happen" mean here? is it on the responsability of the user or the programmer to have this not happen?

tests/test_core.py
```diff
@@ -576,6 +576,11 @@ def test_get_duration_filename():
     assert np.allclose(duration_fn, duration_y)
 
 
+@pytest.mark.xfail(raises=librosa.ParameterError)
+def test_get_duration_fail():
+    librosa.get_duration(y=None, S=None, filename=None)
```
lostanlen: LGTM

librosa/core/audio.py
bmcfee: Making them mutually exclusive would be more complicated to implement.  But the precedence should definitely be documented here.

librosa/core/spectrum.py
bmcfee: I would say it's on the programmer.

To reconstruct what happened in the development of this function, there was an issue that the interpolator would sometimes fail (throw an exception) if the interpolated sample positions landed out of bounds from the observed samples.  This, I think, was only occurring in strange corner cases due to rounding error.  We hack around this by clamping the interpolation points to lie within the sampled range, and then `unique`ing the points to make sure we don't have duplicates.  

This is an ugly hack, and it would be nice to correct it in the future, but I don't think we need to do that right here.

librosa/core/audio.py
lostanlen: making them mutually exclusive has a nice "single dispatch" flavor

i really don't imagine a use case for a callsite of the form
`get_duration(y=my_y, sr=my_sr, S=my_S, filename=my_filename)`
in which, some, but not all, of `my_y`, `my_S` and `my_filename`, would be `None` at runtime in user code.

but this is not a hill i'm willing to die upon. i'm fine with anything as long as it's properly documented

librosa/core/spectrum.py
lostanlen: ok. if it's on the programmer then i'm fine with a rather evasive error message like "redundant sample positions in fast mellin transform"
if a user ever encounters that error, that user will eventually raise an issue and we'll work from there

on the other hand, if the responsability was on the user, we would have needed a much more informative error message, with recommendations on over-sampling, etc.

bmcfee: Agreed.  I think this is how it wound up as an assert in the first place.

librosa/core/audio.py
bmcfee: > i really don't imagine a use case for a callsite of the form
> `get_duration(y=my_y, sr=my_sr, S=my_S, filename=my_filename)`
> in which, some, but not all, of `my_y`, `my_S` and `my_filename`, would be `None` at runtime in user code.

Multiple dispatch over `(y, sr)` and `S` is pervasive in librosa, for better or worse.  A user wouldn't necessarily call it as you've described, but for interdepdent functions (say, `mfcc` and `melspectrogram`), it's handy to have a nice interface for directly passing through the different forms of input.

`get_duration` started life in that form (for the sake of stylistic consistency), and the `filename` option was added later.

bmcfee: Already done! :tada:

PR #793, opened 2018-11-23
Remove confusing reference to "DPWE code".

dpwe:
Remove obsolete explanation for why the FFT output is being conjugated - it's not any more.

The .conj() on the output of fft.fft was removed in 0.6.0; this comment should have been deleted then.  Not sure what RFFT refers to, as the call seems to have always been fft.fft.

Obsolete comments?  In my code base?

Removes an outdated comment that could only confuse readers.

#### Any other comments?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/793)


bmcfee:
Thanks!

lostanlen:
Thank you so much DAn!

PR #799, opened 2018-11-28
Iir sos

fzalkow:
So far, `librosa.core.iirt` is doing iir filtering with [`scipy.signal.filtfilt`](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.signal.filtfilt.html) (using the standard difference equation). It is known that higher order filters can be instable due to numerical imprecisions of floating point numbers. A standard solution is to to use a series of second-order filters instead. Analytically, they are equivalent, but they are more robust against numerical imprecisions. Therefore, I propose to use [`scipy.signal.sosfiltfilt`](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.signal.sosfiltfilt.html).

A motivating example is shown in the documentation of [`scipy.signal.sosfilt`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.sosfilt.html). I came across this when reading the [Introduction to Digital Filters](https://ccrma.stanford.edu/~jos/filters/) of Julius O. Smith III.

What does this mean for `librosa.core.iirt`? For the standard settings, the improvements are insignificant. See the comparison of the magnitude responses, created with [this gist](https://gist.github.com/fzalkow/793dfeb0f6e7ce588889592ef286da46) (upper: magnitude response in db with old approach; middle: magnitude response in db with proposed approach; below: difference between the upper two):

plot1
 
However, if we change the standard settings, e.g. by decreasing the passband ripple from 1 to 0.2 dB and increasing the stopband attenuation from 50 to 100 dB, the instability of the previous approach becomes obvious:

plot2

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/799)

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -690,6 +690,7 @@ def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
           Can be unstable for high-order filters.
         - If `sos`, a series of second-order filters is used for filtering with `scipy.signal.sosfiltfilt`.
           Minimizes numerical precision errors for high-order filters, but is slower.
+        - Raises a `ParameterError` for any other string.
 
     kwargs : additional keyword arguments
         Additional arguments for `librosa.filters.semitone_filterbank()`
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -772,7 +772,7 @@ def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
         if flayout == 'ba':
             cur_filter_output = scipy.signal.filtfilt(cur_filter[0], cur_filter[1],
                                                       y_resampled[cur_sr_idx])
-        else:
+        elif flayout == 'sos':
             cur_filter_output = scipy.signal.sosfiltfilt(cur_filter,
                                                          y_resampled[cur_sr_idx])
 
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -634,7 +634,7 @@ def phase_vocoder(D, rate, hop_length=None):
 
 @cache(level=20)
 def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
-         tuning=0.0, pad_mode='reflect', flayout='ba', **kwargs):
+         tuning=0.0, pad_mode='reflect', flayout=None, **kwargs):
     r'''Time-frequency representation using IIR filters [1]_.
 
     This function will return a time-frequency representation
@@ -720,6 +720,11 @@ def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
     >>> plt.colorbar(format='%+2.0f dB')
     >>> plt.tight_layout()
     '''
+
+    if flayout is None:
+        warnings.warn('Default filter layout for `iirt` is `ba`, but will be `sos` in 0.7.',
+                      FutureWarning)
+        flayout = 'ba'
 
     # check audio input
     util.valid_audio(y)
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -634,7 +634,7 @@ def phase_vocoder(D, rate, hop_length=None):
 
 @cache(level=20)
 def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
-         tuning=0.0, pad_mode='reflect', **kwargs):
+         tuning=0.0, pad_mode='reflect', flayout='ba', **kwargs):
     r'''Time-frequency representation using IIR filters [1]_.
 
     This function will return a time-frequency representation
@@ -642,7 +642,7 @@ def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
     First, `y` is resampled as needed according to the provided `sample_rates`.
     Then, a filterbank with with `n` band-pass filters is designed.
     The resampled input signals are processed by the filterbank as a whole.
-    (`scipy.signal.sosfiltfilt` is used to make the phase linear.)
+    (`scipy.signal.filtfilt` resp. `sosfiltfilt` is used to make the phase linear.)
     The output of the filterbank is cut into frames.
     For each band, the short-time mean-square power (STMSP) is calculated by
     summing `win_length` subsequent filtered time samples.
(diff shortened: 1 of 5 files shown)


stefan-balke:
Hi Frank,

thanks for suggesting this change. If I got you correctly, this only affects higher order filters which could come up when using different pass- and stop-band attenuations...which will make the elliptical filter become longer. So far so good, this change is fine for me, as long as it does not make things significantly slower. Can you comment on this?

Thanks!

fzalkow:
You got it correctly. In fact, it is slower... Mh, not sure, if general numerical stability is worth that?

```python
import librosa

y, sr = librosa.load(librosa.util.example_audio_file())

%timeit librosa.iirt_old(y)  # 3.3 s ± 7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit librosa.iirt_new(y)  # 4 s ± 2.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

stefan-balke:
Well, I would say it depends! :)

If someone adapts the filterbank and ends up with longer filters it totally makes sense.
Maybe it's an option to pass the "convolver" as an `arg` to the function and give details in the docstring?

lostanlen:
hello @fzalkow, and welcome to librosa.
Thank you for pointing this shortcoming of `iirt`. And thank you for a very elegant fix.
The construction of the filter is a cached, so the time difference becomes negligible in the long run.
Ideally I'd like to see a simple heuristic for adapting the method as a function of the parameters.

stefan-balke:
> The construction of the filter is a cached, so the time difference becomes negligible in the long run.

What you mean? It is not about the filter construction but about the actual filtering which was done via `filtfilt()`...there is no caching, or?

lostanlen:
@stefan-balke oh sorry, my bad, I thought that it was the filter design that was slower, not the filtering itself. Then yes, the time difference matters because it scales with the length of the input.

bmcfee:
@fzalkow thanks for doing this!  This is definitely my favorite kind of PR. :grin: 

It does seem to be a pretty significant performance hit, but I'm okay with it if it improves stability.



> Ideally I'd like to see a simple heuristic for adapting the method as a function of the parameters.

I think this could be more trouble than its worth.  It's ultimately a value judgement on the user's behalf: do you care more about a bit of efficiency, or stable filters?  I'm wary of trying to automatically deduce that, and would be generally more amenable to a boolean parameter `sos=True/False`.  However, in this case, I think it's better to just switch it over to sos full stop.

This does present an API-breaking change though, since the filter construction now returns sos format instead of `ba`, as it did before.  Would it be possible to keep `filters._multirate_fb` API-compatible (returning `ba`) and do conversions within `iirt`?  That way we can keep API compatibility and merge this in a minor revision; otherwise we'd have to punt merging until 0.7.

fzalkow:
@lostanlen: Hi! :-) Stefan is right, it is slower because of the filtering, not because of the filter design.

@stefan-balke: As you suggested, I added an kwarg `flayout`.

@bmcfee: Now `filters._multirate_fb` returns `ba` layout by default. Is this okay for a minor revision? Otherwise I can also convert from `ba` inside `iirt`.

lostanlen:
I am in favor of a `FutureWarning`. This change brings unnoticeable results in the default case, but can save the day with user-specified settings. If, for one specific user, `iirt` is the bottleneck of speed, they can always set `flayout` to `ba` after checking manually that it works as expected. But the majority of our users will prefer to err on the side of caution and have a "safe" mode by default.

nils-werner:
Nice! Just two tiny nitpicks:

 - `scipy.signal.sosfiltfilt()` was added to SciPy 0.18.0, but `setup.py` currently says

    ```python
    'scipy >= 0.14.0',
    ```

 - `scipy.signal.iirdesign()` accepts a parameter `output="sos"`, so there is no need to manually do `zpk2sos` afterwards:

    ```python
    cur_filter_zpk = scipy.signal.iirdesign(passband_freqs, stopband_freqs,
                                            passband_ripple, stopband_attenuation,
                                            analog=False, ftype=ftype, output='zpk')
    cur_filter = scipy.signal.zpk2sos(*cur_filter_zpk)
    ```

    can be changed to

    ```python
    cur_filter = scipy.signal.iirdesign(passband_freqs, stopband_freqs,
                                            passband_ripple, stopband_attenuation,
                                            analog=False, ftype=ftype, output='sos')
    ```

fzalkow:
Hi Nils! thanks for the comment!

Concerning the first point: Mh, @bmcfee, if scipy < 0.18 should be supported, we cannot take this solution...

bmcfee:
> Concerning the first point: Mh, @bmcfee, if scipy < 0.18 should be supported, we cannot take this solution...

I'd actually be fine with bumping the version requirement up to 1.0, which has already been out for quite some time.

fzalkow:
Thanks! I am very happy to make a tiny contribution to librosa. :-) If anything else should be done before merging, I am happy to do so.

bmcfee:
Thanks!  I think this is merge-ready once the last round of CI passes.

bmcfee:
Sorry to throw one more request on here, but the coverage report shows that the `sos` path in `iirt` is not covered by tests: https://coveralls.io/builds/20385957/source?filename=librosa%2Fcore%2Fspectrum.py#L776 

Can you add a test to hit this case, just to make sure it runs properly?

fzalkow:
In fact, I already did a test in [a previous commit](https://github.com/librosa/librosa/pull/799/files#diff-2f5f6aa6ab06692483a7e64e57fd9ac3).

Maybe it is because the `else` was not specifically asking for `sos`?

bmcfee:
> Maybe it is because the `else` was not specifically asking for `sos`?

That seems to be checking the filter construction (which is covered), not the transform (`iirt`).

fzalkow:
You are right, this test was only for `semitone_filterbank`. Now I added a test for `iirt`.

bmcfee:
@fzalkow thanks again for this -- any last changes you want to make, or can we merge?

fzalkow:
From my point of view, it's fine. Thanks!

lostanlen:
congratulations @fzalkow on getting this merged! this PR was great. let's have more of these! :)

PR #801, opened 2018-12-04
Rename rmse

### LINKED ISSUE #762: rmse - rms vs energy
This is not really an issue, but I wonder why the rmse feaure:

https://librosa.github.io/librosa/generated/librosa.feature.rmse.html

is named "rms energy". Afaik rms is a 'linear' value (like an amplitude, or any 'field quantity') while energy, like power, is related to suqared amplitudes.

This discussion here 

https://groups.google.com/forum/#!topic/librosa/TIke2HVJHdc

also states that

"energy is the square-root of power." 

I am not quite sure that is correct. Could someone comment on that?

Here are some links that my be useful:

https://en.wikipedia.org/wiki/Root_mean_square

https://en.wikipedia.org/wiki/Energy_(signal_processing)

https://en.wikipedia.org/wiki/Spectral_density

Thanks and BR

nullmightybofo:
#### Reference Issue
 Example: Fixes #762 

Renamed function rmse() to rms() and changed all refernces within librosa.features.spectraly.py accordingly.

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -10,6 +10,7 @@ import scipy.fftpack
 from .. import util
 from .. import filters
 from ..util.exceptions import ParameterError
+from ..util.decorators import moved
 
 from ..core.time_frequency import fft_frequencies
 from ..core.audio import zero_crossings, to_mono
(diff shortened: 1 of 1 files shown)


bmcfee:
> Shouldn't these be covered by the PR?

Ah yes, my bad!  I was just pasting the grep output from my local copy.

nullmightybofo:
That's good, now I learned how to find the references to it fast ;-)

All the tests in test/... still refer to rmse(). Should that stay or also be changed?
Or tests for rms() added? I haven't figured that out.

bmcfee:
Any objection to merging this, and I'll fix the tests later?

lostanlen:
Hey @nullmightybofo ! Thanks again for doing this. Would you mind sharing your full name (first name and last name) so that you're properly credited in the release notes of librosa v0.6.3? That will also allow to produce a DOI for the Zenodo release. Thank you

PR #807, opened 2018-12-28
Improving implementation of "pitch tuning method"

### LINKED ISSUE #806: Maybe a better Implementation of Method "pitch_tuning" in librosa/core/pitch.py
#### Versions

librosa ›
--
 
Maybe a better Implementation of Method "pitch_tuning" in librosa/core/pitch.py

In librosa/core/pitch.py, line 157, "bins = np.linspace(-0.5, 0.5, int(np.ceil(1./resolution)), endpoint=False)"  is written. However, I reckon that it is better to be written like this: "bins = np.linspace(-0.5, 0.5, int(np.ceil(1./resolution))+1)" .

Here is my reason:

image

When you use **bins1** as an input of **np.histogram()**, it has actually 9 bins: [-0.5, -.0.4), ... [0.2, 0.3), [0.3, 0.4]. So any elements > 0.4 in **residuals** will not be counted(see **counts1**). Even if you set the **resolution** as 0.01, any elements > 0.49 in **residuals** will not be counted, which means the range of the **pitch_tuning()**'s return is [-0.5, 0.49], instead of [-0.5, 0.5).

However, look at **bins2** and **counts2**. When put **bins2** in **np.histogram()**, it has actually 10 bins: [-0.5, -.0.4), ... [0.3, 0.4), [0.4, 0.5], and all elements in **residuals** will be counted(because all elements in residuals is between[-0.5, 0.5) ). Now the range of the **pitch_tuning()**'s return is  [-0.5, 0.5), it is more in accordance with the method's description.

monsij:
#### Reference Issue
Closes #806 
This PR changes: 
`bins = np.linspace(-0.5, 0.5, int(np.ceil(1./resolution)), endpoint=False)` to 
`bins = np.linspace(-0.5, 0.5, int(np.ceil(1./resolution))+1)`

### DIFF
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -154,7 +154,7 @@ def pitch_tuning(frequencies, resolution=0.01, bins_per_octave=12):
     # from the next tone up.
     residual[residual >= 0.5] -= 1.0
 
-    bins = np.linspace(-0.5, 0.5, int(np.ceil(1./resolution)), endpoint=False)
+    bins = np.linspace(-0.5, 0.5, int(np.ceil(1./resolution))+1)
 
     counts, tuning = np.histogram(residual, bins)
 


bmcfee:
Thanks for doing this, and so quickly that we got a double PR #808 with the same change! :grin: 

It looks like this change breaks our tuning estimation tests.  Can you look into it and determine whether it's the test that is incorrect?

monsij:
Yeah sure @bmcfee

bmcfee:
I took a quick look into this, and it seems like the only failures are coming up when the tuning is -50c, which is actually an ambiguous case.  The errors (`abs(tuning - est_tuning)`) are coming out nearly a full semitone off: 99c, which is equivalent to 1c since tuning deviation should be computed modulo 1.  The test itself does not compute error deviations with respect to the modulus, so the test fails.

I can think of a couple of things we should do here:

1. Re-write the error comparison in the test to be periodic
2. Don't use an ambiguous test case

bmcfee:
@monsij do you have time to wrap this up in the next week or so?  This is the last issue tagged for 0.6.3, and I'd like to ship that soon.  Thanks!

monsij:
@bmcfee, I'll try my best to do so

PR #808, opened 2018-12-28
a change in pitch.py

Robrib:


bmcfee:
Thanks for doing this -- it looks like #807 beat you to it, with identical changes.  (Such good problems to have!)

Can we consolidate to one PR?

Robrib:
It seems monsij made the same changes at the same time lol
Yeah sure you can consolidate these 2 PR to one @bmcfee

Robrib:
But I want to say apparently I first posted the issue 2 days ago. I have explained the reason for making this change.

PR #814, opened 2019-01-09
Add doc note about bit truncation by `audioread`.

### LINKED ISSUE #813: Doc request: warning about bit truncation for core.load
#### Description
Librosa uses `scipy.io.wavfile.write` for writing wav files and `audioread` for reading audio files. By design, Librosa writes wav files as floating point, which could be half, single, or double precision. This is enforced by a call to `util.valid_audio` and the behavior of `scipy.io.wavfile.write`:
https://github.com/librosa/librosa/blob/master/librosa/output.py#L225
https://github.com/librosa/librosa/blob/master/librosa/util/utils.py#L158
https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.write.html#scipy.io.wavfile.write

The most widely-used precision for wav files is 16-bit integer, but I think the fact that Librosa allows other precisions is OK. Additional precision may be useful to preserving processing that's been applied to a lower-precision signal (e.g., resampling).

Yet `audioread` **always silently truncates audio to a sample width of 2**.
E.g., for integer PCM formats:
https://github.com/beetbox/audioread/blob/master/audioread/rawread.py#L135

And for floating point PCM formats:
https://github.com/beetbox/audioread/blob/master/audioread/ffdec.py#L124

(The latter, which is what is induced by Librosa via `scipy.io.wavfile.write`, uses the FFMPEG decoder to convert/truncate the data to 16-bit PCM)

There's a section in "Advanced I/O Use Cases" that warns users specifically about loading 32-bit floating point audio:
https://github.com/librosa/librosa/blob/master/docs/ioformats.rst

I don't think that this documentation is sufficient, especially considering that the simple of act of writing and loading the same wav file through Librosa will result in bit-truncation for anything over 16-bit int. I recently fell into this trap because I assumed that a write/read cycle would either be lossless or trigger a warning. I should have searched the docs for this specific scenario, but I didn't.

I'd love for the documentation of `core.load` to include a one-line warning about this behavior. It can be similar to the following:

> Warning: This method relies on `audioread`, which silently truncates audio to a sample width of 2. Please refer to "Advanced I/O Use Cases" for alternate methods of loading audio.

#### Versions
I'm using `0.6.2`. This issue is probably relevant to all versions.

woodshop:
#### Reference Issue
Fixes #813

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/814)

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -69,6 +69,13 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
             To use a faster method, set `res_type='kaiser_fast'`.
 
             To use `scipy.signal.resample`, set `res_type='scipy'`.
+            
+        .. note::
+           This uses `audioread`, which may truncate the precision of the
+           audio data to 16 bits.
+
+           See https://librosa.github.io/librosa/ioformats.html for alternate
+           loading methods.
 
 
     Returns


bmcfee:
Thanks again for catching this.  Anyone else want to chime in before we merge?

lostanlen:
Thanks @woodshop !

One small remark: As a non-native English speaker, I did not understand "alternate" in the sense of "alternative". Apparently that's a North American thing.

screen shot 2019-01-10 at 3 11 00 pm

I would be in favor of using the more international term "alternative"

lostanlen:
ok let's merge!

bmcfee:
Thanks! :shipit:

PR #820, opened 2019-01-17
improve audio quality processed by effects.time_stretch() and effects…

vBaiCai:
….pitch_shift()

The origin effects.time_stretch( ) contain a stft operations. **n_fft** of stft is set to 2048 (default), but it's too large in most situation, and it causes some background **noise** in processed audio.
This issus also affects the performance of effects.pitch_shift( ), because pitch_shift( ) relay on time_stretch( ).

I fix it in this way:
Fixed frame-length 40ms is used and **n_fft** is computed by (frame-length * sample_rate / 1000).
The experiments show that my implementation decreases the "noise" in processed audio.
[sample.zip](https://github.com/librosa/librosa/files/2767035/sample.zip)

#### Any other comments?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/820)

### DIFF
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -186,7 +186,7 @@ def percussive(y, **kwargs):
     return y_perc
 
 
-def time_stretch(y, rate):
+def time_stretch(y, sr, rate):
     '''Time-stretch an audio series by a fixed rate.
 
 
@@ -194,6 +194,9 @@ def time_stretch(y, rate):
     ----------
     y : np.ndarray [shape=(n,)]
         audio time series
+
+    sr : number > 0 [scalar]
+        audio sampling rate of `y`
 
     rate : float > 0 [scalar]
         Stretch factor.  If `rate > 1`, then the signal is sped up.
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks for the PR, @vBaiCai .  I see why you would want this, but I have some reservations about the particular changes you're suggesting:

- The `effects` wrapper is mainly intended as a short-cut for quick and dirty manipulations, and it's entirely possible to get the effect that you want (fixed frame length) without modifying the library.  That said, if we can easily extend the API to make this sort of thing possible, I'm generally for it.
- Changing the wrapper so that the frame length is constant in time is not consistent with the rest of the API defaults, which are defined in terms of samples.  If we're going to extend the API, I think we should do it in a way that transparently interfaces with the rest of the code, i.e., provide pass-through parameters for `n_fft` and `hop_length`.
- Anything we do to `time_stretch` should probably also be carried over to `pitch_shift`, since they're essentially the same kind of operation.

None of these are deal-breakers for this PR, I just want to make sure we think carefully about the API when exposing / changing functionality.

As a side note, if you really want high-quality time stretching, you'd be better off looking at `rubberband` (and its wrapper, `pyrubberband`), which are much better at handling transients than librosa's naive phase vocoder implementation.

Issue #818, opened 2019-01-15
use rfft in stft/ifgram/fmt?


bmcfee:
#### Description
In reviewing `core/spectrum.py`, I noticed that we use full FFTs in places where we can almost certainly get away with rffts.  I believe this is a holdover from the initial matlab port, which as noted in #521, involved some unnecessary conjugations.  Places where this comes up:

- STFT: input is assumed to be real, and output is always the non-negative spectrum.  No reason not to use rfft here.
- ISTFT: same as above: we're assumed to be computing a real-valued output, so the conjugate symmetry assumption is enforced anyway.
- IFGRAM: again, same as above, provided we're careful about slicing the phase differential correcly
- FMT: this one's an oddball.  It's technically never stated that we're assuming realness (or conjugate symmetry), but we slice out the non-negative spectrum anyway, so there's no reason not to.

Changing these over to rfft shouldn't impact the outputs.  The runtime will probably be comparable: I don't expect a factor-2 speedup.  However, it would significantly cut down on intermediate storage, which can easily blow up on us with long signals.

As a bonus ask: while implementing this, we should keep an eye out for other optimizations (eg in-place scalings) that could cut down on copies.

#### Versions
librosa <= 0.6.3

bmcfee:
I just took a quick stab at a few different options for this, and here are findings so far.

1. `fftpack.rfft` returns interleaved real values, rather than proper complex values.  So if we use fftpack, the outputs need to be translated into complex values, and this incurs an additional memory copy.
2. `np.fft.rfft` behaves as expected.

I tried a few different methods here, using either numpy or scipy.fftpack, different ways of repacking the real values into complex values, and blocking vs full batch processing.  Each configuration was tested on the included example track at 44100, truncated to different lengths.  I did not vary the frame length.

Before rfft (baseline, 0.6.3 implementation)
```python
In [1]: import librosa                                                                        

In [2]: y, sr = librosa.load(librosa.util.example_audio_file(), sr=None)                      

In [3]: %timeit librosa.stft(y[:sr])                                                          
2.74 ms ± 6.86 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit librosa.stft(y[:2*sr])                                                        
4.12 ms ± 15.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit librosa.stft(y[:4*sr])                                                        
[...]

PR #828, opened 2019-02-16
[WIP] Use rfft where appropriate

bmcfee:
#### Reference Issue
#818 

This PR changes the internal FFT method from `scipy.fftpack.fft` to `numpy.fft.rfft` where appropriate (`stft`, `istft`, `fmt`).  There should be no functionality change, but speed and memory efficiency should be improved.

#### Any other comments?

#818 gives a run-down of some informal benchmarks of various different strategies for this change.  Further improvements may be possibly by monkey-patching the fft module to use `pyfftw`, though this hasn't produced much improvement in my tests so far.

As a quick observation, I'm seeing about a 30% speedup (overall) in the unit tests for `istft`, which include an stft round-trip in their calculation.  STFT and FMT tests are about the same, but the code is a bit simpler and the memory efficiency is indeed better.

That said, `istft` currently involves a separate `irfft` call for each frame, which is probably inefficient.  It could be made faster by doing `irfft` in blocks (similar to stft), and then doing the overlap-add in numba for each block's worth of frames.

TO-DOs:

- [ ] Add documentation/example for fftw monkeypatching
- [ ] Batch-accelerate `istft`, possibly with numba acceleration for overlap-add

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -4,7 +4,7 @@
 import warnings
 
 import numpy as np
-import scipy.fftpack as fft
+import numpy.fft as fft
 import scipy
 import scipy.ndimage
 import scipy.signal
@@ -179,10 +179,12 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
     for bl_s in range(0, stft_matrix.shape[1], n_columns):
         bl_t = min(bl_s + n_columns, stft_matrix.shape[1])
 
-        stft_matrix[:, bl_s:bl_t] = fft.fft(fft_window *
-                                            y_frames[:, bl_s:bl_t],
-                                            axis=0)[:stft_matrix.shape[0]]
-
+        #stft_matrix[:, bl_s:bl_t] = fft.fft(fft_window *
+        #                                    y_frames[:, bl_s:bl_t],
+        #                                    axis=0)[:stft_matrix.shape[0]]
+        stft_matrix[:, bl_s:bl_t] = fft.rfft(fft_window *
+                                             y_frames[:, bl_s:bl_t],
+                                             axis=0)
     return stft_matrix
 
 
@@ -291,13 +293,12 @@ def istft(stft_matrix, hop_length=None, win_length=None, window='hann',
     expected_signal_len = n_fft + hop_length * (n_frames - 1)
     y = np.zeros(expected_signal_len, dtype=dtype)
 
+    # TODO: make this block-sensitive
     for i in range(n_frames):
         sample = i * hop_length
-        spec = stft_matrix[:, i].flatten()
-        spec = np.concatenate((spec, spec[-2:0:-1].conj()), 0)
-        ytmp = ifft_window * fft.ifft(spec).real
+        ytmp = ifft_window * fft.irfft(stft_matrix[:, i])
 
-        y[sample:(sample + n_fft)] = y[sample:(sample + n_fft)] + ytmp
+        y[sample:(sample + n_fft)] += ytmp
 
     # Normalize by sum of squared window
     ifft_window_sum = window_sumsquare(window,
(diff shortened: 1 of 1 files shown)


bmcfee:
Batching and numbafied overlap add in istft brings the time for `pytest -k test_istft` down from 61sec (0.6.3) to 28sec.

bmcfee:
This PR now includes an API to change out the FFT library used throughout spectral transformations to anything which is `numpy.fft` compatible (e.g., `pyfftw.interfaces.numpy_fft`).

The usage is simple enough, as explained in the docstring:
```python
>>> import pyfftw
>>> librosa.set_fftlib(pyfftw.interfaces.numpy_fft)
```
to switch over, and to switch back, a `None` parameter reverts to `numpy.fft`:
```python
>>> librosa.set_fftlib()
```

This will automatically carry over to all stft, istft, ifgram, fmt, autocorrelate, and cqt operations.  We still use `scipy.fftpack` for DCT's in some of the feature transforms (eg mfcc), but I don't see much need for changing that out.

Aside from a couple of simple unit tests, I think this one is done and ready for review.  Any takers?

bmcfee:
Quick follow up: I did a more thorough set of benchmarks using `pyfftw` and `numpy.fft` for `stft` on a number of different signal lengths and FFT sizes, and in all cases on my machine, `numpy.fft` wins.  (Caching enabled, num_threads = 6 on an 8-core machine.)

I still think it's worth including the fftlib backend swapper though, since one could imagine also using `cupy.fft` for GPU-accelerated FFT ops.  (I haven't tested this yet, but it ought to work seamlessly.)

bmcfee:
(Narrator: cupy.fft did not work, but it seems to not work in general.)

bmcfee:
cupy 5.2 also doesn't work, but this seems independent of us.

bmcfee:
@carlthome any thoughts on this?  It's in some ways reimplementing your previous PR #379, but I think we have a clearer picture of how it all fits together now.

librosa/core/fft.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+'''FFT library container'''
```
lostanlen: Fast Fourier Transform

break down the acronym at least once in the file

librosa/core/fft.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def get_fftlib():
+    '''Get the FFT library currently used by librosa
+
+    Returns
+    -------
+    fft : module
+        The FFT library currently used by librosa.
+        Should API-compatible with `numpy.fft`.
```
lostanlen: Must be API-compatible with `numpy.fft`.

librosa/core/spectrum.py
```diff
@@ -172,17 +174,18 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
...
     for bl_s in range(0, stft_matrix.shape[1], n_columns):
         bl_t = min(bl_s + n_columns, stft_matrix.shape[1])
 
-        stft_matrix[:, bl_s:bl_t] = fft.fft(fft_window *
-                                            y_frames[:, bl_s:bl_t],
-                                            axis=0)[:stft_matrix.shape[0]]
-
+        stft_matrix[:, bl_s:bl_t] = fft.rfft(fft_window *
+                                             y_frames[:, bl_s:bl_t],
+                                             axis=0)
```
lostanlen: <3

librosa/core/spectrum.py
```diff
@@ -284,20 +287,29 @@ def istft(stft_matrix, hop_length=None, win_length=None, window='hann',
 
     ifft_window = get_window(window, win_length, fftbins=True)
 
-    # Pad out to match n_fft
-    ifft_window = util.pad_center(ifft_window, n_fft)
+    # Pad out to match n_fft, and add a dummy axis
```
lostanlen: dummy axis -> broadcasting axis

tests/test_core.py
```diff
@@ -1506,3 +1506,21 @@ def test_pcen_ref():
...
     assert np.allclose(Y2, X)
+
+
+def test_get_fftlib():
+    import numpy.fft as fft
+    assert librosa.get_fftlib() is fft
+
+
+def test_set_fftlib():
+
```
lostanlen: extra blank line

bmcfee:
Thanks @lostanlen for the CR.  All comments have been addressed, so this should be ready to :shipit:

Issue #519, opened 2017-02-25
Make aggregation optional in onset_strength_multi


bmcfee:
As a limiting case of subband onset strength calculation, it might be useful to have a way to get an envelope for each input channel. This is more than just diffing and thresholding because of the superflux logic involved in constructing the reference spectrogram.

Currently, this is possible by supplying a full set of band boundaries for aggregation. However, it would be simpler to just make aggregation optional, so that the returned envelope has the same shape as the input spectrogram.

This would probably require a change to the API, since we use `none` to denote default (mean) aggregation. Alternately, we could use a special string `'off'` to disable aggregation.

bmcfee:
Note: while it makes sense to support this in `onset_strength_multi`, it's less clear for `onset_strength`.

It'll take a bit of thinking to work out a consistent API for this.

Perhaps we should take a global pass through the code base, and replace any `None=default callable` arguments with their actual functions.  (Or, perhaps, a dispatch string descriptor ala `get_window`).

bmcfee:
> Perhaps we should take a global pass through the code base, and replace any None=default callable arguments with their actual functions. (Or, perhaps, a dispatch string descriptor ala get_window).

Relatedly: it might be handy to make a repr-wrapper that pretty-prints the function name for default arguments but otherwise acts identically.

bmcfee:
Coming back to this after a mere 18 months: I want to punt on the repr-wrapper and audit for dispatches mentioned above.  Nobody's expressed any opinions on it, and the current API seems good enough.

For the original issue, I think calling with `aggregate=False` would be a reasonable extension of the current API.  This should only be possible in `_multi` though, since by definition the full spectrum onset detector needs to aggregate across bands somehow.  Tagging this for 0.7 now that we have a plan that I like.

PR #831, opened 2019-02-20
Onset multi extension

### LINKED ISSUE #709: Allow user-specified reference spectrum for onset_strength_multi
#### Description

Issue #703 (and corresponding PR #708) generalize PCEN to support pre-computed reference spectra.  The reference spectrum in our PCEN implementation, as in onset strength (superflux) is usually either the input spectrum itself, or a locally max-filtered spectrum (along the frequency axis) to provide robustness against vibrato, scooping, etc.  Allowing the user to override the default reference calculation with their own reference spectrum could be useful if, say, the user wants a frequency-smoothed reference (instead of max-filter), or wants to aggregate reference values across channels, etc.

The question for this issue: do we want to add this functionality to the onset detector as well?  There's really no downside to it, apart from a bit more complexity in the parameter handling logic.  Nobody has asked for this explicitly, but it would make the API consistent across similar functions, and we might be able to factor out some common functionality.

#### Comments

I made this a separate issue, rather than implementing it in #708, because doing it properly would entail a substantial restructuring of `onset_strength_multi` to allow target axes.

bmcfee:
#### Reference Issue
Fixes #709 and #519 

This PR includes two extensions to the onset strength functions:

1. A user can now provide a pre-computed reference spectrum.  This provides a bit more flexibility than just super-flux style max aggregation, and gives a consistent interface with similar functionality in pcen.  #709 
2. In multi-band onsets, aggregation can be completely disabled by calling with `aggregate=False` (#519).  It was previously possible to get this effect by giving `channels=np.arange(len(S))`, but that's rather cumbersome and inefficient.  This is primarily useful for applying the onset strength processing chain to each frequency band for use in some downstream rhythm analysis model.

#### Any other comments?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/831)

### DIFF
--- a/librosa/onset.py
+++ b/librosa/onset.py
@@ -183,6 +183,7 @@ def onset_detect(y=None, sr=22050, onset_envelope=None, hop_length=512,
 
 
 def onset_strength(y=None, sr=22050, S=None, lag=1, max_size=1,
+                   ref=None,
                    detrend=False, center=True,
                    feature=None, aggregate=None,
                    centering=None,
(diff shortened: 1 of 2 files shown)


bmcfee:
Anyone care to give this a look over?  I think it's fine as is, but i'll give it a day or two to gather dust and comments before merging.

Issue #825, opened 2019-02-01
melspectrogram of float32 array returns float64 array


lostanlen:
#### Description
The transformation of an windowed Fourier spectrogram into a mel-frequency spectrogram returns a numPy array in double precision (type `numpy.float64`) even when this windowed Fourier spectrogram is encoded in single precision (type `numpy.float32`). The same happens when computing the mel-frequency spectrogram directly from a single-precision waveform.

Making `melspectrogram` return a numeric `dtype` that is the same as its first positional argument would reduce the memory footprint of librosa and possibly also improve speed. This is all the more crucial that `melspectrogram` is one of the most widely used functions in the library, and that having more than 32 bits of floating point precision is rarely needed in machine listening applications.

The overestimation of precision propagates to all functions that call `melspectrogram`, including, of course, `mfcc`.

#### Steps/Code to Reproduce
```
import librosa

# Load audio file
y, sr = librosa.load(librosa.util.example_audio_file())

# Compute squared modulus of short-term Fourier transform
S = np.abs(librosa.stft(y, dtype='complex64'))**2

# Extract melspectrogram from STFT squared modulus
M_from_S = librosa.feature.melspectrogram(S=S, sr=sr)

# Extract melspectrogram from waveform
M_from_y = librosa.feature.melspectrogram(y=y, sr=sr)

# Print type of all intermediate computations
print(y.dtype, S.dtype, M_from_y.dtype, M_from_S.dtype)
```

#### Expected Results
```float32 float32 float32 float32```

#### Actual Results
```float32 float32 float64 float64```

#### Versions
```
Linux-3.10.0-514.10.2.el7.x86_64-x86_64-with-centos-7.3.1611-Core
Python 3.5.6 |Anaconda, Inc.| (default, Aug 26 2018, 21:41:56) 
[GCC 7.3.0]
NumPy 1.15.2
SciPy 1.1.0
librosa 0.6.2
```

bmcfee:
I'll tack onto this issue that all of our filterbank constructors (mel, chroma, cq_to_chroma, cqt) should have dtype parameters, and we can roll that into the PR.

PR #832, opened 2019-02-21
DType precision for filter bank constructors

bmcfee:
#### Reference Issue
#825 

This PR adds `dtype` options to many of the constructors in `librosa.filters`, and sets defaults to low-precision options (float32 and complex64).  This prevents the downstream processing (mel spectrogram, chroma, cqt) from up-casting inputs from low-precision to high-precision.  If inputs are in high precision arleady, they stay that way as outputs, so there should be no surprises.

#### Any other comments?

I discussed this with @lostanlen offline, and there should be no bad effects of using single precision floats or complex numbers for basis construction.  I expect numerical differences on output to be negligible, especially for machine learning purposes.  For analysis-synthesis purposes, these filters are not (usually) used anyway, so I think we're in the clear.

The outcome of this PR is a significant reduction in memory consumption for most common use cases, and a modest improvement in run time.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/832)

### DIFF
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -209,7 +209,7 @@ def mel(sr, n_fft, n_mels=128, fmin=0.0, fmax=None, htk=False,
 
     dtype : np.dtype
         The data type of the output basis.
-        By default, uses 32-bit floating point.
+        By default, uses 32-bit (single-precision) floating point.
 
     Returns
     -------
@@ -330,7 +330,7 @@ def chroma(sr, n_fft, n_chroma=12, A440=440.0, ctroct=5.0,
 
     dtype : np.dtype
         The data type of the output basis.
-        By default, uses 32-bit floating point.
+        By default, uses 32-bit (single-precision) floating point.
 
     Returns
     -------
@@ -504,7 +504,7 @@ def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, tuning=0.0,
 
     dtype : np.dtype
         The data type of the output basis.
-        By default, uses 64-bit complex.
+        By default, uses 64-bit (single precision) complex floating point.
 
     kwargs : additional keyword arguments
         Arguments to `np.pad()` when `pad==True`.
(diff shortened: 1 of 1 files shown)


librosa/filters.py
```diff
@@ -207,6 +207,10 @@ def mel(sr, n_fft, n_mels=128, fmin=0.0, fmax=None, htk=False,
         (area normalization).  Otherwise, leave all the triangles aiming for
         a peak value of 1.0
 
+    dtype : np.dtype
+        The data type of the output basis.
+        By default, uses 32-bit floating point.
```
lostanlen: By default, uses a single-precision floating-point format, i.e. np.float32.
A double-precision format corresponds to dtype=np.float64

librosa/filters.py
```diff
@@ -494,6 +502,10 @@ def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, tuning=0.0,
         Type of norm to use for basis function normalization.
         See librosa.util.normalize
 
+    dtype : np.dtype
+        The data type of the output basis.
+        By default, uses 64-bit complex.
```
lostanlen: Potentially confusing, because np.complex64 is really two np.float32.

"By default, uses complex-valued, single-precision, floating-point format, i.e. np.complex64.
Double precision can be obtained by setting dtype=np.complex128"

librosa/filters.py
```diff
@@ -703,6 +715,11 @@ def cq_to_chroma(n_input, bins_per_octave=12, n_chroma=12,
         If True, the first chroma bin will start at 'C'
         If False, the first chroma bin will start at 'A'
 
+    dtype : np.dtype
+        The data type of the output basis.
+        By default, uses 32-bit floating point.
```
lostanlen: By default, uses a single-precision floating-point format, i.e. np.float32.
A double-precision format corresponds to dtype=np.float64

bmcfee:
Thanks for the CR!  Changes done.  The only remaining question for me is whether we want to extend this to the elliptical filter constructor (semitone_filter_bank)?  I could see arguments either way -- it'd be nice for consistency, but on the flip side, IIR filters are more numerically sensitive, and scipy's filter design functions don't give us any choice but to use doubles.  We could downcast, but that seems not so great, so I'm leaning toward leaving them as is.

If you're okay with that, I think we're in good shape, and can merge this.

lostanlen:
LGTM

bmcfee:
TY :shipit:

Issue #840, opened 2019-03-07
Include self-loops in recurrence matrix


bmcfee:
#### Description

The recurrence matrix function by default does not include self-loops.    However, there are some cases in which it would be useful to have, e.g., in #741 when doing near-diagonal smoothing.

We should allow setting width=0 to fill in self-loops.

Issue #741, opened 2018-06-28
segment: multi-angle path enhancement


bmcfee:
#### Description

The segment module currently provides basic functionality for doing diagonal filtering and recurrence<->lag matrix conversion.  There are some other basic functions that we could provide for doing general diagonal filtering for path enhancement, and multi-angle filtering and aggregation.

bmcfee:
Tacking on a bit more here, the L/S/Q enhancement methods described by [Serra, 2009](http://iopscience.iop.org/article/10.1088/1367-2630/11/9/093017/meta) would be easy to implement and nice to have.  I took a quick hack at them using numba, and they're pretty quick to compute.  The tricky part would be making them work on sparse input data.

bmcfee:
After a quick chat with @ctralie, I think we're in agreement that the L/S/Q enhancement methods are all special cases of building the [Smith-Waterman](https://en.wikipedia.org/wiki/Smith%E2%80%93Waterman_algorithm) cost matrix.  It would be great if we can build a generic implementation with a flexible API.

bmcfee:
Circling back on the original idea here, I read through the SM toolbox code and [accompanying paper](https://www.audiolabs-erlangen.de/content/05-fau/professor/00-mueller/03-publications/2014_MuellerJiangGrohganz_ToolboxSM_AES.pdf).  As I understand it, the multi-angle enhancement is achieved by resampling one of the input feature sequences *prior* to computing the self-similarity matrix, averaging (smoothing) similarities in time, and then aggregating the resulting self-sims over all the different choices of resampling (tempo ratios).

That approach worked well enough for their specific case, but I think it's not quite general enough to fit the API in librosa.

Instead, I think we can get pretty similar results by convolving the self-similarity matrix with a set of 2-dimensional filters at different orientations.  Slope of 1 = exact time; slope of 2 = twice as fast, slope of 1/2 = half as fast.  These can be obtained by constructing a slope=1 square filter with a smoothing filter on the main diagonal (eg hann), and rotating them by whatever the desired ratios should be.  After convolving, by the set of filters, we can max over all orientations to get a tempo-change-robust path enhancement matrix.  It's not quite the same as SM toolbox, but it's similar in spirit and a bit more flexible since it doesn't rely on knowing the input features and recomputing self-/cross-similarities.

PR #842, opened 2019-03-07
[CR needed] Path enhancement

bmcfee:
#### Reference Issue
This PR implements #741 for multi-angle path enhancement, and #840 for extending the `recurrence_matrix` function to support self-loops.

As discussed in #741, the implementation of multi-angle path enhancement here is a bit different from that of the SM toolbox (and Mueller and Kurth'06).  Rather than resampling the feature sequence at different tempo ratios, the similarity matrix is smoothed directly by convolving with two-dimensional filters at a variety of orientations corresponding to different tempo ratios.  The net effect is similar, but the benefit of this approach is that the similarity matrix calculation is entirely decoupled from the smoothing process.

Also somewhat different from SM toolbox is that the kernel can either be an averaging filter or a zero-mean filter.  The latter is useful for suppressing blocks while enhancing paths, but is disabled by default.  In concert with zero-mean filtering, a clipping parameter is introduced to threshold the output and prevent negative similarities.

TODO:
- [x] add self-loop support to recurrence_matrix
- [x] Clean up docstring formatting

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/842)

### DIFF
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -1256,7 +1256,7 @@ def diagonal_filter(window, n, slope=1.0, angle=None, zero_mean=False):
     window : string, tuple, number, callable, or list-like
         The window function to use for the filter.
 
-        See get_window for details.
+        See `get_window` for details.
 
         Note that the window used here should be non-negative.
 
@@ -1267,9 +1267,9 @@ def diagonal_filter(window, n, slope=1.0, angle=None, zero_mean=False):
         The slope of the diagonal filter to produce
 
     angle : float or None
-        If given, the slope parameter is ignored, 
+        If given, the slope parameter is ignored,
         and angle directly sets the orientation of the filter (in radians).
-        Otherwise, angle is inferred as arctan(slope).
+        Otherwise, angle is inferred as `arctan(slope)`.
 
     zero_mean : bool
         If True, a zero-mean filter is used.
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -23,7 +23,6 @@ Window functions
     window_bandwidth
     get_window
 
-
 Miscellaneous
 -------------
 .. autosummary::
@@ -33,6 +32,7 @@ Miscellaneous
     cq_to_chroma
     mr_frequencies
     window_sumsquare
+    diagonal_filter
 
 Deprecated
 ----------
@@ -46,6 +46,7 @@ import warnings
 import numpy as np
 import scipy
 import scipy.signal
+import scipy.ndimage
 import six
 
 from numba import jit
@@ -68,7 +69,8 @@ __all__ = ['dct',
            'get_window',
            'mr_frequencies',
            'semitone_filterbank',
-           'window_sumsquare']
+           'window_sumsquare',
+           'diagonal_filter']
 
 
 # Dictionary of window function bandwidths
@@ -1241,3 +1243,65 @@ def window_sumsquare(window, n_frames, hop_length=512, win_length=None, n_fft=20
     __window_ss_fill(x, win_sq, n_frames, hop_length)
 
     return x
+
+
+@cache(level=10)
+def diagonal_filter(window, n, slope=1.0, angle=None, zero_mean=False):
+    '''Build a two-dimensional diagonal filter.
+
+    This is primarily used for smoothing recurrence or self-similarity matrices.
+
+    Parameters
+    ----------
+    window : string, tuple, number, callable, or list-like
+        The window function to use for the filter.
+
+        See get_window for details.
+
+    n : int > 0
+        the length of the filter
+
+    slope : float
+        The slope of the diagonal filter to produce
+
+    angle : float or None
+        If given, the slope parameter is ignored, 
+        and angle directly sets the orientation of the filter (in radians).
+        Otherwise, angle is inferred as arctan(slope).
+
+    zero_mean : bool
+        If True, a zero-mean filter is used.
+        Otherwise, a non-negative averaging filter is used.
+
+        This should be enabled if you want to enhance paths and suppress
+        blocks.
+
+
+    Returns
+    -------
+    kernel : np.ndarray, shape=[(m, m)]
+        The 2-dimensional filter kernel
+
+
+    Notes
+    -----
+    This function caches at level 10.
+
+    '''
+
+    if angle is None:
+        angle = np.arctan(slope)
+
+    win = np.diag(get_window(n, window, fftbins=False))
+
+    if not np.isclose(angle, np.pi/4):
+        win = scipy.ndimage.rotate(win, 45 - angle * 180 / np.pi,
+                                   order=5, prefilter=False)
+
+    np.clip(win, 0, None, out=win)
+    win /= win.sum()
+
+    if zero_mean:
+        win -= win.mean()
+
+    return win
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -1258,6 +1258,8 @@ def diagonal_filter(window, n, slope=1.0, angle=None, zero_mean=False):
 
         See get_window for details.
 
+        Note that the window used here should be non-negative.
+
     n : int > 0
         the length of the filter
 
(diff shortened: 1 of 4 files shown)


bmcfee:
This one's feature-complete and ready for CR.  There may be one snag somewhere in the modifications I made to `recurrence_matrix`, which appears to be modifying the structure of the sparse matrix after construction.  I'll look into it and comment here when i have a sense of what's going on.

bmcfee:
Quick note: this doesn't handle distance mode correctly for self=true. Will fix later.

bmcfee:
And for posterity, here's what the multi-angle path enhancement looks like on the sample track after part of it has been sped up by phase-vocoding:
image

bmcfee:
@stefan-balke do you have time for a CR on this one?  (It's fine if not, let me know and I'll reassign.)

stefan-balke:
Hey, yes but currently not top-1 priority. Give me 1 more week to squeeze it in.

stefan-balke:
___
*[tests/test_segment.py, line 340 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaURier1QPQPEMTpRFI:-LaURier1QPQPEMTpRFJ:bzcbgc) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/tests/test_segment.py#L340)):*
> ```Python
>         yield tf, n_segments
> ```

too many empty lines

stefan-balke:
___
*[librosa/segment.py, line 732 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaURzA7Aft5BeBriU87:-LaURzA7Aft5BeBriU88:bzcbgc) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/librosa/segment.py#L732)):*
> ```Python
>         list(1 + np.nonzero(np.diff(clusterer.labels_))[0].astype(int)))
>     return np.asarray(boundaries)
> ```

too many empty lines

stefan-balke:
___
*[librosa/segment.py, line 254 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaUSF4J1FGdYV3z-DF_:-LaUSF4J1FGdYV3z-DFa:b-q3rxtb) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/librosa/segment.py#L254)):*
> ```Python
>     # symmetrize
>     if sym:
>         # Note: this operation produces a CSR matrix!
> ```

Please write-out CSR

stefan-balke:
___
*[librosa/segment.py, line 175 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaUSSDJ0GhaC6Cp3Kxs:-LaUSSDJ0GhaC6Cp3Kxt:b-zdpauq) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/librosa/segment.py#L175)):*
> ```Python
>     ...                          cmap='magma_r')
>     >>> plt.title('Affinity recurrence')
>     >>> plt.tight_layout()
> ```

The plot is not appearing in the current docs. Please add an SSM as example.

stefan-balke:
___
*[librosa/segment.py, line 54 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaUSjgu9oP6V6Qu-aDT:-LaUSjgu9oP6V6Qu-aDU:b-tjokg1) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/librosa/segment.py#L54)):*
> ```Python
> 
> @cache(level=30)
> def recurrence_matrix(data, k=None, width=1, metric='euclidean',
> ```

Call me stupid but I still do not get the difference between a recurrence matrix and an SSM.

Maybe it would help:
* to point to some literature for recurrence matrix computations
* briefly explain the difference to SSMs
* explain how an RM "degenerates" to an SSM
* is this computationally more demanding to do it this way?

However, SSMs are the more well-known use case for me and coming from SSMs to explain RMs would help (hopefully not only me).

stefan-balke:
___
*[librosa/segment.py, line 266 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaUTiQo-tMkwkEmXBnD:-LaUTiQpBeAAutaYbXwK:bajnsva) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/librosa/segment.py#L266)):*
> ```Python
>         if bandwidth is None:
>             bandwidth = np.nanmedian(rec.max(axis=1).data)
>         # Set all the negatives back to 0
> ```

where are those negative values coming from?

stefan-balke:
___
*[librosa/segment.py, line 790 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaUUOvaDrW0mU4YiKrF:-LaUUOvaDrW0mU4YiKrG:bydkfkq) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/librosa/segment.py#L790)):*
> ```Python
>     n_filters : int >= 1
>         The number of different smoothing filters to use.
>         Use an odd integer to ensure that a ratio of 1 is included.
> ```

Hm...this should be checked in the function and raise a warning if not fulfilled.

stefan-balke:
___
*[librosa/segment.py, line 825 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaUUrHr28qIHJTz9v9g:-LaUUrHr28qIHJTz9v9h:bhwmgfw) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/librosa/segment.py#L825)):*
> ```Python
>     >>> y, sr = librosa.load(librosa.util.example_audio_file(), duration=30)
>     >>> chroma = librosa.feature.chroma_cqt(y=y, sr=sr)
>     >>> rec = librosa.segment.recurrence_matrix(chroma, mode='affinity', self=True)
> ```

Is the recurrence matrix now an SSM since `self=True` or is it a RM which includes the main diagonal? We should be careful with the two terms. Maybe we should stick to RM here consistently.

stefan-balke:
Alright, in principal this looks great. However, the use of the term SSM and RM interchangeably is confusing me. Maybe it's time to add a separate SSM function or explain how RM = SSM?

bmcfee:
@stefan-balke thanks for the CR!  I think I've hit all your points and corrections, except for the docstring example, which I think is out of scope for this PR (/ not a real problem in the code base).

I added some language in `recurrence_matrix` that clarifies the relation between recurrence, similarity, and distance.

Apart from that, I think this is good to go once CI passes.

stefan-balke:
___
*[librosa/segment.py, line 54 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaUSjgu9oP6V6Qu-aDT:-LaVk8Fi305jPU6tBOSk:b6q6nk1) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/librosa/segment.py#L54)):*

"recurrence matrix" usually means a binary adjacency matrix, which is how this function began life.  It later grew to include weighted (Sparse) graphs, similarity, and distance calculations.  So I think of RM as a more general term for this kind of thing.
</blockquote></details>

Okay, that helps me in understanding the general concept.
If you could put 2-3 sentences in the docs would be awesome for this PR.

In a future PR, we could extend this a little with some 5x5 toy example matrix...

stefan-balke:
___
*[librosa/segment.py, line 825 at r2](https://reviewable.io/reviews/librosa/librosa/842#-LaUUrHr28qIHJTz9v9g:-LaVmKs75Ngn2RhEWcWf:b-kw05dz) ([raw file](https://github.com/librosa/librosa/blob/067c6f89a03191d67f5575d29b0bb61659a0b129/librosa/segment.py#L825)):*

It depends on the mode, I suppose.  But I'm okay with using `recurrence matrix` pretty liberally to refer to any kind of self-by-self comparison.
</blockquote></details>

Well, if RM on some mode is a SSM, then naming it SSM would be appropriate. Don't get me wrong, I like the concept but it seems to be very versatile and thus a little abstract to get. I think extending the RM docs with more docs will clarify a lot!

stefan-balke:
Yes, we are good here. Thank @bmcfee for putting things together!

Issue #843, opened 2019-03-08
Griffin-Lim, inverse mel, inverse mfcc


bmcfee:
#### Description
This issue is mostly inspired by commentary on #434, and a narrowing of scope.

Specifically, I'd like to add the following functions and a `librosa.feature.inverse` submodule:

- `core.griffinlim` for linear magnitude spectra --> time domain signal
- `feature.inverse.mel_to_stft` invert a mel power spectrogram (with or without dB) --> linear spectrogram
- `feature.inverse.mel_to_audio` invert a mel power spectrogram (with or without dB) --> audio
- `feature.inverse.mfcc_to_mel` invert mfcc -> mel power spectrogram
- `feature.inverse.mfcc_to_audio` -> mfcc to audio

Once GL is in place, the rest can be implemented using least squares / pseudo-inversion of the filters, and the existing `db_to_amplitude` function.

Unit testing this will be a bit of a pain: we can use probe signals centered exactly at mel frequencies for part of it, but it will get pretty dicey for mfcc inversion.

keunwoochoi:
This is great!

PR #844, opened 2019-03-09
Feature inversion

bmcfee:
#### Reference Issue
Implements #843 

This PR adds the following functions:

- `core.griffinlim` (adapted from jongwook's implementation)
- `feature.inverse.mel_to_stft`
- `feature.inverse.mel_to_audio`
- `feature.inverse.mfcc_to_mel`
- `feature.inverse.mfcc_to_audio`

#### Any other comments?

This is still very much a work in progress.  Everything basically works and is more or less documented.  

The major gap to fill right now is in the mel basis inversion.  Ideally, this should be a non-negative, linear least squares problem to recover the STFT magnitudes from the Mel magnitudes.  Right now, it uses a single unconstrained least-squares solve and projects the solution down to the feasible set.  This gets us pretty close, but it's not optimal.  Numpy does not directly provide an nnls solver, and scipy does, but it does not vectorize, so we'd have to solve each column independently.

At the risk of creating some feature bloat, I think there's some justification to providing our own vectorized NNLS solver under `util`.  NNLS occurs in a few places in audio, for example: [NNLS Chroma](https://www.eecs.qmul.ac.uk/~simond/pub/2010/Mauch-Dixon-ISMIR-2010.pdf), [onset detection](http://mac.citi.sinica.edu.tw/~yang/pub/liang15spl.pdf), and estimating stem mixing coefficients from a mix, to name a few.

At the moment, I'm thinking of implementing nnls as a numba-accelerated admm algorithm.  As far as I can tell, all of the required operations are supported by numba (lstsq, thresholding), so it shouldn't be too hard to get it together. (Famous last words, I know).

#### TODO

- [x] Unit tests
- [x] NNLS?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/844)

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -26,7 +26,7 @@ __all__ = ['stft', 'istft', 'magphase', 'iirt',
            'perceptual_weighting',
            'power_to_db', 'db_to_power',
            'amplitude_to_db', 'db_to_amplitude',
-           'fmt', 'pcen']
+           'fmt', 'pcen', 'griffinlim']
 
 
 @cache(level=20)
@@ -1560,6 +1560,110 @@ def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
         return S_out, zf
     else:
         return S_out
+
+
+def griffinlim(S, n_iter=32, hop_length=None, win_length=None, window='hann',
+               center=True, dtype=np.float32, length=None):
+    '''Approximate magnitude spectrogram inversion using the Griffin-Lim algorithm [1]_.
+
+    Given a short-time Fourier transform magnitude matrix (`S`), the algorithm randomly
+    initializes phase estimates, and then alternates forward- and inverse-stft operations
+
+    .. [1] D. W. Griffin and J. S. Lim,
+        "Signal estimation from modified short-time Fourier transform,"
+        IEEE Trans. ASSP, vol.32, no.2, pp.236–243, Apr. 1984.
+
+    Parameters
+    ----------
+    S : np.ndarray [shape=(n_fft / 2 + 1, t), non-negative]
+        An array of short-time fourier transform magnitudes
+
+    n_iter : int > 0
+        The number of iterations to run
+
+    hop_length : None or int > 0
+        The hop length of the STFT.  If not provided, it will default to `n_fft // 4`
+
+    win_length : None or int > 0
+        The window length of the STFT.  By default, it will equal `n_fft`
+
+    window : string, tuple, number, function, or np.ndarray [shape=(n_fft,)]
+        A window specification as supported by `stft` or `istft`
+
+    center : boolean
+        If `True`, the STFT is assumed to use centered frames.
+        If `False`, the STFT is assumed to use left-aligned frames.
+
+    dtype : np.dtype
+        Real numeric type for the time-domain signal.  Default is 32-bit float.
+
+    length : None or int > 0
+        If provided, the output `y` is zero-padded or clipped to exactly `length`
+        samples.
+
+
+    Returns
+    -------
+    y : np.ndarray [shape=(n,)]
+        time-domain signal reconstructed from `S`
+
+    See Also
+    --------
+    stft
+    istft
+    magphase
+    filters.get_window
+
+    Examples
+    --------
+    A basic STFT inverse example
+
+    >>> y, sr = librosa.load(librosa.util.example_audio_file(), duration=5, offset=30)
+    >>> # Get the magnitude spectrogram
+    >>> S = np.abs(librosa.stft(y))
+    >>> # Invert using Griffin-Lim
+    >>> y_inv = librosa.griffinlim(S)
+    >>> # Invert without estimating phase
+    >>> y_istft = librosa.istft(S)
+
+    Wave-plot the results
+
+    >>> import matplotlib.pyplot as plt
+    >>> plt.figure()
+    >>> ax =plt.subplot(3,1,1)
+    >>> librosa.display.waveplot(y, sr=sr, color='b')
+    >>> plt.title('Original')
+    >>> plt.xlabel('')
+    >>> plt.subplot(3,1,2, sharex=ax, sharey=ax)
+    >>> librosa.display.waveplot(y_inv, sr=sr, color='g')
+    >>> plt.title('Griffin-Lim')
+    >>> plt.xlabel('')
+    >>> plt.subplot(3,1,3, sharex=ax, sharey=ax)
+    >>> librosa.display.waveplot(y_istft, sr=sr, color='r')
+    >>> plt.title('Magnitude-only istft')
+    >>> plt.tight_layout()
+    '''
+
+    # Infer n_fft from the spectrogram shape
+    n_fft = 2 * (S.shape[0] - 1)
+
+    # randomly initialize the phase
+    angles = np.exp(2j * np.pi * np.random.rand(*S.shape))
+
+    for _ in range(n_iter):
+        # Invert with our current estimate of the phases
+        inverse = istft(S * angles, hop_length=hop_length, win_length=win_length,
+                        window=window, center=center, dtype=dtype, length=length)
+
+        # Rebuild the spectrogram
+        rebuilt = stft(inverse, n_fft=n_fft, hop_length=hop_length,
+                       win_length=win_length, window=window, center=center)
+
+        # Update our phase estimates
+        angles[:] = np.exp(1j * np.angle(rebuilt))
+
+    return istft(S * angles, hop_length=hop_length, win_length=win_length,
+                 window=window, center=center, dtype=dtype, length=length)
 
 
 def _spectrogram(y=None, S=None, n_fft=2048, hop_length=512, power=1,
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1567,7 +1567,11 @@ def griffinlim(S, n_iter=32, hop_length=None, win_length=None, window='hann',
     '''Approximate magnitude spectrogram inversion using the Griffin-Lim algorithm [1]_.
 
     Given a short-time Fourier transform magnitude matrix (`S`), the algorithm randomly
-    initializes phase estimates, and then alternates forward- and inverse-stft operations
+    initializes phase estimates, and then alternates forward- and inverse-STFT
+    operations.
+    Note that this assumes reconstruction of a real-valued time-domain signal, and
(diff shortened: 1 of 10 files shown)


bmcfee:
PR now includes a numba-accelerated, vectorized non-negative least squares solver.  Qualitatively, the results of mel inversion are substantially improved when using NNLS as compared to a 1-step least-squares and project method.

The NNLS solver is ... somewhat involved.  It's an ADMM method, that in this case, reduces to a sequence of matrix multiplications and thresholding (projection onto feasible set).  The "involved" part here is that for wide bases `A`, the direct update would be expensive, so there's some dynamic checking to see if it would be faster to use the Woodbury identity.  There's also some hackery to guess a good step size for the method based on the singular values of A.  (And, if the input turns out to be a single vector, we fall back on the scipy optimize method.)

bmcfee:
Note: there's currently a bit of a quirk in the nnls solver that force-upcasts all inputs to float64 to prevent some numba mixed-typing errors.  I'd like to get this resolved to be precision-agnostic before merging.

bmcfee:
Unit tests are in place now.  They're relatively strong for `nnls`, less so for griffin-lim and the feature inverters.  Qualitatively, the results sound good, but I'm wary of trying to be too rigorous on the unit tests for these methods.

CR would be welcome at this point!

bmcfee:
Hold up to anyone reviewing this one: I think this can be made faster / more robust by wrapping `fmin_l_bfgs_b` in a more clever way.

bmcfee:
Quick update: lbfgs implementation is substantially more accurate, but slower on long signals.  However, since the problem is separable along columns, we can probably hit a nice trade-off by block-wise processing.  I'll give that a shot and see if it gets closer to admm speed.

Overall, I'd prefer to rely on an existing solver rather than maintain our own here.

bmcfee:
*UPDATE*: nnls solver has been rewritten to use l-bfgs-b.  It's considerably more stable and much less code, though the runtime efficiency isn't always as good.  OTOH the parameters can be directly controlled by the user if they know what they're doing.

librosa/feature/inverse.py
```diff
@@ -0,0 +1,263 @@
...
+    '''
+
+    # Construct a mel basis with dtype matching the input data
+    mel_basis = filters.mel(sr, n_fft, n_mels=M.shape[0],
+                            dtype=M.dtype,
+                            **kwargs)
+
+    # Find the non-negative least squares solution, and apply
+    # the inverse exponent
+    return nnls(mel_basis, M)**(1./power)
```
lostanlen: would it be worth to make a conditional branch for the case power=1? (and potentially power=2?) or is that a premature optimization?

librosa/feature/inverse.py
```diff
@@ -0,0 +1,263 @@
...
+
+    n_mels : int > 0
+        The number of Mel frequencies
+
+    dct_type : None or {1, 2, 3}
+        Discrete cosine transform (DCT) type
+        By default, DCT type-2 is used.
+
+    norm : None or 'ortho'
+        If `dct_Type` is `2 or 3`, setting `norm='ortho'` uses an ortho-normal
```
lostanlen: orthonormal

librosa/feature/inverse.py
bmcfee: Last I checked, numpy is smart enough to special-case those inputs, so it shouldn't matter?  Or are you suggesting to do it in-place to save on memory copies?

lostanlen: no, that's fine. i didn't know that numpy had optimization baked in.

librosa/core/spectrum.py
```diff
@@ -1533,6 +1533,115 @@ def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
...
+        "Signal estimation from modified short-time Fourier transform,"
+        IEEE Trans. ASSP, vol.32, no.2, pp.236–243, Apr. 1984.
+
+    Parameters
+    ----------
+    S : np.ndarray [shape=(n_fft / 2 + 1, t), non-negative]
+        An array of short-time fourier transform magnitudes
+
+    n_iter : int > 0
+        The number of iterations to run
```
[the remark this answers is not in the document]
bmcfee: I don't want to do anything too clever here, but I'm open to suggestions for how to document this without being too detailed for lay users.

librosa/core/spectrum.py
```diff
@@ -1533,6 +1533,115 @@ def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
...
+    >>> librosa.display.waveplot(y, sr=sr, color='b')
+    >>> plt.title('Original')
+    >>> plt.xlabel('')
+    >>> plt.subplot(3,1,2, sharex=ax, sharey=ax)
+    >>> librosa.display.waveplot(y_inv, sr=sr, color='g')
+    >>> plt.title('Griffin-Lim')
+    >>> plt.xlabel('')
+    >>> plt.subplot(3,1,3, sharex=ax, sharey=ax)
+    >>> librosa.display.waveplot(y_istft, sr=sr, color='r')
+    >>> plt.title('Magnitude-only istft')
```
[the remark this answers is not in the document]
bmcfee: I think it's implicit in context, but sure.

librosa/feature/inverse.py
bmcfee: also, `dct_Type` => `dct_type`

librosa/feature/inverse.py
bmcfee: Ah, I spoke too soon.  Python optimizes these cases, but `np.power` does not.  So we can either get it fast (using python ops) or in-place (using `np.power`) but apparently not both.  Curious.

```python
In [11]: x = np.arange(int(1e6), dtype=np.float32)                                            

In [12]: %timeit np.power(x, 1.0)                                                             
11.4 ms ± 33.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [13]: %timeit np.power(x, 1)                                                               
11.6 ms ± 182 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [14]: %timeit np.power(x, 0.5)                                                             
11.3 ms ± 19.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [15]: np.power?                                                                            

In [16]: %timeit np.power(x, 2./3)                                                            
11.4 ms ± 7.19 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [17]: %timeit np.power(x, 1, out=x)                                                        
11.4 ms ± 55.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [18]: %timeit np.power(x, 1., out=x)                                                       
11.4 ms ± 46.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [19]: %timeit np.sqrt(x)                                                                   
342 µs ± 13.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [20]: %timeit x**(1./2)                                                                    
327 µs ± 5.39 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [21]: %timeit x**(1./1)                                                                    
498 µs ± 7.27 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

bmcfee:
Thanks @lostanlen  -- last commit addresses most of your comments.  Some outstanding issues:

1) documentation on `n_iter` for griffin-lim.  Seems to me that less is more here, since it could easily get way too detailed, and ultimately it shouldn't matter too much.  People that really care about this will most likely be using a smarter algorithm (famous last words...)

2) power calculation in `mel_to_stft` has a trade-off between raw speed (using python ops) or memory efficiency (using numpy.power in-place).  Small test cases show python ops winning out, but I think the common case should optimize for memory efficiency, which will most likely be the bottleneck.  Anyway, this isn't a mission-critical function, and we're talking about differences in the milliseconds.

3) nnls CR.. not sure it needs more detailed review; it's functional, and I think the API makes sense.  I don't foresee the need to change the api at all, so any subsequent optimizations we want to make should be easy.

4) usability -- open to suggestions, but I think API-wise it's as simple as it can be.

librosa/core/spectrum.py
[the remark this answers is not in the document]
lostanlen: sure. in that case, i think that the main message that needs to be conveyed to the user is that Griffin-Lim is fundamentally suboptimal for reconstructing natural sounds (i.e. anything outside of a slowly modulated sinusoid), and that increasing the number of iterations quickly reaches a plateau in terms of perceptual audio quality.

lostanlen:
> documentation on n_iter for griffin-lim. Seems to me that less is more here, since it could easily get way too detailed, and ultimately it shouldn't matter too much. People that really care about this will most likely be using a smarter algorithm

ok so then we should be on the nose about the fact that griffin-lim is by no means a smart algorithm, as you put it. adding feature inversion to librosa is going to attract all sorts of use cases for generative models, DAFx, etc. and that is great! but it would be better to not give too many false hopes to new users.



> 4.usability -- open to suggestions, but I think API-wise it's as simple as it can be.

an entry in the "Librosa audio effects and playback" notebook perhaps?

bmcfee:
As discussed offline, people are likely to abuse the vanilla GL implementation as currently written.  It's not much additional work for us to implement the "fast" GL solver described [here](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=6701851) (essentially a momentum update) which appears to consistently improve on GL even with a fixed momentum parameter `alpha=0.99`.  I see no downside to adding this prior to merging.

lostanlen:
@bmcfee thoughts about the idea of putting this in a notebook?

bmcfee:
> @bmcfee thoughts about the idea of putting this in a notebook?

Can you be more specific about what you have in mind?  Something like: NMF on a mel spectrum, followed by resynthesis?  Or something more basic?

bmcfee:
Side note todo: we should allow the user to specify a random seed for phase initialization.

bmcfee:
Implemented FleGLA in a notebook, looks pretty good.  Below is on a 10sec excerpt from the demo track, yaxis is SNR (bigger is better):
image

Notebook implementation is ram-inefficient, will clean up and push later in the week.

bmcfee:
One more example, with a smaller frame overlap (hop_length=1024) and more values for alpha:
image

Consistent with the existing literature, alpha=0.99 seems like a well-behaved default.

bmcfee:
@lostanlen fast griffin-lim is now in, ready for inspection.

bmcfee:
per offline conversation, we decided to punt on designing a new notebook example for griffin-lim in this PR.

I think this is now ready to go.

PR #847, opened 2019-03-11
[CR needed] Soundfile-based loader

### LINKED ISSUE #845: Use soundfile for primary IO, fall back on audioread in emergencies
#### Description
As noted too many times in the issues to keep track of, the current `audioread`-based loader is both inefficient and somewhat finnicky across platforms and with varying back-end codec configurations.
My recommendation has long been to use `soundfile` unless mp3 support is absolutely necessary, and plan to eventually switch over to soundfile in librosa once mp3 support is merged.

That said, there's no reason we couldn't have a soft dependency on `soundfile` as a first-pass attempt to load the data, and only fall back to audioread if soundfile fails for some reason (not installed, missing codec, etc).  This would basically have librosa doing audioread's job (multiplexing over decoding backends), but this is only a temporary situation until mp3 support arrives.  In the meantime, we'd get faster loading and a more stable backend.

What do folks think?

bmcfee:
#### Reference Issue
Fixes #845 

This PR makes `soundfile` the primary IO backend, but will fall back to `audioread` on failure (eg if trying to read an mp3).

#### Any other comments?

This should work transparently, and ease the eventual transition to a purely libsndfile-based loader.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/847)


bmcfee:
Quick benchmark: loading a 4-minute wave file brings the average load time from 92ms down to 8ms.

Speedup is less dramatic on compressed files (eg ogg), but everything seems to work as expected.

bmcfee:
I think this is ready to go, following a CR.

bmcfee:
I've triple checked everything here, and I think things are solid.  The numbers come out bit-for-bit the same (when appropriate), so I think the functionality has no issues.

The one thing I'd like some extra eyes on is the documentation, both the docstring and the readme, and advanced IO examples, just to make sure it's all consistent.  We might want to rewrite a bit of the advanced io examples to use librosa.load directly, now that we can (in some cases), but then again, it might be better to leave it as is because librosa.load obscures the backend depending on codec, which might complicate things a bit.  It's not wrong to leave the docs as is, but I'd like a second opinion before merging.

@faroit @jfsantos @stefan-balke ?  Any takers?

faroit:
>  The numbers come out bit-for-bit the same (when appropriate), so I think the functionality has no issues.

Did you test with different `audioread` backends? E.g. I'm thinking about the small differences in int->float conversion [when using libsndfile](http://blog.bjornroche.com/2009/12/int-float-int-its-jungle-out-there.html)

README.md
```diff
@@ -63,15 +63,24 @@ librosa (0.x.x, /path/to/librosa)
...
 
-#### audioread
+`librosa` uses `soundfile` and `audioread` to load audio files.
+Note that `soundfile` does not currently support MP3, which will cause librosa to
+fall back on the `audioread` library.
 
-*Note that `audioread` needs at least one of the programs to work properly.*
+#### soundfile
+
+If you're using `conda` to install librosa, then all audio coding dependencies (except MP3) will be handled automatically.
```
faroit: I have to double check this, but the last time I checked, the conda-forge builds of libsndfile didn't came with flac support. See issue [here](https://github.com/conda-forge/libsndfile-feedstock/issues/2)

bmcfee:
> Did you test with different `audioread` backends? E.g. I'm thinking about the small differences in int->float conversion [when using libsndfile](http://blog.bjornroche.com/2009/12/int-float-int-its-jungle-out-there.html)

No, but then again, I never did before either, so this wouldn't necessarily reveal any regressions :rofl: 

What I did try was comparing a wav load on master (audioread) to a wav load in the branch, and it checked out fine.  At least, at the level of total scale; bitwise precision is probably out of reach for this kind of comparison anyway.

faroit:
> What I did try was comparing a wav load on master (audioread) to a wav load in the branch, and it checked out fine. At least, at the level of total scale; bitwise precision is probably out of reach for this kind of comparison anyway.

sure. This conversion difference can actually lead to a differences of up to 10^-5. I heard that some people use librosa in their production pipelines, so its just something that people might want to be aware of... what about a note/link in the advanced I/O docs?

bmcfee:
> what about a note/link in the advanced I/O docs?

I think the changelog would be a more appropriate place for this, but definitely.

README.md
```diff
@@ -63,15 +63,24 @@ librosa (0.x.x, /path/to/librosa)
...
-`librosa` uses `audioread` to load audio files.
+#### audioread and MP3 support
 
-To fuel `audioread` with more audio-decoding power (e. g. for reading MP3 files),
+To fuel `audioread` with more audio-decoding power (e.g., for reading MP3 files),
 you can either install *ffmpeg* or *GStreamer*.
 
+*Note that `audioread` needs at least one of the programs to work properly.*
+
 If you are using Anaconda, install *ffmpeg* by calling
```
faroit: actually on osx it also works without anything installed using the system libraries for decoding. So we might rephrase "at least one program"

jfsantos:
I'll give it a look and see if there are any potential issues later today. The dataloader I've been using for PyTorch uses soundfile so I have some experience with it.

README.md
bmcfee: Hmmm... that's deeply unfortunate.  I'll find a way to reword this to more accurately reflect the sad state of codec packaging.

README.md
bmcfee: Interesting -- I'll reword that to be platform-dependent.

bmcfee:
@faroit i looked into it, and it does seem like libsndfile on conda-forge builds with flac and vorbis.  Here's the [recipe](https://github.com/conda-forge/libsndfile-feedstock/blob/master/recipe/meta.yaml)

faroit:
> @faroit i looked into it, and it does seem like libsndfile on conda-forge builds with flac and vorbis. Here's the recipe

thats weird. `conda install -c conda-forge libsndfile` does get me

```
The following NEW packages will be INSTALLED:

  gettext            conda-forge/osx-64::gettext-0.19.8.1-h46ab8bc_1002
  libflac            conda-forge/osx-64::libflac-1.3.1-h0a44026_1002
  libiconv           conda-forge/osx-64::libiconv-1.15-h01d97ff_1005
  libogg             conda-forge/osx-64::libogg-1.3.2-h1de35cc_1001
  libsndfile         conda-forge/osx-64::libsndfile-1.0.28-h0a44026_1000
  libvorbis          conda-forge/osx-64::libvorbis-1.3.5-h1de35cc_1001
```

if I then run `sndfile-info test.flac`                                                                                                                                                   

```
Error : Not able to open input file test.flac.
File : test.flac
Length : 141009181
This version of libsndfile was compiled without FLAC support.

File contains data in an unimplemented format.
```

maybe @mgeier @hbredin or @thomasfillon know more?

jfsantos:
I just checked the code and it seems good to me.

Regarding the whole FLAC/Vorbis thing, @faroit's experiment gives me a different output on Linux (sndfile-info works on both FLAC and Vorbis files). Maybe their Mac build is broken?

bmcfee:
Thanks @jfsantos !

Re: codec dependencies, I think it's a bit upstream of us here, and I've reworded the documentation around it anyway.  Unless there are objections, I think this is good to merge?

stefan-balke:
___
*[docs/ioformats.rst, line 37 at r4](https://reviewable.io/reviews/librosa/librosa/847#-LaUWMJB-yBbxBj2rx2Q:-LaUWMJB-yBbxBj2rx2R:b-7t6jhp) ([raw file](https://github.com/librosa/librosa/blob/06a19b9e4db600169f73bea483e26db51ca6a958/docs/ioformats.rst#L37)):*
> ```reStructuredText
> -----------------
> 
> For large audio signals it could be benificial to not load the whole audio file
> ```

beneficial

bmcfee:
Thanks all @stefan-balke @faroit @jfsantos !

PR #849, opened 2019-03-13
remove one of the two softmask items in the doc

keunwoochoi:
#### Any other comments?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/849)


bmcfee:
Thanks, I never would have noticed that!

PR #852, opened 2019-03-20
[CR wanted] ICQT rewrite

### LINKED ISSUE #802: icqt outputs noise with unusual format
#### Description
icqt (inverse CQT) return loud noise with unusual format when I simply convert ".wav => CQT =>.wav".  
I cannot open output file with Windows10's Groove music app (iSTFT result can be opened), so I try to open with Audacity.  
Audacity enable opening the icqt output .wav file, but the result is loud noise.  

#### Steps/Code to Reproduce
Even example of official document return unintended results.  

```
import librosa

y, sr = librosa.load(librosa.util.example_audio_file(), duration=15)
C = librosa.cqt(y, sr=sr)
y_hat = librosa.icqt(C, sr=sr)

librosa.output.write_wav("./input.wav", y, sr, norm=True)
librosa.output.write_wav("./reconstructed.wav", y_hat, sr, norm=True)
```
basically same as [official sample code](https://librosa.github.io/librosa/generated/librosa.core.icqt.html#librosa.core.icqt)   

#### Expected Results
Above code is simply "wave => CQT => wave" conversion, so I expected to be almost all same wav file.  
Sound could be degraded only a little, but should not be a noise.

#### Actual Results
Loud noise. No music info.  
At the same time, SciPy warning displayed.  
```
FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`.    
In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.  
```
#### Versions
Windows-10-10.0.17134-SP0
Python 3.7.1 (v3.7.1:260ec2c36a, Oct 20 2018, 14:57:15) [MSC v.1915 64 bit (AMD64)]
NumPy 1.15.4
SciPy 1.1.0
librosa 0.6.2

bmcfee:
#### Reference Issue
Fixes #802 

This PR is a complete rewrite of the icqt implementation.  The new implementation is much simpler, and uses `istft` to perform each octave band's inverse transform, rather than populating the waveform manually.

It fixes the numerical stability issues mentioned in #802, as well as a number of other scaling issues.

The `amin` parameter has been deprecated, as it is no longer necessary.

A new `length=` parameter has been added for style compatibility with other inverse transforms.

A warning is now issued if the analysis parameters are not sufficient for reconstruction, namely when the shortest filter length is less than the hop length (non-overlapping filters).

#### Any other comments?

Here are some example inputs / outputs using a triangular-windowed sine sweep from C1 to C8.  All error estimates are restricted to the non-zero portion of the signal.

```
hop_length = 512
oversample = 1
filter_scale = 1
n_octaves = 7
scale=True
sparsity = 1e-3
```
image
RMSE: 0.343, MAE: 0.194
---
```
hop_length = 512
oversample = 5
filter_scale = 1
n_octaves = 7
scale=True
sparsity = 1e-3
```

image

RMSE: 0.018, MAE: 0.011
---
```
hop_length = 512
oversample = 9
filter_scale = 1
n_octaves = 7
scale=True
sparsity = 1e-3
```
image

RMSE: 0.004, MAE: 0.003
---
```
hop_length = 512
oversample = 7
filter_scale = 1
n_octaves = 9
scale=True
sparsity = 1e-3
```
image

RMSE: 0.007, MAE: 0.005
---
```
hop_length = 256
oversample = 5
filter_scale = 1
n_octaves = 7
scale=True
sparsity = 1e-3
```
image
RMSE: 0.007, MAE: 0.011
---

All results are identical with `scale=False`.

At this point, I'm happy with the implementation.  Additional eyeballs for CR would be appreciated, but it's a substantial CR job due to its interdependence with the CQT implementation.  I'd settle for a quick smell test, and a look over the documentation and warnings.  Otherwise I can merge once it passes CI.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/852)

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -165,7 +165,7 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
     # First thing, get the freqs of the top octave
     freqs = cqt_frequencies(n_bins, fmin,
-                            bins_per_octave=bins_per_octave)[-bins_per_octave:]
+                            bins_per_octave=bins_per_octave, tuning=tuning)[-bins_per_octave:]
 
     fmin_t = np.min(freqs)
     fmax_t = np.max(freqs)
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -670,7 +670,9 @@ def icqt(C, sr=22050, hop_length=512, fmin=None, bins_per_octave=12,
         if y is None:
             y = y_oct
         else:
-            y = audio.resample(y, 1, 2, scale=True, fix=False)
+            # Up-sample the previous buffer and add in the new one
+            # Scipy-resampling is fast here, since it's a power-of-two relation
+            y = audio.resample(y, 1, 2, scale=True, res_type='scipy', fix=False)
 
             y[:len(y_oct)] += y_oct
 
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -5,12 +5,11 @@ from __future__ import division
 
 import warnings
 import numpy as np
-from numba import jit
 
 from . import audio
 from .fft import get_fftlib
 from .time_frequency import cqt_frequencies, note_to_hz
-from .spectrum import stft
+from .spectrum import stft, istft
 from .pitch import estimate_tuning
 from .._cache import cache
 from .. import filters
@@ -517,23 +516,13 @@ def pseudo_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
 
 @cache(level=40)
-def icqt(C, sr=22050, hop_length=512, fmin=None,
-         bins_per_octave=12,
-         tuning=0.0,
-         filter_scale=1,
-         norm=1,
-         sparsity=0.01,
-         window='hann',
-         scale=True,
-         amin=1e-6):
+def icqt(C, sr=22050, hop_length=512, fmin=None, bins_per_octave=12,
+         tuning=0.0, filter_scale=1, norm=1, sparsity=0.01, window='hann',
+         scale=True, length=None, amin=util.Deprecated()):
     '''Compute the inverse constant-Q transform.
 
     Given a constant-Q transform representation `C` of an audio signal `y`,
     this function produces an approximation `y_hat`.
-
-    .. warning:: This implementation is unstable, and subject to change in
-                 future versions of librosa.  We recommend that its use be
-                 limited to sonification and diagnostic applications.
 
 
     Parameters
@@ -575,12 +564,15 @@ def icqt(C, sr=22050, hop_length=512, fmin=None,
         If `False`, do not scale the CQT. This is analogous to `norm=None`
         in FFT.
 
-    amin : float or None
-        When applying squared window normalization, sample positions with
-        coefficients below `amin` will left as is.
+    length : int > 0, optional
+        If provided, the output `y` is zero-padded or clipped to exactly
+        `length` samples.
 
-        If `None`, then `amin` is inferred as the smallest valid floating
-        point value.
+    Returns
+
+    amin : float or None [DEPRECATED]
+
+        .. note:: This parameter is deprecated in 0.7.0 and will be removed in 0.8.0.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


librosa/core/constantq.py
```diff
@@ -612,115 +604,80 @@ def icqt(C, sr=22050, hop_length=512, fmin=None,
...
-                                        fmin=fmin_t,
-                                        n_bins=bins_per_octave,
-                                        bins_per_octave=bins_per_octave,
-                                        filter_scale=filter_scale,
-                                        tuning=tuning,
-                                        norm=norm,
-                                        window=window,
-                                        pad_fft=True)
-    n_fft = basis.shape[1]
+    fmin_t = np.min(freqs[-bins_per_octave:])
```
lostanlen: not sure why you need to slice the array here? isn't `freqs` already `bins_per_octave` long from the line above?

librosa/core/constantq.py
```diff
@@ -612,115 +604,80 @@ def icqt(C, sr=22050, hop_length=512, fmin=None,
...
+                                                 bins_per_octave,
+                                                 tuning,
+                                                 filter_scale,
+                                                 norm,
+                                                 sparsity=sparsity,
+                                                 window=window)
 
-    n_trim = basis.shape[1] // 2
+    if min(lengths) < hop_length:
+        warnings.warn('Minimum CQT filter length={:.3f} is less than hop_length={}.\n'
```
lostanlen: I would phrase it the other way around: `hop length is greater than minimum CQT filter length`
because the most common fix is to lower the hop length.

librosa/core/constantq.py
```diff
@@ -612,115 +604,80 @@ def icqt(C, sr=22050, hop_length=512, fmin=None,
...
-                                           filter_scale=filter_scale,
-                                           tuning=tuning,
-                                           window=window)[:, np.newaxis]**0.5
+    # The basis gets renormalized by the effective window length above;
+    # This step undoes that
+    fft_basis = fft_basis.todense() * n_fft / lengths[:, np.newaxis]
 
-    y = None
+    # This step conjugate-transposes the filter
+    inv_basis = np.asanyarray(fft_basis).conj().T
```
lostanlen: `np.asanyarray(fft_basis).H`

librosa/core/constantq.py
```diff
@@ -612,115 +604,80 @@ def icqt(C, sr=22050, hop_length=512, fmin=None,
...
-
-            if y_oct is None:
-                y_oct = y_oct_i
-            else:
-                y_oct += y_oct_i
-
-        # Remove the effects of zero-padding
-        y_oct = y_oct[n_trim:-n_trim] * bscale[i]
 
+        # Apply length corrections
```
lostanlen: length correction makes it sound like it's changing the length.
consider replacing with energy conservation

librosa/core/constantq.py
```diff
@@ -612,115 +604,80 @@ def icqt(C, sr=22050, hop_length=512, fmin=None,
...
+        if scale:
+            C_oct = C_oct / np.sqrt(lengths[-C_oct.shape[0]:, np.newaxis])
+        else:
+            C_oct = C_oct / (lengths[-C_oct.shape[0]:, np.newaxis] * np.sqrt(2**octave))
+
+        # Inverse-project the basis for each octave
+        D_oct = inv_oct.dot(C_oct * n_fft)
```
lostanlen: inconvenient that the multiplicative factor for C_oct is broken down into multiple locations: above and here

librosa/core/constantq.py
```diff
@@ -612,115 +604,80 @@ def icqt(C, sr=22050, hop_length=512, fmin=None,
...
         slice_ = slice(-(octave+1) * bins_per_octave - 1,
                        -(octave) * bins_per_octave - 1)
 
-        # Project onto the basis
-        C_oct = C[slice_] / Cnorm[slice_]
-        basis_oct = basis[-C_oct.shape[0]:]
-
-        y_oct = None
+        # Slice this octave
+        C_oct = C[slice_]
```
lostanlen: not sure what's going on here. why is there only a single subscript if `C` is 2-D? is it something that is taken into account by the `slice` Python function? also, isn't `-C_oct.shape[0]` a constant (independent from `octave`)? why does it need to be estimated at runtime? in any case, this part deserves more comment than the laconic `Slice this octave` :)

librosa/core/constantq.py
```diff
@@ -612,115 +604,80 @@ def icqt(C, sr=22050, hop_length=512, fmin=None,
...
         else:
             # Up-sample the previous buffer and add in the new one
             # Scipy-resampling is fast here, since it's a power-of-two relation
-            y = audio.resample(y, 1, 2, scale=True, res_type='scipy') + y_oct
+            y = audio.resample(y, 1, 2, scale=True, res_type='scipy', fix=False)
+
+            y[:len(y_oct)] += y_oct
+
+    if length:
+        y = util.fix_length(y, length)
```
lostanlen: couldn't we speed up the code by having this happen at the start rather than the end? i.e. truncating the rows of `C` to its `n_cols = ceil(length/hop_length)` first columns (plus maybe some padding), then applying ICQT, and finally doing the waveform truncation to be sample-accurate in terms of length? just an idea

Issue #576, opened 2017-05-17
Add support for scipy polyphase resampling


bmcfee:
#### Description

Scipy 0.18 added a [polyphase resampler](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.signal.resample_poly.html).  We could add support for this to `core.resample`, eg:

```python
y_res = librosa.resample(y, sr_old, sr_new, res_type='polyphase')
```
#### Potential complications

1. This would up the version requirement for scipy from 0.13 to 0.18.  We could potentially make this soft.
2. We'd probably need to expose some new parameters for specifying the window.
3. We'd need a good way to determine the `up` and `down` parameters.  If the sampling rates are integer-valued, we can figure this out from the gcd.  However, the current API does not require integer sampling rates, and I'd like to keep that as flexible as possible.  Perhaps we could only require int-valued rates for this particular mode, and throw a warning / exception if fractional rates are provided.

dpwe:
FWIW, Matlab's resample also requires integer ratios.  I wrote code to find
ratios to get arbitrarily close to a resample ratio expressed as a float (I
was resampling audio by factors very close to 1 to get time alignment
between simultaneous recordings with unsynchronized sample clocks).  See
the first block (calculating the up/down factors P and Q) in:


https://labrosa.ee.columbia.edu/projects/skewview/rewrite_aligned_audio.m.html

  DAn.

bmcfee:
Every time...
![DAn did it](https://i.imgflip.com/1p7ar7.jpg)

lostanlen:
PR #663 (Savitzky-Golay) bumped the scipy requirement from 0.13 to 0.14. Do we still have to make the 0.18  requirement soft to close this issue?

The MATLAB code shared by @dpwe is useful for dealing with floating-point valued `sr_old` and/or `sr_new`

```
% Find p/q s.t. p/q ~= a -- approx 1 part in 1/(a-1)
p0 = floor(1/abs(skew-1));
%disp(['p0=',num2str(p0)]);
if p0 < 2^15
  % exhaustive search for pair of integers closest to desired rate
  p = p0:(2^15);
  q = round(p./skew);
  er = (skew - p./q); 
  [ee,xx] = min(abs(er));
  P = p(xx);
  Q = q(xx);
else
  % too close to unity
  P = 1;
  Q = 1;
end
```

[...]

PR #859, opened 2019-03-28
[CR requested] ICQT resampling mode and polyphase resampling

### LINKED ISSUE #857: Resampling mode in ICQT
#### Description
The ICQT implementation currently has a hard-coded `scipy` (FFT-based) resampling type, built in.  This gives the highest quality of the methods available, but also can be slow depending on signal duration.  We should expose the `res_type` parameter to the API so that users can select which mode they prefer.

bmcfee:
#### Reference Issue
Fixes #857 and #576 

This PR exposes the resampling mode as a parameter to `icqt`.

Additionally, it adds a wrapper for scipy's polyphase resampler.

#### Any other comments?

- polyphase resampling is only supported when the sampling rates are both integer-valued.  The up- and down-sampling ratios are calculated by greatest common divisor.  This is triggered by `res_type='polyphase'`.
- Because `scipy` now has multiple resamplers built in, saying `res_type='scipy'` is no longer informative.  We now support a more informative `res_type='fft'` (equivalent to the old `='scipy'`), though the old version still works.  This should lead to slightly more readable code going forward.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/859)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *.py[co]
 
+__pycache__
+
 # Packages
 *.egg
 *.egg-info
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -269,7 +269,7 @@ def resample(y, orig_sr, target_sr, res_type='kaiser_best', fix=True, scale=Fals
 
             To use a faster method, set `res_type='kaiser_fast'`.
 
-            To use `scipy.signal.resample`, set `res_type='fft'`.
+            To use `scipy.signal.resample`, set `res_type='fft'` or `res_type='scipy'`.
 
             To use `scipy.signal.resample_poly`, set `res_type='polyphase'`.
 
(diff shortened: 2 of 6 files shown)


bmcfee:
I suppose this raises a question of whether we want to expose res_type as a parameter to the forward CQT function as well.  This could be beneficial for speed-critical applications, where one might want to tolerate some loss of precision in favor of a faster transform.

It could get complicated though: the current implementation adapts the resampling mode based on the sampling rate and range of the current octave's filter bank.  If we let the user override this, one proposed API would be as follows:
  - res_type=None (proposed default) ==> retains the current adaptive resampling behavior (switching between "best" and "fast" modes of resampy)
  - res_type=anything else uses the provided mode for all octaves.

Thoughts?  @lostanlen ?

bmcfee:
Update: I went ahead and added the `res_type` API to cqt functions.  Speedup with polyphase resampling is appreciable, but so is the lack of accuracy.  On the other hand, it also gives a nice way to force high-quality downsampling everywhere (by saying `res_type='fft'`), so I think it's a good change overall.

And for good measure, I numba jitted the num_two_factors calculator.  There was no reason not to, but it hardly seemed worth making a separate PR for.

This PR is, I think, done and ready for CR.

bmcfee:
Unless someone wants to chime in on this (@stefan-balke @lostanlen ?), I feel okay with merging this on monday.

PR #863, opened 2019-04-15
resample type option added in pitch_shift()

tae898:
As for resampling, 'kaiser_best' can take twice as much time as 'kaiser_fast'. This can be a bottle neck in some applications where a lot of resampling calls are made. It'd be nice if this is made optional.

#### Any other comments?

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/863)

### DIFF
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -239,7 +239,7 @@ def time_stretch(y, rate):
     return y_stretch
 
 
-def pitch_shift(y, sr, n_steps, bins_per_octave=12, res_type='kaiser_fast'):
+def pitch_shift(y, sr, n_steps, bins_per_octave=12, res_type='kaiser_best'):
     '''Pitch-shift the waveform by `n_steps` half-steps.
 
 
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -260,6 +260,7 @@ def pitch_shift(y, sr, n_steps, bins_per_octave=12, res_type='kaiser_fast'):
     res_type : string
         Resample type.
         Possible options: 'kaiser_best', 'kaiser_fast', and 'scipy'.
+        By default, 'kaiser_best' is used. Refer to core.resample for details.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


bmcfee:
Thanks @tae898 for this!  Changes so far seem good, though I think it should also be applied to `time_stretch` for consistency of API.  Also, we should preserve the current default behavior (`kaiser_best`) so that there are no surprises going forward, and give a documentation example for how to do it quickly.  Can you take care of those requests?

There's also a bit of overlap (in scope) with (WIP) pr #857, where res_type modes have been expanded.  I suspect this one will merge before #857, so I'm happy to smooth over the differences over there.

librosa/effects.py
```diff
@@ -287,13 +290,15 @@ def pitch_shift(y, sr, n_steps, bins_per_octave=12):
     ...                                          bins_per_octave=24)
     '''
 
-    if bins_per_octave < 1 or not np.issubdtype(type(bins_per_octave), np.integer):
+    if bins_per_octave < 1 or not np.issubdtype(type(bins_per_octave),
```
lostanlen: not sure why git diff picks up these two lines? did anything change?

bmcfee: looks like OP's text editor automatically wrapped a line here.  NBD.

tae898: > not sure why git diff picks up these two lines? did anything change?

Oh yeah cuz I made a new line before np.integer to comply with the "80 char per line limit"

tae898:
> Thanks @tae898 for this! Changes so far seem good, though I think it should also be applied to `time_stretch` for consistency of API. Also, we should preserve the current default behavior (`kaiser_best`) so that there are no surprises going forward, and give a documentation example for how to do it quickly. Can you take care of those requests?
> 
> There's also a bit of overlap (in scope) with (WIP) pr #857, where res_type modes have been expanded. I suspect this one will merge before #857, so I'm happy to smooth over the differences over there.

Thanks for the feedback. Btw, what do you mean by that it should also be applied to 'time_stretch"? time_stretch() doesn't call core.resample, but only pitch_shift() does. So I don't think I can add, or there is no need to add, an optional 'res_type' argument to it?

bmcfee:
> Thanks for the feedback. Btw, what do you mean by that it should also be applied to 'time_stretch"? time_stretch() doesn't call core.resample, but only pitch_shift() does.

sorry, you're completely correct!  (I was thinking of a similar pair of functions in a separate project.)

librosa/effects.py
```diff
@@ -239,7 +239,7 @@ def time_stretch(y, rate):
     return y_stretch
 
 
-def pitch_shift(y, sr, n_steps, bins_per_octave=12):
+def pitch_shift(y, sr, n_steps, bins_per_octave=12, res_type='kaiser_fast'):
```
bmcfee: Docstring says `kaiser_best` but parameter still says `kaiser_fast`

tae898: oh yeah sorry what a stupid mistake. Fixed!

bmcfee:
LGTM, thanks again for implementing this! :shipit:

Issue #816, opened 2019-01-11
Allow for streaming per-channel energy normalization (PCEN)


daemon:
#### Description
AFAIK, the current implementation of PCEN (`librosa.core.pcen`) is non-streaming. This is an issue, since applications in, for example, speech recognition and keyword spotting depend on streaming computation of spectrogram features. My [PyTorch PCEN implementation](https://github.com/daemon/pytorch-pcen) supports streaming audio (and trainable weights), but it's probably slower on the CPU than LibROSA's implementation.

lostanlen:
Thanks @daemon for raising this issue and for sharing a PyTorch implementation of PCEN under MIT license! :)

I don't have any experience with streaming, but yes this would be a great addition. I guess that the way we close this issue depends on the decisions we end up taking re: #101.
I agree with @bmcfee that hooking into the initial filter state (zi) is the way to go for the IIR filter. Very neat idea 👍

PR #864, opened 2019-04-16
[CR needed] fixed #851 and #816 to add initialization and streaming support to pcen

### LINKED ISSUE #851: PCEN should not initialize with silence
#### Description
Adapted from https://github.com/librosa/librosa/issues/615#issuecomment-473009124, initially raised by @4p0pt0Z 

When using the current implementation of the PCEN, artifacts appear due to the filter initialization. Currently, the filter is created with
`S_smooth = scipy.signal.lfilter([b], [1, b - 1], ref, axis=axis)`

When the initial conditions for the filter are not specified, initial rest is assumed.
See documentation of `zi` at this address: https://docs.scipy.org/doc/scipy-1.2.1/reference/generated/scipy.signal.lfilter.html

The initialization to zero is equivalent to assuming that the signal is preceded by an infinitely long silence. This is not realistic and causes numerical artifacts in the earliest timestamps of the PCEN-gram, especially in the most active bands.

#### Expected Results vs. Actual Results

Thanks to @f0k and @bmcfee, we already have a plan for a workaround:
* expose the `zi` as a kwarg parameter in the prototype of `pcen`
* `zi` defaults to `None`, in which case it gets automatically set as in steady-state.

This is how we plan to do it: https://docs.scipy.org/doc/scipy-1.2.1/reference/generated/scipy.signal.lfilter_zi.html#scipy.signal.lfilter_zi

Top: current results (with zero initialization of `zi`). Bottom: expected results (with: steady-state initialization of `zi`).

54450887-dbe49c80-4727-11e9-9777-e39f1e059aa8 (1)

#### Versions
```
librosa 0.6.3
scipy 1.2.1
```

bmcfee:
#### Reference Issue
Fixes #851 and #816 

A bug-fix to initialize (by default) the lfilter `zi` to steady-state.

New features: `return_zf` flag to optionally return the final state of the filter, for use as `zi` in a subsequent call to `pcen`.  Also exposing `zi` as a parameter which overrides the steady-state initialization above.

#### Any other comments?

There is a basic unit test that checks the results of this API on a split input against processing the entire input at once, and things line up.  I think that's sufficient for testing here.

I'm a little unsatisfied by some shape hacking I had to do to get the `lfilter_zi` output to line up with multi-dimensional input, and I suspect it won't work for higher-dimensional input (eg stereo tf representations).  If there's a nice way to reshape this automagically according to the input shape, I'm all ears.

Other than that, I think this is good to go.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/864)

### DIFF
--- a/docs/examples/plot_pcen_stream.py
+++ b/docs/examples/plot_pcen_stream.py
@@ -22,7 +22,7 @@ import matplotlib.pyplot as plt
 import soundfile as sf
 
 import librosa as librosa
-import librosa.display as isplay
+import librosa.display as display
 
 
 ######################################################################
--- a/docs/examples/plot_pcen_stream.py
+++ b/docs/examples/plot_pcen_stream.py
@@ -1,0 +1,130 @@
+# coding: utf-8
+# Code source: Brian McFee
+# License: ISC
+"""
+==============
+PCEN Streaming
+==============
+
+This notebook demonstrates how to use `soundfile` streaming IO with `librosa.pcen` 
+to do dynamic per-channel energy normalization on a spectrogram.
+
+This is useful when processing long audio files that are too large to load all at
+once, or when streaming data from a recording device.
+"""
+
+##################################################
+# We'll need numpy and matplotlib for this example
+from __future__ import print_function
+import numpy as np
+import matplotlib.pyplot as plt
+
+import soundfile as sf
+
+import librosa as librosa
+import librosa.display as isplay
+
+
+######################################################################
+# First, we'll start with an audio file that we want to stream
+filename = librosa.util.example_audio_file()
+
+# We can stream in blocks using soundfile
+sr = sf.info(filename).samplerate
+
+
+#####################################################################
+# Next, we'll set up the block reader to work on short segments of 
+# audio at a time.
+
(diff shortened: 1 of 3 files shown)


bmcfee:
Ok, I've fixed up the shape issues and things seem to now work as expected with multidimensional inputs.  This is ready for CR now.

bmcfee:
I'm going to add an advanced example showing how to use the streaming interface here.  Will ping back when ready.

bmcfee:
Example notebook is done.  Final output figure generated by the example looks like:
image

bmcfee:
I think this one's solid and ready to :shipit:, but it would be good to get a second pair of eyeballs on the gallery example just to make sure it's all pedagogically clear.  For convenience, a gist render of the generated notebook is available [here](https://gist.github.com/48ccf2426e09c0b2403b117497ef5c36).

lostanlen:
This is amazing.
I just installed this fork on a fresh miniconda install of librosa. I can reproduce the example file.

On my machine, the elapsed time of both methods are within 10% of each other.
The comments and documentation look great!
Thank you so much @bmcfee 🎉

bmcfee:
Thanks @lostanlen !

PR #868, opened 2019-04-23
fix: dtw(C=C,subseq=True) doesn't work

viktorAndreevichMorozov:
#### Reference Issue
Fixes https://github.com/librosa/librosa/issues/869

Just get dimensions from C ndarray (which is always computed anyway), not from X and Y (thous are optional).

### DIFF
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -226,7 +226,7 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
         wp = np.asarray(wp, dtype=int)
 
         # since we transposed in the beginning, we have to adjust the index pairs back
-        if subseq and (X.shape[1] > Y.shape[1]):
+        if subseq and (C.shape[0] > C.shape[1]):
             wp = np.fliplr(wp)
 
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks for noting this and making the PR.  Please add a bit more description of what the problem was though, since this did not have an existing reference issue.

Also it appears that the proposed change is breaking tests, so please look into that before we do a proper code review.

viktorAndreevichMorozov:
Hello. It is the fix for issue https://github.com/librosa/librosa/issues/869 . With the last commit it passes all tests.

bmcfee:
Thanks!  This looks fine to me, but it would be great if you could add a test case to cover the previously unchecked behavior.  @stefan-balke care to chime in on this?

stefan-balke:
Good catch. But a test would be nice. Provide a ground-truth `C` and let it run. Should give the same result as with providing `X` and `Y`.

viktorAndreevichMorozov:
I wrote the test.

bmcfee:
Perfect, this looks good to me.  Thanks again @viktorAndreevichMorozov !

PR #872, opened 2019-05-02
Stream generator

### LINKED ISSUE #772: On processing long audio files / signals
`<dislaimer>` There are two levels to this possible enhancement discussion, and since this could hypothetically result in implementation changes, figured it made sense here rather than google group – if that's wrong, happy to redirect! `</disclaimer>`

To date, we've gotten pretty far with loading full signals into memory and transforming them without issue, because it's 2018 and everyone has GBs of headroom to play with. Music files, generally, aren't that long, and that makes things easy. 

I'm curious, though, what the preferred course of action would be for, say, hours of audio, when you start to run into resource limitations. There may be others, but I see two kinds of memory issues that crop up:
- the input signal / output signal doesn't fit in memory
- some _intermediary_ representation doesn't fit in memory, but the input & output do, e.g. a matrix product before reduction

There are some simple, albeit unsatisfying, options:
- slice a long recording up into smaller bits at the filesystem level
- process the file in chunks in Python scope

While both are simple in principle, they introduce a layer of bookkeeping I (as a user) would love to not have to think about. I see a few ways to engineer around this, and there are no doubt other (better?) ones:
- A file in / file out API
- buffered iterators and reducers

I suppose a preliminary question is, "is this even `librosa`'s problem?" I'm asking here, so I think it at least could be, but would be keen to kick around thoughts? (and also help implement a solution, if a good one arises).

bmcfee:
#### Reference Issue
Fixes #772

This PR provides a stream generator `librosa.stream()` that can be used to process large audio files in small pieces.

This is done by wrapping `soundfile.blocks` in some helper functions that smooth over some of the details in setting up the framing parameters.  Basically it codifies the example previously introduced in #864 into a core function.

Example usage for mel spectrogram construction over 256 frames at a time:
```python
>>> blocks, sr = librosa.stream(filename, block_length=256)
>>> for y_block in blocks:
...    M_block = librosa.feature.melspectrogram(y_block, sr=sr, center=False)
```
If the `M_blocks` produced above are stitched together, the result will be numerically equivalent to
```python
>>> y, sr = librosa.load(filename, sr=None)
>>> M = librosa.feature.melspectrogram(y=y, sr=sr, center=False)
```

Note that data is streamed at the native sampling rate, so the default `frame_length` and `hop_length` may not be consistent with the rest of the API.

#### Any other comments?

The documentation here is pretty thorough on proper usage, but I'm open to suggestions on changes for the API, specifically around the frame length / hop length parameters.  

Once that's settled, I'll put in some unit tests.

RFC @ejhumphrey @lostanlen @faroit @keunwoochoi

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -300,13 +300,11 @@ def stream(path, block_length, frame_length, hop_length,
     dtype : numeric type
         data type of audio buffers to be produced
 
-    Returns
-    -------
-    stream : generator
-        A generator which produces blocks of audio.
-
-    sr : number > 0
-        The sampling rate of the audio
+    Yields
+    ------
+    y : np.ndarray
+        An audio buffer of (at most) 
+        `block_length * (hop_length-1) + frame_length` samples.
 
     See Also
     --------
@@ -320,20 +318,24 @@ def stream(path, block_length, frame_length, hop_length,
     at a time.  Note that streaming operation requires left-aligned
     frames, so we must set `center=False` to avoid padding artifacts.
 
-    >>> stream, sr = librosa.stream(librosa.util.example_audio_file(),
-    ...                             block_length=256,
-    ...                             frame_length=4096,
-    ...                             hop_length=1024)
+    >>> filename = librosa.util.example_audio_file()
+    >>> sr = librosa.get_samplerate(filename)
+    >>> stream librosa.stream(filename,
+    ...                       block_length=256,
+    ...                       frame_length=4096,
+    ...                       hop_length=1024)
     >>> for y_block in stream:
     ...     D_block = librosa.stft(y_block, center=False)
 
     Or compute a mel spectrogram over a stream, using a shorter frame
     and non-overlapping windows
 
-    >>> stream, sr = librosa.stream(librosa.util.example_audio_file(),
-    ...                             block_length=256,
-    ...                             frame_length=2048,
-    ...                             hop_length=2048)
+    >>> filename = librosa.util.example_audio_file()
+    >>> sr = librosa.get_samplerate(filename)
+    >>> stream = librosa.stream(filename,
+    ...                         block_length=256,
+    ...                         frame_length=2048,
+    ...                         hop_length=2048)
     >>> for y_block in stream:
     ...     m_block = librosa.feature.melspectrogram(y_block, sr=sr,
     ...                                              n_fft=2048,
@@ -353,16 +355,6 @@ def stream(path, block_length, frame_length, hop_length,
     sr = sf.info(path).samplerate
 
     # Construct the stream
-    block_stream = __stream(path, sr, block_length, frame_length, hop_length,
-                            mono, offset, duration, fill_value, dtype)
-
-    return block_stream, sr
-
-
-def __stream(path, sr, block_length, frame_length, hop_length,
-             mono, offset, duration, fill_value, dtype):
-    '''Private function for wrapping sf.blocks in a librosa interface.'''
-
     if offset:
         start = int(offset * sr)
     else:
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -18,8 +18,9 @@ from .._cache import cache
 from .. import util
 from ..util.exceptions import ParameterError
 
-__all__ = ['load', 'to_mono', 'resample', 'get_duration', 'autocorrelate',
-           'lpc', 'zero_crossings', 'clicks', 'tone', 'chirp']
+__all__ = ['load', 'stream', 'to_mono', 'resample', 'get_duration',
+           'autocorrelate', 'lpc', 'zero_crossings',
+           'clicks', 'tone', 'chirp']
 
 # Resampling bandwidths as percentage of Nyquist
 BW_BEST = resampy.filters.get_filter('kaiser_best')[2]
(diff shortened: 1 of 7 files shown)


faroit:
@bmcfee love this idea. Will check the api later this week.

lostanlen:
the bioacoustician in me loves this already

bmcfee:
Some questions to keep in mind when considering the API:

- Should we expose the `overlap` parameter?  Right now it's inferred as `frame_length - hop_length`, so that the frames at the boundaries of blocks behave consistently with batch analysis, but maybe there's a use case for allowing the user to override that?  OTOH, a user of the current API could just as easily pull out a block of one "frame" of some arbitrary length, and then tune the "hop_length" accordingly, eg if they don't want overlap between blocks for some reason.
- The frame parameters (`block_length`, `frame_length`, `hop_length`) are all specified in samples, but the sampling rate may be unknown to the user in advance.  This leads to an awkward chicken-and-egg problem, since the parameters cannot be changed after the stream is constructed.  Some potential solutions here:
    - Leave the parameters as is, and make the user figure out the sampling rate before loading the stream.  We can provide a helper for this, analogous to `get_duration` operating on filenames.
    - Specify the parameters in natural units (unlike everywhere else in the code), and convert accordingly
    - Have the parameters be `None` by default, but then infer the values from the given sampling rate to be (temporally) consistent with the rest of the API.  So if `frame_length=None`, and the native sampling rate is 22050, it gives 2048; if it's 44100, `frame_length` defaults to 4096, and so on.
- Do we want to expose a parameter to have the buffer operate in-place (using `sf.blocks(..., out=)`)?  This can cut down on memory usage, but might have some unexpected consequences for unaware users.

bmcfee:
image

bmcfee:
I took the liberty of adding unit tests, since I think the core of the API here is pretty solid, and the remainders are questions of scope.

The tests are a bit strange due to the need for overlapping blocks.  The simplest way to test it was to frame the blocks, join the frames, and then compared to a framed version of the full audio.  This required downcasting to mono and ignoring any trailing `fill_value`, but I'm okay with that.

librosa/core/audio.py
```diff
@@ -208,6 +210,167 @@ def __audioread_load(path, offset, duration, dtype):
     return y, sr_native
 
 
+def stream(path, block_length, frame_length=2048, hop_length=512,
+           mono=True, offset=0.0, duration=None, fill_value=None,
+           dtype=np.float32):
+    '''Stream audio in fixed-length buffers.
```
ejhumphrey: semantic nit, but maybe important ... should `block_length` be `buffer_length`? I was reading through to see what block_length controlled, figured out it was a frame buffer, and then saw the doc string header (here) called it fixed-length buffer, which grok'ed quickly.

ejhumphrey: this also relates to some q's i've had / thrown around regarding buffering (stream-buffering) time-frequency matrices, where `librosa.util.frame` doesn't operate on ndarrays.rank > 1, but is _also_ a stream / buffer operation 🤔 ... I don't know what the consistent concept is here, but it feels like there's something to converge on?

bmcfee: I would interpret "buffer length" as the number of samples, which is not exactly right here.

I think a better alternative might be `n_frames`?

bmcfee: Aside: the reason that I didn't go with `n_frames` to begin with is that the output is not indexed by frames.  I don't think that it should be (for both technical and aesthetic reasons), so I think it's best to stick to block_length.

Of course, the documentation should be expanded to make it totally clear what a "block" means in this context, and what sort of results the user should expect.

ejhumphrey: the output is _not_ indexed by frames? as a noob, i would expect something like the following:

```
>>> stream, sr = librosa.stream(librosa.util.example_audio_file(), 256)
>>> set([block.shape for block in stream])
{(256, 2048)}
```
librosa is feature-major, right? or is that not true for audio? MFCCs come back `(20, time)` ... if nothing else, a shape tuple in the docstring will help dummies like me

tangentially related, does it make sense that `frame_length` and `hop_length` have default values? had to scroll back to infer my shape example

tests/test_core.py
```diff
@@ -624,7 +624,7 @@ def __test(signal, order, true_coeffs, est_coeffs):
         assert np.allclose(test_coeffs, est_coeffs)
 
     for infile in files(os.path.join('tests', 'data', 'core-lpcburg-*.mat')):
-        test_data = scipy.io.loadmat(infile, squeeze_me=True);
```
ejhumphrey: 😮

ejhumphrey:
overall this looks 👌🔥!! i think i'm struggling a bit when are things time-major or not, and some semantics, but that could be due to my own waning familiarity with librosa than anything fundamental here. that said, perhaps my faux-fresh perspective can be of some use 😛

librosa/core/audio.py
keunwoochoi: Just to clarify, `block_length` is not `n_frames` because..? Is it because interleaved n-channel audio files would have `block_length = n_frames * n_channel`? When I skimmed the code and docstrings I probably thought the same as @ejhumphrey.

But for whatever reason, seems like it's totally fair (or even better) to a new concept (`block`). Probably the docstring could start with `"""0. What is Block?:  .."""`.

tests/test_core.py
bmcfee: The dangling ; was bugging me :D

librosa/core/audio.py
bmcfee: > librosa is feature-major, right? or is that not true for audio?

The convention is that time(-like dimensions) are trailing.  So for audio, it's (samples,) or (channels, samples), and for features it's (features, frames).


> tangentially related, does it make sense that `frame_length` and `hop_length` have default values? had to scroll back to infer my shape example

That's a great question.  In general, things should be consistent across the API assuming a default sampling rate.  Of course, that assumption doesn't apply here, so there's a good case for making these required parameters.



> the output is _not_ indexed by frames?

No, the output is a slice of audio.  I get that this is a little counter-intuitive from a parameterization perspective, but it makes translating current code into the streaming setting much easier (check the examples in the docstring).



> Just to clarify, `block_length` is not `n_frames` because..? Is it because interleaved n-channel audio files would have `block_length = n_frames * n_channel`?

The way I see it, it's because the output is not indexed by frame.  block_length can be thought of as a signal duration measured in frame units.



> But for whatever reason, seems like it's totally fair (or even better) to a new concept (`block`). Probably the docstring could start with `"""0. What is Block?: .."""`.

Yes, I agree with that completely.

bmcfee:
Thanks @ejhumphrey and @keunwoochoi for the feedback!

Does it help if I just say that librosa is always time-minor?  (Only half-kidding there; it's a helpful way for me to think of it at least.  And if it's really confusing, think back to the theano days and how we used to do convnets...)

bmcfee:
To summarize the discussion above, I think the following modifications are in order:

1. Remove default values from `frame_length` and `hop_length`; make them positional arguments.
2. Expand documentation to clearly explain block, frame, and hop relationship.
3. Note in documentation that blocks will be slices of (overlapping) audio, not framed buffers.

Seem reasonable?

keunwoochoi:
> time-minor

or time-last? (from `channel-first` in Keras)

And yes, sounds good. I'm still not confident about Block, I'll not try to figure it out to keep my dummy's criteria for the doc :)

bmcfee:
I'm updating the block-reading section of the docs now, and wondering if it even makes sense to return the sample rate from `stream`, of if that would lead to weird anti-patterns.  If we require people to get the sample rate in advance (ie to set the frame parameters), then there's no need for returning it here.  And in that case, we can simplify `stream` to be a generator directly, rather than returning a generator.

So instead of having

```python
blocks, sr = librosa.stream('filename', [some parameters])

for y in blocks:
    # process y, sr
```
we would have
```python
sr = librosa.get_samplerate('filename')
for y in librosa.stream('filename', [some parameters]):
    # process y, sr
```
The latter seems cleaner to me.  What do y'all think?

lostanlen:
Amazing.
Is it possible that the file will be corrupted after a certain number of blocks? If so, is there a syntax to `catch` that?

bmcfee:
> Is it possible that the file will be corrupted after a certain number of blocks? If so, is there a syntax to `catch` that?

Do you have something specific in mind?  In general, I think that might be out of scope / should be handled at the codec level.

bmcfee:
One more aside here: I think it might be useful to have unit converters for blocks.  As it currently stands, it's totally not obvious how to get the time / sample / frame index or indices for a block.  What might be handy is something like:

```python
for n, block in enumerate(librosa.stream(...)):
    time = librosa.block_to_time(n, [stream parameters])
    E = np.mean(librosa.feature.rms(block))
    print('time={:.3g}: E={:.2g}'.format(time, E))
```
which would print the average RMS over a block starting at `time` seconds. Similarly for `block_to_frame` and `block_to_samples`.  I'm thinking this would just give the frame/sample/time of the left edge of the provided block index (or indices).  And of course, the reverse unit conversions would be provided as well.  The API gets a little dodgy here because blocks are implicitly left-aligned, unlike everything else.  I'm not sure how we want to handle that yet, so thoughts are very welcome here.

bmcfee:
Ok, block unit converters are in.  I think I'm happy with it at this point, and it's ready for a final round of CR.  Thanks again to everyone who chimed in with feedback!

keunwoochoi:
> I can't think of a compelling use case for the reverse conversion anyway

+1.

lostanlen:
@bmcfee i love this new `blocks_to_*` collection of utility functions. These are really handy and will save an inordinate number of headaches and/or off-by-one errors.


> Do you have something specific in mind? In general, I think that might be out of scope / should be handled at the codec level.

No, i don't have anything specific in mind. I never wrote a program for real-time audio analysis, but would be very interested in doing so for the future, especially for doing bioacoustics with a small memory footprint (i.e. not saving the audio to disk, but processing it by buffers). @jongwook did things like that with github.com/marl/crepe, not sure how.

I suppose that it would rely on some external library like `pyAudio` (http://people.csail.mit.edu/hubert/pyaudio/), which itself is a wrapper around `PortAudio`.
My question is: assuming i'm using one of these solutions to link software and hardware, how can i have callbacks at the software level that would inform the programmer/user about acquisition failures?
In other words, could it be that `librosa.stream` would be able to read from a `pyaudio.PyAudio` object (via `stream.read(block_size)`) in addition to reading from disk?
Having a unified API for the two cases would make the maintainment of tools like BirdVoxDetect extremely convenient.

bmcfee:
@lostanlen yeah -- there's a small distinction here between "streaming" (working piece-wise on audio fragments) and "real-time" (pulling data as it comes in).  In principle, the `stream` interface could be extended to support portaudio, and this was indeed the original plan that I alluded to in #772, but as you mention, it raises some other complexities that don't come up in large-file streaming (like dropped frames and callbacks).

This is all to say that I think handling realtime is a bit out of scope for this PR.

bmcfee:
@keunwoochoi @ejhumphrey since y'all seem to have spent the most time already looking through this, do you care for a second pass?  I think it's good to merge, but an extra set of :eyes: would be :100:.

ejhumphrey:
lgtm, the documentation around blocks feels super clear now, and "blocks" as first class concept reads well with the utils. I can't tell how much of this clarity is because I just understand things now, but this looks solid to the extent i can be useful.

👏 👏 👏 well done gg

keunwoochoi:
Looks pretty good for me too and way better than most of the docs in general. That said, when I was reading the docstrings of `stream()`,
```
*blocks* of audio spanning a fixed number of frames at a specified frame length and hop length.
```
I wasn't still 100% sure what it meant because, though I doubted, I wasn't sure whether the definition of *frame* follows the rule in PCM world (https://larsimmisch.github.io/pyalsaaudio/terminology.html for example). No it doesn't, which became obvious later in the docstring, but not really at that point. Probably we could provide even clearer definitions of things, e.g., 
```
* samples: each amplitude at index=n of a discretized audio signal
* frames: contiguous samples (e.g., 1024 samples). For analysis purpose, frames often overlap for e.g., 75%.
* blocks: contiguous frames. 

```

bmcfee:
> Probably we could provide even clearer definitions of things, e.g.,

I think the [glossary](https://librosa.github.io/librosa/glossary.html) already covers this?  Though it's probably due for a general update.  Otherwise, the use of `frame` here is internally consistent, which I think is the best we can hope for.

keunwoochoi:
> frame here is internally consistent

That's true, maybe we can add a link to the glossary here? (+I guess you'd like to add `Blocks` there too). I think I became less confident due to the introduction of a new word. Well, at the same time I was prob overthinking. Actually, the meaning of `Block` is exactly what I guessed when I saw this for the first time like a week ago. So.. overall, great!

bmcfee:
Ok, thanks @keunwoochoi and @ejhumphrey !  I think I'll revise the current doc issue(s) to improve the glossary (#874 ), but otherwise merge this.

PR #884, opened 2019-05-15
Minor fix on string 59

kant:
#### Reference Issue

There is no related issue. 

#### What does this implement/fix? Explain your changes.
There is a typo on string 59. Plus a question: on string 94: `Windows: download binaries from the website`(...) the URL is https://gstreamer.freedesktop.org/ ?

#### Any other comments?
There is no break on code, so, no collateral damage at sight.

PR #885, opened 2019-05-15
[ci skip] Added final URL on string 94

kant:
#### Reference Issue
There is no related issue

Added final URL on string 94

#### Any other comments?
No.


bmcfee:
Thanks @kant -- I'm wondering why you didn't just add this to the existing PR #884 ?

PR #890, opened 2019-05-28
Ifgram window differential

### LINKED ISSUE #888: Window derivative in ifgram
#### Description

The window derivative `d_window` that is calculated in `ifgram()` is appropriate for the default Hann window, but not for other analysis windows.

#### Steps/Code to Reproduce

https://github.com/librosa/librosa/blob/65cbc8291cc265b93236b871727d2a4ec9cf3d06/librosa/core/spectrum.py#L469-L472

These lines in `librosa.core.ifgram()` create an array with the window derivative. They don't use the `window` argument to `ifgram()` to calculate the derivative, but they do produce the (negative) window derivative when the default Hann window is used:

```
import numpy as np

n_fft = 1024

freq_angular = np.linspace(0, 2 * np.pi, n_fft, endpoint=False) 
d_window = np.sin(-freq_angular) * np.pi / n_fft

np.allclose(d_window, -np.gradient(np.hanning(n_fft)), atol=2e-5)
# True
```

But the calculation for `d_window` doesn't lead to the derivative for other windows:

```
np.allclose(d_window, -np.gradient(np.blackman(n_fft)), atol=2e-5)
# False

x = np.arange(n_fft)

plt.plot(x, d_window, c="black")
plt.scatter(x, -np.gradient(np.hanning(n_fft)), s=0.1, c="red")
plt.scatter(x, -np.gradient(np.blackman(n_fft)), s=0.1, c="blue")
plt.show()
```

`d_window` is used for `diff_stft` on these lines:

https://github.com/librosa/librosa/blob/65cbc8291cc265b93236b871727d2a4ec9cf3d06/librosa/core/spectrum.py#L479-L481

This is described in section 2 of the paper referenced in the docstring (Abe, Kobayashi, & Imai 1995), or equation 5.20 of Flandrin, Auger, & Chassande-Mottin (2002).

I might be misunderstanding, but I think that `d_window` should be using `np.gradient()` or another numerical derivative of the `window` argument. Flandrin et al. have an alternative calculation with MATLAB code for the window derivative in section 5.3.1.3 that might be better at the edges than `np.gradient()`.

#### Versions
```
Python 3.6.5
NumPy 1.14.5
SciPy 1.1.0
librosa 0.6.1
```

Abe, T., Kobayashi, T., & Imai, S. (1995). Harmonics tracking and pitch extraction based on instantaneous frequency. 1995 International Conference on Acoustics, Speech, and Signal Processing, 1, 756–759. https://doi.org/10.1109/ICASSP.1995.479804

Flandrin, P., Auger, F., & Chassande-Mottin, E. (2002). Time-Frequency reassignment: from principles to algorithms. In Applications in Time-Frequency Signal Processing (Vol. 10, pp. 179–204). CRC Press.

bmcfee:
#### Reference Issue
Fixes #888 by computing window derivatives directly

#### Any other comments?

Adds `util.cyclic_gradient` too.


librosa/util/utils.py
```diff
@@ -1641,3 +1642,66 @@ def fill_off_diagonal(x, radius, value=0):
...
+    >>> plt.ylim([-0.025, 0.025])
+    >>> plt.show()
+    '''
+    padding = [(0, 0)] * data.ndim
+    padding[axis] = (edge_order, edge_order)
+    data_pad = np.pad(data, padding, mode='wrap')
+
+    grad = np.gradient(data_pad, edge_order=edge_order, axis=axis)
+
+    slices = [slice(None)] * data.ndim
```
lostanlen: would be good to say that these three lines correspond to unpadding, and explain a little bit

librosa/util/utils.py
```diff
@@ -1641,3 +1642,66 @@ def fill_off_diagonal(x, radius, value=0):
...
+    ...          alpha=0.35)
+    >>> plt.plot(x, cyclic_grad, label='cyclic_gradient')
+    >>> plt.plot(x, grad, label='np.gradient', linestyle=':')
+    >>> plt.legend()
+    >>> # Zoom into the first part of the sequence
+    >>> plt.xlim([0, np.pi/16])
+    >>> plt.ylim([-0.025, 0.025])
+    >>> plt.show()
+    '''
+    padding = [(0, 0)] * data.ndim
```
lostanlen: would be good to say that these three lines correspond to padding, and explain a little bit. in particular, it's important to explain that the number of padding samples is equal to edge_order on both sides of the input.

librosa/core/spectrum.py
```diff
@@ -469,7 +469,7 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
     # Window for discrete differentiation
     freq_angular = np.linspace(0, 2 * np.pi, n_fft, endpoint=False)
```
lostanlen: we could move this line down to l 493

librosa/core/spectrum.py
```diff
@@ -469,7 +469,7 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
     # Window for discrete differentiation
```
lostanlen: maybe comment on the importance of cyclic differentiation? or even provide a ref to #888

PR #892, opened 2019-06-01
stabler exponentiation in PCEN + logarithmic limit case

### LINKED ISSUE #891: PCEN range compression is unstable for power -> 0
#### Description
This came up in a private conversation with @bmcfee when developing PCEN (#615), and resurfaced this week during an email discussion with Richard F. Lyon.
The line that computes dynamic range compression in PCEN is numerically unstable for low values of the `power` (i.e. `r`) argument.

https://github.com/librosa/librosa/blob/754fc3c51b96b9d0e5ce5cdec093798464c08612/librosa/core/spectrum.py#L1566

This is all the more concerning that, although RFL's paper recommends `power=0.5`, we've seen lower and lower values of `power` being useful in recent months. I use `power=0.25` for BirdVox, Mark Cartwright uses `power=0.125` for SONYC, and i have even proposed a generalization of PCEN at the limit case `power=0`, which has been surprisingly effective and interpretable for bioacoustic event detection.

After exchanging with RFL and BMcF, the consensus is that we should replace the pow-based exponentiation by an expm1-based exponentiation.

#### Steps/Code to Reproduce
https://gist.github.com/lostanlen/a3c52bb9ab4dd58ea40d5ea2c32871dd

#### Expected vs. Actual Results

librosa-890_a

librosa-890_b

#### Versions
```
Darwin-18.5.0-x86_64-i386-64bit
Python 3.7.3 | packaged by conda-forge | (default, Mar 27 2019, 15:43:19) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
NumPy 1.15.4
SciPy 1.2.1
librosa 0.6.3
```

lostanlen:
#### Reference Issue
Fixes #891, and does more :)

* `expm1`-based dynamic range compression in PCEN
* `bias==0` has to be special-cased to avoid division by zero in `expm1` exponentiation
* i also implemented `power==0` which turned out to be really useful and interpretable in my WASPAA paper
* multiplied by (2**31) in the docstring as per RFL's original code
* cited our IEEE SPL paper on the rationale behind the `time_constant`
* some comments and formatting in the pcen function

#### Any other comments?
I realize that there's a lot going on in this PR. I'm happy to break it into smaller PR's if that makes a clearer git log.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1565,11 +1565,15 @@ def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
         zi = np.empty(shape)
         zi[:] = scipy.signal.lfilter_zi([b], [1, b - 1])[:]
 
+    # Temporal integration
     S_smooth, zf = scipy.signal.lfilter([b], [1, b - 1], ref, zi=zi,
                                         axis=axis)
 
+    # Adaptive gain control
     # Working in log-space gives us some stability, and a slight speedup
     smooth = np.exp(-gain * (np.log(eps) + np.log1p(S_smooth / eps)))
+
+    # Dynamic range compression
     if power == 0:
         S_out = np.log1p(S*smooth)
     elif bias == 0:
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1471,11 +1471,12 @@ def pcen(S, sr=22050, hop_length=512, gain=0.98, bias=2, power=0.5,
     >>> y, sr = librosa.load(librosa.util.example_audio_file(),
     ...                      offset=10, duration=10)
 
-    >>> # We'll use power=1 to get a magnitude spectrum
-    >>> # instead of a power spectrum
+    >>> # We recommend scaling y to the range [-2**31, 2**31[ before applying
+    >>> # PCEN's default parameters. Furthermore, we use power=1 to get a
+    >>> # magnitude spectrum instead of a power spectrum.
     >>> S = librosa.feature.melspectrogram(y, sr=sr, power=1)
     >>> log_S = librosa.amplitude_to_db(S, ref=np.max)
-    >>> pcen_S = librosa.pcen(S)
+    >>> pcen_S = librosa.pcen(S * (2**31))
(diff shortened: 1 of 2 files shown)


bmcfee:
Hang on -- can you explain a bit more (here) about the 2**31 scaling?

bmcfee:
Following up on the changes to the docstring: is the 2**31 thing a general recommendation, or is this recommended when trying to reproduce the original publication?

lostanlen:
it's for reproducing the original publication. the values of delta, alpha, and r that Wang et al recommend correspond to a waveform that takes int32 values.
they don't specify exactly how they scale the norms of their filters (Littlewood-Paley condition for energy conservation), but if anything, the range [-1, 1] associated with librosa's default of 128 filters peaking at 1 is probably too small for the choice of alpha (0.98).
but i admit that it's sketchy and not very clear-cut

bmcfee:
> it's for reproducing the original publication (and also our IEEE SPL publication)

Got it.  It would be helpful to clarify the text of the docstring example along these lines.  If you want to include some general recommendations, that's good too, but those should probably be separate and up in the main text of the docstring rather than the examples section.

lostanlen:
AppVeyor build failed on the branch but succeeded on the PR? Not sure what this means.
Also, Travis CI has been pending for the past five hours.
@bmcfee thank you for accepting these changes. I still need to add a few tests to make sure that we have full coverage of the disjunction.

bmcfee:
> AppVeyor build failed on the branch but succeeded on the PR? Not sure what this means.

I think something is corrupted in the build cache, and don't see an easy way to reset it right now.  I'm not worried about it though, since the PR test passes.

> Also, Travis CI has been pending for the past five hours.

I think there must have been an API outtage when your test ran.  If you click through, everything is green, but the message never made it back to github.  I won't block on that either.



> I still need to add a few tests to make sure that we have full coverage of the disjunction.

Ok, just shout when you think it's ready.  Thanks again!

bmcfee:
Quick update: it looks like the appveyor build error may be due to a problem with the current conda-forge ffmpeg build, so I won't block on this.

lostanlen:
@bmcfee tests are passing! merging

PR #898, opened 2019-06-11
[MRG] more efficient length trimming for inverters

### LINKED ISSUE #802: icqt outputs noise with unusual format

### LINKED ISSUE #854: Length-aware icqt and istft are unnecessarily slow
#### Description

Inverse transforms support a `length=` parameter which ensures that the output has a specific number of samples.  This is primarily used to extend reconstructed signals to match an input signal's length when the transform (eg STFT) discards any samples that don't fill up an entire window at the end of the signal.

If the user specifies a much shorter length than the inversion would otherwise produce, the reconstructed signal is truncated down to the given length.  This results in a lot of wasted computation.

A simple optimization would be to trim the input feature down to the minimum number of frames necessary to reconstruct the target length.  This wouldn't change the result, only make the inversion methods faster for partial reconstruction.

lostanlen:
#### Reference Issue
Fixes #854

Inverse transforms support a `length=` parameter which ensures that the output has a specific number of samples. This is primarily used to extend reconstructed signals to match an input signal's length when the transform (eg STFT) discards any samples that don't fill up an entire window at the end of the signal.

If the user specifies a much shorter length than the inversion would otherwise produce, the reconstructed signal is truncated down to the given length. This results in a lot of wasted computation.

A simple optimization would be to trim the input feature down to the minimum number of frames necessary to reconstruct the target length. This wouldn't change the result, only make the inversion methods faster for partial reconstruction.

During CR of our last ICQT rewrite (#852, which closed #802), i noticed that the kwarg `length` could potentially be used to preemptively trim the CQT input to a smaller number of frames _before_ running inversion, rather than trimming in the waveform domain after inverting the entire CQT.
This PR implements this simple optimization on `istft` as well as `icqt`.

#### Any other comments?
Work in progress. Need to write tests and a performance benchmark.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -295,7 +295,7 @@ def istft(stft_matrix, hop_length=None, win_length=None, window='hann',
     if length is None:
         n_frames = stft_matrix.shape[1]
     else:
-        n_frames = ceil(length / hop_length)
+        n_frames = int(np.ceil(length / hop_length))
 
     expected_signal_len = n_fft + hop_length * (n_frames - 1)
     y = np.zeros(expected_signal_len, dtype=dtype)
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -296,7 +296,7 @@ def istft(stft_matrix, hop_length=None, win_length=None, window='hann',
         n_frames = stft_matrix.shape[1]
     else:
         if center:
-            padded_length = length + int(n_fft//2)
+            padded_length = length + int(n_fft)
         else:
             padded_length = length
         n_frames = int(np.ceil(padded_length / hop_length))
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -292,14 +292,15 @@ def istft(stft_matrix, hop_length=None, win_length=None, window='hann',
     ifft_window = util.pad_center(ifft_window, n_fft)[:, np.newaxis]
 
     # For efficiency, trim STFT frames according to signal length if available
-    if length is None:
-        n_frames = stft_matrix.shape[1]
-    else:
+    if length:
         if center:
             padded_length = length + int(n_fft)
         else:
             padded_length = length
-        n_frames = int(np.ceil(padded_length / hop_length))
+        n_frames = min(
+            stft_matrix.shape[1], int(np.ceil(padded_length / hop_length)))
+    else:
+        n_frames = stft_matrix.shape[1]
 
     expected_signal_len = n_fft + hop_length * (n_frames - 1)
     y = np.zeros(expected_signal_len, dtype=dtype)
@@ -1725,11 +1726,11 @@ def griffinlim(S, n_iter=32, hop_length=None, win_length=None, window='hann',
         rebuilt = stft(inverse, n_fft=n_fft, hop_length=hop_length,
                        win_length=win_length, window=window, center=center,
                        pad_mode=pad_mode)
-        
+
         # Update our phase estimates
         angles[:] = rebuilt - (momentum / (1 + momentum)) * tprev
         angles[:] /= np.abs(angles) + 1e-16
-    
+
     # Return the final phase estimates
     return istft(S * angles, hop_length=hop_length, win_length=win_length,
                  window=window, center=center, dtype=dtype, length=length)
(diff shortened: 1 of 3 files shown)


librosa/core/spectrum.py
```diff
@@ -291,7 +291,12 @@ def istft(stft_matrix, hop_length=None, win_length=None, window='hann',
     # Pad out to match n_fft, and add a broadcasting axis
     ifft_window = util.pad_center(ifft_window, n_fft)[:, np.newaxis]
 
-    n_frames = stft_matrix.shape[1]
+    # For efficiency, trim STFT frames according to signal length if available
+    if length is None:
+        n_frames = stft_matrix.shape[1]
+    else:
+        n_frames = ceil(length / hop_length)
+
```
bmcfee: Is this correctly handling the `center=` parameter?

lostanlen: fixed in  511f626

lostanlen:
`icqt` passes but `istft` doesn't.
Apparently the following block of code does not work as intended


```
    if length is None:
        n_frames = stft_matrix.shape[1]
    else:
        if center:
            padded_length = length + int(n_fft)
        else:
            padded_length = length
        n_frames = int(np.ceil(padded_length / hop_length))
```

i will investigate later this week

lostanlen:
Replaced by 

```python
    # For efficiency, trim STFT frames according to signal length if available
    if length:
        if center:
            padded_length = length + int(n_fft)
        else:
            padded_length = length
        n_frames = min(
            stft_matrix.shape[1], int(np.ceil(padded_length / hop_length)))
    else:
        n_frames = stft_matrix.shape[1]
```

lostanlen:
Travis CI passes! 🎉  There's a failure of the Appveyor build, which looks completely unrelated:
https://ci.appveyor.com/project/bmcfee/librosa/builds/25569620

I made a gist for performance benchmark:
https://gist.github.com/lostanlen/01816b8e3852e783b00aeb479dfc8364

Here is the result.



The blue (resp. orange) curve represents the time taken by istft on master (resp. invlength-854) branch. The x-axis is the value of `length`, normalized by the total length of the example file (~ 1M samples). The y-axis is the computation time, normalized by the computation time for inverting the whole file. The solid line and the shaded areas respectively denote the average and standard deviation of runtime. Measured with `%timeit -r 10`.

This is ready for CR!

bmcfee:
LGTM, thanks!  I'll try a quick rebuild on appveyor, just to keep our badges sane.

bmcfee:
Looks like windows conda is having problems at the moment, but that's not our problem.  I'm happy to merge unless you have anything else to add.

lostanlen:
merge at your convenience!

PR #771, opened 2018-10-22
[In-progress] Cross Similarity

rabitt:
#### Reference Issue

Implements #770 

#### What does this implement/fix? Explain your changes.
Implements an extension of the `recurrence_matrix` function which allows for two different sequences.

Issue #770, opened 2018-10-22
cross-similarity function


rabitt:
I'd love to have a version of `librosa.segment.recurrence_matrix` that compares two arbitrary sequences. A lot of the code in the `recurrence_matrix` function will extend to the two sequence case, happy to make a PR. Any suggestions on how such a function should work within the library?

bmcfee:
This sounds like a good feature to have!   I guess the first question is whether this would be better done as a separate function, or by extending `recurrence_matrix`.

### Arguments for separate function
- It semantically makes more sense, since cross-similarity is in no way "recurrence".
- Not all of the parameters to `recurrence_matrix` apply in the cross-similarity case.  For example, `sym` or `width` only apply in self-similarity.
- Adding a second `data` parameter to `recurrence_matrix` could be awkward.

### Arguments for the same function
- It's a pretty simple extension to the current implementation.

### Verdict?

I'm leaning toward a separate function `cross_similarity`, which would otherwise look a lot like `recurrence_matrix` internally.   But I'm open to counter-arguments.  Anyone else want to chime in?

rabitt:
@bmcfee What about abstracting the common bits into a separate helper function, and then making two separate functions, `cross_similarity` and a simplification of `recurrence_matrix` that rely on the helper?

bmcfee:
@rabitt that also sounds reasonable.  Maybe it's worth prototyping an implementation of `cross_similarity` first and then factoring out the common bits?

rabitt:
`recurrence_matrix` has an axis parameter - in the `cross_similarity` case, do you think it's reasonable to force the data to be in a consistent shape, e.g. `(n_features, n_times)` to avoid having multiple axis parameters?

bmcfee:
yeah, there's precedent for that in [dtw](https://librosa.github.io/librosa/generated/librosa.sequence.dtw.html).

rabitt:
(asking questions as I implement :) ) - 
I guess we want the knn classifier to be relative to (i.e. fit to) the features in the first sequence, and then the connections are built from the features in the second sequence? Should we make this more flexible?

bmcfee:
> I guess we want the knn classifier to be relative to (i.e. fit to) the features in the first sequence, and then the connections are built from the features in the second sequence? Should we make this more flexible?

[...]

PR #904, opened 2019-06-21
[CR needed] Cross similarity

bmcfee:
#### Reference Issue
This PR revises #771 and implements #770 

Adds cross-similarity.  Similar to recurrence_matrix, but intended for comparing two distinct sequences.

#### Any other comments?

I picked this up from @rabitt to try and finish it off.  I've fixed a couple of bugs and unit tests, and tried to work out the docs as best as I could.

I think this needs a very careful CR.  I'm not 100% clear on the semantics of data_from and data_to here.  We've gone back and forth on this a bit, and I wonder if it would be clearer to state it as `data_ref = (D, N)` and `data = (D, M)`, with the resulting xsimilarity being `xsim = (N, M)`.  In this sense, `xsim[i,j] = 1` if `data_ref[:, i]` is one of the k neighbors of `data[:, j]`.  This seems a bit clearer to me than _from and _to, what do yall think?


librosa/segment.py
```diff
@@ -50,6 +52,193 @@
...
+            k = 2 * np.ceil(np.sqrt(n_to - 1))
+        else:
+            k = 2
+
+    if bandwidth is not None:
+        if bandwidth <= 0:
+            raise ParameterError('Invalid bandwidth={}. '
+                                 'Must be strictly positive.'.format(bandwidth))
+
+    k = int(k)
```
lostanlen: i would prefer to see this integer conversion happening directly on line 184

librosa/segment.py
```diff
@@ -50,6 +52,193 @@
...
+
+    # Build the neighbor search object
+    try:
```
lostanlen: can you put a comment explaining why this try-catch statement is necessary? what could go wrong with `algorithm='auto'`?

librosa/segment.py
```diff
@@ -50,6 +52,193 @@
...
+                                                 algorithm='auto')
+    except ValueError:
+        knn = sklearn.neighbors.NearestNeighbors(n_neighbors=min(n_from-1, k + 2),
+                                                 metric=metric,
+                                                 algorithm='brute')
+
+    knn.fit(data_from)
+
+    # Get the knn graph
+    if mode == 'affinity':
```
lostanlen: here you need to explain why `'affinity'` falls back to sklearn's `'distance'`.

librosa/segment.py
```diff
@@ -50,6 +52,193 @@
...
+        # Get the links from point i
+        links = xsim[i].nonzero()[1]
+
+        # Order them ascending
+        idx = links[np.argsort(xsim[i, links].toarray())][0]
+
+        # Everything past the kth closest gets squashed
+        xsim[i, idx[k:]] = 0
+
+    xsim = xsim.tocsr()
```
lostanlen: `# Convert to Compressed Sparse Row (CSR) matrix type`

(the CSR acronym is a bit obscure)

bmcfee:
After giving it some thought, I went ahead and reorganized the input parameters for `cross_similarity`.  Concretely, the semantics are as follows:

- `cross_similarity(data, data_ref)` computes the k nearest neighbors from `data_ref` for each point (column) in `data`.  This parameter ordering mirrors that of similar methods elsewhere, eg softmasking with `X` and `X_ref`.  Intuitively, the first parameter should be considered as the *variable*, and the second is treated as a (fixed) parameter set.
- If `data = (d, n)` and `data_ref = (d, n_ref)`, then `xsim = (n_ref, n)`.  This preserves our usual dimension ordering where the last index corresponds to samples.  This in turn makes it easy to interpret `xsim` as akin to a feature matrix.
- I've rewritten the automatic `k`-selection logic to just be `2 * ceil(sqrt(n_ref))`.  The equation used in `recurrence_matrix`, which was the basis for this one, was inspired by that setting, but had some extra cruft in there to prevent counting self-loops and suppressed pseudo-neighbors within the `width` range.  This is not relevant for cross-similarity, so I got rid of it.
- Each column of `xsim` should have exactly `k` non-zeros.

I think this all makes sense, though it does change how @rabitt's original implementation worked.  Commentary would be most welcome at this point.

librosa/segment.py
bmcfee: Yes, except this way we can gracefully handle floating point inputs `k`.

librosa/segment.py
```diff
@@ -50,6 +52,193 @@
            'path_enhance']
 
 
+@cache(level=30)
+def cross_similarity(data_from, data_to, k=None, metric='euclidean',
+                     sparse=False, mode='connectivity', bandwidth=None):
+    '''Compute cross similarity between two data sequences.
+
+
+    `xsim[i, j]` is non-zero if `data_to[:, j]` is a k-nearest neighbor
```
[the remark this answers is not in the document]
bmcfee: We consistently use non-zero elsewhere, so I'm inclined to not change it here.

librosa/segment.py
lostanlen: OK!

lostanlen:
merge at your convenience! 🎉

bmcfee:
Upon further reflection, my proposed change to cross-similarity is inconsistent with `recurrence_matrix` in just about every way:

1. it's column-normalized instead of row-normalized
2. it returns csc instead of csr

Because `recurrence_matrix` was always producing square outputs, it never occurred to me to think too carefully about which format made more sense.  The docstring for `recurrence_matrix` is similarly unclear on the precise semantics.  I think now that going column-major is more consistent with the rest of the library, for the reasons detailed above.

What to people think about using this as an opportunity to modify the semantics of `recurrence_matrix` to be column-oriented?  Some things to keep in mind:

1. This won't change anything except memory layout when `sym=True` is used in recurrence matrix.
2. This will change the semantics of `decompose.nn_filter`, which expects row-sparse and row-normalized data.

Qualitatively, I don't think this will really change anything for users that don't directly manipulate asymmetric recurrence matrices, though I could be wrong about that.

rabitt:
> What to people think about using this as an opportunity to modify the semantics of recurrence_matrix to be column-oriented? Some things to keep in mind:
> 
> This won't change anything except memory layout when sym=True is used in recurrence matrix.
> This will change the semantics of decompose.nn_filter, which expects row-sparse and row-normalized data.
> Qualitatively, I don't think this will really change anything for users that don't directly manipulate asymmetric recurrence matrices, though I could be wrong about that.

Agree 100%. I think it makes more intuitive sense too.

bmcfee:
Ok great.  In practice, the only change here is that `recurrence_matrix` now transposes the output before returning.

I have this working locally, but there's still a bunch of integration work to make sure downstream processing functions as expected.

bmcfee:
Ok, I think this is all good to go.  The recurrence matrix is now column-major, and everything works as expected.

I did identify some opportunities for optimization #907 , but am punting those for a future release.

Anyone want to give this a last once over before I merge?

bmcfee:
:shipit:

PR #920, opened 2019-07-10
Check boundary condition for non-subseq dtw

fzalkow:
For non-subsequence dynamic time warping, currently the boundary condition is not always satisfied. See the following minimal example:

```
import numpy as np
import librosa

x = np.array([1, 2, 3, 4, 5])
y = np.array([1, 1, 1, 2, 4, 5, 6, 5, 5])
D, wp = librosa.sequence.dtw(x, y)

plt.imshow(D, aspect='auto', cmap='gray_r', origin='lower')
plt.plot(wp[:, 1], wp[:, 0], 'ro-')
```
dtw

This is because the backtracking loop currently stops when the beginning of the first sequence is reached. That makes sense for subsequence DTW, but not for usual DTW. I fixed that by adding a `subseq` keyword for the backtracking function and use a different loop stop criterion based on that. Now it looks like this:
dtw2

### DIFF
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -224,10 +224,10 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
         if subseq:
             # search for global minimum in last row of D-matrix
             wp_end_idx = np.argmin(D[-1, :]) + 1
-            wp = __dtw_backtracking(D_steps[:, :wp_end_idx], step_sizes_sigma)
+            wp = __dtw_backtracking(D_steps[:, :wp_end_idx], step_sizes_sigma, subseq)
         else:
             # perform warping path backtracking
-            wp = __dtw_backtracking(D_steps, step_sizes_sigma)
+            wp = __dtw_backtracking(D_steps, step_sizes_sigma, subseq)
 
         wp = np.asarray(wp, dtype=int)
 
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks @fzalkow !

Before moving forward with this, it's worth noting that the documentation for `dtw` isn't totally clear on what the `subseq` flag does.  @stefan-balke can you confirm / clarify that the current behavior is not what we expect?

fzalkow:
The DTW documentation refers to Meinard Müller's *Fundamentals of Music Processing*, where it is described. For usual DTW, we assume the boundary condition (Equation 3.16). This implies that the first index pair of the warping path must be `(0, 0)`. Intuitively, that means that we have to align the beginnings of both sequences.

This condition is relaxed for Subsequence DTW (Equation 7.25), where the first index pair of the warping path can be any `(0, m)`for `m ∈ {0, …, M-1}`, where `M` refers to the length of the smaller sequence. Intuitively, that means that we can align the beginning of the smaller sequence to any element in the longer sequence.

bmcfee:
@fzalkow thanks for your patience on this one.  I've taken the time to give it some thought, and agree that the change you propose is sensible.  I assume that this does not change the behavior of our implementation when `subseq=True`?

The PR generally looks good to me, but it would be great if you could add a unit test that covers this case for both values of `subseq` -- the example you gave above would be fine for this purpose.  This way we'd be able to detect regressions in the future.

Apart from that, I'm not concerned about the py3.5 error on travis, since it's an unrelated issue with the conda environment, and the changes proposed here are not version-dependent.

fzalkow:
I am happy to contribute!

Btw, you are right, the behavior for `subseq=True` does not change.

lostanlen:
Nice work @fzalkow!

lostanlen:
Closes #937 (an issue number that i just added for the sake of publication in v0.7.1 release notes)

stefan-balke:
thanks @fzalkow

PR #922, opened 2019-07-12
adds the ability to pass in an array of samples to be placed in for c…

mross1080:
#### Reference Issue
This is a PR for a new feature so new Issue exists right now, though I can create one if that's where discussion is better hosted.  

What my new functionality allows for is to pass in an array of Samples to the Click Function, which would then generate a sequentially generated series of Clicks with custom samples.  

Why would we want this?  Well basically when I was going through the tutorials I realized that I wanted to combine the functionality displayed in two of the tutorials.
https://musicinformationretrieval.com/pitch_transcription_exercise.html
and 
https://musicinformationretrieval.com/onset_detection.html

Where every time a click is placed, instead of the same sample being placed at each index of the Onset, we can choose a custom sample for each click.  So for example I have a function that takes human speech and I'm extracting beats and pitch so that it is played back using the corresponding Piano sample, which is now possible with the addition of the click_samples keyword.  

Screen Shot 2019-07-12 at 4 05 07 PM

Let me know what ya'll think, I'm happy to make changes or adjustments based on what you think.  Curious if you think others have use for this functionality.  Thanks for all the amazing work on this Lib!

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -1027,7 +1027,7 @@ def zero_crossings(y, threshold=1e-10, ref_magnitude=None, pad=True,
 
 
 def clicks(times=None, frames=None, sr=22050, hop_length=512,
-           click_freq=1000.0, click_duration=0.1, click=None, length=None):
+           click_freq=1000.0, click_duration=0.1, click=None, length=None, click_samples=[]):
     """Returns a signal with the signal `click` placed at each specified time
 
     Parameters
@@ -1114,6 +1114,10 @@ def clicks(times=None, frames=None, sr=22050, hop_length=512,
         # Convert times to positions
         positions = time_to_samples(times, sr=sr)
 
+    if click_samples is not []:
+        for sample in click_samples:
+            util.valid_audio(sample, mono=True)
+
     if click is not None:
         # Check that we have a well-formed audio buffer
         util.valid_audio(click, mono=True)
@@ -1148,7 +1152,9 @@ def clicks(times=None, frames=None, sr=22050, hop_length=512,
     click_signal = np.zeros(length, dtype=np.float32)
 
     # Place clicks
-    for start in positions:
+    for index, start in enumerate(positions):
+        if click_samples is not []:
+            click = click_samples[index]
         # Compute the end-point of this click
         end = start + click.shape[0]
 


bmcfee:
Thanks for the PR.

I'd like to hear from other devs on this, but I think this functionality may be a little too niche for inclusion in the main library.

As an aside, it's dangerous to use `[]` as a default parameter, since python parameters are pass-by-reference.  It's safer to use `None` and then replace `None` with a newly constructed list at run-time.

lostanlen:
Honestly, i like it. Beyond OP's application in melodic synth command, adapting the click track may be useful for design stimuli in music psychology experiments.
Perhaps @koudyk and @blairkan can comment on this?

bmcfee:
> adapting the click track may be useful for design stimuli in music psychology experiments.

I guess I can see a use-case there, but I think extending the API for clicks in this way might make it too obscure.  The click API currently takes a list of times (or frames, or whatever) indicating event positions, and the proposed change would cycle through a list of different clicks in a round-robin fashion, which is not immediately transparent from the API.

I think what would be more useful is an API that could take labeled event times `(t[i], label[i])` at time `t[i]` plays the sound indexed by the label `clicks[label[i]]`, where `clicks` is now a dictionary rather than an ordered list.  It would take a little more effort to use on behalf of the user, but I think the API would be much more clear.

That would take us pretty far afield from a standard click-track generator though, so maybe it should just be a different function?  OTOH, I'm still not entirely convinced that we need to add this function, since it could be simulated with just a couple more lines, eg:

```python
track = np.zeros(duration)
for class in labels:
    track += librosa.clicks(times[labels == class], clicks[class])
```
(or something to that effect)

koudyk:
I see how it might be useful in music/speech psychology, for example for making perceptual stimuli with only a few features (e.g., onsets and pitch) based on a complex sound. But another user might want to vary other click features, like velocity, duration, timbre, etc. Bmcfee's suggestion of using labelled event times seems most intuitive to me, especially if the `clicks` dictionary can have more features than pitch.  But I might be out of my depth here - I'm not too familiar with the inner workings of librosa!

lostanlen:
Right. Also note that `clicks` is probably the wrong name for that feature, because it can be many other things that clicks: spoken words, musical tones, jingles for prompting the subject, noise bursts for "resetting" short-term memory, etc.
I'm thinking of the experiments of Claire Pelofi for example, on the effect of melodic priming in the perception of the tritone paradox.
So although this might help for automated stimulus design, we're probably out of scope of `clicks`, which was originally intended as a quick way to evaluate the performance of a beat tracker.

blairkan:
I do use a separate audio channel of clicks to send stimulus timing information directly to the response acquisition system during experiments, but this might not be what you are talking about. If you are talking more generally about timed sequences of audio content, I think this could be useful but sometimes people prefer to implement this in an experimental design package like [e-Prime](https://pstnet.com/products/e-prime/) or [PsychoPy](https://www.psychopy.org/index.html). I'm sorry I haven't been following librosa as much as I should -- just let me know if I'm misunderstanding the topic/question!

bmcfee:
Circling back on this  -- it seems like the consensus here (if any) is that we should probably leave clicks alone, but that it might also make sense to have a more general additive synthesis function for doing sample composition.

As an aside, we do have something a bit like this in [amen](https://github.com/algorithmic-music-exploration/amen/blob/master/amen/synthesize.py), but it's A) much more complex, and B) tied in with amen's object model.  So I do see value in having a more basic additive synth function in librosa.

@mross1080 are you still up for this / have any interest in adapting your design a bit in light of the discussion above?

PR #926, opened 2019-07-23
[MRG] Time-frequency reassignment

### LINKED ISSUE #394: Log-Freq Representations
Hey all,

I'm not convinced by the output of all the CQT transforms for **analysis tasks**, maybe I'm too stupid. I'd like to add more log-freq representations to librosa.
Candidates would be:
- Straight binning approach (well, maybe we can leave this one out)
- Version using interpolation
- Reassigned Spectrogram
- Pitch-Filterbank form Müller et al. (using fixed coeffs from the chroma toolbox)

As I said, these are all for analysis tasks, since the representations are not invertible

My goal would be to have a resource where we could compare all these representations and immediately see, discuss the pros and cons...

Any thoughts on this?

scjs:
#### Reference Issue
#394 

[EDIT by @lostanlen: closes #394]

This adds a function `reassigned_spectrogram` to calculate a time-frequency reassigned spectrogram. Time and frequency reassignment vectors are calculated in two separate functions `reassign_times` and `reassign_frequencies`.

The new `reassign_frequencies` function is similar to the existing `ifgram`, but the frequency and magnitude clipping are factored out into `reassigned_spectrogram` for consistency with the new `reassign_times`. The `reassign_frequencies` function passes the existing `ifgram` tests against MATLAB.

#### Any other comments?
A suggestion in #394 was to allow the reassignment functions to optionally accept a complex STFT input, like other spectrogram-based representations. However, reassignment requires calculating two auxiliary STFTs, except for one or two special cases (which are not implemented as special cases here). Accepting STFT input would make all three functions more efficient, but they would still need the original time-series and STFT parameters to calculate the auxiliary STFTs. Also, accepting STFT input would make it possible to accidentally provide an STFT that had been calculated with different parameters than the ones that are passed to the reassignment functions. That would produce incorrect results that could be hard to notice. For those reasons, STFT input isn't currently included as an option in this PR.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -521,7 +521,8 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
 
 
 def reassign_frequencies(y, sr=22050, n_fft=2048, hop_length=None,
-                         win_length=None, window="hann", dtype=np.complex64):
+                         win_length=None, window="hann", center=True,
+                         dtype=np.complex64, pad_mode="reflect"):
     """Instantaneous frequencies based on a spectrogram representation.
 
     The reassignment vector is calculated using equation 5.20 in Flandrin,
@@ -534,6 +535,12 @@ def reassign_frequencies(y, sr=22050, n_fft=2048, hop_length=None,
     window.
 
     See `reassigned_spectrogram` for references.
+
+    It is recommended to use `pad_mode="wrap"` or else `center=False`, rather
+    than the defaults. Frequency reassignment assumes that the energy in each
+    FFT bin is associated with exactly one signal component. Reflection padding
+    at the edges of the signal invalidates the reassigned estimates in the edge
+    frames.
 
     Parameters
     ----------
@@ -564,8 +571,17 @@ def reassign_frequencies(y, sr=22050, n_fft=2048, hop_length=None,
 
         .. see also:: `filters.get_window`
 
+    center : boolean
+        - If `True`, the signal `y` is padded so that frame
+          `S[:, t]` is centered at `y[t * hop_length]`.
+        - If `False`, then `S[:, t]` begins at `y[t * hop_length]`
+
     dtype : numeric type
         Complex numeric type for `S`. Default is 64-bit complex.
+
+    pad_mode : string
+        If `center=True`, the padding mode to use at the edges of the signal.
+        By default, STFT uses reflection padding.
 
     Returns
     -------
@@ -609,14 +625,14 @@ def reassign_frequencies(y, sr=22050, n_fft=2048, hop_length=None,
     window = get_window(window, win_length, fftbins=True)
     window = util.pad_center(window, n_fft)
 
-    # center is always False for consistency with `reassign_times`
     S_h = stft(
         y=y,
         n_fft=n_fft,
         hop_length=hop_length,
         window=window,
-        center=False,
+        center=center,
         dtype=dtype,
+        pad_mode=pad_mode,
     )
 
     # cyclic gradient to correctly handle edges of a periodic window
@@ -627,8 +643,9 @@ def reassign_frequencies(y, sr=22050, n_fft=2048, hop_length=None,
         n_fft=n_fft,
         hop_length=hop_length,
         window=window_derivative,
-        center=False,
+        center=center,
         dtype=dtype,
+        pad_mode=pad_mode,
     )
 
     # equation 5.20 of Flandrin, Auger, & Chassande-Mottin 2002
@@ -643,7 +660,8 @@ def reassign_frequencies(y, sr=22050, n_fft=2048, hop_length=None,
 
 
 def reassign_times(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
-                   window="hann", dtype=np.complex64):
+                   window="hann", center=True, dtype=np.complex64,
+                   pad_mode="reflect"):
     """Time reassignments based on a spectrogram representation.
 
     The reassignment vector is calculated using equation 5.23 in Flandrin,
@@ -656,6 +674,12 @@ def reassign_times(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
     by the time offset from the window center.
 
     See `reassigned_spectrogram` for references.
+
+    It is recommended to use `pad_mode="constant"` or else `center=False`,
+    rather than the defaults. Time reassignment assumes that the energy in each
+    FFT bin is associated with exactly one impulse event. Reflection padding
+    at the edges of the signal invalidates the reassigned estimates in the edge
+    frames.
 
     Parameters
     ----------
@@ -686,8 +710,17 @@ def reassign_times(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
 
         .. see also:: `filters.get_window`
 
+    center : boolean
+        - If `True`, the signal `y` is padded so that frame
+          `D[:, t]` is centered at `y[t * hop_length]`.
+        - If `False`, then `D[:, t]` begins at `y[t * hop_length]`
+
     dtype : numeric type
         Complex numeric type for `S`. Default is 64-bit complex.
+
+    pad_mode : string
+        If `center=True`, the padding mode to use at the edges of the signal.
+        By default, STFT uses reflection padding.
 
     Returns
     -------
@@ -738,15 +771,14 @@ def reassign_times(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
     if hop_length is None:
         hop_length = int(win_length // 4)
 
-    # center is always False because padding will duplicate impulse components
-    # at the edges, which invalidates reassignment for edge frames
     S_h = stft(
         y=y,
         n_fft=n_fft,
         hop_length=hop_length,
         window=window,
-        center=False,
+        center=center,
         dtype=dtype,
+        pad_mode=pad_mode,
     )
 
     # calculate window weighted by time
@@ -765,8 +797,9 @@ def reassign_times(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
         n_fft=n_fft,
         hop_length=hop_length,
         window=window_time_weighted,
-        center=False,
+        center=center,
         dtype=dtype,
+        pad_mode=pad_mode,
     )
 
     # equation 5.23 of Flandrin, Auger, & Chassande-Mottin 2002
@@ -774,8 +807,14 @@ def reassign_times(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
     # Meyer, & Ainsworth 1998 pp. 283-284
     correction = np.real(S_th / S_h)
 
+    if center:
+        pad_length = None
+
+    else:
+        pad_length = n_fft
+
     times = time_frequency.frames_to_time(
-        np.arange(S_h.shape[1]), sr=sr, hop_length=hop_length, n_fft=n_fft
+        np.arange(S_h.shape[1]), sr=sr, hop_length=hop_length, n_fft=pad_length
     ).reshape((1, -1))
 
     times = times + correction
@@ -784,8 +823,9 @@ def reassign_times(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
 
 
 def reassigned_spectrogram(y, sr=22050, n_fft=2048, hop_length=None,
-                           win_length=None, window="hann", ref_power=1e-6,
-                           clip=True, dtype=np.complex64):
+                           win_length=None, window="hann", center=True,
+                           ref_power=1e-6, clip=True, dtype=np.complex64,
+                           pad_mode="reflect"):
     """Time-frequency reassigned spectrogram.
 
     The reassignment vectors are calculated using equations 5.20 and 5.23 in
@@ -802,6 +842,10 @@ def reassigned_spectrogram(y, sr=22050, n_fft=2048, hop_length=None,
     window multipled by the time offset from the window center. See [2]_ for
     additional algorithms, and [3]_ and [4]_ for history and discussion of the
     method.
+
+    It is recommended to use `center=False` rather than the default `True`, to
+    avoid invalidating the reassigned estimates in the edge frames (see
+    `reassign_frequencies` and `reassign_times`).
 
     .. [1] Flandrin, P., Auger, F., & Chassande-Mottin, E. (2002).
     Time-Frequency reassignment: From principles to algorithms. In Applications
@@ -850,6 +894,11 @@ def reassigned_spectrogram(y, sr=22050, n_fft=2048, hop_length=None,
 
         .. see also:: `filters.get_window`
 
+    center : boolean
+        - If `True`, the signal `y` is padded so that frame
+          `D[:, t]` is centered at `y[t * hop_length]`.
+        - If `False`, then `D[:, t]` begins at `y[t * hop_length]`
+
     ref_power : float >= 0 or callable
         Minimum power threshold for estimating time-frequency reassignments.
         Any bin with `np.abs(S[f, t])**2 < ref_power` will be returned as
@@ -864,6 +913,10 @@ def reassigned_spectrogram(y, sr=22050, n_fft=2048, hop_length=None,
 
     dtype : numeric type
         Complex numeric type for `S`. Default is 64-bit complex.
+
+    pad_mode : string
+        If `center=True`, the padding mode to use at the edges of the signal.
+        By default, STFT uses reflection padding.
 
     Returns
     -------
@@ -921,6 +974,9 @@ def reassigned_spectrogram(y, sr=22050, n_fft=2048, hop_length=None,
 
     """
 
+    if not six.callable(ref_power) and ref_power < 0:
+        raise ParameterError("ref_power must be non-negative or callable.")
+
     freqs, S = reassign_frequencies(
         y,
         sr=sr,
@@ -928,7 +984,9 @@ def reassigned_spectrogram(y, sr=22050, n_fft=2048, hop_length=None,
         hop_length=hop_length,
         win_length=win_length,
         window=window,
+        center=center,
         dtype=dtype,
+        pad_mode=pad_mode,
     )
 
     times, S = reassign_times(
@@ -938,23 +996,22 @@ def reassigned_spectrogram(y, sr=22050, n_fft=2048, hop_length=None,
         hop_length=hop_length,
         win_length=win_length,
         window=window,
+        center=center,
         dtype=dtype,
+        pad_mode=pad_mode,
     )
 
     mags = np.abs(S)
 
+    # remove estimates outside the spectrogram
+    if clip:
+        freqs[(freqs < 0) | (freqs > sr / 2.0)] = np.nan
+        times[(times < 0) | (times > len(y) / float(sr))] = np.nan
+
+    # remove estimates with low support
     if six.callable(ref_power):
         ref_power = ref_power(mags ** 2)
 
-    elif ref_power < 0:
-        raise ParameterError("ref_power must be non-negative or callable.")
-
-    # remove estimates outside the spectrogram
-    if clip:
-        freqs[(freqs < 0) | (freqs > sr // 2)] = np.nan
-        times[(times < 0) | (times > len(y) / sr)] = np.nan
-
-    # remove estimates with low support
     if ref_power > 0:
         mags_nan = np.less(mags, ref_power ** 0.5, where=~np.isnan(mags))
 
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -22,7 +22,7 @@ from ..filters import get_window, semitone_filterbank
 from ..filters import window_sumsquare
 
 __all__ = ['stft', 'istft', 'magphase', 'iirt', 'ifgram',
-           'reassign_frequencies', 'reassign_times', 'reassigned_spectrogram',
+           'reassigned_spectrogram',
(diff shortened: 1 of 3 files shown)


bmcfee:
Wow, thanks for this! I'm at a conference this week and won't have time for a careful review for a few days, but am acknowledging receipt of the PR.

bmcfee:
Ok, I've given this a quick once over and it looks great!  I have a few minor points that I'll drop in the CR, but the main question right now is whether this should replace `ifgram`?  I'm fine with deprecating that implementation, or if we can do an API-preserving rename, that'd be even better.  But it should take some careful thought.

librosa/core/spectrum.py
```diff
@@ -519,6 +520,450 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
...
+    window = get_window(window, win_length, fftbins=True)
+    window = util.pad_center(window, n_fft)
+
+    # center is always False for consistency with `reassign_times`
+    S_h = stft(
+        y=y,
+        n_fft=n_fft,
+        hop_length=hop_length,
+        window=window,
+        center=False,
```
bmcfee: I see where you're coming from here, but I think it's better to have consistency with the entire API (center=True by default) and document the boundary effects.  In general, this should be a parameter that the user can control.

librosa/core/spectrum.py
```diff
@@ -519,6 +520,450 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
...
+        hop_length=hop_length,
+        window=window_derivative,
+        center=False,
+        dtype=dtype,
+    )
+
+    # equation 5.20 of Flandrin, Auger, & Chassande-Mottin 2002
+    # the sign of the correction is reversed in some papers - see Plante,
+    # Meyer, & Ainsworth 1998 pp. 283-284
+    correction = -np.imag(S_dh / S_h)
```
bmcfee: This division makes me nervous.  Is there a more numerically stable way to compute this?

librosa/core/spectrum.py
```diff
@@ -519,6 +520,450 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
...
+        hop_length = int(win_length // 4)
+
+    # center is always False because padding will duplicate impulse components
+    # at the edges, which invalidates reassignment for edge frames
+    S_h = stft(
+        y=y,
+        n_fft=n_fft,
+        hop_length=hop_length,
+        window=window,
+        center=False,
```
bmcfee: Same comment as before; consistency is more important than absolute correctness in this case.

librosa/core/spectrum.py
```diff
@@ -519,6 +520,450 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
...
+        hop_length=hop_length,
+        window=window_time_weighted,
+        center=False,
+        dtype=dtype,
+    )
+
+    # equation 5.23 of Flandrin, Auger, & Chassande-Mottin 2002
+    # the sign of the correction is reversed in some papers - see Plante,
+    # Meyer, & Ainsworth 1998 pp. 283-284
+    correction = np.real(S_th / S_h)
```
bmcfee: Same comment as before; can we do this in a more numerically stable way?

librosa/core/spectrum.py
```diff
@@ -519,6 +520,450 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
...
+    times, S = reassign_times(
+        y,
+        sr=sr,
+        n_fft=n_fft,
+        hop_length=hop_length,
+        win_length=win_length,
+        window=window,
+        dtype=dtype,
+    )
+
```
bmcfee: There's a lot of redundant calculation here.  Can we refactor this to improve efficiency?

librosa/core/spectrum.py
```diff
@@ -519,6 +520,450 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
...
+    mags = np.abs(S)
+
+    if six.callable(ref_power):
+        ref_power = ref_power(mags ** 2)
+
+    elif ref_power < 0:
+        raise ParameterError("ref_power must be non-negative or callable.")
```
bmcfee: This check should come before heavy-weight calculation so that failures are fast.

librosa/core/spectrum.py
```diff
@@ -519,6 +520,450 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
...
+
+    # remove estimates outside the spectrogram
+    if clip:
+        freqs[(freqs < 0) | (freqs > sr // 2)] = np.nan
+        times[(times < 0) | (times > len(y) / sr)] = np.nan
```
bmcfee: Both of these checks should be floating point divisions.  `sr` need not be an integer.

tests/test_core.py
```diff
@@ -17,6 +17,7 @@
 import scipy.io
 import six
 import pytest
+import unittest.mock
```
bmcfee: This import fails on python 2.  Please use the `six.moves` alias for this until we can drop py2.7 support in the 0.8 series.

librosa/core/spectrum.py
scjs: Done in https://github.com/librosa/librosa/pull/926/commits/9450b92e021de9f34cb5044957d9501c61098f20

librosa/core/spectrum.py
scjs: Done in https://github.com/librosa/librosa/pull/926/commits/9450b92e021de9f34cb5044957d9501c61098f20

librosa/core/spectrum.py
scjs: Done in https://github.com/librosa/librosa/pull/926/commits/9450b92e021de9f34cb5044957d9501c61098f20

tests/test_core.py
scjs: I didn't see a `six.moves` alias for `mock`. (I looked [here](https://six.readthedocs.io/#module-six.moves) in case I missed it.) `mock` was a [separate package](https://pypi.org/project/mock/) in Python 2 that was added to the standard library in Python 3. I could add `mock` to the tests requirements and use `import mock` instead?

librosa/core/spectrum.py
scjs: I put notes about this in the function docstring summaries. Is that the right place for it?

librosa/core/spectrum.py
scjs: I'm not sure about how to improve the redundant calculation of the basic STFT. This algorithm requires three STFTs: the basic one (`S_h`), plus one calculated with the window derivative (`S_dh`, for frequency reassignment) and one calculated with a time-weighted window (`S_th`, for time reassignment).

Two other ways that I thought about organizing `reassigned_spectrogram` so that the basic STFT is only calculated once:
- Put the calculation of both auxiliary windows and all three STFTs into `reassigned_spectrogram` (and probably remove the two duplicate support functions). This disadvantages are that it would be a big function and tests would be harder to write. For users who only want frequency reassignment or time reassignment, this function would calculate one unneeded STFT.
- Allow the two support functions to accept the original STFT as input. In `reassigned_spectrogram`, the STFT would only be calculated in the first function, and then passed to the other. My thoughts on this are in the first PR comment.

What do you think?

bmcfee: >     * Allow the two support functions to accept the original STFT as input.

I like this option, and don't see any obvious downsides to it.  I agree that 3 different stft's are necessary, but we're doing 4 now, and that's my main objection.

tests/test_core.py
bmcfee: That's strange -- I would expect six to have an alias in this case.

As an alternative, you could do the following:

```python
try:
    import mock
except ImportError
    from unittests import mock
```
which would work for our purposes here.  I'm not used to using mock for this kind of test though -- is it necessary in your opinion?  Or could there be a more direct way to test this?

librosa/core/spectrum.py
scjs: Done in https://github.com/librosa/librosa/pull/926/commits/30695ee22631138fbc10207738c5421f65bd1be6

tests/test_core.py
scjs: As is, `reassigned_spectrogram` only does clean-up on the results of the two support functions. My preference is to mock those two functions so that the tests for `reassigned_spectrogram` can just test its functionality separately, without depending on the correctness of the support functions.

It could be tested using data that is known to violate the reassignment assumptions, where some of the numerical estimates from `reassign_*` are expected to be NaN or outside the spectrogram limits. But those estimates are the most likely to be unstable if the support functions change in the future.

I ran into this when I compared this implementation to the [tfr](https://github.com/bzamecnik/tfr) package, which uses a different reassignment algorithm. Because of that, I prefer mocking the support functions to have fixed return values, and only testing that `reassigned_spectrogram` does the right thing with those values. What do you think?

bmcfee: Ok, that's reasonable, thanks for the explanation.  I think we should just go with the try..except for py2/3 support, and that will make it easy to transition to 3-only in the 0.8 series.

scjs: Done in https://github.com/librosa/librosa/pull/926/commits/a444584bf0483490aaa900b94d35f2ec51a137b3

bmcfee:
Thanks @scjs for all your work on this!  I'm not concerned about the 3.5 test failure, since that's due to an upstream problem with the conda environment and not dependent on changes in this PR.  This looks good to me otherwise.

The last point here is to determine the status of `ifgram` given the new implementations.  I think we should definitely deprecate it, since the new tfr methods are a bit cleaner and provide the same functionality.  I guess the only real question is whether doing a straight deprecation rename and removing the old implementation is feasible.  There are two parameters in `ifgram` which are missing from the new methods that prevent a direct API-compatible replacement: `norm=` and `ref_power`.  The former is used to (optionally) normalize the spectrogram (`*= 2 / window.sum()`), the latter is used to threshold bins with low energy and not compute instantaneous frequency there because the division would be unstable.

My thoughts on these:

- deprecate `ifgram` and suggest to use the new methods instead.
- `norm=` option probably shouldn't have been there in the first place, so I don't mind keeping it out of the new methods.
- `ref_power` is implemented in the full `reassigned_spectrogram` function, but not in the time- and frequency-specific versions.  This should be consistent across all three.  I think it might also be worth having a separate parameter to control whether the unstable points are filled with `np.nan` (which is reasonable, and done here already in the PR) or the expected barycenters (fft frequencies and frame centers, respectively).  I think there are good arguments for both, and we shouldn't limit ourselves to one or the other.

scjs:
Thanks for all the feedback!

I haven't really been thinking of `reassign_frequencies` as a full replacement for `ifgram`. My thought was that the `reassigned_spectrogram` function would be the basic function for spectrogram reassignment, with the actual calculations factored out in `reassign_*` support functions for testability and use in other functions. So it would be `reassigned_spectrogram` that supersedes `ifgram`, and a redundant `ifgram`-like function that cleans up the results in a specific way would only be needed if many users will want frequency but not time reassignment.

As an alternative, would it make sense to add documentation to `reassign_*` to warn users that the results are not sanity-checked, and point them to `reassigned_spectrogram` for more plottable/usable results? And maybe move `reassign_*` somewhere else, or mark them as private.

But I also don't mind adding `ref_power` (and `clip`) to `reassign_*` if that seems better. The major downside just being testability, like in the comments above about mock.

I'm not sold yet on returning the bin centers for unstable points. What's an example use case where a user would want the bin centers for bins that don't meet a threshold? I'd wonder whether users would pick this option (return bin centers) just to avoid having to deal with NaNs, and that would cause hard-to-notice problems for later analysis tasks (e.g., estimating frequency-wise or time-wise gradients of the reassigned estimates to locate line or impulse components).

bmcfee:
> I haven't really been thinking of `reassign_frequencies` as a full replacement for `ifgram`. My thought was that the `reassigned_spectrogram` function would be the basic function for spectrogram reassignment, with the actual calculations factored out in `reassign_*` support functions for testability and use in other functions. So it would be `reassigned_spectrogram` that supersedes `ifgram`, and a redundant `ifgram`-like function that cleans up the results in a specific way would only be needed if many users will want frequency but not time reassignment.

That's entirely reasonable.  On the other hand, it doesn't make sense to have multiple implementations kicking around.  If we remove `ifgram`, then users will have to use either `reassign_frequencies` directly (no post-processing) or `reassigned_spectrogram`, which would incur additional work if they don't want time reassignment.

One solution here might be to do something like `numpy.linalg.svd`, and have flags `return_frequencies=False|True` and `return_times=False|True` that control whether these reassignments are computed and returned.  Then `ifgram` becomes `reassigned_spectrogram(..., return_times=False)`.  We still get post-processing, and eliminate unnecessary computation, with a slight increase in API complexity.  I'd be okay with this.



> And maybe move `reassign_*` somewhere else, or mark them as private.

If we do the above, I'd say that they should become `__private`.



> I'm not sold yet on returning the bin centers for unstable points. What's an example use case where a user would want the bin centers for bins that don't meet a threshold?

The way I see it, bins would fall below the threshold if they're quiet.  In this case, it's just as valid to put the center of mass at its expectation than anywhere else, so it becomes a definitional convention for 0/0: do you say it's `nan` (not observed) or revert to the "prior" (analysis frequency) due to lack of contradictory evidence.  I see the latter being useful primarily for plumbing through to downstream models, where one nan could easily propagate outward and become very messy.

I'm fine with nan being the default, or not, but I do think it's worth keeping both options around.  There's also the case for backwards compatibility with `ifgram` (once it gets dep-removed), in which case center frequency defaults become necessary.

bmcfee:
One last point on this, following offline discussion with @lostanlen -- is there a better name than `reassigned_spectrogram`?  (Not saying that there is, but we should give it some thought.)

- `reassigned_spectrogram`
  - Pros: salient for searching purposes, pretty close to the literature
  - Cons: not obvious to newbies what it means, kind of a lot to type
- `tfreassign`
  - Pros: shorter, maybe easier to remember
  - Cons: still somewhat opaque, `tf` is getting overloaded (tensorflow?  time-frequency?)
- something else?  `tfr`?

scjs:
> One solution here might be to do something like numpy.linalg.svd, and have flags return_frequencies=False|True and return_times=False|True that control whether these reassignments are computed and returned. Then ifgram becomes reassigned_spectrogram(..., return_times=False). We still get post-processing, and eliminate unnecessary computation, with a slight increase in API complexity. I'd be okay with this.

That solution sounds good to me. I can add those flags to `reassigned_spectrogram` and rename the two support functions to be private.

Would it make sense to have the flags be `reassign_frequencies` and `reassign_times` (or `correct_*`) rather than `return_*`? When they're set to `False`, the function would return a 2-D array of bin or frame centers in place of calculating the corresponding reassignments. It will be unneeded for some people, and it's different from `ifgram`, but it's not much extra calculation and it means that the function would always return the same number of objects representing the same things.

I'll leave deprecating `ifgram` to you, if you have a process for it.

I can add `fill` or `replace_nan` as an option, defaulting to `False`. At the end of `reassigned_spectrogram`, it will optionally replace the divide-by-zero and thresholded `np.nan` results with the centers. The NaN bins will also include estimates that are outside the spectrogram as well as the low-energy ones. Do you think those should be handled differently? I think they will tend to be low-energy also, but `ifgram` set them to the spectrogram boundaries rather than bin/frame centers.

For the function name, I agree with your pros and cons, but I don't have a preference. Fulop & Fitz call it a 'time-corrected instantaneous frequency spectrogram', so it could be `tcifgram`. The full term is more descriptive but the acronym is opaque.

You also had a note about numerical stability for the low-energy bins. The reassignment calculation could be rearranged to `(S_dh * np.conj(S_h)) / np.square(np.abs(S_h))` if that seems any better, but otherwise I'm not sure.

lostanlen:
I still prefer `reassigned_spectogram`. Yes, it's a lot to type, but it will nicely autocomplete from `librosa.rea` + TAB. Wikipedia calls it "reassigned spectrogram" (https://en.wikipedia.org/wiki/Reassignment_method#/media/File:Short-window_reassigned_spectrogram_of_speech.png) and says that "TFR" means "Time-frequency representation". The collision with "TF" as "TensorFlow" is also inconvenient.

bmcfee:
> Would it make sense to have the flags be `reassign_frequencies` and `reassign_times` (or `correct_*`) rather than `return_*`? When they're set to `False`, the function would return a 2-D array of bin or frame centers in place of calculating the corresponding reassignments. It will be unneeded for some people, and it's different from `ifgram`, but it's not much extra calculation and it means that the function would always return the same number of objects representing the same things.

Interesting suggestion.  I like the idea of having the same number of outputs always, and I'm okay with it not being 100% API-compatible with ifgram (people will have about a year to adapt).

My main concern here is for cpu and memory consumption when the outputs aren't needed.  It would be wasteful to allocate an STFT-sized array of static frequencies (or times), but I bet we could fake it by using a 1-D array and playing some stride tricks to give it the same shape as the STFT without over-allocating.  If that's possible, then I support the `reassign_*` option.  If not, then we should revert to the `return_*` option.  I'll look into it this afternoon.


> I'll leave deprecating `ifgram` to you, if you have a process for it.

Yup, we have a whole system in place for this kind of thing.



>  can add `fill` or `replace_nan` as an option, defaulting to `False`. At the end of `reassigned_spectrogram`, it will optionally replace the divide-by-zero and thresholded `np.nan` results with the centers.

Great!



> The NaN bins will also include estimates that are outside the spectrogram as well as the low-energy ones. Do you think those should be handled differently? I think they will tend to be low-energy also, but `ifgram` set them to the spectrogram boundaries rather than bin/frame centers.

Good point.  I think right now in `ifgram` we clip those to the valid ranges for real-valued band-limited signals (ie, `np.clip(freqs, 0, sr/2.0)`, and that's a sensible thing to do here as well since we know the values aren't outside that range.  That's categorically different from a numerical instability problem.




> You also had a note about numerical stability for the low-energy bins.

Yeah, I'm okay with leaving it as it is for now.



Re: naming, @lostanlen raises some valid points.  As I said, I just wanted to feel out whether there's a better naming option, and it's totally possible that this is the best one.

scjs:
What about [`numpy.broadcast_to`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.broadcast_to.html)?

bmcfee:
> What about [`numpy.broadcast_to`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.broadcast_to.html)?

That doesn't appear to work:
```python
ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (1025,) and requested shape (1025,100)
```

but I'm pretty sure you can use `as_strided` like in this example: https://stackoverflow.com/questions/23695851/python-repeating-numpy-array-without-replicating-data

scjs:
This doesn't seem to copy data:

```python
freqs_1D = librosa.fft_frequencies()
freqs_2D = np.broadcast_to(freqs_1D.reshape((-1, 1)), (1025, 100))
```

bmcfee:
> ```python
> freqs_1D = librosa.fft_frequencies()
> freqs_2D = np.broadcast_to(freqs_1D.reshape((-1, 1)), (1025, 100))
> ```

Ah, good call.  I had forgot to reshape the input.  A slightly nicer looking version of this is 

```python
freqs_2d = np.broadcast_to(freqs_1d[:, np.newaxis], (len(freqs_1d), n_frames))
```

For the record, ipython's `%whos` magic does not correctly report the memory usage here, but checking `len(freqs_2d.data)` does.  :+1:

bmcfee:
Quick update: the travis build for 3.5 should be fixed now.  If you rebase this PR, tests should pass cleanly.

librosa/core/spectrum.py
```diff
@@ -519,6 +520,623 @@ def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
...
+        if center:
+            pad_length = None
+
+        else:
+            pad_length = n_fft
+
+        bin_freqs = time_frequency.fft_frequencies(sr=sr, n_fft=n_fft)
+
+        frame_times = time_frequency.frames_to_time(
+            frames=np.arange(S.shape[1]),
```
bmcfee: You could do `frames_like(S, ...)`

bmcfee:
Btw, don't worry about the times_like comment. I've noticed this in a few spots elsewhere in the library, so I'll probably make a separate pr to fix them all at once.

I think this is good to go, unless there's anything else you want to add?

scjs:
Sounds good! Thanks for all the comments and help getting this ready. I'm done with it on my side, so go ahead and merge if it looks good to you.

lostanlen:
I just had a very cursory overview of this PR and this looks fantastic to me. 

The only thing that surprises me is that functions `__reassign_times` and `__reassign_frequencies` are documented and contain examples despite beginning with two underscores. This means that, by default, they won't be `import`-ed. Double leading underscores are the convention for "private" methods, which most often are undocumented and whose API is prone to change at any version. OTOH, it would be inconvenient to rename them to `reassign_times` and `reassign_frequencies` because then `librosa.rea` + TAB would not autocomplete to `reassigned_spectrogram`, which is the function 99% of users want. So i guess we should just go on with it, and take the examples and docs in `__reassign_*` as a bonus for code readability (which is one of the core missions of the librosa codebase).

TL;DR: LGTM 🎉 

Pinging @helenacuesta: here's the new librosa v0.7.1 feature i told you about over ice cream last Saturday 🍨

bmcfee:
:shipit: !

stefan-balke:
This is great news. thank @scjs!
Should be "easy" now to setup some salience representation similar to MELODIA. Actually, since we have Viterbi also now, all building are there...

Issue #509, opened 2017-02-11
Default sample rate for librosa.core.load is not the native sample rate


mcartwright:
Currently, the default sample rate is 22050 Hz when loading an audio file. In order to use the native sample rate, `sr` must be set to `None`. I often forget to set this to `None` when I want the native sample rate, and I have noticed many students do the same. 

Have others experienced confusion over this? 
Should this change in a later release so that the native sample rate is the default?

bmcfee:
As an aside: we've sometimes discussed wholly doing away with I/O and codec functionality in librosa, and encouraging people to use pysoundfile instead of `librosa.load`.  (Or, audioread if folks really want mp3 support.)

If, at some point, we do deprecate `librosa.load`, this issue would resolve itself.

mcartwright:
I think this could be a good solution as I have had other issues with students when writing sound files (e.g. by default librosa uses the dtype of the array for the output, but a lot of audio players don't support float32 encodings, etc.).

bmcfee:
An alternate solution is to use [presets](https://presets.readthedocs.io/en/latest/) to change the default parameters to load.  I wouldn't recommend this for teaching purposes, but it is sometimes useful when you have a fixed environment and don't want to copy parameters everywhere.

mcartwright:
Yeah, agreed. That's a decent solution for someone like me who is simply forgetful, but not a great solution for teaching. 

Also, it helps for the input side, but not the output, which also has its issues as mentioned above.

keunwoochoi:
Agree that the default should be `None`, seems like that's how people expect it would work when they don't specify `sr`. (people: {me, my friend 1, my friend 2})

bmcfee:
Maybe it's time to start thinking about big changes for a 1.0.0 release...

carlthome:
I too think it would be nice of `sr=None` was the default.

dpwe:
I've had this feeling to, but there are further consequences: librosa uses
a default window length for stft etc in samples, not seconds. This is fine
if the sample rate is essentially always 22050, but is much less pleasant
if you're trying to be sample rate agnostic.

My preference would be for native sample rate by default, and window sizes
etc defined in seconds. But *that* gets ugly because you want 2^R fft
sizes, and it's not trivial to enforce that transparently.

The 22kHz-by-default is quite a neat work around in practice.

  DAn.

bmcfee:
[...]

PR #930, opened 2019-08-02
CQT tuning correction, tuning standardization, deprecate A440=

### LINKED ISSUE #925: inconsistent/double calculation of tuning in cqt
#### Description
When the `tuning` argument of `librosa.cqt()` is `None`, it estimates tuning by [calling `estimate_tuning()`](https://github.com/librosa/librosa/blob/0.7.0/librosa/core/constantq.py#L174-L175) to find the tuning and [calls `cqt_frequencies()`](https://github.com/librosa/librosa/blob/0.7.0/librosa/core/constantq.py#L178-L179).

1. The call to `estimate_tuning()` uses its default argument `bins_per_octave=12` always, but the call to `cqt_frequencies()` uses the `bins_per_octave` argument given to `cqt()`. This makes [the frequency correction term](https://github.com/librosa/librosa/blob/0.7.0/librosa/core/time_frequency.py#L957) different from the intended tuning, by a factor of `12 / bins_per_octave`.

2. Furthermore, `cqt()` calls `__cqt_filter_fft` with `fmin_t` in which the [tuning correction is already applied](https://github.com/librosa/librosa/blob/0.7.0/librosa/core/constantq.py#L178-L181), and in `constant_q` [the correction is applied once again](https://github.com/librosa/librosa/blob/0.7.0/librosa/filters.py#L512-L513). So the tuning correction is twice larger than it should be.

#### Steps/Code to Reproduce

This example uses a sound that is 42 cents above the standard A440 tuning. `estimate_tuning` returns 0.43 (which is close enough).

```
import numpy as np
import librosa

sr = 22050
fmin = librosa.note_to_hz('Ab4')
freq = librosa.note_to_hz('A4') * 2 ** (42 / 1200)  # 42 cent above A440
t = np.linspace(0, 1, sr, endpoint=False)
y = np.sin(2 * np.pi * freq * t)
print(np.abs(librosa.cqt(y, sr, fmin=fmin, n_bins=7, bins_per_octave=36, tuning=None))[:, 10])
```

#### Expected Results
If the tuning is correctly identified, the CQT spectrum would be symmetric around the peak.

#### Actual Results
The spectrum is not symmetric and the peak lies somewhere between the 4th and 5th bin.

`[ 0.05483014  0.67865918  5.63108114 22.61531804 19.71375346  3.43137636
  0.56649648]`

#### Workaround

To apply a certain `tuning` value, I can instead give `(tuning / 2) * (bins_per_octave / 12)`, dividing by 2 to compensate for the double calculation and multiplying by `bins_per_octave / 12` to compensate for the incorrect `estimate_tuning()` call.

So if I run the following instead:

`print(np.abs(librosa.cqt(y, sr, fmin=fmin, n_bins=7, bins_per_octave=36, tuning=0.63))[:, 10])`

I get a symmetric spectrum:

`[ 0.07743703  0.08895902 12.56089353 24.92553951 12.64337768  0.22712318
  0.11059439]`

#### Versions

```
Linux-4.15.0-54-generic-x86_64-with-debian-buster-sid
Python 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21) 
[GCC 7.3.0]
NumPy 1.16.4
SciPy 1.3.0
librosa 0.7.0
```

I think it'd be better if tuning is always defined in terms of 'fractions of a bin' and avoid using the term 'cents' altogether, as it's confusing with the music theory term (a 100th of a semitone)

bmcfee:
#### Reference Issue
Fixes #925 

This PR addresses the following issues:

- [x] redundant tuning estimation in CQT
- [x] incorrect bin resolution of CQT tuning estimation
- [x] deprecation of `tuning=` parameter in CQT basis constructors
- [x] standardize `tuning=` parameter to be in "fractions of a bin" rather than "cents" across the library
- [x] deprecate `A440=` parameters in favor of `tuning=` parameters across the library.

#### Any other comments?

As mentioned in the discussion of #925, the approach to tuning correction going forward is to simplify the CQT basis constructors to not support tuning.  Rather, it is the job of the caller (ie, `librosa.cqt`) to adjust the `fmin` parameter as a function of tuning prior to requesting the basis.  This simplifies all of the downstream logic, and clarifies exactly the interpretation of `fmin` when tuning is non-zero.

A couple of quirks that popped up in implementing this:

- Should we support `tuning=` in the display module, ie in specshow?  If we don't, then the bin centers for `cqt_hz` decoration will be incorrect.  However, the user may not have the tuning value on-hand if it's automatically calculated and not returned.

- Can we find a more accurate reference for `filters.constant_q`?  The function has generalized substantially from its first implementation, and I don't think the ref we have now is appropriate.

I'll signal for a CR once the last two bullet points are done.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -54,7 +54,7 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     bins_per_octave : int > 0 [scalar]
         Number of bins per octave
 
-    tuning : None or float in `[-0.5, 0.5)`
+    tuning : None or float
         Tuning offset in fractions of a bin.
 
         If `None`, tuning will be automatically estimated from the signal.
@@ -313,7 +313,7 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     bins_per_octave : int > 0 [scalar]
         Number of bins per octave
 
-    tuning : None or float in `[-0.5, 0.5)`
+    tuning : None or float
         Tuning offset in fractions of a bin.
 
         If `None`, tuning will be automatically estimated from the signal.
@@ -461,7 +461,7 @@ def pseudo_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     bins_per_octave : int > 0 [scalar]
         Number of bins per octave
 
-    tuning : None or float in `[-0.5, 0.5)`
+    tuning : None or float
         Tuning offset in fractions of a bin.
 
         If `None`, tuning will be automatically estimated from the signal.
@@ -566,7 +566,7 @@ def icqt(C, sr=22050, hop_length=512, fmin=None, bins_per_octave=12,
     fmin : float > 0 [scalar]
         Minimum frequency. Defaults to C1 ~= 32.70 Hz
 
-    tuning : float in `[-0.5, 0.5)` [scalar]
+    tuning : float [scalar]
         Tuning offset in fractions of a bin.
 
         The minimum frequency of the CQT will be modified to
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -55,9 +55,12 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         Number of bins per octave
 
     tuning : None or float in `[-0.5, 0.5)`
-        Tuning offset in fractions of a bin (cents).
+        Tuning offset in fractions of a bin.
 
         If `None`, tuning will be automatically estimated from the signal.
+
+        The minimum frequency of the resulting CQT will be modified to
+        `fmin * 2**(tuning / bins_per_octave)`.
 
     filter_scale : float > 0
         Filter scale factor. Small values (<1) use shorter windows
@@ -172,11 +175,14 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         fmin = note_to_hz('C1')
 
     if tuning is None:
-        tuning = estimate_tuning(y=y, sr=sr)
+        tuning = estimate_tuning(y=y, sr=sr, bins_per_octave=bins_per_octave)
+
+    # Apply tuning correction
+    fmin = fmin * 2.0**(tuning / bins_per_octave)
 
     # First thing, get the freqs of the top octave
     freqs = cqt_frequencies(n_bins, fmin,
-                            bins_per_octave=bins_per_octave, tuning=tuning)[-bins_per_octave:]
+                            bins_per_octave=bins_per_octave)[-bins_per_octave:]
 
     fmin_t = np.min(freqs)
     fmax_t = np.max(freqs)
@@ -207,7 +213,6 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         fft_basis, n_fft, _ = __cqt_filter_fft(sr, fmin_t,
                                                n_filters,
                                                bins_per_octave,
-                                               tuning,
                                                filter_scale,
                                                norm,
                                                sparsity,
@@ -235,7 +240,6 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
     fft_basis, n_fft, _ = __cqt_filter_fft(sr, fmin_t,
                                            n_filters,
                                            bins_per_octave,
-                                           tuning,
                                            filter_scale,
                                            norm,
                                            sparsity,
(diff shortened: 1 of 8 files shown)


bmcfee:
Aside from the reference issue, I think this one's all done.  @jongwook got time to look this over and check that your reported example now behaves as expected?

@lostanlen got time to give it a once-over for CR?

jongwook:
I confirm that my reported example now works as intended (spectrum somewhat symmetric around the peak).

librosa/core/constantq.py
```diff
@@ -55,10 +55,13 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         Number of bins per octave
 
     tuning : None or float in `[-0.5, 0.5)`
-        Tuning offset in fractions of a bin (cents).
+        Tuning offset in fractions of a bin.
 
         If `None`, tuning will be automatically estimated from the signal.
 
+        The minimum frequency of the resulting CQT will be modified to
+        `fmin * 2**(tuning / bins_per_octave)`.
```
lostanlen: What is the use case for setting `tuning` as a nonzero value that isn't estimated from `y`?

librosa/core/constantq.py
```diff
@@ -172,11 +175,14 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
         fmin = note_to_hz('C1')
 
     if tuning is None:
-        tuning = estimate_tuning(y=y, sr=sr)
+        tuning = estimate_tuning(y=y, sr=sr, bins_per_octave=bins_per_octave)
+
+    # Apply tuning correction
+    fmin = fmin * 2.0**(tuning / bins_per_octave)
```
lostanlen: There is nothing preventing the user to set `tuning` outside the recommended range `[-0.5, 0.5)`. Bug or feature?

librosa/core/time_frequency.py
```diff
@@ -844,9 +846,19 @@ def hz_to_octs(frequencies, A440=440.0):
     ----------
     frequencies   : number >0 or np.ndarray [shape=(n,)] or float
         scalar or vector of frequencies
-    A440          : float
+
+    tuning        : number in [-0.5, 0.5)
```
lostanlen: i don't understand the requirement of having tuning bounded by 0.5
if i have `bins_per_octave=12` and i want to process baroque music at A415, that difference is slightly above one semitone.
`np.log2(415/440)*12 == -1.012706247484469` 

So the tuning deviation is outside of the recommended range. The code won't fail though, because there is nothing preventing `tuning` to be in this recommended range. Am i missing something crucial here?

librosa/core/time_frequency.py
```diff
@@ -844,9 +846,19 @@ def hz_to_octs(frequencies, A440=440.0):
...
+        Tuning deviation from A440 in bins per octave.
+
+    bins_per_octave : int > 0
+        Number of bins per octave.
+
+    A440          : float <DEPRECATED>
```
lostanlen: I applaud the deprecation. It's embarassing to set `A440=415`

librosa/core/time_frequency.py
```diff
@@ -844,9 +846,19 @@ def hz_to_octs(frequencies, A440=440.0):
     ----------
     frequencies   : number >0 or np.ndarray [shape=(n,)] or float
         scalar or vector of frequencies
-    A440          : float
+
+    tuning        : number in [-0.5, 0.5)
+        Tuning deviation from A440 in bins per octave.
```
lostanlen: given that our default sample rate is 22050 (#509), setting the librosa pitch standard to be A441, not 440, would be nicely consistent. (a 441 Hz tone at 22050 Hz sample rate has exactly 50 samples per period). But it may not satisfy the principle of least surprise. I could have it either way.

librosa/core/constantq.py
bmcfee: Perhaps you have a different tuning estimator?

librosa/core/constantq.py
bmcfee: Indeed.  Per your point below, I guess there isn't really a good reason to require this limit.

librosa/core/time_frequency.py
bmcfee: Great point.  The +-1/2 bound is an artifact of the estimator, which will only estimate tuning deviations in that range.  But there's no technical requirement for the bound, so I'm happy to adjust the docs accordingly.

librosa/core/time_frequency.py
bmcfee: Hah.  Well, since we're deprecating this parameter anyway, I don't think it much matters, but I'd prefer to keep things in line with [standards](https://en.wikipedia.org/wiki/A440_(pitch_standard)).

bmcfee:
Thanks @jongwook and @lostanlen !  I've revised the docstrings around tuning parameters to drop the +- 0.5 constraint; I think it's good to go.

librosa/core/time_frequency.py
lostanlen: we're deprecating `A440`, but we're not deprecating the fact that the tuning deviation is computed with respect to 440 Hz, not 441 Hz. But sure, let's keep it in line with ISO 16.

PR #934, opened 2019-08-03
CQT Griffin-Lim

### LINKED ISSUE #912: Griffin-Lim for CQT
#### Description

The GL implementation added in 0.7 #843 operates on linear spectrograms.  At some point in the discussion, we raised the idea of having it work on CQT as well, but dropped the idea to keep things simple.

I just prototyped a CQT-based GL offline and it sounds pretty good.  It seems like a useful kind of thing to have in the library.  The only question for me is whether we wrap it into the existing implementation, or just have a separate function that's nearly identical, but with different arguments and calls `cqt/icqt` instead of `stft/istft`.  I'm leaning toward the latter, since the method is simple enough that a little duplication of code won't be catastrophic to maintain.

What do folks think?

bmcfee:
#### Reference Issue
Implements #912 
[EDIT by @lostanlen: closes #912]

This PR adds `griffinlim_cqt` for reconstructing audio from CQT magnitude spectra.

The algorithm is identical to the existing griffinlim, except that stft is replaced by cqt, so the parameters of the function have to change as well.

#### Any other comments?

Tests here are pretty weak, but it seems to work.  I'm not sure what a reasonable trade-off of test coverage and runtime efficiency is here, but i'd be okay with keeping the tests here pretty light.

### DIFF
--- a/librosa/core/__init__.py
+++ b/librosa/core/__init__.py
@@ -38,6 +38,7 @@ Spectral representations
     fmt
 
     griffinlim
+    griffinlim_cqt
 
     interp_harmonics
     salience
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -17,7 +17,8 @@ from .. import filters
 from .. import util
 from ..util.exceptions import ParameterError
 
-__all__ = ['cqt', 'hybrid_cqt', 'pseudo_cqt', 'icqt']
+__all__ = ['cqt', 'hybrid_cqt', 'pseudo_cqt',
+           'icqt', 'griffinlim_cqt']
 
 
 @cache(level=20)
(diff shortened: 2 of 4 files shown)


bmcfee:
And for reference, here's what the docstring example (identical otherwise to the existing griffinlim example) renders:

image

lostanlen:
i installed this branch on a fresh conda environment, with latest stable dependencies. I can reproduce the docstring figure, but i'm seeing this warning at every iteration

```
/Users/vl238/miniconda3/envs/gl/lib/python3.7/site-packages/scipy/sparse/lil.py:504: FutureWarning: future versions will not create a writeable array from broadcast_array. Set the writable flag explicitly to avoid this warning.
  if not i.flags.writeable or i.dtype not in (np.int32, np.int64):
```

Here's my config:

```
Darwin-18.6.0-x86_64-i386-64bit
Python 3.7.3 (default, Mar 27 2019, 16:54:48) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
NumPy 1.17.0
SciPy 1.3.0
librosa 0.7.0
```

lostanlen:
Another problem is that `y_inv` has a single-precision `dtype` (`float32`) even when `C` has a double-precision `dtype` (`float64`).

MWE: 


```python
y, sr = librosa.load(librosa.util.example_audio_file(), duration=5, offset=30, sr=None)
y = y.astype('float64')
C = np.abs(librosa.cqt(y=y, sr=sr, bins_per_octave=36, n_bins=7*36))
y_inv = librosa.griffinlim_cqt(C, n_iter=1, sr=sr, bins_per_octave=36)
print((C.dtype, y_inv.dtype))
```

Expected results: `(dtype('float64'), dtype('float64'))`
Actual results: `(dtype('float64'), dtype('float32'))`

lostanlen:
After some sleuthing, it appears that the type instability can be traced down to `istft` returning `np.float32` by default (`dtype` kwarg) rather than inferring from `stft_matrix`.
Related: #825

bmcfee:
> /Users/vl238/miniconda3/envs/gl/lib/python3.7/site-packages/scipy/sparse/lil.py:504: FutureWarning: future versions will not create a writeable array from broadcast_array. Set the writable flag explicitly to avoid this warning.
>   if not i.flags.writeable or i.dtype not in (np.int32, np.int64):

That's real strange.  It must be coming from the `sparsify_rows` in the basis construction.  Can you open a new issue for that?



> Another problem is that `y_inv` has a single-precision `dtype` (`float32`) even when `C` has a double-precision `dtype` (`float64`).

Your sleuthing is correct.  `istft` supports dtype, but `icqt` does not.  That's easily fixable.

The trickier thing to fix is interpretation of `dtype` in griffin-lim functions, because forward and inverse transforms generally have different dtypes (complex and real, respectively).  I suppose the consistent thing to do in this situation is push the `dtype` variable only into the inverse calls (istft and icqt), unless there's an easy way to cast a real dtype to its equivalent-precision complex type?

bmcfee:
... serves me right for responding before finishing :coffee: , griffinlim already handles dtype in the right way.  All we need to do is adapt `icqt` and `griffinlim` accordingly.  That's an easy fix.

bmcfee:
> After some sleuthing, it appears that the type instability can be traced down to `istft` returning `np.float32` by default (`dtype` kwarg) rather than inferring from `stft_matrix`.

The dtype kwarg takes precedence over implicit dtype inference, so this behavior is correct.  However, icqt and griffinlim didn't expose the dtype parameter, so it was uncontrolled -- this is fixed now.

lostanlen:
We should make sure to remember to mention the new `dtype` kwarg in `icqt` in the release notes of v0.7.1. Not sure if it's worth raising an issue about this, but IMO, that new keyword needs a bullet point of its own, as it's completely independent from the addition of `griffinlim_cqt` as a new feature.

bmcfee:
Absolutely.  I usually go through the PRs / comment thread when compiling release notes, so I think there shouldn't be a problem here.

PR #940, opened 2019-08-06
HTK-compatibility features: liftering and pre-emphasis

### LINKED ISSUE #902: Features for replicating HTK: pre-emphasis and liftering
#### Description
In looking at #804, I noticed that HTK's implementation of MFCCs has a few steps that we don't currently implement, notably pre-emphasis and liftering.  These would be easy to add, but it's worth considering how exactly to structure the API.

Liftering I think makes the most sense as an optional parameter to `feature.mfcc`, which could be 0 by default.

Pre-emphasis is a little thornier from the API perspective.  HTK seems to implement this inside the STFT (eg as a window function), but I could see arguments for doing it as an effect (in the time domain) or as a feature enhancement (in the spectral domain).  The effect view keeps the API relatively simple: load, pre-emph, feature-extract.  But maybe there are good arguments for doing it in the spectral domain?

bmcfee:
#### Reference Issue
Implements #902
[EDIT by @lostanlen: closes #902] 

- [x] MFCC liftering
- [x] effects.preemphasis

### DIFF
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -573,7 +573,7 @@ def split(y, top_db=60, ref=np.max, frame_length=2048, hop_length=512):
 
 
 def preemphasis(y, coef=0.97, zi=None, return_zf=False):
-    '''Pre-emphasize (high-pass filter) an audio signal:
+    '''Pre-emphasize an audio signal with a first-order auto-regressive filter:
 
         y[n] -> y[n] - coef * y[n-1]
 
@@ -584,7 +584,11 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
         Audio signal
 
     coef : positive number
-        Pre-emphasis coefficient.  Larger values provide sharper filtering.
+        Pre-emphasis coefficient.  Typical values of `coef` are between 0 and 1.
+
+        At the limit `coef=0`, the signal is unchanged.
+
+        At `coef=1`, the result is the first-order difference of the signal.
 
     zi : number
         Initial filter state
@@ -622,10 +626,11 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
     >>> plt.tight_layout();
 
 
-    Apply pre-emphasis in pieces for block streaming
+    Apply pre-emphasis in pieces for block streaming.  Note that the second block
+    initializes `zi` with the final state `zf` returned by the first call.
 
     >>> y_filt_1, zf = librosa.effects.preemphasis(y[:1000], return_zf=True)
-    >>> y_filt_2, zf = librosa.effects.preemphasis(y[1000:], zi=y_filt_2, return_zf=True)
+    >>> y_filt_2, zf = librosa.effects.preemphasis(y[1000:], zi=zf, return_zf=True)
     >>> np.allclose(y_filt, np.concatenate([y_filt_1, y_filt_2]))
     True
 
@@ -635,9 +640,9 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
     if zi is None:
         zi = scipy.signal.lfilter_zi(b, 1)
 
-    yout, z_f = scipy.signal.lfilter(b, 1, y, zi=zi)
+    y_out, z_f = scipy.signal.lfilter(b, 1, y, zi=zi)
 
     if return_zf:
-        return yout, z_f
+        return y_out, z_f
 
-    return yout
+    return y_out
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -635,12 +635,14 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
     True
 
     '''
-    b = [1, -coef]
+    b = np.asarray([1.0, -coef], dtype=y.dtype)
+    a = np.asarray([1.0], dtype=y.dtype)
 
     if zi is None:
-        zi = scipy.signal.lfilter_zi(b, 1)
+        zi = scipy.signal.lfilter_zi(b, a)
 
-    y_out, z_f = scipy.signal.lfilter(b, 1, y, zi=zi)
+    y_out, z_f = scipy.signal.lfilter(b, a, y,
+                                      zi=np.asarray(zi, dtype=y.dtype))
 
     if return_zf:
         return y_out, z_f
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -583,8 +583,8 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
     y : np.ndarray
         Audio signal
 
-    coef : number in [0, 1)
-        Pre-emphasis coefficient.  Larger values 
+    coef : positive number
+        Pre-emphasis coefficient.  Larger values provide sharper filtering.
 
     zi : number
         Initial filter state
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -29,9 +29,11 @@ Miscellaneous
     remix
     trim
     split
+    preemphasis
 """
 
 import numpy as np
+import scipy.signal
 
 from . import core
 from . import decompose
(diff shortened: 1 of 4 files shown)


librosa/effects.py
```diff
@@ -568,3 +570,74 @@ def split(y, top_db=60, ref=np.max, frame_length=2048, hop_length=512):
 
     # Stack the results back as an ndarray
     return edges.reshape((-1, 2))
+
+
+def preemphasis(y, coef=0.97, zi=None, return_zf=False):
```
lostanlen: how about using a verb (`preemphasize`) rather than a noun (`preemphasis`)?
granted, all the other effects (`remix`, `split`, `stretch`, `shift`, `trim`) are simultaneously verbs and nouns, but i think we should use verbs, in the spirit of sklearn's `fit`, `transform`, `predict`, etc.

librosa/effects.py
```diff
@@ -568,3 +570,74 @@ def split(y, top_db=60, ref=np.max, frame_length=2048, hop_length=512):
...
+    '''Pre-emphasize (high-pass filter) an audio signal:
```
lostanlen: '''Pre-emphasize an audio signal with a first-order autoregressive filter:

(note that there is nothing preventing `coef` to be negative, in which case the filter becomes low-pass. bug or feature?)

librosa/effects.py
```diff
@@ -568,3 +570,74 @@ def split(y, top_db=60, ref=np.max, frame_length=2048, hop_length=512):
...
+        y[n] -> y[n] - coef * y[n-1]
+
+
+    Parameters
+    ----------
+    y : np.ndarray
+        Audio signal
+
+    coef : positive number
+        Pre-emphasis coefficient.  Larger values provide sharper filtering.
```
lostanlen: Typically in the range (0, 1)
At the limit coef=0, the output signal is unchanged.
At the limit coef=1, the output is the first-order derivative of the input.

librosa/effects.py
```diff
@@ -568,3 +570,74 @@ def split(y, top_db=60, ref=np.max, frame_length=2048, hop_length=512):
...
+    >>> plt.title('Original signal')
+    >>> plt.colorbar()
+    >>> plt.subplot(2,1,2)
+    >>> librosa.display.specshow(S_preemph, y_axis='log', x_axis='time')
+    >>> plt.title('Pre-emphasized signal')
+    >>> plt.colorbar()
+    >>> plt.tight_layout();
+
+
+    Apply pre-emphasis in pieces for block streaming
```
lostanlen: Note that, after the first block, we re-use the final state of the recursive filter `zf` as initial state for the second block (`zi=zf`).

librosa/effects.py
```diff
@@ -568,3 +570,74 @@ def split(y, top_db=60, ref=np.max, frame_length=2048, hop_length=512):
...
+
+    >>> y_filt_1, zf = librosa.effects.preemphasis(y[:1000], return_zf=True)
+    >>> y_filt_2, zf = librosa.effects.preemphasis(y[1000:], zi=y_filt_2, return_zf=True)
```
lostanlen: ⚠️  I'm pretty sure you meant `zi=zf`

lostanlen: also, you don't need `return_zf=True` for this second block, although i can see why you're keeping the method prototype for consistency

librosa/effects.py
```diff
@@ -568,3 +570,74 @@ def split(y, top_db=60, ref=np.max, frame_length=2048, hop_length=512):
...
+    >>> np.allclose(y_filt, np.concatenate([y_filt_1, y_filt_2]))
+    True
+
+    '''
+    b = [1, -coef]
+
+    if zi is None:
+        zi = scipy.signal.lfilter_zi(b, 1)
+
+    yout, z_f = scipy.signal.lfilter(b, 1, y, zi=zi)
```
lostanlen: `yout` -> `y_out` (for consistency with docstring, and easier to "segment" for the reader)

librosa/feature/spectral.py
```diff
@@ -1586,6 +1586,14 @@ def mfcc(y=None, sr=22050, S=None, n_mfcc=20, dct_type=2, norm='ortho', **kwargs
 
         Normalization is not supported for `dct_type=1`.
 
+    lifter : number >= 0
+        If `lifter>0`, apply *liftering* (cepstral filtering) to the MFCCs:
+
+        `M[n, :] <- M[n, :] * (1 + sin(pi * (n + 1) / lifter)) * lifter / 2`
+
+        Setting `lifter >= 2 * n_mfcc` emphasizes the higher-order coefficients.
```
lostanlen: I would love to see a recommended range of values. I have a gut feeling that `lifter=1`  (it cancels one every other MFCC).
"of the order of `n_mfcc`"?

Perhaps the `lifter` should automatically scale with `n_mfcc`? or is that not explicit enough?

librosa/feature/spectral.py
```diff
@@ -1674,7 +1682,14 @@ def mfcc(y=None, sr=22050, S=None, n_mfcc=20, dct_type=2, norm='ortho', **kwargs
     if S is None:
         S = power_to_db(melspectrogram(y=y, sr=sr, **kwargs))
 
-    return scipy.fftpack.dct(S, axis=0, type=dct_type, norm=norm)[:n_mfcc]
+    M = scipy.fftpack.dct(S, axis=0, type=dct_type, norm=norm)[:n_mfcc]
+
+    if lifter > 0:
+        M *= 1 + (lifter / 2) * np.sin(np.pi * np.arange(1, 1 + n_mfcc) / lifter)[:, np.newaxis]
+    elif lifter < 0:
```
lostanlen: note that `lifter=np.nan` will return `M` unchanged. That's undocumented, untested, and probably not what the user will intend.

librosa/effects.py
bmcfee: I don't have strong feelings about this -- most of our functions (but not all) are named for what they return (chroma, melspectrogram, etc).  I expect most people would search for this by the noun than by the verb, but that's just a guess.

librosa/feature/spectral.py
```diff
@@ -1674,7 +1682,14 @@ def mfcc(y=None, sr=22050, S=None, n_mfcc=20, dct_type=2, norm='ortho', **kwargs
     if S is None:
         S = power_to_db(melspectrogram(y=y, sr=sr, **kwargs))
 
-    return scipy.fftpack.dct(S, axis=0, type=dct_type, norm=norm)[:n_mfcc]
+    M = scipy.fftpack.dct(S, axis=0, type=dct_type, norm=norm)[:n_mfcc]
+
+    if lifter > 0:
+        M *= 1 + (lifter / 2) * np.sin(np.pi * np.arange(1, 1 + n_mfcc) / lifter)[:, np.newaxis]
```
lostanlen: if `lifter` is very large, `M` is unchanged.
but if `lifter` overflows, `M` is changed to a constant array of `np.inf` values.

If would be in favor of turning this condition into `0 < lifter < np.inf`

librosa/effects.py
bmcfee: good catch.  And yeah, it's not needed, but it's good style.

librosa/feature/spectral.py
bmcfee: I thought about normalizing the liftering coefficient by `n_mfcc`, but decided against it for a more transparent relation to prior implementations.  Maybe that's important?

librosa/feature/spectral.py
bmcfee: hmmm, good catch.  I'll fix it.

librosa/effects.py
lostanlen: yeah, me neither. i hate to bikeshed names, so i'll let you choose. i just wanted to have this voiced out

bmcfee: Ok.  I'll leave this as is then.

librosa/feature/spectral.py
bmcfee: I don't see a great need to explicitly handle infiities here.  I can't imagine someone accidentally using an infinite lifter coefficient, and if they do it on purpose, then it's on them.

lostanlen: OK, here's what i propose. Turn the `np.sin` into a sine cardinal: `np.sinc`. 
The infinity case will nicely fall back to `np.sinc(0) == 1.0`, and the result will be well-defined.
It will also be more numerically stable for large finite values of the `lifter/n_mfcc` ratio.

lostanlen: Also, please don't forget the `dtype=y.dtype` in `np.arange`

librosa/effects.py
```diff
@@ -568,3 +570,79 @@ def split(y, top_db=60, ref=np.max, frame_length=2048, hop_length=512):
...
+    Apply pre-emphasis in pieces for block streaming.  Note that the second block
+    initializes `zi` with the final state `zf` returned by the first call.
+
+    >>> y_filt_1, zf = librosa.effects.preemphasis(y[:1000], return_zf=True)
+    >>> y_filt_2, zf = librosa.effects.preemphasis(y[1000:], zi=zf, return_zf=True)
+    >>> np.allclose(y_filt, np.concatenate([y_filt_1, y_filt_2]))
+    True
+
+    '''
+    b = [1, -coef]
```
lostanlen: Consider replacing by

`b = np.array([1, -coef], dtype=y.dtype)`

for type stability. 

(in numpy, `float32` get promoted to `float64` when multiplied by integers)

bmcfee: Good catch.  I'll do that and add test cases.

librosa/feature/spectral.py
bmcfee: +1 for dtype.

Re sinc: the lifter isn't really a sinc, so the equivalent filter looks like 

`1 + np.sinc(np.arange(1, N+1) / L) * np.arange(1, N+1) * np.pi / 2`

Comparing the two using `float32` and `lifter=1e80` still agree down to 6 decimal places, so I'm not sure there's much benefit here.

librosa/effects.py
bmcfee: It looks like `lfilter` does not preserve dtype, even when the inputs are all consistently typed.

lostanlen:
LGTM! :shipit:

Issue #941, opened 2019-08-07
File objects work with librosa.load but not librosa.stream


carlthome:
#### Description
Loading a WAV file via tf.gfile.GFile works with `librosa.load` but not `librosa.stream`. Interestingly the same error appears regardless of if it is a bucket path or a local path.

#### Steps/Code to Reproduce
```py
import librosa as lr
import tensorflow as tf

path = 'gs://my-bucket/example.wav'

# Loading the entire file works as expected.
with tf.gfile.GFile(path, 'rb') as f:
    lr.load(f, 48000)

# The streaming method crashes.
with tf.gfile.GFile(path, 'rb') as f:
    assert f.seekable() is True
    for waveforms in lr.stream(f, 128, 1024, 512):
        print(waveforms.shape)
```
#### Expected Results
No errors.

#### Actual Results
> RuntimeError: Error opening : File contains data in an unknown format.

#### Versions
```
import platform; print(platform.platform())...
Linux-4.9.0-9-amd64-x86_64-with-debian-9.9
Python 3.5.3 (default, Sep 27 2018, 17:25:39) 
[GCC 6.3.0 20170516]
NumPy 1.17.0
SciPy 1.3.0
librosa 0.7.0
```

carlthome:
And super curiously, this will work fine

```py
import soundfile as sf

with tf.gfile.GFile(path, 'rb') as f:
    for waveform in sf.blocks(f, blocksize=1024):
        pass
```

so perhaps it's something with the overlapping blocks.

lostanlen:
Thanks for reporting. `librosa.stream` is a wrapper around `sf.blocks` so you're hitting at the problem here. I don't know the inner workings of `tf.gfile.GFile`, so i can't comment further for the time being. But it's something worth investigating.

carlthome:
Setting `overlap` in `sf.blocks` still works so I'll have to dig a bit deeper in how _exactly_  librosa.stream and soundfile.blocks differ. :thinking:

lostanlen:
The plot thickens!

For future reference, our call site is at:
https://github.com/librosa/librosa/blob/6e537b476cfdfe2a34124a3aa1d2916d3c64ebbf/librosa/core/audio.py#L213

The called function is at:
https://github.com/bastibe/SoundFile/blob/161e930da9c9ea76579b6ee18a131e10bca8a605/soundfile.py#L319

carlthome:
Found the problem.

[...]

PR #942, opened 2019-08-07
Add seek call after sf.info

carlthome:
## Reference Issue
Fixes https://github.com/librosa/librosa/issues/941

Rewinds the read position in `librosa.stream` in case the input is a file handle.

```py
import librosa as lr

# Works as expected.
next(lr.stream(lr.util.example_audio_file(), 1, 1024, 512))

# Fixed by this pull request:
next(lr.stream(open(lr.util.example_audio_file(), 'rb'), 1, 1024, 512))
```

#### Any other comments?
Perhaps a unit test should be added for this as well? I'm also curious if this is really something `librosa` should handle or if it's actually a mistake in `soundfile.info`?

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -4,7 +4,7 @@
 
 import os
 import six
-import logging
+import warnings
 
 import soundfile as sf
 import audioread
@@ -23,8 +23,6 @@ __all__ = ['load', 'stream', 'to_mono', 'resample',
            'get_duration', 'get_samplerate',
            'autocorrelate', 'lpc', 'zero_crossings',
            'clicks', 'tone', 'chirp']
-
-logger = logging.getLogger(__name__)
 
 # Resampling bandwidths as percentage of Nyquist
 BW_BEST = resampy.filters.get_filter('kaiser_best')[2]
@@ -144,7 +142,7 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
 
         # If soundfile failed, try audioread instead
         if isinstance(path, six.string_types()):
-            logger.warning('PySoundFile failed. Trying audioread instead.')
+            warnings.warn('PySoundFile failed. Trying audioread instead.')
             y, sr_native = __audioread_load(path, offset, duration, dtype)
         else:
             raise exc
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -4,6 +4,7 @@
 
 import os
 import six
+import logging
 
 import soundfile as sf
 import audioread
@@ -22,6 +23,8 @@ __all__ = ['load', 'stream', 'to_mono', 'resample',
            'get_duration', 'get_samplerate',
            'autocorrelate', 'lpc', 'zero_crossings',
            'clicks', 'tone', 'chirp']
+
+logger = logging.getLogger(__name__)
 
 # Resampling bandwidths as percentage of Nyquist
 BW_BEST = resampy.filters.get_filter('kaiser_best')[2]
@@ -139,8 +142,8 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
 
     except RuntimeError as exc:
         # If soundfile failed, fall back to the audioread loader
-        if hasattr(path, 'name'):
-            path = path.name
+        logger.warning('PySoundFile failed. Trying audioread instead.',
+                       exc_info=exc)
         y, sr_native = __audioread_load(path, offset, duration, dtype)
 
     # Final cleanup for dtype and contiguity
@@ -890,7 +893,7 @@ def __lpc(y, order):
         # fwd_pred_error = f_{M-1,k}       (we have advanced M)
         # den <- DEN_{M}                   (lhs)
         #
-        
+
         q = dtype(1) - reflect_coeff**2
         den = q*den - bwd_pred_error[-1]**2 - fwd_pred_error[0]**2
 
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -4,6 +4,7 @@
 
 import os
 import six
+import sys
 import warnings
 
 import soundfile as sf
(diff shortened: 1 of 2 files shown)


bmcfee:
> Perhaps a unit test should be added for this as well? I'm also curious if this is really something `librosa` should handle or if it's actually a mistake in `soundfile.info`?

I think we should have a unit test for this regardless of if it's a bug (or otherwise undocumented behavior in soundfile).

librosa/core/audio.py
```diff
@@ -353,6 +353,10 @@ def stream(path, block_length, frame_length, hop_length,
     # Get the sample rate from the file info
     sr = sf.info(path).samplerate
 
+    # If the input is a file handle, rewind its read position after `sf.info`
+    if not isinstance(path, str) and not isinstance(path, int):
+        path.seek(0)
+
```
bmcfee: I wonder if there's a more direct way to detect if seeking is supported here?  Checking for `str` type is brittle (esp in py2, where str and unicode may not be the same).

It might be more robust to just do this seek if `hasattr(path, 'seek')`

carlthome:
@bmcfee I added the code path to `test_stream`. Do you think it's fine to skip a `with` statement or explicit `close` call? I believe it's closed on garbage collection but not sure about pytest.

bmcfee:
> Do you think it's fine to skip a `with` statement or explicit `close` call? I believe it's closed on garbage collection but not sure about pytest.


Thanks for doing this!

I think the best way to handle setup/teardown is with a pytest fixture.  Something like what's described here: https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

librosa/core/audio.py
```diff
@@ -139,6 +139,8 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
 
     except RuntimeError as exc:
         # If soundfile failed, fall back to the audioread loader
+        if hasattr(path, 'name'):
```
carlthome: If a file cannot be parsed by `soundfile`, the resulting error leads to this `audioread` fallback and `audioread will subsequently complain that it expects a string and not a file. This new if check is a hacky way of getting the underlying path from an `open()` call but not all file like objects have a name so this is very fragile!

Probably better to check why `soundfile` failed and handle distinct cases differently? Also, should a warning be logged here to the user because now it's failing silently which is confusing (sometimes loading is super fast, sometimes it's slow, depending on if e.g. `sudo apt install libsndfile1` was run on the user machine or not).

bmcfee: > Probably better to check why `soundfile` failed and handle distinct cases differently?

Good point.  In the first cut of this, I was really only concerned with missing codec errors, but soundfile can throw a bunch of typeerrors or valueerrors that are not handled here.  I think it's fine to propagate soundfile errors upstream if we can't compensate for them by falling back on audioread.



> This new if check is a hacky way of getting the underlying path from an `open()` call but not all file like objects have a name so this is very fragile!

I don't think we should do this.  If the user supplies an open file handle, and soundfile can't work with it, we shouldn't try to make audioread do it.  In this case, it should fail noisily.

I think the logic should be that audioread fallback is only triggered if the input is a path string, and otherwise any exceptions get reraised.



> Also, should a warning be logged here to the user because now it's failing silently which is confusing

Yes, excellent point!

carlthome: >if the input is a path string

This is pretty hard to check in a a Python 2 _and_ 3 compatible way.

tests/test_core.py
```diff
@@ -2030,6 +2040,10 @@ def test_stream(block_length, frame_length, hop_length, mono, offset,
     # Load the reference data.
     # We'll cast to mono here to simplify checking
 
+    # File objects have to be reset before loading
+    if hasattr(path, 'seek'):
```
carlthome: Not sure how to setup a pytest.fixture in a cleaner way than this :disappointed: 

Gladly take tips!

librosa/core/audio.py
bmcfee: `isinstance(path, six.string_types())` ought to do the trick. :)

tests/test_core.py
bmcfee: That seems good to me.  It's a little gnarly, but that should be expected since we're mixing different types in the fixture.

librosa/core/audio.py
carlthome: Ooh, convenient.

carlthome:
Believe this is ready for merge, @bmcfee and @lostanlen

librosa/core/audio.py
```diff
@@ -23,6 +24,8 @@
            'autocorrelate', 'lpc', 'zero_crossings',
            'clicks', 'tone', 'chirp']
 
+logger = logging.getLogger(__name__)
```
bmcfee: We don't use logging anywhere else in the library -- why do this instead of warnings?

librosa/core/audio.py
```diff
@@ -138,8 +141,13 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
...
     except RuntimeError as exc:
-        # If soundfile failed, fall back to the audioread loader
-        y, sr_native = __audioread_load(path, offset, duration, dtype)
+
+        # If soundfile failed, try audioread instead
+        if isinstance(path, six.string_types):
+            logger.warning('PySoundFile failed. Trying audioread instead.')
+            y, sr_native = __audioread_load(path, offset, duration, dtype)
+        else:
+            raise exc
```
bmcfee: It would be good to re-raise this exception so the call stack is preserved.  This can be done using `six.reraise`: https://six.readthedocs.io/#six.reraise

librosa/core/audio.py
carlthome: `warnings` is nice! Just went with `logging` by habit.

librosa/core/audio.py
carlthome: Wow, cool!

bmcfee:
LGTM!  Thanks for all your work on this one!

Issue #911, opened 2019-07-02
Minimum python 3 version


bmcfee:
We previously decided #896 to drop legacy python support starting in v0.8.  One question left open is what the minimum python version we *will* support is going to be.

For the record, numpy officially lists 3.5, 3.6, and 3.7, having dropped 3.4.  Our current test suite does not cover 3.4, and I see no reason to keep it around.

The safest choice is probably 3.5, but I want to make a properly informed decision about this.  For the record, here are links to the "what's new" docs for each of the current and forthcoming python3 versions:

- [3.5](https://docs.python.org/3/whatsnew/3.5.html) (2015)
- [3.6](https://docs.python.org/3/whatsnew/3.6.html) (2016)
- [3.7](https://docs.python.org/3/whatsnew/3.7.html) (2018)
- [3.8](https://docs.python.org/3.8/whatsnew/3.8.html) (TBA)

The main question, I think, boils down to whether we want to make 3.6 the minimum  Advantages to that (relative to 3.5):

- `asyncio` stability
- f-strings
- type hints
- async generators and comprehensions

Where the async stuff mainly could be beneficial for extending the streaming interface going forward.  F-strings are a nice-to-have, but not necessary.  Type hints, also, might be nice in some cases, but most of our type requirements have to do with numpy arrays and value checks, so I don't think they can really be enforced syntactically.

bmcfee:
As an aside, 3.5 is expected to hit EOL in [september 2020](https://devguide.python.org/#status-of-python-branches).  Since I expect 0.8 will probably land just around then --- we usually do point releases at the beginning of summer --- we may want to simply leapfrog up to 3.6+.

PR #949, opened 2019-08-13
Accelerated lag<->recurrence, util.shear, deprecated roll_sparse

### LINKED ISSUE #839: numba accelerate lag matrix
#### Description
numba 0.43 (currently an RC) adds support for numpy roll, which we currently use in a tight loop within the lag<->recurrence functions.  When roll support is released, we could accelerate lag computations (for dense matrices) by numba-wrapping the roll loop.

For sparse matrices, we'd probably have to keep things as is.

### LINKED ISSUE #907: roll_sparse acceleration
#### Description

`util.roll_sparse` does basically the same thing as `np.roll`, but for sparse matrices.  It's primary use case for us is in translating between recurrence and lag matrices with sparse structure.  The current implementation is somewhat inefficient, as it performs full format conversions.

I think we can do this better by working directly on the data structures themselves by manipulating the `.indices` field of csr/csc matrices.  lil/dok/diag might be tricker / impossible to accelerate.

bmcfee:
#### Reference Issue
Fixes #839 and #907 

[EDIT by @lostanlen: fixes #907]

This PR contains a rewrite of the lag<->recurrence conversion functions.

The new version is somewhere between 2 and 6000x faster, but functionally equivalent.

To support this, I added a new function `util.shear` which shears a 2-d array along a given axis at a given slope (positive or negative).  For dense matrices, it uses a numba-accelerated roll loop.  For sparse matrices, it operates directly on compressed index data structures to shear the matrix without moving data.

The old `roll_sparse` function is now deprecated and will be removed in 0.8.

#### Any other comments?

We now require numba >= 0.43 for `np.roll` support.

### DIFF
--- a/librosa/segment.py
+++ b/librosa/segment.py
@@ -518,6 +518,7 @@ def recurrence_to_lag(rec, pad=True, axis=-1):
     --------
     recurrence_matrix
     lag_to_recurrence
+    util.shear
 
     Examples
     --------
@@ -549,40 +550,30 @@ def recurrence_to_lag(rec, pad=True, axis=-1):
 
     sparse = scipy.sparse.issparse(rec)
 
-    roll_ax = None
     if sparse:
-        roll_ax = 1 - axis
-        lag_format = rec.format
-        if axis == 0:
-            rec = rec.tocsc()
-        elif axis in (-1, 1):
-            rec = rec.tocsr()
+        fmt = rec.format
 
     t = rec.shape[axis]
 
-    if sparse:
-        if pad:
+    if pad:
+        if sparse:
             kron = np.asarray([[1, 0]]).swapaxes(axis, 0)
-            lag = scipy.sparse.kron(kron.astype(rec.dtype), rec, format='lil')
+            if axis == 0:
+                rec_fmt = 'csr'
+            else:
+                rec_fmt = 'csc'
+            rec = scipy.sparse.kron(kron.astype(rec.dtype), rec, format=rec_fmt)
         else:
-            lag = scipy.sparse.lil_matrix(rec)
-    else:
-        if pad:
             padding = [(0, 0), (0, 0)]
             padding[(1-axis)] = (0, t)
-            lag = np.pad(rec, padding, mode='constant')
-        else:
-            lag = rec.copy()
+            rec = np.pad(rec, padding, mode='constant')
 
-    idx_slice = [slice(None)] * lag.ndim
-
-    for i in range(1, t):
-        idx_slice[axis] = i
-        lag[tuple(idx_slice)] = util.roll_sparse(lag[tuple(idx_slice)], -i, axis=roll_ax)
+    lag = util.shear(rec, factor=-1, axis=axis)
 
     if sparse:
-        return lag.asformat(lag_format)
-    return np.ascontiguousarray(lag.T).T
+        lag = lag.asformat(fmt)
+
+    return lag
 
 
 def lag_to_recurrence(lag, axis=-1):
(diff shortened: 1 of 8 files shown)


librosa/util/utils.py
```diff
@@ -1708,3 +1715,101 @@ def cyclic_gradient(data, edge_order=1, axis=-1):
...
+    >>> librosa.util.shear(E, factor=-1, axis=0)
+    array([[1., 0., 0.],
+           [1., 0., 0.],
+           [1., 0., 0.]])
+    >>> librosa.util.shear(E, factor=1, axis=-1)
+    array([[1., 0., 0.],
+           [0., 0., 1.],
+           [0., 1., 0.]])
+    '''
+
```
lostanlen: should we check that `factor` is integer here?
calling `shear` with non-integer `factor` is going to lead to an inscrutable error in `np.roll` via `__shear_*`

librosa/segment.py
```diff
@@ -549,40 +550,30 @@ def recurrence_to_lag(rec, pad=True, axis=-1):
...
-            rec = rec.tocsr()
+        fmt = rec.format
 
     t = rec.shape[axis]
 
-    if sparse:
-        if pad:
+    if pad:
+        if sparse:
             kron = np.asarray([[1, 0]]).swapaxes(axis, 0)
```
lostanlen: i don't love the fact that kron is both a variable and a function here

librosa/segment.py
```diff
@@ -549,40 +550,30 @@ def recurrence_to_lag(rec, pad=True, axis=-1):
...
         else:
-            lag = scipy.sparse.lil_matrix(rec)
-    else:
-        if pad:
             padding = [(0, 0), (0, 0)]
             padding[(1-axis)] = (0, t)
-            lag = np.pad(rec, padding, mode='constant')
-        else:
-            lag = rec.copy()
+            rec = np.pad(rec, padding, mode='constant')
```
lostanlen: wait, `mode=constant`? isn't the `kron`-based padding in the `sparse` case equivalent to `mode=wrap`?
why is the padding mode depending on the matrix type here?

librosa/segment.py
```diff
@@ -658,25 +649,12 @@ def lag_to_recurrence(lag, axis=-1):
...
-        roll_ax = 1 - axis
-    else:
-        rec = lag.copy()
-        roll_ax = None
 
-    idx_slice = [slice(None)] * lag.ndim
-    for i in range(1, t):
-        idx_slice[axis] = i
-        rec[tuple(idx_slice)] = util.roll_sparse(lag[tuple(idx_slice)], i, axis=roll_ax)
+    rec = util.shear(lag, factor=+1, axis=axis)
```
lostanlen: amazing!

librosa/segment.py
bmcfee: no; the padding here (sparse or not) is zero-padding.  The idea is to pad out with zeros so that circular shear does not conflate positive and negative lag.  This was not done in, eg, Serra, Mueller, Grosche & Arcos, but it makes a huge difference when you do any analysis on the sheared matrix (eg svd as in my paper with DAn in icassp 2014).

librosa/util/utils.py
bmcfee: Yeah, fair enough.

bmcfee:
>  I just have a doubt regarding constant vs kron-based padding.

TLDR They're doing the same thing; it just happened to be easier to implement sparse padding by a kronecker product with `[1, 0]` since scipy.sparse doesn't have a pad function.

librosa/segment.py
lostanlen: ah, i see. the `(0, 1)` in `kron` acts as a multiplier to get a zero-multiplied replication of the data, i.e. zero padding. neat!

bmcfee:
Build errors are due to the 3.5 conda environment being a mess again.  Le sigh.

bmcfee:
Re: py35 environment

The problem is that numba in the mainline anaconda distribution is still at 0.39 for python 3.5.

The numba channel has up to date packages, but I'm not sure how to coherently manage that wrt the rest of our conda dependency chain, either at distribution or at testing.

On pypi, this shouldn't be an issue, but pip is unable to resolve the environment for 3.5 with the packages already installed via conda.

A short-term fix would be to update the build environment to run `conda install -c numba numba` instead of mainline anaconda.  This will work, but it might obscure the distinction between testing and deployment.

Maybe we need to update the install instructions to point to the latest numba install for 3.5 users?

lostanlen:
or maybe ... https://github.com/librosa/librosa/issues/911 👀

bmcfee:
> or maybe ... #911 eyes

Yeah, this issue will go away next year, if not sooner when the next anaconda mainline distribution rolls out.  But for now, I put in the hack for travis and updated the install instructions accordingly.

PR #950, opened 2019-08-14
Allow initial phase estimates for griffinlim methods

### LINKED ISSUE #948: Ability to provide initial phase to Griffin-Lim
FEATURE REQUEST

Librosa has a Griffin-Lim implementation to reconstruct audio from spectrogram magnitudes. As far as I can see, it takes only the magnitudes and estimates the phase from scratch in the process:

https://librosa.github.io/librosa/generated/librosa.core.griffinlim.html

But in tasks such as source separation, often the solution converges much more quickly/becomes better when using the mixture phase for the estimation of the source audio output. So it would be cool if another parameter could be added to the function allowing you to optionally pass in an initial guess of the phase:

```
y, sr = librosa.load(librosa.util.example_audio_file(), duration=5, offset=30)
# Get the magnitude spectrogram
mag, phase = librosa.core.magphase(librosa.stft(y))
# Invert using Griffin-Lim starting with phase as initial phase estimate
y_inv = librosa.griffinlim(mag, initial_phase=phase)
```

bmcfee:
#### Reference Issue
Fixes #948 

This PR adds a new parameter `init=` to Griffin-Lim methods.  Supported modes are `None` and `'random'` (default).  If `init=None`, then the initial phase estimate is taken from the spectrogram input.  If the input is a magnitude spectrogram, this is equivalent to initializing with all-zeros.  However, if you have a good phase guess already, you can pass in a complex-valued `S` and the phase will be taken directly.

Under the hood, this is implemented by setting `angles[:] = 1` when `init is None`, so that `S * angles` preserves phase.  This avoids any tricky type-based dispatching, and lets us keep the algorithm exactly as it was before.

As an added bonus, this turns our GL implementation into an any-time algorithm, and computation can be resumed from a previous stopping point.  For instance:

```python
y = griffinlim(S, ..., n_iter=10)

# Is y good enough?  No?  Keep going...
S = librosa.stft(y, ...)
y = griffinlim(S, ..., n_iter=10, init=None)
```

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -851,7 +851,7 @@ def __num_two_factors(x):
 def griffinlim_cqt(C, n_iter=32, sr=22050, hop_length=512, fmin=None, bins_per_octave=12, tuning=0.0,
                    filter_scale=1, norm=1, sparsity=0.01, window='hann', scale=True,
                    pad_mode='reflect', res_type='kaiser_fast', dtype=np.float32,
-                   length=None, momentum=0.99, random_state=None):
+                   length=None, momentum=0.99, init='random', random_state=None):
     '''Approximate constant-Q magnitude spectrogram inversion using the "fast" Griffin-Lim
     algorithm [1]_ [2]_.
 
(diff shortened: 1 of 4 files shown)


lostanlen:
nice!
i'm unclear about the semantics of `init` being either `"random"` (a string) or `None`. 
If only two values are allowed, why should they be of different type? Why not a boolean `random_init=True`?

bmcfee:
> If only two values are allowed, why should they be of different type? Why not a boolean `random_init=True`?

I want to leave open the possibility of other initialization schemes in the future, and this API seems like the simplest way to do that without adding too much complexity.

lostanlen:
LGTM!

PR #953, opened 2019-08-14
Deprecated ifgram

bmcfee:

This PR deprecates `ifgram` in favor of the new `reassigned_spectrogram` function merged in #926 .

It also fixes a few spots of documentation that were not updated in the #926 merge but should have been.  No functionality changes here.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -18,6 +18,7 @@ from .audio import resample
 from .._cache import cache
 from .. import util
 from ..util.exceptions import ParameterError
+from ..util.decorators import deprecated
 from ..filters import get_window, semitone_filterbank
 from ..filters import window_sumsquare
 
@@ -132,7 +133,7 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
     --------
     istft : Inverse STFT
 
-    ifgram : Instantaneous frequency spectrogram
+    reassigned_spectrogram : Time-frequency reassigned STFT
 
 
     Notes
@@ -406,6 +407,7 @@ def __overlap_add(y, ytmp, hop_length):
         y[sample:(sample + n_fft)] += ytmp[:, frame]
 
 
+@deprecated('0.7.1', '0.8.0')
 def ifgram(y, sr=22050, n_fft=2048, hop_length=None, win_length=None,
            window='hann', norm=False, center=True, ref_power=1e-6,
            clip=True, dtype=np.complex64, pad_mode='reflect'):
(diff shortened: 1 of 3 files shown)


librosa/core/spectrum.py
```diff
@@ -92,7 +93,7 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
     --------
     istft : Inverse STFT
 
-    ifgram : Instantaneous frequency spectrogram
+    reassigned_spectrogram : Time-frequency reassigned STFT
```
lostanlen: STFT magnitudes after time-frequency reassignment


(time-frequency is established in the literature as an adjective, not an adverb. also it's unclear that phases are no longer available)

bmcfee: Er, reassigned_spectrogram does return phase?

bmcfee: !! you're right, nevermind.

Why did we do that?  Shouldn't reassigned_spectrogram return the stft instead of magnitudes?

lostanlen: no, it doesn't. which is why we call it a "reassigned spectrogram", not a "reassigned STFT".
my suggestion is to insist on the fact that reassigned_spectrogram returns STFT magnitudes, not complex-valued coefficients

bmcfee: > my suggestion is to insist on the fact that reassigned_spectrogram returns STFT magnitudes, not complex-valued coefficients

We could also just change it to return complex valued coefficients, so that feature parity is maintained with the deprecated ifgram.  Since it hasn't been released yet, we still have some flexibility to change things.

I guess I really dropped the ball on CR for that one -- is it standard for TFR to produce magnitudes instead of proper STFTs?  I can't imagine a good reason for that, but maybe my imagination isn't very good.

lostanlen: Yes, it's standard for time-frequency reassignment to only return the magnitudes. The phase is ill-defined after time-frequency reassignment. The TFR is not a bilinear time-frequency distribution in the sense of the Cohen's class because its smoothing kernel with respect to the Wigner-Ville is not positive.

bmcfee: Ok, TIL! Thanks for clarifying.

bmcfee:
I think this one is now good to go-- @lostanlen?

lostanlen:
Right. Again, i don't love the "time-frequency reassigned spectrogram" terminology which makes it sound like "time-frequency" is an adverb. If you really want to avoid the pleonasm: (`reassigned_spectrogram: reassigned spectrogram`), i propose something like `STFT magnitudes after time-frequency reassignment`.

PR #957, opened 2019-08-15
times_like audit

### LINKED ISSUE #954: Audit for times_like usage
#### Description

As mentioned in the CR for #926, there are a few points in the code that do old-style explicit frames-to-time conversion like `frames_to_time(np.arange(len(foo)))` rather than the new-style `times_like(foo)` introduced in 0.6.  We should do a quick audit for both code and docstring examples to make sure everything is in good style.

bmcfee:
#### Reference Issue
Fixes #954 

This PR mostly consists of modernization of docstring examples to use `times_like(X)` instead of the older style of `frames_to_time(np.arange(len(X)))` (or something morally equivalent).

There were a couple of spots in the code (display module, mostly) that also used the old style, and these have been modernized as well.

#### Any other comments?

In addition to modernization, I noticed a few badly formatted docstrings while building docs to test the changes in this PR.  I've fixed those here, even though they're out of the specific scope of #954.

No major CR needs here, IMO.  It should be good to go once CI passes.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -34,24 +34,25 @@ __all__ = ['stft', 'istft', 'magphase', 'iirt', 'ifgram',
 @cache(level=20)
 def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
          center=True, dtype=np.complex64, pad_mode='reflect'):
-    """Short-time Fourier transform (STFT). [1, chapter 2]
-    
+    """Short-time Fourier transform (STFT). [1]_ (chapter 2)
+
     The STFT represents a signal in the time-frequency domain by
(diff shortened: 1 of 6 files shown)


librosa/core/spectrum.py
```diff
@@ -34,24 +34,25 @@
 @cache(level=20)
 def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
          center=True, dtype=np.complex64, pad_mode='reflect'):
-    """Short-time Fourier transform (STFT). [1, chapter 2]
-    
+    """Short-time Fourier transform (STFT). [1]_ (chapter 2)
```
lostanlen: um, my bad. sorry about that

PR #963, opened 2019-08-16
support more x_axis values in waveplot

### LINKED ISSUE #737: Display axis decorators: seconds, ms

### LINKED ISSUE #962: waveplot xlabel lacks physical units and fails for `x_axis='ms'`
#### Description
This came up while reviewing #960, but is largely unrelated.
It's in the wake of PR #760, which fixed #737. Thanks again to @kyungyunlee for her great work.

#### Steps/Code to Reproduce
The `waveplot` docstring:

```python
y, sr = librosa.load(librosa.util.example_audio_file(), duration=10)
plt.figure()
plt.subplot(3, 1, 1)
librosa.display.waveplot(y, sr=sr)
plt.title('Monophonic')
```

#### Expected Results
I would hope to see an xlabel that says "Time (s)" rather than "Time".

waveplot_after

#### Actual Results
waveplot_before

### But there's more!
`waveplot` only allows `x_axis` to be: {'time', 'off', 'none'}. So: there no way to show miliseconds, and there's no way to know what unit `time` maps adaptively choses (this is a side effect of #760).

Example:

```python
y, sr = librosa.load(librosa.util.example_audio_file(), duration=10)
plt.figure()
plt.subplot(3, 1, 1)
librosa.display.waveplot(y, sr=sr, x_axis='ms')
plt.title('Monophonic')
```

### Expected results

waveplot_ms

[EDIT: um, actually, this is _not_ the expected result ...]

### Actual results

```python

ParameterError                            Traceback (most recent call last)
<ipython-input-8-e629e1e7e58d> in <module>
      3 plt.figure()
      4 plt.subplot(3, 1, 1)
----> 5 librosa.display.waveplot(y, sr=sr, x_axis='ms')
      6 plt.title('Monophonic')
      7 plt.xlabel("Time (s)")

~/librosa-dev/librosa/librosa/display.py in waveplot(y, sr, max_points, x_axis, offset, max_sr, ax, **kwargs)
    488         axes.set_xticks([])
    489     else:
--> 490         raise ParameterError('Unknown x_axis value: {}'.format(x_axis))
    491 
    492     return out

ParameterError: Unknown x_axis value: ms
```

I can prepare a PR right now.

#### Versions
librosa 0.7.0
matplotlib 3.1.0

lostanlen:
#### Reference Issue
Fixes #962

See #962 for a description of the issue.
This was a very easy fix. It's essentially a re-use of PR #760, which cl0sed #737.

#### Any other comments?

The addition of this new feature doesn't affect the test coverage. I guess i don't need to add more tests, do i?

Also please let me know if i'm trespassing backtick etiquette in the docstring.


lostanlen:
ah, `Image files did not match.`
Will make new images tomorrow

librosa/display.py
```diff
@@ -386,8 +386,18 @@ def waveplot(y, sr=22050, max_points=5e4, x_axis='time', offset=0.0,
...
-        If 'time', the x-axis is given time tick-marks.
+    x_axis : str or None
+        Display of the x-axis ticks and tick markers. Accepted values are:
+        - 'time' : markers are shown as milliseconds, seconds,
+          minutes, or hours.
+          Values are plotted in units of seconds.
+        - 's' : markers are shown as seconds.
+        - 'ms' : markers are shown as milliseconds.
+        - 'lag' : like time, but past the halfway point counts
+          as negative values.
```
bmcfee: I'm not sure that lag formatters make sense for waveplot?  I guess it doesn't hurt to include, but it seems strange to me.

lostanlen: that's true. But one application for which it could be useful would be to show window functions.

bmcfee: Good point, I hadn't considered that.

lostanlen:
No, this is wrong. This image shows "ms" instead of "s":
https://github.com/librosa/librosa/blob/be84495c81bb5361d5d67af838b041c4d8758438/tests/baseline_images/test_display/test_sharex_specshow_ms.png

And this other one shows scientific notation: https://github.com/librosa/librosa/blob/be84495c81bb5361d5d67af838b041c4d8758438/tests/baseline_images/test_display/test_sharex_waveplot_ms.png

:/

bmcfee:
> No, this is wrong. This image shows "ms" instead of "s":
> https://github.com/librosa/librosa/blob/be84495c81bb5361d5d67af838b041c4d8758438/tests/baseline_images/test_display/test_sharex_specshow_ms.png

Ah, tricky problem.  What's going on here is that specshow is rendered first (correctly) with `ms` units, and waveplot is called with the default x_axis parameter.  This clobbers the ticks of the first plot, but not the axis label.  I don't think there's an elegant solution to this.



> And this other one shows scientific notation: https://github.com/librosa/librosa/blob/be84495c81bb5361d5d67af838b041c4d8758438/tests/baseline_images/test_display/test_sharex_waveplot_ms.png

It's supposed to though.  ms format uses `'{:.3g}'`, so this part seems correct (if ugly) to me.  However, the axis label is still incorrect like the other example, for the same reason.

What do you want to do about these test cases?  As far as I can tell, everything is working correctly, and the only error is that axis labels are out of sync.

lostanlen:
OK, so IIUC this is not stricto sensu a test failure, just a weird usage pattern? In this case, i would vote for amending these tests so that they give a more genuine idea of what's supposed to happen from the user perspective.

bmcfee:
Ok, I revised the tests a little.  Misleading axis labels are removed, so now it's just making sure that the default units from subplot 1 are changed to those of subplot 2.

lostanlen:
LGTM! :shipit: 
thank you so much for your help

PR #966, opened 2019-08-16
multidimensional framing and contiguous stacking

### LINKED ISSUE #944: multi-dimensional framing
#### Description
In machine listening applications, it is very common to train a classifier on short audio clips and then deploy it on full-duration audio recordings. In this case, what is necessary to do is to divide the long-duration spectrogram into time-frequency buffers, loop over these buffers, compute a prediction on each of them, and aggregate the predictions over time into a single time series.

Some examples:
* @jtcramer et al. Open-L3: https://github.com/marl/openl3
* @lostanlen et al. BirdVoxDetect: https://github.com/BirdVox/birdvoxdetect
* @rabitt et al. deep salience: https://github.com/rabitt/ismir2017-deepsalience
* dozens, possibly hundreds more among librosa users.

Currently, this is not possible in librosa. We have the cherished `librosa.util.frame` for monophonic audio, but nothing for matrices.

My current code implementation BirdVoxDetect is very opaque, and probably suboptimal, but works miraculously:

```python
        X_shape = (n_strides, clip_length, n_freqs)
        X_stride = (itemsize*n_freqs*stride_length, itemsize*n_freqs, itemsize)
        X_pcen = np.lib.stride_tricks.as_strided(
            np.ravel(np.copy(pcen).T),
            shape=X_shape,
            strides=X_stride,
            writeable=False)
        # Prepare for Keras training
        X_pcen = np.transpose(X_pcen, (0, 2, 1))[:, :, :, np.newaxis]
```

Writing this, with some help from @jongwook, accelerated BirdVoxDetect by an order of magnitude. This is what makes me think that this feature would be vital for machine listening applications of librosa. I would expect to call this with the following prototype:

```python
S_buffers = librosa.util.buffer(S=S, frame_length, hop_length)
```
where `S` is a matrix of size `(n_freqs, n_times)` and `S_buffer` a matrix of size `(n_freqs, frame_length, n_times/hop_length)`.

Doing it with the current implementation of `librosa.util.frame` would break numpy contiguity, and thus defeat the purpose of having a view-based in-place implementation.

Bonus points if it works on multivariate representations like HCQT or multiscale PCEN.

@ all: Let me know what you think!

#### Versions
librosa 0.7.0

### LINKED ISSUE #961: Constructing multi-channel audio buffers
#### Description
In testing for #960 , we discovered that waveplot tests are passing when they should fail.  The cause is subtle: waveplot makes stereo envelopes by framing each channel separately, which means that a stereo signal needs to be C-contiguous for this to work.  This was implicit behavior before, and #960 ensures that multi-channel audio is F-contiguous.  However, the test still passes because the fixture for stereo waveplot constructs stereo signals by vstacking two mono signals, which ends up producing C-contiguous data.

The point of this story is that vstacking mono signals to get stereo sort of works, but it's not the "right" way to do it: the byte ordering gets messed up.  There isn't an obviously easy way to do this properly, so we should make one.

We already have `to_mono` for downmixing stereo (or general multichannel) to mono by averaging.  It stands to reason that we should have an easy way to go the opposite direction: combine multiple input signals to a single multi-channel array with proper ordering.

The obvious name here is `to_stereo`, but that's not quite right if we want multichannel.  `to_multichannel` is a bit of a mouthful, but could work.  Anyone have other suggestions?  Maybe `upmix`, and we can have an alias of `downmix = to_mono`?  I worry about people expecting fancier downmixing behaviors though (eg 5.1 -> stereo).

bmcfee:
#### Reference Issue
- [x] Fixes #944 
- [x] Fixes #961 
- [x] partially addresses #967 

This PR contains a rewrite of `util.frame` to support multidimensional framing and target axes.

The documentation has been expanded, and unit tests rewritten as well.

#### Checklist

- [x] rewrite frame
- [x] implement `stack` to fix problems with display plots that we identified in the run-up to this issue.
- [x] update old code to make better use of `frame` (eg in waveplot, possibly elsewhere)
- [x] update old code for contiguity


bmcfee:
Ok, I think this one's ready for a CR now.

lostanlen:
thanks!
i can't review today, i'm sorry. But i will ASAP

bmcfee:
Note: this pr does not finish #967 -- it just makes sure that concatenation-type operations preserve the expected ordering.

I think a larger PR for #967 is in order that would add contiguity tests to all feature extractors and transformations.

librosa/display.py
```diff
@@ -451,6 +455,10 @@ def waveplot(y, sr=22050, max_points=5e4, x_axis='time', offset=0.0,
     target_sr = sr
     hop_length = 1
 
+    # Pad an extra channel dimension, if necessary
+    if y.ndim == 1:
+        y = util.stack([y], axis=0)
```
lostanlen: Neat! So this turns `(n_samples,)` into `(1, n_samples)` without breaking F-contiguity? Is there a reason why this couldn't be achieved with `.reshape(1, -1)`?

librosa/display.py
```diff
@@ -460,17 +468,11 @@ def waveplot(y, sr=22050, max_points=5e4, x_axis='time', offset=0.0,
 
         hop_length = sr // target_sr
 
-        if y.ndim == 1:
-            y = __envelope(y, hop_length)
-        else:
-            y = np.vstack([__envelope(_, hop_length) for _ in y])
+    # Reduce by envelope calculation
+    y = __envelope(y, hop_length)
```
lostanlen: a net reduction in LoCs! and less special-casing 🎉

librosa/util/utils.py
```diff
@@ -35,80 +35,168 @@
...
+def frame(x, frame_length=2048, hop_length=512, axis=-1):
+    '''Slice a data array into (overlapping) frames.
 
     This implementation uses low-level stride manipulation to avoid
-    redundant copies of the time series data.
+    making a copy of the data.  The resulting frame representation
+    is a new view of the input data.
+
+    For example, a one-dimensional input `x = [0, 1, 2, 3, 4, 5, 6]`
+    can be framed with frame-length 3 and hop length 2 in two ways.
```
lostanlen: frame length

librosa/util/utils.py
```diff
@@ -35,80 +35,168 @@
...
+        `[[0, 1, 2],
+          [2, 3, 4],
+          [4, 5, 6]]`
+
+    where each row `x_frames[i]` contains a contiguous slice of the input.
+
+    This generalizes to higher dimensional inputs, as shown in the examples below.
+    In general, the framing operation increments by 1 the number of dimensions,
+    adding a new "frame axis" either to the end of the array (`axis=-1`)
+    or the beginning of the array (`axis=0`).
```
lostanlen: super clear!

librosa/util/utils.py
```diff
@@ -35,80 +35,168 @@
...
 
     if hop_length < 1:
         raise ParameterError('Invalid hop_length: {:d}'.format(hop_length))
 
-    if not y.flags['C_CONTIGUOUS']:
-        raise ParameterError('Input buffer must be contiguous.')
+    n_frames = 1 + (x.shape[axis] - frame_length) // hop_length
+    strides = np.asarray(x.strides)
+
+    if axis == -1:
```
lostanlen: what about `axis=x.ndims`? would that be a synonym of `axis=-1`?

librosa/util/utils.py
```diff
@@ -35,80 +35,168 @@
...
 
+    Frame a stereo signal:
+
+    >>> y, sr = librosa.load(librosa.util.example_audio_file(), mono=False)
+    >>> y.shape
+    (2, 1355168)
+    >>> frames = librosa.util.frame(y, frame_length=2048, hop_length=64)
+    (2, 2048, 21143)
+
+    Carve an STFT into fixed-length patches of 32 frames with no overlap
```
lostanlen: i would prefer an example with 50% overlap. For the user, getting it to work without overlap is simpler (`np.reshape`), as it does not involve `stride_tricks`. Does that make sense?

librosa/util/utils.py
```diff
@@ -35,80 +35,168 @@
...
+    See Also
+    --------
+    np.asfortranarray : Convert data to F-contiguous representation
+    np.ascontiguousarray : Convert data to C-contiguous representation
+    np.ndarray.flags : information about the memory layout of a numpy `ndarray`.
 
     Examples
     --------
-    Extract 2048-sample frames from `y` with a hop of 64 samples per frame
+    Extract 2048-sample frames from monophonic `y` with a hop of 64 samples per frame
```
lostanlen: is that a realistic use case? 97% overlap?

librosa/util/utils.py
```diff
@@ -35,80 +35,168 @@
...
 
-    # Vertical stride is one sample
-    # Horizontal stride is `hop_length` samples
-    y_frames = as_strided(y, shape=(frame_length, n_frames),
-                          strides=(y.itemsize, hop_length * y.itemsize))
-    return y_frames
+    elif axis == 0:
+        if not x.flags['C_CONTIGUOUS']:
+            raise ParameterError('Input array must be C-contiguous '
+                                 'for framing along axis={}'.format(axis))
```
lostanlen: you can just write `'axis=0'` rather than `'axis={}'.format(axis)` in this case

librosa/util/utils.py
```diff
@@ -35,80 +35,168 @@
...
+    strides = np.asarray(x.strides)
+
+    if axis == -1:
+        if not x.flags['F_CONTIGUOUS']:
+            raise ParameterError('Input array must be F-contiguous '
+                                 'for framing along axis={}'.format(axis))
 
-    # Compute the number of frames that will fit. The end may get truncated.
-    n_frames = 1 + int((len(y) - frame_length) / hop_length)
+        shape = list(x.shape)[:-1] + [frame_length, n_frames]
```
lostanlen: maybe convert explicitly to tuple?

librosa/util/utils.py
```diff
@@ -1815,3 +1903,109 @@ def shear(X, factor=1, axis=-1):
...
+      UPDATEIFCOPY : False
+    '''
+
+    shapes = {arr.shape for arr in arrays}
+    if len(shapes) > 1:
+        raise ParameterError('all input arrays must have the same shape')
+    elif len(shapes) < 1:
+        raise ParameterError('at least one input arrays must be provided for stack')
+
+    shape_in = shapes.pop()
```
lostanlen: very nice that this is checked through a `set` construction. i should use `set`'s more often.

librosa/util/utils.py
```diff
@@ -1815,3 +1903,109 @@ def shear(X, factor=1, axis=-1):
...
+      ALIGNED : True
+      WRITEBACKIFCOPY : False
+      UPDATEIFCOPY : False
+    '''
+
+    shapes = {arr.shape for arr in arrays}
+    if len(shapes) > 1:
+        raise ParameterError('all input arrays must have the same shape')
+    elif len(shapes) < 1:
+        raise ParameterError('at least one input arrays must be provided for stack')
```
lostanlen: one input array

librosa/core/spectrum.py
```diff
@@ -1471,7 +1471,7 @@ def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
         y_resampled.append(resample(y, sr, cur_sr))
 
     # Compute the number of frames that will fit. The end may get truncated.
-    n_frames = 1 + int((len(y) - win_length) / float(hop_length))
+    n_frames = 1 + int((len(y) - win_length) // float(hop_length))
```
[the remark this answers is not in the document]
bmcfee: just a style change that caught my eye while editing

librosa/display.py
bmcfee: ah good catch; reshape would be more efficient here

librosa/util/utils.py
bmcfee: 3ms hop gets used in onset detection, yeah

librosa/util/utils.py
bmcfee: tuple wouldn't allow mutation here, and it gets (implicitly) cast below anyway

librosa/util/utils.py
bmcfee: this is stolen directly from np.stack :grin:

librosa/util/utils.py
bmcfee: technically yes, but -1 is a more standard convention for this sort of thing (see eg np.stack)

librosa/util/utils.py
bmcfee: true -- it doesn't hurt though

bmcfee:
> I guess the only question i have is whether we should support `axis=x.ndim` as a synonym for `axis=-1`.

I don't think that we should.  axis=-1 is clear enough, and I don't think we need more than one way to do it here.

bmcfee:
Regression here: `stack([y]) != y[np.newaxis, :]` .. will look into later

bmcfee:
> Regression here: `stack([y]) != y[np.newaxis, :]` .. will look into later

Aha, the problem is that a `newaxis` index makes a stride of 0.  Frame takes the product of strides to compute the stride of the new axis, which screws everything up.

Solution here should be to make frame take the product of non-zero strides.  I think this leads to correct behavior, but it's subtle.

bmcfee:
I think this one's :shipit: -- @lostanlen?

librosa/util/utils.py
lostanlen: you're right. upon reflection, it would be a bad idea to support more cases than `0` and `-1`.

PR #971, opened 2019-08-18
Beat and tempo priors

### LINKED ISSUE #880: Allow beat / tempo models to accept arbitrary priors
#### Description
The beat tracker / tempo estimation implementation API allows a user to specify a location and scale of a log-normal distribution that is used as a prior over tempo.  The inference then estimates a posterior pseudo-probability by multiplying the prior probabilities against the tempogram / averaged tempogram to pick the maximum.

In implementing #739, the reference implementation uses a uniform prior (not log-normal) over a tempo range for a similar purpose.  It would be great to have a unified API for both methods so that we can have consistency by default and backward compatibility when needed.

I think the easiest way to achieve this is to provide a `prior=` parameter to the tempo estimator (and downstream functions) which accepts a `scipy.stats` model object that can override the default location/scale parameters.  This gets us out of having to provide interfaces for every different choice of priors.  Observe:

```python
# With a log-normal prior, not-quite equivalent to current implementation
>>> prior = scipy.stats.lognorm(1, loc=np.log(120.0))
>>> librosa.beat.tempo(tempogram, prior=prior)
# With a uniform prior U[30, 300] equiv to the PLP reference
>>> prior = scipy.stats.uniform(30, 300)
>>> librosa.beat.tempo(tempogram, prior=prior)
```

I think this is a pretty easy change that can be rolled into #739 and give us a bit more flexibility.

bmcfee:
#### Reference Issue
Fixes #880 

This PR adds the ability to specify arbitrary prior distributions in tempo estimation.  This is done by constructing a `scipy.stats` frozen RV object, eg as follows:

```python
>>> prior = scipy.stats.uniform(60, 240)
```
and passing it as a parameter to the estimator:

```python
>>> tempo = librosa.beat.tempo(y=y, sr=sr, prior=prior)
```

To-do:

- [x] tests

#### Any other comments?

This PR also converts the tempo inference to log-probability space.  This should make things a bit more numerically stable going forward.

A few things that cropped up in the development of this PR:

1. The tempo estimator's prior distribution is *almost* log-normal, but not quite.  It's off by a critical scaling factor, and always has been.  I decided against "fixing" this here, for the sake of backwards compatibility of defaults, but it really should be changed in the next major revision.
2. The tempo estimation in general has poor frequency resolution exactly in the range we care about:
```python
In [5]: librosa.tempo_frequencies(384)[:40]                                                                                                                                                   
Out[5]: 
array([          inf, 2583.984375  , 1291.9921875 ,  861.328125  ,
        645.99609375,  516.796875  ,  430.6640625 ,  369.140625  ,
        322.99804688,  287.109375  ,  258.3984375 ,  234.90767045,
        215.33203125,  198.76802885,  184.5703125 ,  172.265625  ,
        161.49902344,  151.99908088,  143.5546875 ,  135.99917763,
        129.19921875,  123.046875  ,  117.45383523,  112.34714674,
        107.66601562,  103.359375  ,   99.38401442,   95.703125  ,
         92.28515625,   89.10290948,   86.1328125 ,   83.35433468,
         80.74951172,   78.30255682,   75.99954044,   73.828125  ,
         71.77734375,   69.83741554,   67.99958882,   66.25600962])
In [6]: librosa.fourier_tempo_frequencies()[:40]                                                                                                                                              
Out[6]: 
array([  0.        ,   6.72912598,  13.45825195,  20.18737793,
        26.91650391,  33.64562988,  40.37475586,  47.10388184,
        53.83300781,  60.56213379,  67.29125977,  74.02038574,
        80.74951172,  87.4786377 ,  94.20776367, 100.93688965,
       107.66601562, 114.3951416 , 121.12426758, 127.85339355,
       134.58251953, 141.31164551, 148.04077148, 154.76989746,
       161.49902344, 168.22814941, 174.95727539, 181.68640137,
       188.41552734, 195.14465332, 201.8737793 , 208.60290527,
       215.33203125, 222.06115723, 228.7902832 , 235.51940918,
       242.24853516, 248.97766113, 255.70678711, 262.43591309])
```
As you can see, those bin spacings are huge under default parameters, and are probably the cause of some sub-optimal tempo estimation performance.  This also should be fixed in a future iteration, but it's out of scope for this PR.

3. The tempo inference in general is not a well-formed probabilistic model.  Rather, we treat (log) tempogram energies as if they were unnormalized probabilities, add the log-prior, and do MAP inference.  This could be improved substantially by defining a proper statistical model of autocorrelation, but again, this is out of scope for this issue.

### DIFF
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -320,12 +320,12 @@ def tempo(y=None, sr=22050, onset_envelope=None, hop_length=512, start_bpm=120,
     bpms = core.tempo_frequencies(tg.shape[0], hop_length=hop_length, sr=sr)
 
     # Weight the autocorrelation by a log-normal distribution
-    prior = np.exp(-0.5 * ((np.log2(bpms) - np.log2(start_bpm)) / std_bpm)**2)
+    logprior = -0.5 * ((np.log2(bpms) - np.log2(start_bpm)) / std_bpm)**2
 
     # Kill everything above the max tempo
     if max_tempo is not None:
         max_idx = np.argmax(bpms < max_tempo)
-        prior[:max_idx] = 0
+        logprior[:max_idx] = -np.inf
 
     # Really, instead of multiplying by the prior, we should set up a
     # probabilistic model for tempo and add log-probabilities.
@@ -334,7 +334,7 @@ def tempo(y=None, sr=22050, onset_envelope=None, hop_length=512, start_bpm=120,
     # it would also make time aggregation much more natural
 
     # Get the maximum, weighted by the prior
-    best_period = np.argmax(tg * prior[:, np.newaxis], axis=0)
+    best_period = np.argmax(np.log(tg + 1e-10) + logprior[:, np.newaxis], axis=0)
 
     tempi = bpms[best_period]
     # Wherever the best tempo is index 0, return start_bpm
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -51,7 +51,7 @@ def beat_track(y=None, sr=22050, onset_envelope=None, hop_length=512,
         sampling rate of `y`
 
     onset_envelope : np.ndarray [shape=(n,)] or None
-        ((optional) pre-computed onset strength envelope.
+        (optional) pre-computed onset strength envelope.
 
     hop_length : int > 0 [scalar]
         number of audio samples between successive `onset_envelope` values
@@ -71,7 +71,6 @@ def beat_track(y=None, sr=22050, onset_envelope=None, hop_length=512,
 
     prior      : scipy.stats.rv_continuous [optional]
         An optional prior distribution over tempo.
-
         If provided, `start_bpm` will be ignored.
 
     units : {'frames', 'samples', 'time'}
@@ -97,8 +96,7 @@ def beat_track(y=None, sr=22050, onset_envelope=None, hop_length=512,
     Raises
     ------
     ParameterError
-        if neither `y` nor `onset_envelope` are provided
-
+        if neither `y` nor `onset_envelope` are provided,
         or if `units` is not one of 'frames', 'samples', or 'time'
 
     See Also
@@ -245,7 +243,6 @@ def tempo(y=None, sr=22050, onset_envelope=None, hop_length=512, start_bpm=120,
     prior : scipy.stats.rv_continuous [optional]
         A prior distribution over tempo (in beats per minute).
         By default, a pseudo-log-normal prior is used.
-
         If given, `start_bpm` and `std_bpm` will be ignored.
 
     Returns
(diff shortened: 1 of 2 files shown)


bmcfee:
Tests are in place, and I took the opportunity to modernize the test structure a little bit.

The "tests" here are extremely light: mostly just checking that nothing breaks, but nothing really verifies that the prior distributions get used.  I couldn't think of a simple way to do this, which is not a great sign, but OTOH the altered code here is extremely simple so it probably doesn't matter.

librosa/beat.py
```diff
@@ -50,7 +51,7 @@ def beat_track(y=None, sr=22050, onset_envelope=None, hop_length=512,
         sampling rate of `y`
 
     onset_envelope : np.ndarray [shape=(n,)] or None
-        (optional) pre-computed onset strength envelope.
+        ((optional) pre-computed onset strength envelope.
```
lostanlen: spurious opening paren

librosa/beat.py
```diff
@@ -68,6 +69,11 @@ def beat_track(y=None, sr=22050, onset_envelope=None, hop_length=512,
         (optional) If provided, use `bpm` as the tempo instead of
         estimating it from `onsets`.
 
+    prior      : scipy.stats.rv_continuous [optional]
+        An optional prior distribution over tempo.
+
```
lostanlen: spurious newline

librosa/beat.py
```diff
@@ -235,6 +242,12 @@ def tempo(y=None, sr=22050, onset_envelope=None, hop_length=512, start_bpm=120,
         Aggregation function for estimating global tempo.
         If `None`, then tempo is estimated independently for each frame.
 
+    prior : scipy.stats.rv_continuous [optional]
+        A prior distribution over tempo (in beats per minute).
+        By default, a pseudo-log-normal prior is used.
+
```
lostanlen: spurious newline

librosa/beat.py
```diff
@@ -320,30 +354,24 @@ def tempo(y=None, sr=22050, onset_envelope=None, hop_length=512, start_bpm=120,
...
-    # Really, instead of multiplying by the prior, we should set up a
-    # probabilistic model for tempo and add log-probabilities.
-    # This would give us a chance to recover from null signals and
-    # rely on the prior.
-    # it would also make time aggregation much more natural
+        logprior[:max_idx] = -np.inf
 
     # Get the maximum, weighted by the prior
-    best_period = np.argmax(tg * prior[:, np.newaxis], axis=0)
+    best_period = np.argmax(np.log(tg + 1e-10) + logprior[:, np.newaxis], axis=0)
```
lostanlen: i'm not a big fan of `np.log(x + 1e-10)`
consider replacing by `np.log1p(1e10 * x)`, which is the same up to an additive constant (not affecting the argmax), and more numerically stable for small `x`.

librosa/beat.py
```diff
@@ -476,19 +521,21 @@ def plp(y=None, sr=22050, onset_envelope=None, hop_length=512,
...
 
-    # TODO: in the future, generalize this to support arbitrary priors
     if tempo_min is not None:
         ftgram[tempo_frequencies < tempo_min] = 0
     if tempo_max is not None:
         ftgram[tempo_frequencies > tempo_max] = 0
 
     # Step 3: Discard everything below the peak
-    ftmag = np.abs(ftgram)
+    ftmag = np.log(1e-10 + np.abs(ftgram))
```
lostanlen: same remark as above re `log1p`

librosa/beat.py
bmcfee: good catch

bmcfee:
Thanks for the CR @lostanlen -- I'll circle back on this when I have a moment.

bmcfee:
Ok, fixed up the issues from @lostanlen's CR.  I think this is :shipit: once CI passes.

PR #969, opened 2019-08-18
Compat fix: Fix core.load fallback condition

r9y9:
Property fallback to audioread when OSError is raised in soundfile due to missing libsndfile binary dependency (like https://github.com/bastibe/SoundFile/issues/118)

#### Reference Issue

Supports https://github.com/librosa/librosa/issues/845
Ref: https://github.com/librosa/librosa/pull/847

#### What does this implement/fix? Explain your changes.

Fallback to old audioeread backend when `OSError` is raised in sndfile.

#### Any other comments?

I know the notice below, but I'd prefer a non-breaking approach if possible.

https://github.com/librosa/librosa/tree/b1e41e83a9d177d7ca83bf7e0160c5e3c4177e3b#soundfile
> If you're using pip on a Linux environment, you may need to install libsndfile manually. Please refer to the SoundFile installation documentation for details.

PR #972, opened 2019-08-19
Make soundfile optional at librosa import-time.

### LINKED ISSUE #970: soundfile raises OSError due to missing libsndfile dependency
#### Description
Originally in: https://github.com/bastibe/SoundFile/issues/118
Related: #847 which closed #845

#### Steps/Code to Reproduce
Described by @r9y9 in #969

#### Versions
https://github.com/librosa/librosa/tree/b1e41e83a9d177d7ca83bf7e0160c5e3c4177e3b#soundfile
librosa 0.7.0

r9y9:
- Remove global `import soundfile as sf` in audio.py, as it raises `OSError` at import-time when libsndfile is not available on a system. See https://github.com/bastibe/SoundFile/blob/161e930da9c9ea76579b6ee18a131e10bca8a605/soundfile.py#L139-L151 for details.
- Add `import soundfile as sf` to functions that actually require
soundfile
- Handle OSError as well as RuntimError possibly thrown by soundfile

#### Reference Issue

Fixes #970
Improved version of #969

Trying to fix import-time failture described in https://github.com/librosa/librosa/pull/969#issuecomment-522383302 when librosa is property installed by libsndfile is not available.

#### Any other comments?

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -7,7 +7,6 @@ import six
 import sys
 import warnings
 
-import soundfile as sf
 import audioread
 import numpy as np
 import scipy.signal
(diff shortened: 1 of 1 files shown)


bmcfee:
Thanks for doing this @r9y9 , but I don't think we should make soundfile an optional dependency.  If it fails at import time, this is a serious error, and the user needs to fix it.

Apologies if this comes across as inconsistent with my approving the previous PR, but an error at import time is categorically different from an error at runtime.  I had misunderstood the problem in my original review.

I now think that we should revert the previously merged PR to only catch `RuntimeError`.  However, I'd be okay with wrapping the soundfile import with a try so that we can catch oserrors and provide a more useful error message.  Really, this is a problem that should be handled by soundfile upstream of us, but I'm happy to provide a more user-friendly error message if that's helpful.

r9y9:
Closing in favor of #970

PR #974, opened 2019-08-20
[WIP] YIN

### LINKED ISSUE #527: YIN fundamental frequency estimator
Can anyone please tell how to get Fundamental frequency (F0) feature using Librosa?

thank you!

lostanlen:
#### Reference Issue
Fixes #527 (eventually)

This is a comprehensive reimplementation of the YIN algorithm (de Cheveigné and Kawahara, JASA 2002). It's a complex and long-awaited new feature, so i imagine that it's going to require meticulous rounds of CR.

To write this function, i re-read the YIN paper as well as the MATLAB code:
https://github.com/lemonzi/matlab/blob/master/yin/yin2.m

Thankfully, this paper is very well written so it was easy to link it to relevant parts of the code.

### Any other comments?

- the code is in a relatively mature state, but it's still a work in progress. I haven't written the tests yet. I want to create a consensus in the API before i do.

- i tried my best to follow existing naming conventions in the librosa API. Hence `y`, `sr`, `frame_length`, `hop_length`, `fmin`, `fmax`, and `pad_mode`.

- However, `threshold_1` and `threshold_2` could use better names. Ideas appreciated.

- Likewise for the booleans `cumulative` and `interpolate`. Do we even support them as kwargs? If so, i need to document them.

- Perhaps the docstring should be longer and give some historical / methodological context around YIN? Contributions appreciated.

- The parabolic interpolation step could potentially be faster by allocating less intermediate arrays. But maybe that's a premature optimization.

- I did not implement part 6: "best local estimate". It would considerably slow down the code; de Cheveigné and Kawahara report a relatively small effect; and i doubt many people use it anyway. Long term, we would rather use pYIN rather then part 6.

- This implementation works splendidly well as long as there are ten periods or more in the window. Below that number, it's quite numerically unstable. Could it be due to the boxcar window? de Cheveigné and Kawahara write that "there is [...] little reason not to use a square window", but i doubt it.

- Likewise for padding. I am using `pad_mode=reflect`, but really, there is little hope of recovering the instantaneous frequency at the edges.

- What should the tests be? Tests on synthetic tones or tests vs the MATLAB code?

### DIFF
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -336,3 +336,166 @@ def piptrack(y=None, sr=22050, S=None, n_fft=2048, hop_length=None,
                                   + dskew[idx[:, 0], idx[:, 1]])
 
     return pitches, mags
+
+
+def yin(y, sr=22050, frame_length=512, hop_length=None, fmin=40, fmax=None,
+        cumulative=False, interpolate=True,
+        threshold_1=0.1, threshold_2=0.2, pad_mode='reflect'):
+    '''Fundamental frequency (F0) estimation. [1]_
+
+
+    .. [1] De Cheveigné, A., & Kawahara, H. (2002).
+        "YIN, a fundamental frequency estimator for speech and music."
+        The Journal of the Acoustical Society of America, 111(4), 1917-1930.
+
+    Parameters
+    ----------
+    y : np.ndarray [shape=(n,)]
+        audio time series
+
+    sr : number > 0 [scalar]
+        sampling rate of `y` in Hertz
+
+    frame_length : int [scalar]
+        length of the frame in samples.
+        By default, frame_length=512 corresponds to a time scale of 23 ms at
+        a sampling rate of 22050 Hz. This is close to the default of the
+        original publication.
+        We recommend setting `frame_length` to a power of two for optimizing
+        the speed of the fast Fourier transform (FFT) algorithm.
+
+    hop_length : None or int > 0 [scalar]
+        number of audio samples between adjacent YIN predictions.
+        If `None`, defaults to `frame_length // 4`.
+
+    fmin: number > 0 [scalar]
+        minimum frequency in Hertz
+
+    fmax: None or number > 0 [scalar]
+        maximum frequency in Hertz. If `None`, defaults to fmax = sr / 4.0
+
+    threshold_1: number >= 0 [scalar]
+        absolute threshold for peak estimation
+
+    threshold_2: number >= 0 [scalar]
+        absolute threshold for aperiodicity estimation
+        NB: we recommend setting threshold_2 >= threshold_1.
+
+    pad_mode: string or function
+        This argument is passed to `np.pad` for centering frames before
+        computing autocorrelation.
+        By default (`pad_mode="reflect"`), `y` is padded on both sides with its
+        own reflection, mirrored around its first and last sample respectively.
+
+    Returns
+    -------
+    f0: np.ndarray [shape=(n_frames,)]
+        time series of fundamental frequencies in Hertz.
+        Null values represent the absence of a perceptible fundamental frequency.
+
+    See also
+    --------
+    piptrack: sinusoidal peak interpolation
+
+    Examples
+    --------
+    Computing a fundamental frequency curve from an audio input:
+
+    >>> y = librosa.chirp(440, 880, duration=7.0)
+    >>> yin(y)
+    array([  0.        ,   0.        , 222.62295321, ..., 877.12362404,
+       885.76086659, 886.80491703])
+    '''
+    # Set the maximal frequency.
+    if fmax is None:
+        fmax = sr / 4
+
+    # Set the default hop, if it's not already specified.
+    if hop_length is None:
+        hop_length = int(frame_length // 4)
+
+    # Check that audio is valid.
+    util.valid_audio(y, mono=True)
+
+    # Pad the time series so that frames are centered.
+    y = np.pad(y, int(frame_length // 2), mode=pad_mode)
+
+    # Compute autocorrelation in each frame.
+    y_frames = util.frame(y, frame_length=frame_length, hop_length=hop_length)
+    n_frames = y_frames.shape[1]
+    acf_frames = librosa.autocorrelate(y_frames, axis=0)
+
+    # Difference function.
+    min_period = np.maximum(int(np.floor(sr/fmax)), 2)
+    max_period = np.minimum(int(np.ceil(sr/fmin)), frame_length-1)
+    boxcar_window = scipy.signal.windows.boxcar(frame_length)
+    energy = np.convolve(y*y, boxcar_window, mode="same")
+    energy_frames = util.frame(
+        energy, frame_length=frame_length, hop_length=hop_length)
+    energy_0 = energy_frames[0, :]
+    energy_tau = energy_frames[(min_period-1):(max_period+1), :]
+    acf_tau = acf_frames[(min_period-1):(max_period+1), :]
+    yin_frames = energy_0 + energy_frames - 2*acf_frames
+
+    # Cumulative mean normalized difference function.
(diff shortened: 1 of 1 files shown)


bmcfee:
😮 thanks for doing this! I'll give it some cycles later on today.

librosa/core/pitch.py
```diff
@@ -336,3 +336,166 @@ def piptrack(y=None, sr=22050, S=None, n_fft=2048, hop_length=None,
...
+    sr : number > 0 [scalar]
+        sampling rate of `y` in Hertz
+
+    frame_length : int [scalar]
+        length of the frame in samples.
+        By default, frame_length=512 corresponds to a time scale of 23 ms at
+        a sampling rate of 22050 Hz. This is close to the default of the
+        original publication.
+        We recommend setting `frame_length` to a power of two for optimizing
+        the speed of the fast Fourier transform (FFT) algorithm.
```
bmcfee: I get the motivation for keeping close to the original implementation, but internal consistency is more important.  Can we bump this up to 2048?

librosa/core/pitch.py
```diff
@@ -336,3 +336,166 @@ def piptrack(y=None, sr=22050, S=None, n_fft=2048, hop_length=None,
...
+    Computing a fundamental frequency curve from an audio input:
+
+    >>> y = librosa.chirp(440, 880, duration=7.0)
+    >>> yin(y)
+    array([  0.        ,   0.        , 222.62295321, ..., 877.12362404,
+       885.76086659, 886.80491703])
+    '''
+    # Set the maximal frequency.
+    if fmax is None:
+        fmax = sr / 4
```
bmcfee: 4.0 to force floating point division in python 2

librosa/core/pitch.py
```diff
@@ -336,3 +336,166 @@ def piptrack(y=None, sr=22050, S=None, n_fft=2048, hop_length=None,
...
+
+    # Compute autocorrelation in each frame.
+    y_frames = util.frame(y, frame_length=frame_length, hop_length=hop_length)
+    n_frames = y_frames.shape[1]
+    acf_frames = librosa.autocorrelate(y_frames, axis=0)
+
+    # Difference function.
+    min_period = np.maximum(int(np.floor(sr/fmax)), 2)
+    max_period = np.minimum(int(np.ceil(sr/fmin)), frame_length-1)
+    boxcar_window = scipy.signal.windows.boxcar(frame_length)
```
bmcfee: generalize this to support `window=` option like everything else; use `filters.get_window` to do the window construction rather than calling directly.

librosa/core/pitch.py
```diff
@@ -336,3 +336,166 @@ def piptrack(y=None, sr=22050, S=None, n_fft=2048, hop_length=None,
...
+    energy = np.convolve(y*y, boxcar_window, mode="same")
```
bmcfee: scipy signal convolve might be better here, since it can figure out whether fft-based convolution will be faster.

librosa/core/pitch.py
```diff
@@ -336,3 +336,166 @@ def piptrack(y=None, sr=22050, S=None, n_fft=2048, hop_length=None,
...
+            (yin_frames[:-2, :]+yin_frames[2:, :]-2*yin_frames[1:-1, :]) / 2
+        parabola_b = (yin_frames[2:, :]-yin_frames[:-2, :]) / 2
+        parabolic_shifts = -parabola_b / (2*parabola_a)
+        parabolic_values =\
+            yin_frames[1:-1, :] - parabola_b*parabola_b/(4*parabola_a)
+        is_trough = np.logical_and(
+            yin_frames[1:-1, :]<yin_frames[:-2, :],
+            yin_frames[1:-1, :]<yin_frames[2:, :])
+        yin_frames = yin_frames[1:-1, :]
+        yin_frames[is_trough] = parabolic_values[is_trough]
```
bmcfee: is it necessary to do this by hand, or could polyfit be used instead?

librosa/core/pitch.py
```diff
@@ -336,3 +336,166 @@ def piptrack(y=None, sr=22050, S=None, n_fft=2048, hop_length=None,
...
+    fmax: None or number > 0 [scalar]
+        maximum frequency in Hertz. If `None`, defaults to fmax = sr / 4.0
+
+    threshold_1: number >= 0 [scalar]
+        absolute threshold for peak estimation
+
+    threshold_2: number >= 0 [scalar]
+        absolute threshold for aperiodicity estimation
+        NB: we recommend setting threshold_2 >= threshold_1.
+
```
bmcfee: how about `threshold_1` => `peak_threshold` and `threshold_2` => `periodicity_threshold`?

bmcfee:
> * the code is in a relatively mature state, but it's still a work in progress. I haven't written the tests yet. I want to create a consensus in the API before i do.
> 

Notes in my first CR.  Generally, I think it's better if we can make the API and default parameters internally consistent with the rest of the library.  If we want to show how to recover a good approximation to the original paper / implementation, that can be done in a notebook or docstring example.

> * i tried my best to follow existing naming conventions in the librosa API. Hence `y`, `sr`, `frame_length`, `hop_length`, `fmin`, `fmax`, and `pad_mode`.
> 
> * However, `threshold_1` and `threshold_2` could use better names. Ideas appreciated.
> 

`peak_threshold` and `periodicity_threshold`?  A bit wordy, but they get the point across.

> * Likewise for the booleans `cumulative` and `interpolate`. Do we even support them as kwargs? If so, i need to document them.
> 

I don't have strong opinions on this one, but it doesn't seem to hurt to include them.

> * Perhaps the docstring should be longer and give some historical / methodological context around YIN? Contributions appreciated.
> 

Let's hold off on that for now.

> * The parabolic interpolation step could potentially be faster by allocating less intermediate arrays. But maybe that's a premature optimization.
> 

I'm wondering if this could be simplified to use scipy's polynomial interpolation instead?  It might take a little bit of clever array shaping, but it would be preferable to having a bunch of grungy interpolation code IMO.

> * I did not implement part 6: "best local estimate". It would considerably slow down the code; de Cheveigné and Kawahara report a relatively small effect; and i doubt many people use it anyway. Long term, we would rather use pYIN rather then part 6.
> 

Indeed.  Our API plan should make it easy to extend to pyin.

> * This implementation works splendidly well as long as there are ten periods or more in the window. Below that number, it's quite numerically unstable. Could it be due to the boxcar window? de Cheveigné and Kawahara write that "there is [...] little reason not to use a square window", but i doubt it.
> 

We should definitely support user-specified windows.  Ditto for padding modes.  Edge behavior is not well-defined anyway, so I'm not going to worry about it.

> * Likewise for padding. I am using `pad_mode=reflect`, but really, there is little hope of recovering the instantaneous frequency at the edges.
> 
> * What should the tests be? Tests on synthetic tones or tests vs the MATLAB code?

I think synthetic tones would make sense here.

librosa/core/pitch.py
lostanlen: i think polyfit would be slower.
this code vectorizes several dozens/hundreds of parabolic interpolations in the same linalg operations.
polyfit would have to be called in a loop
also note that only the troughs are interpolated. (this is in compliance with the MATLAB code)

librosa/core/pitch.py
lostanlen: should we default to `hann`?

bmcfee: That sounds good to me.

librosa/core/pitch.py
bmcfee: Ok then.  Related question: is there any redundancy between this calculation and the interpolation implemented in piptrack?  Or are they totally different?

lostanlen:
In the interest of releasing v0.7.1 soon, i suggest we bump this to the v0.8 milestone?

bmcfee:
I'm okay with that.

lostanlen:
Looking at this again in the wake of ISMIR. I will address the CR, add tests, and reconvene.

lostanlen:
superseded by #1063

PR #983, opened 2019-09-09
griffinlim_cqt should use filter_scale argument

voodoohop:
Previously griffinlim_cqt was not passing the filter_scale argument on to icqt and cqt.

griffinlim_cqt has an argument called filter_scale which was not passed on to the downstream inverse constant-q functions


bmcfee:
Good catch, thanks for noticing it!

bmcfee:
LGTM!  Anything else to add here, or should we merge?

PR #994, opened 2019-09-20
Optimize stft - skip window multiplication for identity windows

### LINKED ISSUE #986: Acceleration for window-less STFT
#### Description

The STFT implementation operates by chunking the input signal into blocks, broadcast-multiplying the window across the block, and then computing the (r)fft.  See https://github.com/librosa/librosa/blob/master/librosa/core/spectrum.py#L236-L237

This invokes a temporary allocation/copy for the intermediate multiplication result.  In general, this is unavoidable, and the desired behavior for the default window configuration.

However, we sometimes want to compute an STFT without a window, eg, the CQT relies on several calls to `stft` with a rectangular (all-ones) window, as windowing is implemented by the filter basis.  In this case, the intermediate windowed result is identical to the input signal, but it still invokes an array copy, which is inefficient.

The suggested change is to detect and bypass the all-ones window multiplication and operate on the signal directly.  This should provide a modest bump in throughput.

ZiyaoWei:
#### Reference Issue
Fixes #986.

This change skips window multiplication during stft when the window is the identity, thus avoids a possibly expensive array copy operation.

#### Any other comments?
Microbenchmarking is hard - I created a script so I can test out the speed improvement. I tried interleaving the new/old function calls to average out external factors.

This is the output:

```
Original:
Max: 0.09029326800000015s
Min: 0.04929946500000071s
Avg: 0.054472541203333356s

Optimized:
Max: 0.10773746899999992s
Min: 0.045211178999999824s
Avg: 0.04973186985666671s
Optimized wins 272/300 times (0.9066666666666666%)
```

On average the optimized version takes ~91.30% the time when tested with random 1000000 element arrays.

The execution times in a scatterplot, red are the original and blue the optimized:

bench

The correctness are covered by existing tests.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -222,10 +222,11 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
     n_columns = int(util.MAX_MEM_BLOCK / (stft_matrix.shape[0] *
                                           stft_matrix.itemsize))
 
-    if window == "ones" :
+    if window in ['boxcar', 'box', 'ones', 'rect', 'rectangular']:
         for bl_s in range(0, stft_matrix.shape[1], n_columns):
             bl_t = min(bl_s + n_columns, stft_matrix.shape[1])
-            stft_matrix[:, bl_s:bl_t] = fft.rfft(y_frames[:, bl_s:bl_t], axis=0)
+            stft_matrix[:, bl_s:bl_t] = fft.rfft(y_frames[:, bl_s:bl_t],
+                                                 axis=0)
     else:
         fft_window = get_window(window, win_length, fftbins=True)
 
@@ -241,6 +242,7 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
                                                  y_frames[:, bl_s:bl_t],
                                                  axis=0)
     return stft_matrix
+
 
 @cache(level=30)
 def istft(stft_matrix, hop_length=None, win_length=None, window='hann',
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -201,14 +201,6 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
     if hop_length is None:
         hop_length = int(win_length // 4)
 
-    fft_window = get_window(window, win_length, fftbins=True)
-
-    # Pad the window out to n_fft size
-    fft_window = util.pad_center(fft_window, n_fft)
-
-    # Reshape so that the window can be broadcast
-    fft_window = fft_window.reshape((-1, 1))
-
     # Check audio is valid
     util.valid_audio(y)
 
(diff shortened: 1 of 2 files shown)


lostanlen:
Thank you for starting this. Especially, many thanks for the thorough benchmark.

(1) please make sure to cover all synonyms of `ones`: `box`, `boxcar`, `ones`, `rect`, `rectangular`. 

(2) i wonder if this branch can be made even faster by `librosa.util.frame`?

(3) MATLAB regression tests are failing

ZiyaoWei:
Thanks for the review! @lostanlen comments:

1) Didn't realize there are so many aliases, thanks!

2) Don't know the internal of librosa - will take a look!

3) Weird - I can't replicate this locally, and if you look at the numbers printed out in cI they seem to be the same - I might push a couple "debug" commits to this branch and try to figure out what's going on.

lostanlen:
Re (2), forget what i said about `frame`. It's already framed, actually.

We'd have to check, but my gut feeling is that we don't necessarily need `bl_s` and `bl_t` in the transform. I would surmise that `fft.rfft` is clever enough to figure out how to optimally chunk this operation in order to make the most efficient use of CPU vectorization while avoiding cache overflow. This machine-specific optimization is the entire premise of the FFTW library. So, at least, it's worth trying a simplification of the branch down to

`stft_matrix = fft.rfft(y_frames, axis=0)`

and seeing how fast that is, and whether it produces out-of-memory errors.

lostanlen:
Are you saying it's _faster_ to do a chunking heuristic before calling `fft.fft` rather than calling `fft.fft` over the entire matrix? Independently of FFT backend? Wow.

librosa/core/spectrum.py
```diff
@@ -230,12 +222,25 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
     n_columns = int(util.MAX_MEM_BLOCK / (stft_matrix.shape[0] *
                                           stft_matrix.itemsize))
 
-    for bl_s in range(0, stft_matrix.shape[1], n_columns):
-        bl_t = min(bl_s + n_columns, stft_matrix.shape[1])
+    if window in ['boxcar', 'box', 'ones', 'rect', 'rectangular']:
+        for bl_s in range(0, stft_matrix.shape[1], n_columns):
```
lostanlen: can you try removing the `bl_s` loop?

lostanlen:
@ZiyaoWei hi! any updates regarding this PR? Last time we checked, you reported an acceleration from using a manual for loop rather than letting numpy do the vectorization, which sounded a bit surprising.

The acceleration is quite small for the moment. I think that an acceleration of at least 2x would be necessary to warrant making a special branch for this. Rectangular windows ("identity functions")  can be useful in bioacoustics for detecting sinusoids in noise, but they are almost never used in speech or music processing.

bmcfee:
I just tried this out locally (independent reimplementation), and get similar behavior.  The block-wise acceleration is indeed faster.  I get about a 6ms boost by skipping the window multiply for the demo track (39ms down to 33ms average time).

I'm inclined to think this optimization isn't worth it.  But it's good to have done this exercise!

bmcfee:
Any objection to closing this PR / issue?

lostanlen:
I am in favor of closing. `librosa.stft` already does the right thing and anyone wanting to try optimizing the boxcar STFT by skipping the window multiplication can obtain it very simply: (assuming `center=False`)

```python
def stft_rect(y, n_fft, hop_length, win_length, center=True, pad_mode="reflect"):
    if center:
        y = np.pad(y, int(n_fft // 2), mode=pad_mode)
    y_frames = librosa.util.frame(y, win_length, hop_length)
    fft = librosa.get_fftlib()
    stft_matrix = fft.rfft(y_frames, n=n_fft, axis=0)
```

PR #998, opened 2019-09-23
Load() type checking

### LINKED ISSUE #996: `librosa.load()` silently ignores the `duration` parameter if it is a string
#### Description
When using `soundfile`, `librosa.load()` silently ignores the `duration` parameter if it is a string.
I haven't tested other input types, but passing `"4"` as the `duration` should either work as expected or raise an exception rather than silently being ignored,

#### Steps/Code to Reproduce
1. Ensure you're loading a format supported by `soundfile` (eg. WAV or FLAC)
2. Pass a string as the `duration` parameter of `librosa.load()`:
```python
audio, sr = librosa.load('file.flac', duration="4")  # Note that `duration` is a string
print((len(audio) / sr) == 4)
```
This generally comes out `False` since the `duration` parameter is ignored and the whole file is loaded.

#### Expected Results
The duration of the loaded audio should be 4 seconds (if the string is castable to an int), OR a `TypeError` (or some other exception) should be raised.

#### Versions
```python
Windows-10-10.0.18362-SP0
Python 3.7.3 (default, Apr 24 2019, 15:29:51) [MSC v.1915 64 bit (AMD64)]
NumPy 1.16.4
SciPy 1.3.1
librosa 0.7.0
```

konoikon:
### Reference Issue
Fixes #996 

Included `if` statements that perform type and value checking. Ensures that parameters `duration`, `offset` and `sr` are of the right respective type and non-negative.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -128,14 +128,10 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
     try:
         with sf.SoundFile(path) as sf_desc:
             sr_native = sf_desc.samplerate
-            if offset:
-                if not np.isscalar(offset) or offset <= 0.0:
-                    raise ParameterError("Parameter 'offset' must have a positive float value.")
+            if offset and util.utils.valid_num(offset, positive=True):
                 # Seek to the start of the target read
                 sf_desc.seek(int(offset * sr_native))
-            if duration is not None:
-                if not np.isscalar(duration) or duration <= 0.0:
-                    raise ParameterError("Parameter 'duration' must have a positive float value.")
+            if duration is not None and util.utils.valid_num(duration, positive=True):
                 frame_duration = int(duration * sr_native)
             else:
                 frame_duration = -1
@@ -156,9 +152,7 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
     if mono:
         y = to_mono(y)
 
-    if sr is not None:
-        if not np.isscalar(sr) or sr <= 0:
-            raise ParameterError("Parameter 'sr' must have a positive integer value")
+    if sr is not None and util.utils.valid_num(sr, positive=True):
         y = resample(y, sr_native, sr, res_type=res_type)
 
     else:
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -168,10 +168,10 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
 
 
 def __audioread_load(path, offset, duration, dtype):
-    '''Load an audio buffer using audioread.
+    """Load an audio buffer using audioread.
 
     This loads one block at a time, and then concatenates the results.
-    '''
+    """
 
     y = []
     with audioread.audio_open(path) as input_file:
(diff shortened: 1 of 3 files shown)


konoikon:
During `pytest` I saw that there are two test cases where `duration=1` and `duration=3` respectively. Should we allow duration being an integer? If yes, we should change the docstring.

bmcfee:
Thanks for starting this!

As a general rule, one shouldn't use exact type checking in python, since it wouldn't handle subclasses: this is what the `isinstance` operator is for.

With numpy data, we can be a bit more specific, and use `np.isscalar` to test if an input is scalar (not an array) and `np.isrealobj` to test that it's real-valued.  This is more safe than checking for float, int, or anything else.

konoikon:
Thank you for your comment! So, I created a helper function in `util` to check the validity of the parameters and used the native `ParameterError` when something's wrong.

I also wrote tests that have fully covered my additional lines.

One question: in `core/audio.py` I cannot use my util function as `util.valid_num()` but I have to use `util.utils.valid_num()`. It does not get resolved otherwise. Any ideas about why?

konoikon:
Hey @bmcfee ! Any updates?

bmcfee:
Hey @konoikon sorry to lag here, been a hectic couple of weeks.  Last I heard, @lostanlen was taking a look at this one generally, so I don't want to duplicate his efforts.

Per your question though:

>  in `core/audio.py` I cannot use my util function as `util.valid_num()` but I have to use `util.utils.valid_num()`. It does not get resolved otherwise. Any ideas about why?

That's strange.  Can you provide a traceback showing the failure?

lostanlen:
Hello again @konoikon. Upon reflection, i think that it's best to close this PR without merging.
We really appreciate your commitment to improving librosa, but in this case, it's not clear that this addition of code would be necessary. First of all, the current implementation of `valid_num` is not complete, because there are some numeric values for which it returns `False`. One example is `np.float32(1.0)`. Making it correct would require more conditional branches, involving numpy-specific utilities, etc. None of this is particularly obvious. Then, we'd also have to bikeshed the cases of `NaN` and `inf`. So that would require even more testing, more documentation, and so forth. It would incur a lot of maintenance in the long term for a relatively small benefit.

Generally speaking, i think that it would be a risky direction for librosa to offer automatic type checks and type conversions in its API. I can only speak from personal experience here; but as a contributor to the Julia language, i know the level of meticulosity it takes to build a fully functional type system. There is no ambition of such kind in librosa. It is documented that `duration` has to be `None` or a `float`. If the user happens to store duration as a `string` at some point of their workflow (because they read it from a text file), i think that it's reasonable to assume that the onus of type conversion should be on them. I don't really see a use case where the type of `duration` would be unknown at runtime.

I propose to close the issue #996 as well as this PR without merging it. I apologize if this is a frustration to you, @konoikon, especially given that it's your first contribution to librosa. But i think that it's for the benefit of the project to keep a "no-frills" approach in terms of API. Some other issues are up for grabs, and you are welcome to turn to them.

We can reconvene in the future (post-0.7.1) if this issue becomes more dire. But for now, i think we can consider that the current implementation of `load` is WAI.

I hope this helps!

PR #1004, opened 2019-10-08
onset and chroma do not use kwarg-getting

### LINKED ISSUE #989: onset_strength_multi uses nfft=2048 and hop_length=512 by default instead of preset values
Correct me if I'm wrong but I believe it happens [here](https://github.com/librosa/librosa/blob/5e48c6ffd4d3f84f03fbc4ba0f728cd7d52f8d2c/librosa/onset.py#L537-L538). The default values are hard coded inside the function instead of in the list of key word arguments like usual. So when setting a preset using the presets library with e.g. sr=44100, nfft=4096 and hop_length=1024, the new default values are not used.

bmcfee:
#### Reference Issue
Fixes #989 

This PR promotes spectrogram parameters in `onset` module to function-level parameters, rather than being pulled out of `kwargs`.

Similarly, `n_chroma` is now a proper argument to `feature.chroma_stft`.

This should resolve issues related to preset-overrides for default parameters.

The documentation has been updated accordingly.  These changes do not constitute a substantial change to the API, though some keyword parameter orderings have shifted.

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -1059,7 +1059,8 @@ def zero_crossing_rate(y, frame_length=2048, hop_length=512, center=True,
 # -- Chroma --#
 def chroma_stft(y=None, sr=22050, S=None, norm=np.inf, n_fft=2048,
                 hop_length=512, win_length=None, window='hann', center=True,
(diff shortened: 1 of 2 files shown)


beasteers:
Lgtm 👍🏻

Issue #952, opened 2019-08-14
RFC: more example tracks?


bmcfee:
It's been mentioned on numerous occasions that our current example track, catchy as it may be, is not great for demonstrating many of the functions of librosa.  In offline discussions, @lostanlen and I have talked about extending the included audio content to have several tracks, which could be used to demonstrate different functionality.  So I want to kick this out to the community: what do you want to see in our examples?  By this I mean: please recommend specific recordings.

To prevent this discussion from becoming an infinite bike shed, I'm going to lay down some ground rules for inclusion:

1. Content must be CC-licensed or public domain.
2. Total content should not bloat the package too much; I think 10MB is a reasonable upper bound, and this functionally limits us to between 5 and 10 total recordings.
3. No single track should be too long.  It's okay to have some very short examples.
4. The total collection should be diverse in terms of style, instrumentation, polyphony, etc.  If possible, I'd like to include some non-western recordings as well.
5. Any lyrical content should not be offensive, for some reasonable definition of offensive which is compatible with our [CoC](https://github.com/librosa/librosa/blob/master/CODE_OF_CONDUCT.md).  I don't expect any problems here, but I'll reserve executive privilege here to veto anything that could be problematic.
6. Familiarity would be a bonus, for making the examples and documentation more immediately accessible.

With all of that out of the way, let's talk about things not currently demonstrated by our current example.  We don't have to hit all of these, and I'm sure I'm missing some, but we should aim to hit most of them.

* Monophonic audio: we should have at least one solo instrument recording that can be used to demonstrate things like pitch tracking.  Maybe a raga or makam could be good here?  This could also be good for demonstrating onsets.
* Interesting harmony: the current example is pretty boring, harmonically speaking.  Maybe a jazz recording would be appropriate here?  Maybe something with some key changes as well.
* Non-percussive rhythmic elements: something classical (strings) would be nice to have for demonstrating onset and beat tracking with soft attacks.
* Different time signatures: examples with 4/4, 3/4, and maybe a 5/4 or 7/8 would be nice for demonstrating some of the rhythmic features (tempogram, fmt)
* Vocals and instrumentals: we should have at least one track with vocals.
* Non-musical audio: do we need/want this?  Speech?  Environmental sound?  Librosa gets used for these things, so it might be worth considering their inclusion.

Some other discussions that we can have around this:

- Should we have a multi-track / stem set as one of the examples?
- What kind of genre coverage should we strive for?
[...]

PR #1008, opened 2019-10-16
 add liftering support to MFCC inversion 

### LINKED ISSUE #1006: MFCC inversion should support liftering
#### Description
#940 implemented liftering in the MFCC extractor, but we didn't add inverse-liftering to the inverse-mfcc function.

This would be relatively easy to do, as inverse-liftering would divide out the lifter (raised sine) before applying the inverse-DCT transform using this calculation: https://github.com/librosa/librosa/blob/master/librosa/feature/spectral.py#L1697

Note that for some settings of `lifter`, this could result in divide-by-zero errors, and it should raise a warning in that case.

SuperKogito:
#### Reference Issue

fixes #1006 

I added a new input variable called `inv_lifter` to [mfcc_to_mel](https://github.com/librosa/librosa/blob/master/librosa/feature/inverse.py#167)  and [mfcc_to_audio](https://github.com/librosa/librosa/blob/master/librosa/feature/inverse.py#217), which is used in a similar fashion to `lifter` in  [mfcc](https://github.com/librosa/librosa/blob/master/librosa/feature/spectral.py#1571).
- if `inv_lifter` is positive, then inverse the liftering by mfcc division.
- if `inv_lifter` is zero, then no inverse liftering is performed.
- if `inv_lifter` is negative, then raise a `ParameterError`.

The docs are adjusted accordingly.

### DIFF
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -164,7 +164,7 @@ def mel_to_audio(M, sr=22050, n_fft=2048, hop_length=512, win_length=None,
                       pad_mode=pad_mode)
 
 
-def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
+def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, inv_lifter=0):
     '''Invert Mel-frequency cepstral coefficients to approximate a Mel power
     spectrogram.
 
@@ -195,12 +195,19 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
     ref : number or callable
         Reference power for (inverse) decibel calculation
 
+    inv_lifter : number >= 0
+        If `inv_lifter>0`, apply *inverse-liftering* (inverse cepstral filtering):
+        `M[n, :] <- M[n, :] / (1 + sin(pi * (n + 1) / inv_lifter)) * inv_lifter / 2`
 
     Returns
     -------
     M : np.ndarray [shape=(n_mels, n)]
         An approximate Mel power spectrum recovered from `mfcc`
 
+    Warns
+    --------
+    RuntimeWarning
+        divide by zero encountered during inverse-liftering.
 
     See Also
     --------
@@ -208,13 +215,18 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
     melspectrogram
     scipy.fftpack.dct
     '''
+    if inv_lifter > 0:
+        mfcc /= 1 + (inv_lifter / 2) \
+                * np.sin(np.pi * np.arange(1, 1 + n_mels, dtype=mfcc.dtype) / inv_lifter)[:, np.newaxis]
+    elif inv_lifter != 0:
+        raise ParameterError('MFCC to MEL inv_lifter={} must be a non-negative number'.format(inv_lifter))
 
     logmel = scipy.fftpack.idct(mfcc, axis=0, type=dct_type, norm=norm, n=n_mels)
 
     return db_to_power(logmel, ref=ref)
 
 
-def mfcc_to_audio(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, **kwargs):
+def mfcc_to_audio(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, inv_lifter=0, **kwargs):
     '''Convert Mel-frequency cepstral coefficients to a time-domain audio signal
 
     This function is primarily a convenience wrapper for the following steps:
@@ -260,6 +272,7 @@ def mfcc_to_audio(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, **kwargs)
     core.griffinlim
     scipy.fftpack.dct
     '''
-    mel_spec = mfcc_to_mel(mfcc, n_mels=n_mels, dct_type=dct_type, norm=norm, ref=ref)
+    mel_spec = mfcc_to_mel(mfcc, n_mels=n_mels, dct_type=dct_type, norm=norm,
+                           ref=ref, inv_lifter=inv_lifter)
 
     return mel_to_audio(mel_spec, **kwargs)
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 '''Feature inversion'''
 
+import warnings
 import numpy as np
 import scipy.fftpack
 
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -224,7 +224,7 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, lifter=0):
             * np.sin(np.pi * np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype) / lifter)[:, np.newaxis]
 
         # avoid division of zero by zero errors
-        mfcc = np.nan_to_num(mfcc, nan=1.)
+        mfcc[np.isnan(mfcc)] = 1.
 
     elif lifter != 0:
         raise ParameterError('MFCC to MEL lifter={} must be a non-negative number'.format(lifter))
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -221,10 +221,13 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, lifter=0):
         n_mfcc = mfcc.shape[0]
         idx = np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype)
         lifter_sine = 1 + (lifter / 2) * np.sin(np.pi * idx / lifter)[:, np.newaxis]
-        lifter_sine = np.where(lifter_sine == 0, np.finfo(float).eps, lifter_sine)
-        mfcc = np.where(mfcc == 0, np.finfo(float).eps, mfcc) 
-        liftered_mfcc = mfcc / lifter_sine
-        mfcc = liftered_mfcc
+
+        # check lifter array for critical values
+        if (lifter_sine == 0).any():
+            warnings.warn("Warning: lifter array includes critial values that will likely to invoke underflow.")
+
+        # lifter mfcc values
+        mfcc = mfcc / lifter_sine
 
     elif lifter != 0:
         raise ParameterError('MFCC to mel lifter={} must be a non-negative number'.format(lifter))
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -165,7 +165,6 @@ def mel_to_audio(M, sr=22050, n_fft=2048, hop_length=512, win_length=None,
                       window=window, center=center, dtype=dtype, length=length,
                       pad_mode=pad_mode)
 
-
 def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, lifter=0):
     '''Invert Mel-frequency cepstral coefficients to approximate a Mel power
     spectrogram.
@@ -208,9 +207,8 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, lifter=0):
 
     Warns
     --------
-    RuntimeWarning
-        overflow encountered in power or
-        divide by zero encountered during inverse liftering.
+    UserWarning
+        due to critical values in lifter array that invokes underflow.
 
     See Also
     --------
@@ -221,21 +219,21 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, lifter=0):
     if lifter > 0:
         n_mfcc = mfcc.shape[0]
         idx = np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype)
-        lifter_sine = 1 + (lifter / 2) * np.sin(np.pi * idx / lifter)[:, np.newaxis]
-
-        # check lifter array for critical values
-        if (lifter_sine == 0).any():
-            warnings.warn("Warning: lifter array includes critial values that will likely to invoke underflow.")
-
+        lifter_sine = 1 + lifter * 0.5 * np.sin(np.pi * idx / lifter)[:, np.newaxis]
+        
+        # raise a UserWarning if lifter array includes critical values
+        if np.any(np.abs(lifter_sine) < np.finfo(lifter_sine.dtype).eps):
+            warnings.warn(message="lifter array includes critial values that may invoke underflow.", 
+                          category=UserWarning)
+        
         # lifter mfcc values
         mfcc = mfcc / lifter_sine
-
+        
     elif lifter != 0:
-        raise ParameterError('MFCC to mel lifter={} must be a non-negative number'.format(lifter))
-
+        raise ParameterError('MFCC to mel lifter must be a non-negative number.')
+    
     logmel = scipy.fftpack.idct(mfcc, axis=0, type=dct_type, norm=norm, n=n_mels)
     return db_to_power(logmel, ref=ref)
-
 
 def mfcc_to_audio(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, lifter=0, **kwargs):
     '''Convert Mel-frequency cepstral coefficients to a time-domain audio signal
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -221,7 +221,9 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, lifter=0):
         n_mfcc = mfcc.shape[0]
         idx = np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype)
         lifter_sine = 1 + (lifter / 2) * np.sin(np.pi * idx / lifter)[:, np.newaxis]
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks for getting this started!  It generally looks good, but I have a few comments:

1. Can you change the parameter name from `inv_lifter` to `lifter` so that it matches with the forward transformation?
2. Your implementation applies inverse-liftering in-place (`mfcc /= ...`) which will modify the input data because it's passed by reference.  Please change it to operate on a copy instead.
3. Can you add test coverage for the new functionality in `mfcc_to_mel`?  I don't think we need a full forward-backward comparison, but at least check the following cases:
  a. warnings are triggered when appropriate
  b. the lifter parameter is used properly -- this can be tested with, say, an all-ones mfcc vector, so that `idct(ones * lifter)` can be pre-computed for a couple of different lifter values.

SuperKogito:
I was not sure if you meant a shallow or deep copy so I just followed what is used in other librosa scripts. There was also an error in the previous implementation causing a size mismatch that I now solved .

Point 3.b. is still not covered.
I need first to find some pre-defined cases as you mentioned so I can test if the lifter is used properly. 
I will try to get it done over the weekend.

tests/test_features.py
```diff
@@ -958,10 +958,21 @@ def test_mfcc_to_mel(n_mfcc, n_mels, dct_type):
                                             dct_type=dct_type,
                                             lifter=10**-3)
 
-    # check if mfcc_to_mel it works correctly with lifter
-    librosa.feature.inverse.mfcc_to_mel(mfcc, n_mels=n_mels,
-                                        dct_type=dct_type,
-                                        lifter=1)
+    # check if mfcc_to_mel works correctly with lifter
+    ones = np.ones(mfcc.shape)
+    negate_lifter_vec = 1 + (lifter / 2) * np.sin(np.pi * np.arange(1, 1 +
```
lostanlen: this nonsyntactic line break is difficult to parse for the human reader. it breaks at the worst possible place in the equation, at the sixth level of nesting.

SuperKogito: the line breaks are automatically done using [autopep8](https://github.com/hhatto/autopep8). Although, this is practical but I admit that it can result in some weird line breaks. [yapf](https://github.com/jpadilla/yapf-online) seems to provide better results. Shortening the variable name and using yapf results in: 
```
    lifter_vec = 1 + (lifter / 2) * np.sin(np.pi * np.arange(
        1, 1 + n_mfcc, dtype=mfcc.dtype) / lifter)[:, np.newaxis]
```
If you think this variant is better, I will edit the code accordingly. Otherwise, I suggest a no line break at all in this position.

lostanlen: you can also assign a temporary variable for `np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype)`

librosa/feature/inverse.py
```diff
@@ -195,26 +196,44 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
     ref : number or callable
         Reference power for (inverse) decibel calculation
 
+    lifter : number >= 0
+        If `lifter>0`, apply *inverse-liftering* (inverse cepstral filtering):
```
lostanlen: `*inverse-liftering*` -> `inverse liftering`
(no need for `*` nor hyphen)

librosa/feature/inverse.py
```diff
@@ -195,26 +196,44 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
...
     Returns
     -------
     M : np.ndarray [shape=(n_mels, n)]
         An approximate Mel power spectrum recovered from `mfcc`
 
+    Warns
+    --------
+    RuntimeWarning
+        overflow encountered in power or
+        divide by zero encountered during inverse-liftering.
```
lostanlen: inverse liftering (no hyphen)

librosa/feature/inverse.py
```diff
@@ -195,26 +196,44 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
...
     --------
     mfcc
     melspectrogram
     scipy.fftpack.dct
     '''
+    if lifter > 0:
+        mfcc = mfcc.copy()
+        n_mfcc = mfcc.shape[0]
+        mfcc /= 1 + (lifter / 2) \
+            * np.sin(np.pi * np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype) / lifter)[:, np.newaxis]
```
lostanlen: declare a different variable `liftered_mfcc`, so that there's no need need for `.copy()`

lostanlen: also define the constant `np.sin(np.pi * np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype) / lifter)` on a separate line, with an informative name (`lifter_sine` ?)

librosa/feature/inverse.py
```diff
@@ -195,26 +196,44 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
...
 
-    logmel = scipy.fftpack.idct(mfcc, axis=0, type=dct_type, norm=norm, n=n_mels)
+        # avoid division of zero by zero errors
+        mfcc[np.isnan(mfcc)] = 1.
+
+    elif lifter != 0:
+        raise ParameterError('MFCC to MEL lifter={} must be a non-negative number'.format(lifter))
```
lostanlen: MEL -> mel

librosa/feature/inverse.py
```diff
@@ -195,26 +196,44 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
...
     '''
+    if lifter > 0:
+        mfcc = mfcc.copy()
+        n_mfcc = mfcc.shape[0]
+        mfcc /= 1 + (lifter / 2) \
+            * np.sin(np.pi * np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype) / lifter)[:, np.newaxis]
 
-    logmel = scipy.fftpack.idct(mfcc, axis=0, type=dct_type, norm=norm, n=n_mels)
+        # avoid division of zero by zero errors
+        mfcc[np.isnan(mfcc)] = 1.
```
lostanlen: is this really needed? can't you catch this before applying the division?
i would prefer to avoid multiple sequential assignments on the same array. they tend to be error-prone. ideally, the output is computed in a single pass.

librosa/feature/inverse.py
```diff
@@ -244,6 +263,10 @@ def mfcc_to_audio(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, **kwargs)
     ref : number or callable
         Reference power for (inverse) decibel calculation
 
+    lifter : number >= 0
+        If `lifter>0`, apply *inverse-liftering* (inverse cepstral filtering):
```
lostanlen: again, no asterisks, no hyphen please

SuperKogito:
Well, I already mentioned that I used a linter last time (autopep8 and yapf). So if I get it correctly, I should cancel the linter changes and take into account your new remarks. Or is it enough to fix the sections you mentioned?

lostanlen:
@SuperKogito I would say it's fine. We'll keep your linter changes: there aren't that many. Please focus on addressing my CR comments. In the future, though, we'd appreciate if you put linter changes in a separate commit (if not a separate PR altogether).

librosa/feature/inverse.py
SuperKogito: This is not my favorite line and I agree with your concern. I replaced it by the following `np.where(mfcc == 0, np.finfo(float).eps, mfcc) ` which should help avoid the zero by zero division errors.

SuperKogito: This change resulted in a failed build :( any suggestions on how to fix this?

SuperKogito: I fixed it.

bmcfee:
Thanks again to @SuperKogito and @lostanlen for the work on this PR/CR. We've all been busy for the past couple of weeks, but this seems pretty close to finished and I'd hate to see it languish from lack of attention.

@SuperKogito is there anything else you need from us to finish up the requested changes?

SuperKogito:
@bmcfee From my side, I think this is done. However, I was waiting for the feedback from @lostanlen and you, especially about the last changes I implemented to handle division by zero and the FloatingPointError. If you are okay with those, then this PR is ready for merging.

librosa/feature/inverse.py
```diff
@@ -195,26 +196,44 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
...
     scipy.fftpack.dct
     '''
+    if lifter > 0:
+        n_mfcc = mfcc.shape[0]
+        idx = np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype)
+        lifter_sine = 1 + (lifter / 2) * np.sin(np.pi * idx / lifter)[:, np.newaxis]
+        lifter_sine = np.where(lifter_sine == 0, np.finfo(float).eps, lifter_sine)
+        mfcc = np.where(mfcc == 0, np.finfo(float).eps, mfcc) 
+        liftered_mfcc = mfcc / lifter_sine
+        mfcc = liftered_mfcc
```
bmcfee: This part has a few issues that need to be fixed before we can merge.

1. Line 224: lifter_sine has exact zeros replaced by float epsilons.  Two things here: we'll almost never get exact zeros, and if you are going to replace these by machine epsilons, it should match the data type of `lifter_sine`, not be the default `float`.  You can get this in short-hand by calling `librosa.util.tiny(lifter_sine)`, but I'm not convinced this is a safe operation in general.
2. Why the in-filling of `mfcc` at line 225?  Dividing `0/lifter_sine` ought to be safe, but in general the `0/0` problem is undefined here, but I think any infilling should only be applied if we have a lifter that could zero out values.  If `lifter_sine` is always sufficiently positive, then we can safely infer that `mfcc==0` implies that the 0 value came pre-liftering, so post-liftering should stay 0 rather than becoming `eps/eps ~= 1`.  The same critique as above applies though, exact checks for 0 should be avoided.

Overall, I'm wary of trying too hard to prevent errors here.  Instead, I think it's safer to do the following:

1. Build the lifter array.
2. Check the lifter array for values likely to invoke underflow, and warn if they occur.
3. `mfcc = mfcc / lifter_sine` without any additional pre- or post-processing of `mfcc`.

bmcfee:
> However, I was waiting for the feedback from @lostanlen and you, especially about the last changes I implemented to handle division by zero and the FloatingPointError.

Ah, sorry, github was doing that thing again where it's not giving the most recent view of the PR state.  I went in and added my thoughts to the CR, let me know what you think.

librosa/feature/inverse.py
```diff
@@ -195,26 +197,47 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
...
     See Also
     --------
     mfcc
     melspectrogram
     scipy.fftpack.dct
     '''
+    if lifter > 0:
+        n_mfcc = mfcc.shape[0]
+        idx = np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype)
+        lifter_sine = 1 + (lifter / 2) * np.sin(np.pi * idx / lifter)[:, np.newaxis]
```
bmcfee: This will break in python 2.7 because it will use integer division instead of float.  You can fix this by saying `lifter * 0.5` instead of dividing by 2.

librosa/feature/inverse.py
```diff
@@ -195,26 +197,47 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
...
 
-    logmel = scipy.fftpack.idct(mfcc, axis=0, type=dct_type, norm=norm, n=n_mels)
+        # check lifter array for critical values
+        if (lifter_sine == 0).any():
```
bmcfee: Can we soften this to `np.abs(lifter_sine) < np.finfo(lifter_sine.dtype).eps`?  That way we can detect unstable points that aren't exactly on 0.

librosa/feature/inverse.py
```diff
@@ -195,26 +197,47 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0):
...
+            warnings.warn("Warning: lifter array includes critial values that will likely to invoke underflow.")
```
bmcfee: This will default to type `UserWarning` but the doc says `RuntimeWarning`.  I don't have a strong opinion about which is appropriate here, but it should be consistent.

SuperKogito:
I am sorry that this PR is getting quite messy and ugly..I am facing some trouble running tests offline as I mentioned in  #1026. I agree with all your points and I am working on fixing the code but I keep making these stupid small mistakes (pushes and new PRs) every time I try to test or work around #1026.

bmcfee:
> .I am facing some trouble running tests offline as I mentioned in #1026.

Is this still causing troubles for you?  Is there anything else we can do to help wrap this one up?

bmcfee:
Your PR is showing a merge conflict because your editor has reformatted a bunch of unrelated code.  Could you revert those changes?  Thanks!

SuperKogito:
@bmcfee I am working on it as we speak. I will make sure to reverse the formatting back, but first I need to fix the test that breaks with a `FloatingPointError: divide by zero encountered in true_divide`. I just managed to finally replicate the travis testing environment and its errors on my pc.  That's why it took me so long to respond.

However, I have a question concerning the raised `UserWarning` when this case occurs.  How should I go about that? right now I am testing by catching the warning but right after the test breaks due to the `FloatingPointError`, should I catch the error too or mark the test as failed to pass? is there a better option . You can take a look at my error logs [here](https://travis-ci.org/SuperKogito/librosa/jobs/630039825).

bmcfee:
> However, I have a question concerning the raised `UserWarning` when this case occurs. How should I go about that? right now I am testing by catching the warning but right after the test breaks due to the `FloatingPointError`, should I catch the error too or mark the test as failed to pass?

I think it should be safe to divide by `lifter_sine + util.tiny(mfcc)` instead of `lifter_sine`.  If `lifter_sine` is near 0, then the corresponding mfcc dimension should also be, so you'd get `0 / (0 + eps) -> 0`.  If the mfcc dimension is not near 0, then it'll blow up, but that's why we have the warning.  If `lifter_sine` is far from 0, then this should have no measurable effect.

SuperKogito:
Using `util.tiny()` is a great suggestion, and it seems to work like a charm. I tried to reverse the linter effects as much as I can but I am not sure how successful I was.

I thought it would be good to provide some feedback concerning my experience to help improve future contributor experiences: 
- To me the code was very readable & understandable and the communication with you guys and your guidance was great (really impressed by that and the welcoming culture and patience that you displayed) but replicating the development environment was very tricky especially since I worked on multiple projects on parallel and I used different devices to contribute. This might have been in part my fault and I might have not made the best choices when it comes to updating packages and my env but I still think that the `pytest` version requirement should be highlighted. Overall, it would be better if one can just get an anaconda development environment to be set-up automatically. In my case, every time I fixed a dependency something else was broken, that is why I did most of my last changes online and therefore I had to make a lot  of commits in comparison to the quantity of changes.
- I coded online and mostly relied on the Travis tests to check my code. However, in some cases some snippets did not display the same behaviour. For example: it took me a while to replicate the `FloatingPointError` on my machine. Instead I was getting a `RuntimeWarning` but using `np.seterr(divide='raise')` I got the error to show up. This might be useful for others to know.
- I also found the Travis tests kinda slow, therefore I think it is worthy taking a look at **distributed testing** using `pytest-xdist`.

bmcfee:
> Using `util.tiny()` is a great suggestion, and it seems to work like a charm. I tried to reverse the linter effects as much as I can but I am not sure how successful I was.

Thanks, this all looks great to me!  There are a few dangling linter side effects, but I can live with that.  We do a pre-release audit for this kind of thing anyway, and it doesn't affect correctness.



> I thought it would be good to provide some feedback concerning my experience to help improve future contributor experiences:

Thanks for the feedback!  I'm glad we got this worked through, and appreciate your comments.  Some responses inline:



>     * but replicating the development environment was very tricky especially since I worked on multiple projects on parallel and I used different devices to contribute. This might have been in part my fault and I might have not made the best choices when it comes to updating packages and my env but I still think that the `pytest` version requirement should be highlighted.

Yeah, I don't have a great solution for this in general.  We do have all the requirements specified in the setup.py so that `pip install -e .[tests]` should do the trick (as stated in the contributing docs), but if there's anything that could be improved there, I'd be happy to revise things.



>     * I coded online and mostly relied on the Travis tests to check my code. However, in some cases some snippets did not display the same behaviour. For example: it took me a while to replicate the `FloatingPointError` on my machine. Instead I was getting a `RuntimeWarning` but using `np.seterr(divide='raise')` I got the error to show up. This might be useful for others to know.

This is kind of the point of having CI running on multiple versions / environments, so I take this as a success. :grin: 

It is problematic to *only* get testing through CI though, since it slows down progress substantially.  Again, if there's anything that could be made clearer in `CONTRIBUTING.md`, I think that would be the best place to clarify things.


>     * I also found the Travis tests kinda slow,

Agreed, but I don't think this is Travis's fault.  (Unlike AppVeyor, travis builds get scheduled and run pretty quickly.)

We have a long-standing issue to rewrite the test suite using more modern pytest (>4).  One side-effect of this rewrite would be better (re)use of test fixtures, which ought to bring down the running time of tests substantially.  It's currently at around 15min per build, and I'd like to get that down to under 5 if possible.

This is a huge undertaking though, and probably will take a couple of solid weeks of work by someone who knows the test suite well. (IE, probably me, though I'd be extremely happy for someone else to do it!)

It's also a little tangled up in #952, which if that ends up including some very short excerpts, could be helpful for accelerating certain parts of the test suite.

bmcfee:
At any rate -- this PR looks great now, thanks for all the effort!  I'm happy to merge if @lostanlen wants to give the :+1:

bmcfee:
:shipit: thanks again!

PR #1016, opened 2019-11-10
Added variable-Q filter creation

cwitkowitz:
#### Reference Issue

N/A

This adds a new parameter "gamma" to the constant-q filterbank creation (and by extension the calculation of filter lengths). "Gamma" allows us to lower the q-factor towards lower frequencies to a variable degree, while still maintaining an approximately constant q-factor at higher frequencies. Setting "gamma" equal to zero means that the filters created will be constant-q, and leaves the current functionality unchanged. The variable q-factor relaxes the need for large filters at low frequencies that the constant q-factor imposes.

See for more information:
Schörkhuber, Christian, et al. "A Matlab toolbox for efficient perfect reconstruction time-frequency transforms with log-frequency resolution." Audio Engineering Society Conference: 53rd International Conference: Semantic Audio. Audio Engineering Society, 2014.

#### Any other comments?
Forgive me if there is a better way to implement this (such as with separate functions), or if I have missed part of the process of making a contribution or anything required to make the change.

### DIFF
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -399,7 +399,7 @@ def __float_window(window_spec):
 
 
 @cache(level=10)
-def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, tuning=Deprecated(),
+def constant_q(sr, fmin=None, n_bins=84, bins_per_octave=12, gamma = 0, tuning=Deprecated(),
                window='hann', filter_scale=1, pad_fft=True, norm=1,
                dtype=np.complex64, **kwargs):
     r'''Construct a constant-Q basis.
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks @cwitkowitz for getting this one started!  As it turns out, I was recently experimenting with some vqt stuff locally, and have a basic implementation going (with some snagS).  Changing the basis constructor isn't quite enough to implement VQT though, since we need to modify the recursive downsampling step to allow for varying Q in each octave.

My implementation has some issues relating to normalization and parametrization though, which is why I haven't bothered to put it into a PR yet.  Since you seem to be interested in this, what do you think about combining our efforts?  If you're up for it, I can open a PR for my implementation and we can polish it up together.

cwitkowitz:
Sounds good to me!

I recognize that my changes do not actually implement the VQT. For my Master's thesis, I needed a way to create a filterbank corresponding to a CQT/VQT, with each band explicitly defined. I did not need to go any further than the changes I have added here, which gave me such filterbanks.

I am again working on a project where I just want the filterbank, not an actual VQT implementation, so what I added would be sufficient for my use-case. However, I would love to assist in bringing a full-blown variable-q transform to librosa. I knew doing that would require much more effort, and I am glad to here you've been working on it already.

PR #1018, opened 2019-11-11
[CR] Variable-Q transform

bmcfee:

This PR extends CQT to add support for variable-Q transforms.  It's a work in progress, but the core functionality seems to work already.  A new function (`vqt`) has been added that is similar to CQT, but allows for different filterbanks to be used in each downsampled octave.

I'm opening this PR now to prevent duplication of effort between this and #1016 .

As I mentioned above, I don't consider this "done" yet, for a few reasons:

- The reference papers for CQT and VQT use slightly different conventions for determining filter bandwidths (low-frequency aligned vs centered), and we should reconcile this before merging.
- There may be an issue of normalization / gain correction that we need to sort out.  @lostanlen has thoughts about this (discussed offline), but we haven't had time to thoroughly work through it all.  Getting this done may require introducing some new normalization or scaling parameters, or changing the behavior of cqt in a backwards-incompatible way.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -173,6 +173,8 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
     len_orig = len(y)
 
+    alpha = (2.0**(1. / bins_per_octave) - 1)
+
     if fmin is None:
         # C1 by default
         fmin = note_to_hz('C1')
@@ -515,6 +517,9 @@ def pseudo_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
     if tuning is None:
         tuning = estimate_tuning(y=y, sr=sr, bins_per_octave=bins_per_octave)
+
+    if gamma is None:
+        gamma = 24.7 * alpha / 0.108
 
     # Apply tuning correction
     fmin = fmin * 2.0**(tuning / bins_per_octave)
@@ -892,7 +897,6 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
     fmax_t = np.max(freqs)
 
     # Determine required resampling quality
-    alpha = (2.0**(1. / bins_per_octave) - 1)
     Q = float(filter_scale) / alpha
     filter_cutoff = fmax_t * (1 + 0.5 * filters.window_bandwidth(window) / Q) + 0.5 * gamma
     nyquist = sr / 2.0
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -915,6 +915,7 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
     if auto_resample and res_type != 'kaiser_fast':
 
         # Do the top octave before resampling to allow for fast resampling
+        # TODO: scale gamma as well
         fft_basis, n_fft, _ = __cqt_filter_fft(sr, fmin_t,
                                                n_filters,
                                                bins_per_octave,
@@ -961,7 +962,7 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
             my_sr /= 2.0
             my_hop //= 2
 
-        fft_basis, n_fft, _ = __cqt_filter_fft(my_sr, fmin_t * 2.0**(-i),
+        fft_basis, n_fft, _ = __cqt_filter_fft(my_sr, fmin_t * 2.0**-i,
                                                n_filters,
                                                bins_per_octave,
                                                filter_scale,
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -121,6 +121,7 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
 
     See Also
     --------
+    vqt
     librosa.core.resample
     librosa.util.normalize
 
@@ -167,124 +168,13 @@ def cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
            [  4.896e-08,   5.407e-07, ...,   9.176e-08,   1.051e-07]])
     '''
 
-    # How many octaves are we dealing with?
-    n_octaves = int(np.ceil(float(n_bins) / bins_per_octave))
-    n_filters = min(bins_per_octave, n_bins)
-
-    len_orig = len(y)
-
-    alpha = (2.0**(1. / bins_per_octave) - 1)
-
-    if fmin is None:
-        # C1 by default
-        fmin = note_to_hz('C1')
-
-    if tuning is None:
-        tuning = estimate_tuning(y=y, sr=sr, bins_per_octave=bins_per_octave)
-
-    # Apply tuning correction
-    fmin = fmin * 2.0**(tuning / bins_per_octave)
-
-    # First thing, get the freqs of the top octave
-    freqs = cqt_frequencies(n_bins, fmin,
-                            bins_per_octave=bins_per_octave)[-bins_per_octave:]
-
-    fmin_t = np.min(freqs)
-    fmax_t = np.max(freqs)
-
-    # Determine required resampling quality
-    Q = float(filter_scale) / (2.0**(1. / bins_per_octave) - 1)
-    filter_cutoff = fmax_t * (1 + 0.5 * filters.window_bandwidth(window) / Q)
-    nyquist = sr / 2.0
-
-    auto_resample = False
-    if not res_type:
-        auto_resample = True
-        if filter_cutoff < audio.BW_FASTEST * nyquist:
-            res_type = 'kaiser_fast'
-        else:
-            res_type = 'kaiser_best'
-
-    y, sr, hop_length = __early_downsample(y, sr, hop_length,
-                                           res_type,
-                                           n_octaves,
-                                           nyquist, filter_cutoff, scale)
-
-    cqt_resp = []
-
-    if auto_resample and res_type != 'kaiser_fast':
-
-        # Do the top octave before resampling to allow for fast resampling
-        fft_basis, n_fft, _ = __cqt_filter_fft(sr, fmin_t,
-                                               n_filters,
-                                               bins_per_octave,
-                                               filter_scale,
-                                               norm,
-                                               sparsity,
-                                               window=window)
-
-        # Compute the CQT filter response and append it to the stack
-        cqt_resp.append(__cqt_response(y, n_fft, hop_length, fft_basis, pad_mode))
-
-        fmin_t /= 2
-        fmax_t /= 2
-        n_octaves -= 1
-
-        filter_cutoff = fmax_t * (1 + 0.5 * filters.window_bandwidth(window) / Q)
-
-        res_type = 'kaiser_fast'
-
-    # Make sure our hop is long enough to support the bottom octave
-    num_twos = __num_two_factors(hop_length)
-    if num_twos < n_octaves - 1:
-        raise ParameterError('hop_length must be a positive integer '
-                             'multiple of 2^{0:d} for {1:d}-octave CQT'
-                             .format(n_octaves - 1, n_octaves))
-
-    # Now do the recursive bit
-    fft_basis, n_fft, _ = __cqt_filter_fft(sr, fmin_t,
-                                           n_filters,
-                                           bins_per_octave,
-                                           filter_scale,
-                                           norm,
-                                           sparsity,
-                                           window=window)
-
-    my_y, my_sr, my_hop = y, sr, hop_length
-
-    # Iterate down the octaves
-    for i in range(n_octaves):
-
-        # Resample (except first time)
-        if i > 0:
-            if len(my_y) < 2:
-                raise ParameterError('Input signal length={} is too short for '
-                                     '{:d}-octave CQT'.format(len_orig,
-                                                              n_octaves))
-
-            my_y = audio.resample(my_y, 2, 1,
-                                  res_type=res_type,
-                                  scale=True)
-            # The re-scale the filters to compensate for downsampling
-            fft_basis[:] *= np.sqrt(2)
-
-            my_sr /= 2.0
-            my_hop //= 2
-
-        # Compute the cqt filter response and append to the stack
-        cqt_resp.append(__cqt_response(my_y, n_fft, my_hop, fft_basis, pad_mode))
-
-    C = __trim_stack(cqt_resp, n_bins)
-
-    if scale:
-        lengths = filters.constant_q_lengths(sr, fmin,
-                                             n_bins=n_bins,
-                                             bins_per_octave=bins_per_octave,
-                                             window=window,
-                                             filter_scale=filter_scale)
-        C /= np.sqrt(lengths[:, np.newaxis])
-
-    return C
+    # CQT is the special case of VQT with gamma=0
+    return vqt(y=y, sr=sr, hop_length=hop_length,
+               fmin=fmin, n_bins=n_bins, gamma=0,
+               bins_per_octave=bins_per_octave,
+               tuning=tuning, filter_scale=filter_scale,
+               norm=norm, sparsity=sparsity, window=window, scale=scale,
+               pad_mode=pad_mode, res_type=res_type)
 
 
 @cache(level=20)
@@ -751,7 +641,7 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
         sampling rate of `y`
 
     hop_length : int > 0 [scalar]
-        number of samples between successive CQT columns.
+        number of samples between successive VQT columns.
 
     fmin : float > 0 [scalar]
         Minimum frequency. Defaults to C1 ~= 32.70 Hz
@@ -784,7 +674,7 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
 
         If `None`, tuning will be automatically estimated from the signal.
 
-        The minimum frequency of the resulting CQT will be modified to
+        The minimum frequency of the resulting VQT will be modified to
         `fmin * 2**(tuning / bins_per_octave)`.
 
     filter_scale : float > 0
@@ -806,10 +696,10 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
         See `filters.get_window` for details.
 
     scale : bool
-        If `True`, scale the CQT response by square-root the length of
+        If `True`, scale the VQT response by square-root the length of
         each channel's filter.  This is analogous to `norm='ortho'` in FFT.
 
-        If `False`, do not scale the CQT. This is analogous to
+        If `False`, do not scale the VQT. This is analogous to
         `norm=None` in FFT.
 
     pad_mode : string
@@ -830,8 +720,8 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
 
     Returns
     -------
-    CQT : np.ndarray [shape=(n_bins, t), dtype=np.complex or np.float]
-        Constant-Q value each frequency at each time.
+    VQT : np.ndarray [shape=(n_bins, t), dtype=np.complex or np.float]
+        Variable-Q value each frequency at each time.
 
     Raises
     ------
@@ -839,7 +729,7 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
         If `hop_length` is not an integer multiple of
         `2**(n_bins / bins_per_octave)`
 
-        Or if `y` is too short to support the frequency range of the CQT.
+        Or if `y` is too short to support the frequency range of the VQT.
 
     See Also
     --------
@@ -917,7 +807,6 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
     if auto_resample and res_type != 'kaiser_fast':
 
         # Do the top octave before resampling to allow for fast resampling
-        # TODO: scale gamma as well
         fft_basis, n_fft, _ = __cqt_filter_fft(sr, fmin_t,
                                                n_filters,
                                                bins_per_octave,
@@ -942,7 +831,7 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
     num_twos = __num_two_factors(hop_length)
     if num_twos < n_octaves - 1:
         raise ParameterError('hop_length must be a positive integer '
-                             'multiple of 2^{0:d} for {1:d}-octave VQT'
+                             'multiple of 2^{0:d} for {1:d}-octave CQT/VQT'
                              .format(n_octaves - 1, n_octaves))
 
     # Now do the recursive bit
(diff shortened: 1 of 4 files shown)


bmcfee:
Tagging @cwitkowitz

librosa/filters.py
```diff
@@ -642,7 +643,7 @@ def constant_q_lengths(sr, fmin, n_bins=84, bins_per_octave=12,
         raise ParameterError('Filter pass-band lies beyond Nyquist')
 
     # Convert frequencies to filter lengths
-    lengths = Q * sr / freq
+    lengths = Q * sr / (freq + gamma)
```
cwitkowitz: I do not think this is the correct way to calculate the lengths.

The paper defines the bandwidth as: alpha * f_k + gamma,
where we can think of alpha as the inverse of Q (my understanding is that it is arbitrary which convention is used, but I could be wrong - it could also affect reconstruction).

This code currently implements: alpha * (f_k + gamma)

bmcfee: Ah good catch!  I believe you're correct -- must have gotten messed up in the translation from their notation to ours.

bmcfee:
One question that we should sort out sooner rather than later: does it make sense to have an entirely separate function for vqt, or should we expand cqt to support vqt as a general case?

Arguments for two functions:

- Keeps CQT simple
- Easy to discover by documentation

Arguments for one function:

- Minimizes redundancy in the API
- Easier to maintain

Thoughts?

lostanlen:
Wow! A historic day for librosa!


> There may be an issue of normalization / gain correction that we need to sort out. @lostanlen has thoughts about this (discussed offline), but we haven't had time to thoroughly work through it all. Getting this done may require introducing some new normalization or scaling parameters, or changing the behavior of cqt in a backwards-incompatible way.

Yes. It's a dilemma:
(1) if we don't renormalize, then there is no energy conservation, which means that measuring loudness (after `perceptual_weighting`) will be numerically wrong in ways that are difficult to interpret
(2) if we do renormalize, then the maximum amplitude of a single sine wave (or slow AM/FM additive model) will not be consistent across the CQT axis, which will cause problem for CNNs in the time--frequency domain after max-pooling

I'm on the side of (1), but ideally we should have a utility to recover the littlewood-paley sum so that we can estimate dB from variable-Q transform.


Another thing to note is that `variable-Q` looks a bit like a pun on `constant-Q`, and the VQT acronym is not really established. The received terminology is "Nonstationary Gabor transform" (NSGT), but the problem is that it's restrictive on the type of wavelet (Gabor means Gaussian envelope, whereas we use a Hann envelope by default), so technically it isn't right either. But that's something to keep in mind for the docs.

bmcfee:
> I'm on the side of (1), but ideally we should have a utility to recover the littlewood-paley sum so that we can estimate dB from variable-Q transform.

Another option is to add a new scaling mode, analogous to how `fft` has `mode='ortho'`, in case people want that behavior.  Or, we could do as suggested elsewhere (offline?) and build in the scaling to any perceptual weighting function, but that seems a little messy from the API perspective.  I'm not sure there's a good solution overall here.



> The received terminology is "Nonstationary Gabor transform" (NSGT), but the problem is that it's restrictive on the type of wavelet (Gabor means Gaussian envelope, whereas we use a Hann envelope by default), so technically it isn't right either.

Indeed, that's why I tend to prefer VQT over NSGT, even though it's a relatively non-informative name.  (Technically, a DFT has variable Q as well...)

librosa/core/constantq.py
```diff
@@ -728,10 +730,257 @@ def icqt(C, sr=22050, hop_length=512, fmin=None, bins_per_octave=12,
...
+        number of samples between successive CQT columns.
+
+    fmin : float > 0 [scalar]
+        Minimum frequency. Defaults to C1 ~= 32.70 Hz
+
+    n_bins : int > 0 [scalar]
+        Number of frequency bins, starting at `fmin`
+
+    gamma : number > 0 [scalar]
+        Bandwidth offset for determining filter lengths.
```
lostanlen: I would love to see a more thorough documentation of the role of gamma. 

* What is the unit of `gamma`?
* What happens in the default case `gamma=None`? 
* What happens for very large `gamma`? 
* What are some typical values? (from the Schörkhuber group)

librosa/core/constantq.py
```diff
@@ -728,10 +730,257 @@ def icqt(C, sr=22050, hop_length=512, fmin=None, bins_per_octave=12,
...
+
+    tuning : None or float
+        Tuning offset in fractions of a bin.
+
+        If `None`, tuning will be automatically estimated from the signal.
+
+        The minimum frequency of the resulting CQT will be modified to
+        `fmin * 2**(tuning / bins_per_octave)`.
+
+    filter_scale : float > 0
```
lostanlen: How does this play with the `gamma`?

librosa/core/constantq.py
cwitkowitz: Although it was not explicitly stated, I think we can infer that the units of gamma are hertz, since it is added as a constant offset to bandwidth.

I don't think it makes sense to default gamma to None. It should instead be defaulted to zero, where it would just give us the standard CQT. Similarly, with regard to the question Brian posed earlier, my opinion is that the CQT and VQT should be combined into one function, if the octave-wise iterative processing does not require significant changes (I need to read the paper again to better understand how they account for gamma in their implementation and reconstruction).

A very large gamma is not advised in my opinion. In the paper, I don't recall them going past gamma=25, but even that gives very poor frequency resolution at lower frequencies. However, it should be noted once again that they were using an alternate convention for calculating bandwidth, and my assessment of poor frequency resolution at lower frequencies for gamma=25 was made while using the convention currently adopted by librosa.

I think the range [0, 5] is where I would usually place the parameter for practical scenarios. Besides gamma=0 giving the CQT, there is one other special case where gamma can be set to make bandwidths proportional, not to be confused with equal, to the ERB scale. I think special case gamma would be somewhere in this range, but depends on the number of bins per octave.

bmcfee: >     * What happens in the default case `gamma=None`?

Nothing yet, but the idea was to automatically fill a value based on ERB.  I don't think there's much sense in having 0 be the default if the two functions remain separate (vqt and cqt), but of course, if we merge them, then `gamma=0` is a totally fine (and expected) default.

Otherwise, I agree with @cwitkowitz's comments.

cwitkowitz: True, if the functions remain separate, the ERB setting would be a good default for gamma='None'.

librosa/core/constantq.py
bmcfee: They're basically independent.  `filter_scale` is a multiplier on the Q factor, allowing for longer (or shorter) filters than would be dictated by the bandwidth calculation.  This was introduced very (very) long ago to match a feature of mcvicar's implementation.

lostanlen: yes but isn't this going to lead us astray of ERB inadvertently?
if we tinker with the `filter_scale`, then this changes the bandwidth in inverse proportion, and thus changes the frequency where the ERB scales transitions from constant-bw to constant-Q. (because physiologically speaking, the ERB is an upper bound on the latency of the auditory system)
Ideally i'd like to see an equation in the documentation where all these elements are discussed in conjunction.

bmcfee: Yeah, that's fair.  I'd actually be okay with deprecating this parameter and simplifying the whole thing.

lostanlen: the way i like to think about it is the following: independently of the number of bins per octave and from the choice of wavelet, there are two degrees of freedom:
(1) `max_Q`: maximum quality factor. Defaults to `bins_per_octave`.
(2) `min_bandwidth`: minimum bandwidth, defaulting to zero.

If `max_Q < bin_per_octave`, there is an error.

The bandwidth of the filter at any frequency `f` is equal to:
(i) if `f < min_bandwidth`, then `f` (i.e. the quality factor is 1) 
(ii) otherwise, if `f` ranges between `min_bandwidth` and `max_Q*min_bandwidth`, then `min_bandwidth`
(iii) if `(f/max_Q) > min_bandwidth`, then `f/max_Q` (i.e. the quality factor is max_Q)

So there are actually three regimes: constant-Q, then constant-bw, then constant-Q again. The priorities are: Q>=1 first; bw > min_bw unless impossible; Q=max_Q unless impossible.

(in practice, we want a definition of bandwidth that is a little bit more whizzy than f/Q, involving 2^(1/Q)-2^(-1/Q) etc. but that doesn't change the narrative. also, we might want to smoothly interpolate between these regimes, rather than having sharp breakpoints)

just my 2c

lostanlen: e.g. see here, where i expressed `min_bandwidth` in terms of its inverse (`max_scale`, measured in samples) for the convenience of the user.

https://github.com/lostanlen/WaveletScattering.jl/blob/master/src/spec.jl#L9

lostanlen: here is where the construction happens
https://github.com/lostanlen/WaveletScattering.jl/blob/master/src/spec1d.jl#L55

bmcfee: > involving 2^(1/Q)-2^(-1/Q)

Okay, let's resurface that conversation now, since we previously had it OTR.

Our current CQT filterbank uses `2**(1/bpo) - 1` to determine bandwidth, as given in the S&K 2010 paper.  However, the later paper that I'm basing this vqt implementation on uses `2**(1/bpo) - 2**(-1/bpo)`, effectively centering the frequency range rather than edge-aligning it.  I can see arguments for both, but should we expose an interface to control this?  Or is it way too far in the weeds for users to care about?

lostanlen: For `bpo >> 1` and wavelets with a quasi-symmetric profile in the Fourier domain (i.e. almost every use case, including the default), it doesn't really make a difference. If we were making Gammatones with Q=2, then i would argue otherwise.

I don't see a compelling reason to break the current implementation

cwitkowitz:
I am starting to think that our implementation might be cleaner if we choose instead to combine the CQT and VQT functions in constantq.py.

From what I can tell, the only differences (assuming gamma=0 for CQT and gamma>0 or gamma=None for VQT) between the two functions is the generation and/or rescaling of the filters for each iteration (as opposed to reusing the same bases by simply downsampling them). I think we could probably make a few tweaks so that we have the VQT function looking similar to the way it does now, with the CQT function simply calling VQT with gamma=0.

I think it would be easier to test this way as well.

Thoughts?

bmcfee:
> I am starting to think that our implementation might be cleaner if we choose instead to combine the CQT and VQT functions in constantq.py.

I raised this point earlier [here](https://github.com/librosa/librosa/pull/1018#issuecomment-552956253).  I'm still not convinced that they should be the same function, since that could confuse the API.  Probably the best solution is to factor out the commonalities and put some light-weight API on top so that we still have `cqt` and `vqt`, but minimize redundant code.

I agree with your assessment that the only real difference is the rescaling logic for each octave.  If we can generalize this to support `gamma=0` directly, and not introduce additional branching paths (conditionals within the downsampling loop), that would be ideal.

cwitkowitz:
I think that's the right way to move forward. What we have now seems to work well, so I wouldn't be opposed to adding some test cases and inserting it into master for now. I just feel like the functions are too similar to leave by themselves. As I mentioned in my current pull request for the VQT default gamma, maybe we should consider the implementation presented in the original VQT paper http://www.cs.tut.fi/sgn/arg/CQT/ (which we would have to port over to python), as it is my understanding that it covers all the CQT/VQT bases, and has a generalized inverse.

If we want to just finish what we have now, I need to add another comment addressing the ERB constants, and then I could essentially copy the CQT test cases for the VQT function. Then, I believe we could call it a pull request (correct me if I am wrong).

bmcfee:
> I think that's the right way to move forward.

:+1:  but let's get everything stable and correct first, then worry about refactoring.



> As I mentioned in my current pull request for the VQT default gamma, maybe we should consider the implementation presented in the original VQT paper http://www.cs.tut.fi/sgn/arg/CQT/ (which we would have to port over to python), as it is my understanding that it covers all the CQT/VQT bases, and has a generalized inverse.

Indeed -- this is more or less what I had in mind for the default gamma setting.  My main concern with implementing it exactly as stated is that their frequency binning (alpha, eq 11) is slightly different from ours: centered (theirs) vs left-aligned (ours, inherited from earlier work on cqt toolbox) in the frequency domain, and this leads to slightly different calculations for gamma.  There are reasonable arguments to be had for both ways of specifying the bins, and I could envision supporting both modes down the road.  @lostanlen and I whiteboarded this back in november, but didn't really come to a conclusion about how to proceed.  I think this is ultimately a separate issue, but whatever we implement here should be consistent with how the bandwidths are set.

bmcfee:
(sorry, did not mean to close PR)

bmcfee:
Just rebased this one to current master so we can get tests running again.

Aside from @cwitkowitz 's other pr on this #1045 , there are still a few points we've yet to resolve.  The big one IMO has to do with centering vs left-aligning frequency ranges, since that's a huge difference between us and the reference implementation / paper.

Arguments for keeping it left-aligned:
- No changes to the code (we already do this)
- Semantics of `fmin` are clear (does not depend on Q or other parameters)

Arguments for centering:
- Consistency with literature
- Arguably, consistency (in the abstract) with other parts of the library (eg frame alignment)

Other minor things to work out:

- deprecate `filter_scale`?
- Separate or same functions?
  - On this, I'd be okay with having everything implemented in `cqt` with default `gamma=0`, and then making `vqt` a thin wrapper on `cqt` that sets the default `gamma=None` (implemented via #1045) and otherwise does no work.  Sound good?

bmcfee:
One slight hiccup to the idea of merging `cqt` and `vqt` as one function with a light wrapper: any exceptions / warnings raised could get confusing since they explicitly say CQT or VQT in the current implementation.  I'd be okay with rewording the exceptions to avoid this problem.

The other minor barrier to merging implementations is that the vqt is going to be a little slower than the cqt, since it's building new bases for each octave.  I don't think this constitutes a huge performance hit, but it's worth considering.

As a quick anecdote: example track at 44.1K gives these timings on my machine:
```python
In [11]: %timeit librosa.cqt(y=y, sr=sr, bins_per_octave=5*12, n_bins=5*12*7)                                                                                                                 
1.57 s ± 7.81 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [14]: %timeit librosa.vqt(y=y, sr=sr, bins_per_octave=5*12, n_bins=5*12*7, gamma=0)                                                                                                        
1.59 s ± 31.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
So, not appreciably different for a decently representative use case.

bmcfee:
Well that was a gnarly merge, but I think we're back in business.

bmcfee:
Current status:

- [x] consolidate CQT and VQT implementations.  CQT now wraps to VQT; exceptions are updated accordingly.
- [ ] inverse VQT.  This will follow the same pattern as the forward transform, and I anticipate no major headaches now that we have the forward path done.
- [ ] griffinlim_vqt?  Or should we just add a gamma= parameter to `cqt` and document that it works for both?

Things left to decide:

- Should we deprecate `filter_scale`?
- Do we want to implement a normalization mode following this comment:

> I'm on the side of (1), but ideally we should have a utility to recover the littlewood-paley sum so that we can estimate dB from variable-Q transform.

This seems like a useful thing to have in general (and also for CQT), but maybe it doesn't need to be part of this PR?

librosa/core/constantq.py
```diff
@@ -723,10 +617,274 @@ def icqt(C, sr=22050, hop_length=512, fmin=None, bins_per_octave=12,
...
+
+        If `gamma=0`, produces the constant-Q transform.
+
+        If 'gamma=None', gamma will be calculated such that filter bandwidths are equal to a
+        constant fraction of the equivalent rectangular bandwidths (ERB). This is accomplished
+        by solving for the gamma which gives B_k = alpha * f_k + gamma = C * ERB(f_k), where
+        B_k is the bandwidth of filter k with center frequency f_k, alpha is the inverse of
+        what would be the constant Q-factor, and C = alpha / 0.108 is the constant fraction
+        across all filters. Here we use ERB(f_k) = 24.7 + 0.108 * f_k, the best-fit curve derived
+        from experimental data in [2]_.
```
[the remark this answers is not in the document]
bmcfee: Looked a bit more into the Glasberg & Moore paper and found some details:

image

This makes me think the ERB estimate here is probably not well suited for our default (Hann) windows.  I get that this was a psychoacoustics study, but I'm wondering how relevant these bandwidth settings are for VQT with filter shapes substantially different from their model.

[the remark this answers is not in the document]
cwitkowitz: Good point. A few questions:
Where in librosa is ERB returned for a given window function? Are you referring to ENBW?
What exactly is the difference between ERB and ENBW? Is ERB just the ENBW for the auditory filterbank?

Essentially, we have:
B_k = alpha * f_k + gamma = C * ERB(f_k)
where C is a constant, meaning
gamma = C * ERB(f_k) - alpha * f_k

We need to fix C, such that gamma has no dependence of frequency f_k.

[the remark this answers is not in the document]
cwitkowitz: What exactly do the values for WINDOW_BANDWIDTHS in filters.py mean?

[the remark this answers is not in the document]
cwitkowitz: Given the consolidation of CQT/VQT, do we still need to duplicate all test cases, or can we simply collapse them now?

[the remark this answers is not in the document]
cwitkowitz: So I guess the question then becomes, how can we define the ERB as a function of center frequency and window type?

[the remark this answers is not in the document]
bmcfee: > So I guess the question then becomes, how can we define the ERB as a function of center frequency and window type?

Exactly.  I'm a bit stumped on this, tbh.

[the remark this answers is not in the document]
cwitkowitz: I have been pondering this for a bit. The point of this default parameter selection was only to fix gamma such that filter bandwidth varies with that of ERB (which no one necessarily claims will help - it is just a means for interpretation of gamma). We make each bandwidth a constant fraction of the ERB, which is not the same as the ERB itself. As long as the bandwidth of a filter using each window function increases linearly, I think the default does what we want it to, i.e. increase bandwidth with the same rate as ERB.

I am no expert in psychoacoustics, and this seems like a pretty deep topic, so feel free to correct me or dispute my claim.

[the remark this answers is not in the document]
bmcfee: > I am no expert in psychoacoustics, and this seems like a pretty deep topic, so feel free to correct me or dispute my claim.

Nor am I :grin:  Maybe it's time to bring in the heavies ... @dpwe got any thoughts on this?

I think I generally agree that this is probably okay.  We just need to be a bit careful about how the documentation is worded.

bmcfee:
Pulling IVQT and GL-VQT out to a separate issue #1161 to be resolved at a later date.

This brings the scope of todos on this PR down to:
 
>     * Should we deprecate `filter_scale`?
> 

I think I'm convinced that we should keep it.  It's rarely used, but it does no harm and it could be useful in some contexts (eg efficiently computing a subset of frequency responses from a high-res CQT without building the whole thing).

>     * Do we want to implement a normalization mode following this comment:
> > I'm on the side of (1), but ideally we should have a utility to recover the littlewood-paley sum so that we can estimate dB from variable-Q transform.

This one, I'm less sure of, mainly because I don't know how to do it.

Otherwise, I think this is ready for a final CR.

lostanlen:
> > * Do we want to implement a normalization mode following this comment:
> > > I'm on the side of (1), but ideally we should have a utility to recover the littlewood-paley sum so that we can estimate dB from variable-Q transform.

> This one, I'm less sure of, mainly because I don't know how to do it.

I'm now 100% inclined towards (1), i.e. no fancy normalization, aftering having read the excellent paper by Cheuk et al. on nnAudio. They took great care in reproducing the librosa CQT output exactly.
https://arxiv.org/abs/1912.12055

Clearly, deep learning practitioners they care a lot about Diracs and sine sweeps, and very little about energy conservation. 
It's also less code, less docs, and less tests on our end.

bmcfee:
> Clearly, deep learning practitioners they care a lot about Diracs and sine sweeps, and very little about energy conservation.
> It's also less code, less docs, and less tests on our end.

:laughing: 

Seriously though, I was envisioning this as a helper function like perceptual-weighting that you could use to normalize a cqt/vqt output after the fact.  But I agree that it's not critical at this point / could be added later.

bmcfee:
Any objection to merging this?  It's been through quite a bit of preliminary review already.

cwitkowitz:
Looks ready to me. I have a few questions.

Is there any plan to do anything regarding the bandwidth convention in the future?
What about the original implementation in the VQT paper?
Are we satisfied with the current wrapped solution instead?

From what I gather, future work so far consists only of adding an inverse VQT and griffinlim VQT.

bmcfee:
> Is there any plan to do anything regarding the bandwidth convention in the future?

No plan as of now.  The most likely path that I see is that we add a switch to change from left-aligned to centered (in frequency) filters.  I guess it won't matter much until we get around to implementing asymmetric windows.

> What about the original implementation in the VQT paper?

In terms of the filter centering, or did you mean something else?

> Are we satisfied with the current wrapped solution instead?
> 

As in, `cqt(...) = vqt(..., gamma=0)`?  Or something else?

I'm fine with it.

> From what I gather, future work so far consists only of adding an inverse VQT and griffinlim VQT.

Those are the main things, yes.  GL will be easy once ivqt is in place.

cwitkowitz:
> In terms of the filter centering, or did you mean something else?

I remember considering that implementation as a possible way to combine cqt/vqt efficiently. However, we decided to avoid it for now because of the alternate bandwidth convention. I'm not sure that there is any strong benefit to do it one way or the other. I'm just wondering where you stood on that.

> As in, `cqt(...) = vqt(..., gamma=0)`? Or something else?
>
> I'm fine with it.

Yes, that's what I'm referring to. Again my reason for asking this stems from the original implementation proposed in the VQT paper. I'm just wondering whether we get any speed benefits or anything else from the original implementation.

bmcfee:
> Yes, that's what I'm referring to. Again my reason for asking this stems from the original implementation proposed in the VQT paper. I'm just wondering whether we get any speed benefits or anything else from the original implementation.

Ah.  I benchmarked it earlier on some representative examples, and the performance hit relative to pure CQT was extremely small, so IMO it's worth it to have a simpler code base.

I also have some performance improvements in mind #1144  that should make up the difference.

bmcfee:
> I remember considering that implementation as a possible way to combine cqt/vqt efficiently. However, we decided to avoid it for now because of the alternate bandwidth convention. I'm not sure that there is any strong benefit to do it one way or the other. I'm just wondering where you stood on that.

Sorry, forgot to respond to that earlier.

I think it's more or less independent of the cqt/vqt combination, since both can (maybe should?) be defined with centered bandwidth calculations.  FWIW the original CQT toolbox paper used left-aligned, and then the implementation later moved to centered.  Point is, I think both cqt and vqt can be implemented either way, and it doesn't change much, but we should provide a consistent interface.  But this can come later IMO.

cwitkowitz:
Makes sense. Well, I have no objections to merging.

bmcfee:
I'll merge this tomorrow, unless any one objects.

PR #1020, opened 2019-11-19
Fixed calculation of maximum filter cutoff

cwitkowitz:
#### Reference Issue

N/A

In the current code, gamma is added directly to the maximum frequency to calculate filter cutoff.

This results in

filter_cutoff = f_k + gamma + 0.5 * B_k + gamma * 0.5 * B_k / f_k

, whereas my change implements

filter_cutoff = f_k + 0.5 * B_k + 0.5 * gamma

, which I believe is what we want instead.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -878,9 +878,9 @@ def vqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84, gamma=None,
     fmax_t = np.max(freqs)
 
     # Determine required resampling quality
-    Q = float(filter_scale) / (2.0**(1. / bins_per_octave) - 1)
-
-    filter_cutoff = (fmax_t + gamma) * (1 + 0.5 * filters.window_bandwidth(window) / Q)
+    alpha = (2.0**(1. / bins_per_octave) - 1)
+    Q = float(filter_scale) / alpha
+    filter_cutoff = fmax_t * (1 + 0.5 * filters.window_bandwidth(window) / Q) + 0.5 * gamma
     nyquist = sr / 2.0
 
     auto_resample = False


bmcfee:
This LGTM, but I'd like to get @lostanlen's perspective on it, especially with respect to the interpretation of the gamma parameter (and whatever units it implicitly carries).

lostanlen:
> This LGTM, but I'd like to get @lostanlen's perspective on it, especially with respect to the interpretation of the gamma parameter (and whatever units it implicitly carries).

Thanks. Yes, the unit of this parameter needs to be clearly specified and documented.

Also, i'm a little bit uncomfortable at the thought that, for large `gamma`, the ratio of center frequency to bandwidth might fall below 1. (and even below 1/2 ?? )

cwitkowitz:
I'm not sure which line you are referring to @lostanlen , but I can imagine it is the second filter cutoff calculation. I am not entirely sure yet, as the varied Q at that point in the function might reflect gamma, though I am still trying to understand the current implementation. I will (slowly but surely) move through the rest of the function to do a full review. My change here is by no means exhaustive.

cwitkowitz:
With respect to large gamma, maybe we could enforce an upper boundary that prevents this case. As I mentioned in the main pull request, I believe there is little utility in setting gamma to be large. Support for large gamma would mainly be for the sake of completeness in my opinion.

bmcfee:
> Support for large gamma would mainly be for the sake of completeness in my opinion.

I agree with that; large gamma is mainly useful for approximating STFT-like behavior in the bottom end.

At any rate, is there anything else to do on this PR or should I fold it into #1018 ?

cwitkowitz:
I think you can fold it in for now. I need to review the rest of the current implementation, but will be preoccupied for the next few weeks. For future changes, should I just commit them to your current PR? If so, I will need the appropriate permissions.

bmcfee:
> I think you can fold it in for now.

Ok, thanks!

> For future changes, should I just commit them to your current PR? If so, I will need the appropriate permissions.


I think this flow works well enough.  I'm a little wary of messing with permissions here, mainly because the main librosa repo was migrated over from my personal account at some point, and it's never been clear how github handles the permissions on the backend.  (Not that I'm worried about you doing anything malicious :grin: just paranoid in general.)

Issue #1007, opened 2019-10-14
Perceptual weighting is limited to dB(A) standard


beasteers:
Since there's A weighting, we could also support C weighting 🤷‍♀️

lostanlen:
Yes, i'm down for adding B, C, D weighting in addition to A.
The way i would see it would be to add a kwarg to `perceptual_weighting`, named `letter` (or `standard`, or `curve` ...) that would default to `A` for backwards compatibility and that could take values `B`, `C`, `D`. 
See: 
'https://github.com/librosa/librosa/blob/master/librosa/core/spectrum.py#L1747

Also worth reading if you haven't already done so: #463

bmcfee:
I like this proposal.  Historically, we only implemented A-weighting to reproduce @mattmcvicar 's chroma pipeline.

Filling out @lostanlen's suggestion a bit, how about the following:

- Generalize `core.A_weighting` to support `B,C,D,Z` modes, and rename it to something like `loudness_curve`.  We could also use `None` to support a flat response
- Add a deprecation-rename for `core.A_weighting` to be removed in 0.9.
- Add the corresponding parameter to `perceptual_weighting`

beasteers:
Sgtm! 😊

lostanlen:
@bmcfee do you have an opinion on the best name for the kwarg? (with A/B/C/D/Z values)

nullmightybofo:
> [...]
>     * Generalize `core.A_weighting` to support `B,C,D,Z` modes, and rename it to something like `loudness_curve`.  We could also use `None` to support a flat response
> [...]

May I suggest a different term than `loudness_curve`? That term makes me think of a time-series of e.g. per-frame loudness values (though the context should make it clear that this is not what is meant).

Wouldn't  `frequency_weighting` or `weighting_curve` be more appropriate?

https://en.wikipedia.org/wiki/A-weighting

bmcfee:
> Wouldn't `frequency_weighting` or `weighting_curve` be more appropriate?

I like both of those, thanks.

PR #1022, opened 2019-11-26
mu-law companding

### LINKED ISSUE #1034: μ-law companding
#### Description

From: 
https://en.wikipedia.org/wiki/%CE%9C-law_algorithm

"The μ-law algorithm is a companding algorithm, primarily used in 8-bit PCM digital telecommunication systems in North America and Japan. Companding algorithms reduce the dynamic range of an audio signal. Companding algorithms reduce the dynamic range of an audio signal. In analog systems, this can increase the signal-to-noise ratio (SNR) achieved during transmission; in the digital domain, it can reduce the quantization error (hence increasing signal to quantization noise ratio). These SNR increases can be traded instead for reduced bandwidth for equivalent SNR."

Addressed by PR #1022

bmcfee:
#### Reference Issue

Related to #1007 , but different.
[EDIT by @lostanlen: closes #1034]

This PR implements mu-law compression and expansion, with and without quantization.

#### Any other comments?

A couple of things to consider here:

- Should we also add A-law companding?  Does anyone even use it?
- Would it make sense to separate the quantization parameter from the compression parameter?  Right now, it quantizes to `lg(1+mu)` bits, which derives from the common setting of `mu=255` and 8-bit quantization.  But there's technically no reason that these things have to be tied together.
- Should we worry about `dtype` inference based on the number of quantization levels?  Numpy doesn't make it totally trivial to infer the minimum dtype for a given value range, but it's something we could do with a bit of work.  It just seems wasteful to use `int64` to hold 8-bit quantized values.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -23,7 +23,8 @@ from ..util.exceptions import ParameterError
 __all__ = ['load', 'stream', 'to_mono', 'resample',
            'get_duration', 'get_samplerate',
            'autocorrelate', 'lpc', 'zero_crossings',
-           'clicks', 'tone', 'chirp']
+           'clicks', 'tone', 'chirp',
+           'mu_compress', 'mu_expand']
 
 # Resampling bandwidths as percentage of Nyquist
 BW_BEST = resampy.filters.get_filter('kaiser_best')[2]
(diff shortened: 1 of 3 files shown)


librosa/core/audio.py
```diff
@@ -1347,3 +1348,177 @@ def chirp(fmin, fmax, sr=22050, length=None, duration=None, linear=False, phi=No
...
+
+    if mu <= 0:
+        raise ParameterError('mu-law compression parameter mu={} '
+                             'must be strictly positive.'.format(mu))
+
+    if np.any(x < -1) or np.any(x > 1):
+        raise ParameterError('mu-law input x={} must be in the '
+                             'range [-1, +1].'.format(x))
+
+    x_comp = np.sign(x) * np.log1p(mu * np.abs(x)) / np.log1p(mu)
```
lostanlen: `log1p` 😍

lostanlen:
Thanks for doing this. It looks great but i'm curious as to the practical purpose. Given that all of this operates on IEEE 754 arithmetic, i doubt it reflects the effect of mu law in practice (e.g. audio sensing with limited resources)

Suppose that i have a WAV file in PCM (~fixed-point) format, encoded with a bit depth of 16 (or 24, or 32, etc.)
Suppose that i want to simulate how 8-bit compression affects the audio quality of that file upon listening to it.
How do i do this in librosa?

bmcfee:
> Suppose that i want to simulate how 8-bit compression affects the audio quality of that file upon listening to it.
> How do i do this in librosa?

```python
z = librosa.mu_compress(y, quantize=True)
y_compand = librosa.mu_expand(z, quantize=True)
```

I don't think we need to go as far as to add an explicit `compand` shortcut function, but above should do the trick,

bmcfee:
One more question raised offline: should quantization be the default behavior? Or are there more compelling use cases for continuous mu compression?

lostanlen:
LGTM. I would prefer `quantize=True` by default though. The main application of mu-law is to have it quantized, be it for storage or for multinomial modeling à la WaveNet. Furthermore, the quantization is actually non-trivial, with a call to `np.digitize` and a `-int(mu + 1)//2` offset.

bmcfee:
Ok, I'm onboard with quantizing by default.

Final question: i put this in the audio submodule, because it seemed like the most logical place.  But I could also see it living under `effects` alongside preemphasis and other time-domain transformations.  What do you think about migrating it there?

lostanlen:
I would prefer `audio` over `effects`. Although it is true that back-and-forth companding and expansion does result in an audible effect, it's not the main purpose of why we have it here.

bmcfee:
> I would prefer `audio` over `effects`. Although it is true that back-and-forth companding and expansion does result in an audible effect, it's not the main purpose of why we have it here.
> 
> Please merge at your convenience

Fair enough, thanks again!

Should we bump this up to the 0.7.2 release then?

lostanlen:
Sure. I just bumped it to 0.7.2

bmcfee:
whoops, forgot to set `quantize=True` to default.  Fixing and then will merge again.

PR #1031, opened 2019-12-04
various DTW bugfixes

### LINKED ISSUE #1014: numba's DTW segfaults on NaN similarity
#### Description

The example code plot_music_sync.py from [https://librosa.github.io/librosa/auto_examples/plot_music_sync.html#sphx-glr-download-auto-examples-plot-music-sync-py](https://librosa.github.io/librosa/auto_examples/plot_music_sync.html#sphx-glr-download-auto-examples-plot-music-sync-py) runs fine with sir_duke_fast.mp3 and sir_duke_slow.mp3. But encounter segmentation fault when using other mp3 files.
#### Steps/Code to Reproduce
1. Replace 
```
x_1, fs = librosa.load('audio/sir_duke_fast.mp3')
```
to 
```
x_1, fs = librosa.load('audio/BFJazz_-_Stargazer_from_Texas.mp3')
```
in plot_music_sync.py.

2. Run plot_music_sync.py again and segmentation fault occurs.

3. The exception seems to be thrown from this line:
```
D, wp = librosa.sequence.dtw(X=x_1_chroma, Y=x_2_chroma, metric='cosine')
```

#### Expected Results

Python should exit without Segmentation fault: 11
#### Actual Results

/Users/igneus/.pyenv/versions/3.7.4/lib/python3.7/site-packages/librosa/core/audio.py:146: UserWarning: PySoundFile failed. Trying audioread instead.
  warnings.warn('PySoundFile failed. Trying audioread instead.')
/Users/igneus/.pyenv/versions/3.7.4/lib/python3.7/site-packages/librosa/core/audio.py:146: UserWarning: PySoundFile failed. Trying audioread instead.
  warnings.warn('PySoundFile failed. Trying audioread instead.')
Segmentation fault: 11
#### Versions

Darwin-19.0.0-x86_64-i386-64bit
Python 3.7.4 (default, Sep 26 2019, 17:15:53)
[Clang 10.0.1 (clang-1001.0.46.4)]
NumPy 1.17.3
SciPy 1.3.1
librosa 0.7.1

### LINKED ISSUE #1029: DTW with global constraints is destructive to cost-matrix input
#### Description

When calling DTW with a precomputed cost matrix and `global_constraints=True`, the cost matrix is passed by reference and corrupted by the `fill_off_diagonal` call which populates the cost matrix with `np.inf` outside the Sakoe-Chiba band.

#### Steps/Code to Reproduce
```python
import librosa
import numpy as np

# Generate one frame of random STFT-like energy
C = np.ones((50, 50))

librosa.dtw(C=C, global_constraints=True)
```

#### Expected Results
`C` should be unmodified after call to `dtw`.

#### Actual Results
`C` has nans created by `dtw`.

#### Versions
```
Linux-5.0.0-32-lowlatency-x86_64-with-debian-buster-sid
Python 3.7.5 (default, Oct 25 2019, 15:51:11) 
[GCC 7.3.0]
NumPy 1.17.4
SciPy 1.3.0
librosa 0.7.1
```

bmcfee:
#### Reference Issue

- #1014 
- #1029 (fixed)

[EDIT by lostanlen: closes #1014, closes #1029]

This PR (in progress) is to fix issues in our DTW implementation, specifically relating to forced full-sequence alignment and errors arising from NaN/Inf in the distance matrix.

### DIFF
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -163,7 +163,9 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
     c_is_transposed = False
 
     # calculate pair-wise distances, unless already supplied.
+    C_local = False
     if C is None:
+        C_local = True
         # take care of dimensions
         X = np.atleast_2d(X)
         Y = np.atleast_2d(Y)
@@ -195,6 +197,9 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
 
     if global_constraints:
         # Apply global constraints to the cost matrix
+        if not C_local:
+            # If C was provided as input, make a copy here
+            C = np.copy(C)
         fill_off_diagonal(C, band_rad, value=np.inf)
 
     # initialize whole matrix with infinity values
@@ -351,6 +356,7 @@ def __dtw_backtracking(D_steps, step_sizes_sigma, subseq):  # pragma: no cover
     # Stop criteria:
     # Setting it to (0, 0) does not work for the subsequence dtw,
     # so we only ask to reach the first row of the matrix.
+
     while (subseq and cur_idx[0] > 0) or (not subseq and cur_idx != (0, 0)):
         cur_step_idx = D_steps[(cur_idx[0], cur_idx[1])]
 
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -146,12 +146,40 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
     >>> plt.show()
     '''
     # Default Parameters
+    default_steps = np.array([[1, 1], [0, 1], [1, 0]], dtype=np.int)
+    default_weights_add = np.zeros(3, dtype=np.float)
+    default_weights_mul = np.ones(3, dtype=np.float)
+
     if step_sizes_sigma is None:
-        step_sizes_sigma = np.array([[1, 1], [0, 1], [1, 0]])
-    if weights_add is None:
-        weights_add = np.zeros(len(step_sizes_sigma))
-    if weights_mul is None:
-        weights_mul = np.ones(len(step_sizes_sigma))
+        # Use the default steps
+        step_sizes_sigma = default_steps
+
+        # Use default weights if none are provided
+        if weights_add is None:
+            weights_add = default_weights_add
+
+        if weights_mul is None:
+            weights_mul = default_weights_mul
+    else:
+        # If we have custom steps but no weights, construct them here
+        if weights_add is None:
+            weights_add = np.zeros(len(step_sizes_sigma), dtype=np.float)
+
+        if weights_mul is None:
+            weights_mul = np.ones(len(step_sizes_sigma), dtype=np.float)
+
+        # Make the default step weights infinite so that they are never
+        # preferred over custom steps
+        default_weights_add.fill(np.inf)
+        default_weights_mul.fill(np.inf)
+
+        # Append custom steps and weights to our defaults
+        step_sizes_sigma = np.concatenate((default_steps, step_sizes_sigma))
+        weights_add = np.concatenate((default_weights_add, weights_add))
+        weights_mul = np.concatenate((default_weights_mul, weights_mul))
+
+    if np.any(step_sizes_sigma < 0):
+        raise ParameterError('step_sizes_sigma cannot contain negative values')
 
     if len(step_sizes_sigma) != len(weights_add):
         raise ParameterError('len(weights_add) must be equal to len(step_sizes_sigma)')
@@ -166,6 +194,8 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
     c_is_transposed = False
 
     # calculate pair-wise distances, unless already supplied.
+    # C_local will keep track of whether the distance matrix was supplied
+    # by the user (False) or constructed locally (True)
     C_local = False
     if C is None:
         C_local = True
@@ -221,8 +251,8 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
     # initialize step matrix with -1
     # will be filled in calc_accu_cost() with indices from step_sizes_sigma
     D_steps = np.zeros(D.shape, dtype=np.int)
-    # TODO: replace these with steps that correspond to left (first row)
-    # and up (first column)
+
+    # these steps correspond to left- (first row) and up-(first column) moves
     D_steps[0, :] = 1
     D_steps[:, 0] = 2
 
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -244,6 +244,9 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
         else:
             # perform warping path backtracking
             wp = __dtw_backtracking(D_steps, step_sizes_sigma, subseq)
+            if wp[-1] != (0, 0):
+                # TODO: what should we do here?
+                pass
 
         wp = np.asarray(wp, dtype=int)
 
@@ -327,7 +330,7 @@ def __dtw_calc_accu_cost(C, D, D_steps, step_sizes_sigma,
     return D, D_steps
 
 
-#@jit(nopython=True, cache=True)
+@jit(nopython=True, cache=True)
 def __dtw_backtracking(D_steps, step_sizes_sigma, subseq):  # pragma: no cover
     '''Backtrack optimal warping path.
 
@@ -374,6 +377,10 @@ def __dtw_backtracking(D_steps, step_sizes_sigma, subseq):  # pragma: no cover
         # save tuple with minimal acc. cost in path
         cur_idx = (cur_idx[0] - step_sizes_sigma[cur_step_idx][0],
                    cur_idx[1] - step_sizes_sigma[cur_step_idx][1])
+
+        # If we run off the side of the cost matrix, break here
+        if min(cur_idx) < 0:
+            break
 
         # append to warping path
         wp.append((cur_idx[0], cur_idx[1]))
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -116,9 +116,12 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
     Raises
     ------
     ParameterError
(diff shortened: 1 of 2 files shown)


librosa/sequence.py
```diff
@@ -208,7 +220,11 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
 
     # initialize step matrix with -1
     # will be filled in calc_accu_cost() with indices from step_sizes_sigma
-    D_steps = -1 * np.ones(D.shape, dtype=np.int)
+    D_steps = np.zeros(D.shape, dtype=np.int)
+    # TODO: replace these with steps that correspond to left (first row)
+    # and up (first column)
+    D_steps[0, :] = 1
+    D_steps[:, 0] = 2
```
bmcfee: @fzalkow @stefan-balke Any ideas for how to address this one?

The old version (-1) initializes all steps to whatever the last entry in the `step_sizes` array is, with the assumption that every element of `D_steps` will be overwritten before any backtracking is attempted.  However, this does not always happen, eg when there are `Inf` values in the cost matrix.

The current change (not intended to be a permanent fix) sets the initial step values in the first row and column to be horizontal and vertical moves respectively, but this is brittle and will fail if alternate step sizes are provided.  I'd like to have a robust solution that works in general, but don't see how to achieve that without adding restrictions on the step sizes.  Any thoughts?

librosa/sequence.py
```diff
@@ -228,6 +244,9 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
         else:
             # perform warping path backtracking
             wp = __dtw_backtracking(D_steps, step_sizes_sigma, subseq)
+            if wp[-1] != (0, 0):
+                # TODO: what should we do here?
+                pass
```
bmcfee: @fzalkow I added an escape condition to the backtracking loop to avoid array bound exceptions, but we still need to decide what the behavior should be if a full alignment is requested but cannot be constructed.  I'm not totally clear on this TBH.

fzalkow: What do you think about giving a warning:
```python
warnings.warn('No valid warping path was computed! The path starts at {} instead of [0 0].'.format(wp[-1]), RuntimeWarning)
```

librosa/sequence.py
fzalkow: Instead of 1 and 2, we can use `most_horizontal_step_idx` and `most_vertical_step_idx` (see my PR #1033). Still, it may cover not all cases with arbitrary `infs` in the cost matrix.

bmcfee: I agree in general, but I don't think steepest/shallowest slope is a strong enough condition.  What we want to do here is prevent the back-pointers from going off the edge of the array, which means the first/row column has to be purely horizontal or vertical moves.

Alternately, we could rewrite the data structure such that `D_steps` contains the absolute position instead of the index into the step_sizes array.  This would allow us to populate the steps matrix with technically illegal moves that form a coherent path, and then overwrite them when better solutions are available.  Doing this would require a bit of numpy hackery though, and possibly a custom structure dtype for tuples, but it might be the cleanest solution in the long run.

librosa/sequence.py
bmcfee: Should it be a warning or an exception?

librosa/sequence.py
fzalkow: I changed my PR #1033 so that now we check the accumulated cost matrix before doing backtracking. If we additionally forbid nans in the cost matrix (infs are fine!), I can see no option for the backtracking to access anything outside the array edges.

librosa/sequence.py
fzalkow: With the inf-check for the accumulated cost matrix, I also don't see when this case occurs... So an exception is maybe more appropriate.

librosa/sequence.py
bmcfee: @fzalkow I gave this some thought, and I have an idea that could work.

Instead of replacing the default step sizes (and costs/weights) with user-supplied values, we can keep the defaults ([1,1], [0,1], [1,0]) and append user-supplied steps to that list.  If that happens, we set the weights for the defaults to infinity so that they are never preferred over the user-supplied steps.  It's a little hacky, but I think it nicely solves our problems with initializing the step matrix.

We'd still have the potential problem of running off the edge of the array while backtracking, so that special-case handler still needs to be there.

What do you think?

bmcfee: Implemented this and added a safety check for negative step parameters.  Tests pass, and i'm reasonably happy with the solution.  Please give it a look over!

fzalkow: I don't know if it is a good idea that non-supplied step sizes can be used. For example, in the following case, I would expect DTW not to turn any result (and in particular, not to use any non-supplied step sizes), because there is no valid solution given the supplied step sizes.

```python
step_sizes_sigma = np.array([[1, 1], [1, 2], [2, 1]])
C = np.ones((4, 2))
D, wp = librosa.sequence.dtw(C=C, step_sizes_sigma=step_sizes_sigma)
```

In the this current branch, full DTW returns no valid warping path for this example (not using cell `(0, 0)`), because of the comparison between costs of the added infinity rows/columns and the actual infinity values of the `D` matrix.

I would rather be in favor of raising an exception as in PR #1033.

Just my opinion, of course!

bmcfee: > I don't know if it is a good idea that non-supplied step sizes can be used.

The whole point is that they won't be used -- any path with a finite cost will take precedent.  All this proposed change does is ensure that the step matrix is always well-formed.


>  would expect DTW not to turn any result (and in particular, not to use any non-supplied step sizes), because there is no valid solution given the supplied step sizes.

Agreed.  The current branch does not yet implement the error-handling logic, so the call succeeds.  I think I see what you're saying with the example though; there's no explicit logic to distinguish valid (finite-cost) paths from invalid (infinite-cost) paths.  I think we could easily add that to the loop condition in the backtracker and that should resolve everything.

fzalkow: > The whole point is that they won't be used -- any path with a finite cost will take precedent. All this proposed change does is ensure that the step matrix is always well-formed.

I am not sure if we have full control over that. When initializing the `D` matrix, additional infinity columns and rows are created outside of the valid range of `D` ([code link](https://github.com/librosa/librosa/blob/90ef6b2736c127bfb7e314b8b839fd5b494f52ce/librosa/sequence.py#L243)). Now, if steps with infinite cost are introduced, then a comparison of infinity to infinity may be performed (going outside the valid range or using a non-supplied step).

Sorry for bringing it up again. :-) PR #1033 proposes not to do any backtacking when the upper right corner of the `D` matrix has infinite cost. Shouldn't that solve the overall issue?

bmcfee: >  then a comparison of infinity to infinity may be performed (going outside the valid range or using a non-supplied step).

The step value is only updated if the proposed cost is less than the initial cost; `inf < inf` is False, so the initial value will remain in place, guaranteeing pointer safety.



> PR #1033 proposes not to do any backtracking when the upper right corner of the `D` matrix has infinite cost. Shouldn't that solve the overall issue?

I think it does in common cases, but I'm more worried at this point about correctly handling non-standard step sizes.  For example, what if a user wants to forbid diagonal moves and use `step_sizes_sigma = [[0, 1], [1, 0]]`?  In this case, initializing steps with `1, 2` would lead to segfaults again.  More generally, i'd really like to avoid anything that involves scanning through the step sizes matrix to find "good" initialization values.

fzalkow: > > PR #1033 proposes not to do any backtracking when the upper right corner of the `D` matrix has infinite cost. Shouldn't that solve the overall issue?
> 
> I think it does in common cases, but I'm more worried at this point about correctly handling non-standard step sizes. For example, what if a user wants to forbid diagonal moves and use `step_sizes_sigma = [[0, 1], [1, 0]]`? In this case, initializing steps with `1, 2` would lead to segfaults again.

As far as I see, it would not lead to segfaults. If there is a valid solution for these step sizes (meaning `D[-1, -1]` is not infinity), then all used entries in `D_steps` are overwritten before backtracking. If there is no valid solution, then the backtracking won't be performed. Maybe I'm overlooking something here?

bmcfee: Hrm, I suppose you're right.  It still makes me nervous having bad values in the step matrix though, even if there's logic elsewhere that (indirectly) prevents them from being used as pointer offsets.

I'm still thinking over how this relates to #1033 though -- I still don't think they're addressing the same issue exactly, but I'm having a difficult time articulating what exactly the difference is.

bmcfee: Coming back to this after some time away, I think this discussion got confusing because we're trying to solve too many problems simultaneously.  Specifically, there are (at least) two issues to resolve:

1. initialization and invalid backpointers
2. handling nans/infs in the cost matrix

The two are related, as bad handling of nans/infs can lead the backtracker into states that expose invalid backpointers, which causes segfaults and initiated this thread in the first place.

The initialization behavior in the current master branch (fill with -1) is most certainly wrong, and is a prime example of using a sentinel value that's easily mistaken for real data: negative values are valid indices in numpy arrays, but we were using it as if it was not.

The fix as implemented so far in this PR ensures that backpointers are always valid (with respect to the steps matrix), even when a user supplies their own steps matrix.  In the latter case, the weights on the default steps are modified such that costs become infinite, implying that these steps are never selected in the optimization.  Of course, these infinite-cost steps are still "invalid" from the perspective of DTW since they use steps other than those defined by the user; all we're doing here is buying some safety from the backtracking routine to avoid segmentation faults in case it lands up in one of these situations, eg with a cost matrix (`C`) like:

```
0   inf 0    0
0   0    inf 0
0   0    inf 0
```
Here, no finite, full-sequence warping path exists, but it's not trivial to detect this situation until you get to the backtracking phase.

The accumulated cost should propagate these infs forward so they become easily detectable by checking the end-point of the warping path, as @fzalkow has implemented in #1033.  I think both changes are necessary to fully resolve the issue in this PR.

librosa/sequence.py
bmcfee: I think you're correct.  I haven't been able to construct an example that would trigger this behavior that also somehow ends up with a finite warping path cost.

I'll leave this in as an exception, just in case there is some configuration that could trigger this, but I think it's otherwise safe.

bmcfee:
I'll give this one more spin through CI, but I think it's good to merge.  It includes @fzalkow's changes from #1033 , so we can close that one out once this merges.

bmcfee:
If there are no objections, I'll merge this one tomorrow (monday, 01/13) so we can prep the 0.7.2 release.

PR #1033, opened 2019-12-06
DTW: Assert certain cost matrix shape for given step sizes

fzalkow:
#### Reference Issue
Related: #1014 and #1031

For certain non-default step sizes, a valid warping path has a restricted slope. Specifically, this is the case if the step sizes don't contain `[1, 0]` and `[0, 1]`. The steepest vertical step and the flattest horizontal step restrict the possible ratios of the lengths of `X` and `Y`. For example, if we use the step sizes `[[1, 1], [1, 2], [2, 1]]`, then the length of `X` has to be shorter than twice the length of `Y`, and vice versa. This PR asserts for the length of the sequences.

The figure below illustrates the new behavior (created with [this gist](https://gist.github.com/fzalkow/d86a7a6e0fbd18a4218dcece467700bf)) for the step sizes `[[1, 1], [1, 2], [2, 1]]`. You see an alignment for various random sequences. The columns show an increase in the length of `Y` and the rows show an increase in the length of `X`. If a cell in the figure grid is left empty, then an assertion was raised.

The outermost grid cells (far off the diagonal) correspond to a maximal difference between the lengths of `X` and `Y`. For those grid cells the warping path is of maximal slope (either only using `[1, 2]` or only using `[2, 1]`). Previous to this PR, it would have returned a segfault in the cases where now an assertion is raised.

test

### DIFF
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -190,6 +190,23 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
         raise ParameterError('For diagonal matching: Y.shape[1] >= X.shape[1] '
                              '(C.shape[1] >= C.shape[0])')
 
+    # given the step sizes, check of the shape of C is valid
+    with np.errstate(divide='ignore'):
+        step_sizes_ratio = step_sizes_sigma[:, 0] / step_sizes_sigma[:, 1]
+
+    most_horizontal_step_idx = np.argmin(step_sizes_ratio)
+    most_vertical_step_idx = np.argmax(step_sizes_ratio)
+    most_horizontal_step = step_sizes_sigma[most_horizontal_step_idx]
+    most_vertical_step = step_sizes_sigma[most_vertical_step_idx]
+
+    if subseq is False and most_horizontal_step[0] != 0:
+        assert (C.shape[0] / most_horizontal_step[0]) > (C.shape[1] / most_horizontal_step[1]),\
+            'Sequence lengths does not match step sizes'
+
+    if most_vertical_step[1] != 0:
+        assert (C.shape[0] / most_vertical_step[0]) < (C.shape[1] / most_vertical_step[1]),\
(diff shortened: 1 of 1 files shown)


fzalkow:
Oops, I see it is more complicated than I thought before. I will think again...

fzalkow:
A found a more easy solution: Instead of computing slopes, we just check if the necessary cells in the accumulated cost matrix are filled (= non-infs). That means checking the upper right corner in case of full dtw, and all cells in the upper row in case of subseq dtw.

librosa/sequence.py
```diff
@@ -216,6 +216,11 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
                                       weights_mul, weights_add,
                                       max_0, max_1)
 
+    if subseq is False and np.isinf(D[-1, -1]):
```
lostanlen: You can make this a one-liner with some Boolean fiddling:

`if np.isinf(D[-1, -1]) and (not subseq or np.all(np.isinf(D[-1, :])))`

fzalkow: I found the two if-clauses more easy to understand. Have not a strong opinion in the case, though... Are there any librosa-guidelines for short code vs easy to understand code?

bmcfee: I'm not a fan of strict guidelines here.  In this specific case, the if conditions are more readable as @fzalkow originally wrote them.  However, both conditions lead to the same (duplicated) code block (the runtime error), which isn't great style.  Because of that, I'd prefer @lostanlen's suggestion since it reduces branching paths and redundant code.

lostanlen: tagging @bmcfee

fzalkow: Ok, thanks for your input. Merged the lines and added a comment for understandability.

bmcfee:
These changes have been integrated in #1031  and merged, so I'll close this out.

PR #1041, opened 2019-12-13
Fix librosa.feature.rms #1040

### LINKED ISSUE #1040: librosa.feature.rms inconsistency - y vs S
#### Description

When calling `librosa.feature.rms`, the result using a wave `y` and the result using the spectrogram `S` are different.  
This is because the algorithm is not correctly implemented to satisfy the Parseval's equation.  

texclip20191213224740

Specifically, the direct current bin `S[0]` (and only if `n_fft` is odd, the sr/2 bin `S[-1]`) has twice as much influence as the other bins.

#### Steps/Code to Reproduce

```python
import numpy as np
import librosa

# set paramters
np.random.seed(0)
frame_length = 2048
hop_length = 512
center = True

# make a wave and its spectrogram
y = np.random.rand(3000)  # wave, DC component is about 0.5 because the range of np.random.rand is [0, 1)
S = librosa.magphase(librosa.stft(y,  # spectrogram
        n_fft=frame_length,
        hop_length=hop_length,
        window=np.ones,
        center=center))[0]

# calculate RMS
rms1 = librosa.feature.rms(y=y, frame_length=frame_length, hop_length=hop_length, center=center)
rms2 = librosa.feature.rms(S=S, frame_length=frame_length, hop_length=hop_length, center=center)
print(rms1)
print(rms2)
print(np.allclose(rms1, rms2))
```

#### Expected Results

`rms1` and `rms2` are almost the same.

#### Actual Results

```
[[0.57401353 0.58105711 0.58252662 0.58571533 0.58668334 0.58252178]]
[[0.75760764 0.76713306 0.76892614 0.7742523  0.77579546 0.77086043]]
False
```

#### Versions

Darwin-18.7.0-x86_64-i386-64bit
Python 3.7.3 (default, May 28 2019, 02:11:32) 
[Clang 10.0.1 (clang-1001.0.46.3)]
NumPy 1.17.4
SciPy 1.3.3
librosa 0.7.1

onkyo14taro:
#### Reference Issue

Fixes #1040

I changed the calculation algorithm of `librosa.feature.rms` when giving `S` (not `y`).
In order to satisfy Parseval's equation, the negative frequency bins' power is correctly handled.
Since the calculation formula is slightly different when `n_fft` is odd and even, the new argument `odd_n_fft` is added.
This argument specifies whether or not the `n_fft` used when computing `S` is odd.
(`odd_n_fft` does nothing when `y` is given.)

Here is a sample snippet.
```python
import librosa

frame_length_even = 2048
frame_length_odd = 2049
hop_length = 512
center = True

# make a wave and its spectrogram 
y = np.random.rand(3000)  # wave 
S_even = librosa.magphase(librosa.stft(y,  # spectrogram 
        n_fft=frame_length_even, 
        hop_length=hop_length, 
        window=np.ones))[0]
S_odd = librosa.magphase(librosa.stft(y,  # spectrogram 
        n_fft=frame_length_odd, 
        hop_length=hop_length, 
        window=np.ones))[0] 

# calculate RMS 
rms_even_y = librosa.feature.rms(y=y, frame_length=frame_length_even, hop_length=hop_length, center=center)
rms_even_S = librosa.feature.rms(S=S_even, center=center)
rms_odd_y = librosa.feature.rms(y=y, frame_length=frame_length_odd, hop_length=hop_length, center=center)
rms_odd_S = librosa.feature.rms(S=S_odd, center=center, odd_n_fft=True)
rms_odd_S_mistake = librosa.feature.rms(S=S_odd, center=center)  # mistake
print(rms_even_y, ': rms_even_y')  
print(rms_even_S, ': rms_even_S')  
print(rms_odd_y, ': rms_odd_y')  
print(rms_odd_S, ': rms_odd_S') 
print(rms_odd_S_mistake, ': rms_odd_S_mistake')
```

Output:
```
[[0.56467811 0.56795336 0.57112578 0.57611183 0.57552133 0.57208155]] : rms_even_y
[[0.5646781  0.56795336 0.57112576 0.57611182 0.57552133 0.57208155]] : rms_even_S
[[0.56471934 0.56805793 0.57113983 0.57632674 0.57539677 0.572001  ]] : rms_odd_y
[[0.56471935 0.56805794 0.57113982 0.57632676 0.57539679 0.57200101]] : rms_odd_S
[[0.56488239 0.56820956 0.57133744 0.57660026 0.57567047 0.5722761 ]] : rms_odd_S_mistake
```

#### Any other comments?
I changed the test code `test_rms()` in `test_features.py`.

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -766,7 +766,7 @@ def spectral_flatness(y=None, S=None, n_fft=2048, hop_length=512,
 
 
 def rms(y=None, S=None, frame_length=2048, hop_length=512,
-        center=True, pad_mode='reflect', odd_n_fft=False):
+        center=True, pad_mode='reflect'):
     '''Compute root-mean-square (RMS) value for each frame, either from the
     audio samples `y` or from a spectrogram `S`.
 
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -865,12 +865,13 @@ def rms(y=None, S=None, frame_length=2048, hop_length=512,
         # power spectrogram
         x = np.abs(S) ** 2
 
-        # Recover negative frequency bins' power
-        end = None if frame_length % 2 else -1
-        x[1:end] *= 2
+        # Adjust the DC and sr/2 component
+        x[0] *= 0.5
+        if frame_length % 2 == 0:
+            x[-1] *= 0.5
 
         # Calculate power
-        power = np.sum(x, axis=0, keepdims=True) / frame_length**2
+        power = 2 * np.sum(x, axis=0, keepdims=True) / frame_length**2
     else:
         raise ValueError('Either `y` or `S` must be input.')
 
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -800,12 +800,6 @@ def rms(y=None, S=None, frame_length=2048, hop_length=512,
         Padding mode for centered analysis.  See `np.pad` for valid
         values.
 
-    odd_n_fft : bool
-        This parameter is used on frequency-domain input (`S`).
-        If an odd number is given as `n_fft` for calculating `S`,
-        set `odd_n_fft = True`.
-        By defaults, `odd_n_fft = False`.
-
     Returns
     -------
     rms : np.ndarray [shape=(1, t)]
@@ -858,19 +852,25 @@ def rms(y=None, S=None, frame_length=2048, hop_length=512,
         # Calculate power
         power = np.mean(np.abs(x)**2, axis=0, keepdims=True)
     elif S is not None:
-        x, _ = _spectrogram(y=y, S=S,
-                            n_fft=frame_length,
-                            hop_length=hop_length)
+        # Check the frame length
+        if S.shape[0] != frame_length // 2 + 1:
+            raise ParameterError(
+                    'Since S.shape[0] is {}, '
+                    'frame_length is expected to be {} or {}; '
+                    'found {}'.format(
+                            S.shape[0],
+                            S.shape[0] * 2 - 2, S.shape[0] * 2 - 1,
+                            frame_length))
+
         # power spectrogram
-        x = np.abs(x) ** 2
+        x = np.abs(S) ** 2
 
         # Recover negative frequency bins' power
-        end = None if odd_n_fft else -1
+        end = None if frame_length % 2 else -1
         x[1:end] *= 2
 
         # Calculate power
-        n_fft = x.shape[0]*2 - 1 if odd_n_fft else x.shape[0]*2 - 2
-        power = np.sum(x, axis=0, keepdims=True, dtype=np.float64) / n_fft**2
+        power = np.sum(x, axis=0, keepdims=True) / frame_length**2
     else:
         raise ValueError('Either `y` or `S` must be input.')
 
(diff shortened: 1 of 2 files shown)


librosa/feature/spectral.py
```diff
@@ -766,7 +766,7 @@ def spectral_flatness(y=None, S=None, n_fft=2048, hop_length=512,
 
 
 def rms(y=None, S=None, frame_length=2048, hop_length=512,
-        center=True, pad_mode='reflect'):
+        center=True, pad_mode='reflect', odd_n_fft=False):
```
bmcfee: I'm not sure we need a separate parameter here -- there's already a parameter for frame length which can be used to specify it exactly.  Probably, if a user has a non-standard frame length, they have it in a variable somewhere, and it's easier to pass that along and then infer whether it's even or odd, rather than force the user to make that calculation externally.

librosa/feature/spectral.py
```diff
@@ -849,19 +855,26 @@ def rms(y=None, S=None, frame_length=2048, hop_length=512,
...
+
+        # Recover negative frequency bins' power
+        end = None if odd_n_fft else -1
+        x[1:end] *= 2
 
-        # FFT introduces a scaling of n_fft to energy calculations
-        norm = 2 * (x.shape[0] - 1)
+        # Calculate power
+        n_fft = x.shape[0]*2 - 1 if odd_n_fft else x.shape[0]*2 - 2
+        power = np.sum(x, axis=0, keepdims=True, dtype=np.float64) / n_fft**2
```
bmcfee: power should not explicitly set the dtype here; it should inherit from the dtype of the input.

librosa/feature/spectral.py
onkyo14taro: Thank you for your advice, your indication is the optimal solution. I removed `odd_n_fft` and fixed the code.

librosa/feature/spectral.py
onkyo14taro: Thank you for pointing it out. I fixed it not to specify dtype.

librosa/feature/spectral.py
bmcfee: Thanks!  The parameter is still in the call signature though.

librosa/feature/spectral.py
```diff
@@ -849,19 +849,32 @@ def rms(y=None, S=None, frame_length=2048, hop_length=512,
...
+                            S.shape[0],
+                            S.shape[0] * 2 - 2, S.shape[0] * 2 - 1,
+                            frame_length))
+
+        # power spectrogram
+        x = np.abs(S) ** 2
+
+        # Recover negative frequency bins' power
+        end = None if frame_length % 2 else -1
+        x[1:end] *= 2
```
bmcfee: I think this could be done with fewer multiply adds if we scale `x[0] *= 0.5` (and `x[-1]` for even length) and then multiply `power = 2 * np.sum(...)`.

librosa/feature/spectral.py
onkyo14taro: Sorry, I missed it. I removed unnecessary argument `odd_n_fft`.

librosa/feature/spectral.py
onkyo14taro: I never came up with such a smart calculation. Thank you, I have reflected the proposed algorithm.

bmcfee:
LGTM!  Thanks for all your work on this @onkyo14taro !

:shipit:

onkyo14taro:
@bmcfee 
Thank you for much advice! Your advice deepened my thoughts on API and algorithm design.

PR #1050, opened 2020-01-06
Mel filterbank normalization

### LINKED ISSUE #1030: Normalise Mel filters to constant energy of 1
There is some scaling introduced by the Mel filter bank when calculating the Mel spectrogram. As stated in the documentation setting `norm = 1` does give constant energy per filter, but their energy is not 1. In fact, if you sum the filters you don't get a value of 1, but rather an average of 0.0929 with the following distribution (for `n_mel = 128`):

This is due to the Mel frequency bandwidth of the triangles used for normalisation, but I'm thinking it might make sense to have the sum of the filter weights equal to 1 when normalised, maybe as an additional option.

bmcfee:
#### Reference Issue

Fixes #1030 

This PR initiates a deprecation (redefinition) cycle for the `norm` parameter in `filters.mel`.

A new setting `norm='slaney'` is now the default normalization mode.  `norm=1` is currently equivalent, but will trigger a warning that this situation will change in version 0.8.0.

Similarly, `norm=np.inf` is currently a no-op, and equivalent to `norm=None`.  Using `np.inf` will also trigger a future warning that behavior will change in 0.8.0.

In 0.8 and onwards, `norm=p` (for numeric `p`) will use `util.normalize` to perform numerically stable, discrete frequency normalization, rather than the continuous frequency normalization currently used.

`norm='slaney'` will remain the default (for at least one more cycle), and use continuous frequency normalization.

#### Any other comments?

Should be good to go, assuming CI passes.

### DIFF
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -111,7 +111,7 @@ WINDOW_BANDWIDTHS = {'bart': 1.3334961334912805,
 
 @cache(level=10)
 def mel(sr, n_fft, n_mels=128, fmin=0.0, fmax=None, htk=False,
-        norm=1, dtype=np.float32):
+        norm='slaney', dtype=np.float32):
     """Create a Filterbank matrix to combine FFT bins into Mel-frequency bins
 
     Parameters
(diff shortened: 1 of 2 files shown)


bmcfee:
I think this is ready to go, anyone care to look it over?

PR #1054, opened 2020-01-09
length of generated tones

### LINKED ISSUE #1053: librosa.core.tone returns wrong length
Here is my console output:

> >>> import librosa
> >>> wav = librosa.core.tone(440,44100,length=1740)
> >>> len(wav)
> 1741

I expect the output to be 1740.

bmcfee:
#### Reference Issue
Fixes #1053 

This PR fixes an edge case in tone generation, where the length of output signals did not match up exactly.

A unit test has been added to cover the case.


bmcfee:
I don't think this one needs a CR; will merge once CI passes.

PR #1055, opened 2020-01-09
removed dct_type=None from mfcc docs

### LINKED ISSUE #1051: MFCC function throws error when dct_type=None
#### Description

From my vague understanding of MFCC, the DCT is needed to compute the final values. However, according to the docs at [https://librosa.github.io/librosa/generated/librosa.feature.mfcc.html](https://librosa.github.io/librosa/generated/librosa.feature.mfcc.html), you should be able to compute the MFCC (or get some kind of return) with dct_type set to none.

> dct_type:None, or {1, 2, 3}

At this time, setting dct_type to none throws an error. Is there a way to get MFCCs (or something like it) without the DCT step? If not, maybe remove None as an option in the docs?

#### Steps/Code to Reproduce

Example:

```
import librosa

# Make some random data
S = (np.random.random(1024) * 2) - 1

mfccs = librosa.feature.mfcc(y=S, sr=16000, dct_type=None)
print(mfccs)
```

#### Expected Results

Array of MFCC (or MFCC-like) values.

#### Actual Results

TypeError: %d format: a number is required, not NoneType

#### Versions

Windows-10-10.0.18362-SP0
Python 3.7.5 (default, Oct 31 2019, 15:18:51) [MSC v.1916 64 bit (AMD64)]
NumPy 1.17.4
SciPy 1.3.1
librosa 0.7.1

bmcfee:
#### Reference Issue
Fixes #1051 

Documentation erroneously listed `None` as a valid `dct_type` in MFCC calculation.  This PR removes that option, which was never supported.

### DIFF
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -184,7 +184,7 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm='ortho', ref=1.0, lifter=0):
     n_mels : int > 0
         The number of Mel frequencies
 
-    dct_type : None or {1, 2, 3}
+    dct_type : {1, 2, 3}
         Discrete cosine transform (DCT) type
         By default, DCT type-2 is used.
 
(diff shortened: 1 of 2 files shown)


bmcfee:
No CR needed here, will merge once CI passes.

PR #1057, opened 2020-01-09
Fix power_to_db: #1056

### LINKED ISSUE #1056: [Bug] `power_to_db` when using complex spectrogram
#### Description

When using complex spectrogram as the argument, the output of `librosa.power_to_db` becomes half the expected value.

#### Steps/Code to Reproduce

```python
import librosa
import numpy as np

# complex spectrogram
S_complex = np.array([10], dtype=np.complex64)

# amplitude spectrogram
S_amplitude = np.abs(S_complex)

# power spectrogram
S_power = np.abs(S_complex) ** 2

print('power_to_db (power) :\t\t', librosa.power_to_db(S_power))
print('power_to_db (complex):\t\t', librosa.power_to_db(S_complex))
print('amplitude_to_db (amplitude) :\t', librosa.amplitude_to_db(S_amplitude))
print('amplitude_to_db (complex):\t', librosa.amplitude_to_db(S_complex))
```

#### Expected Results

The all are 20.
10 log_10(10**2) = 20 log_10(10) = 20

#### Actual Results

```
power_to_db (power) :		 [20.]
power_to_db (complex):		 [10.]
amplitude_to_db (amplitude) :	 [20.]
amplitude_to_db (complex):	 [20.]
```

#### Versions

Darwin-18.7.0-x86_64-i386-64bit
Python 3.7.3 (default, May 28 2019, 02:11:32)
[Clang 10.0.1 (clang-1001.0.46.3)]
NumPy 1.17.4
SciPy 1.3.3
librosa 0.7.1

onkyo14taro:
#### Reference Issue

Fixes #1056 

When using complex spectrogram as the argument of `librosa.power_to_db`, complex spectrogram is correctly converted into power spectrogram.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1602,7 +1602,7 @@ def power_to_db(S, ref=1.0, amin=1e-10, top_db=80.0):
         warnings.warn('power_to_db was called on complex input so phase '
                       'information will be discarded. To suppress this warning, '
                       'call power_to_db(np.abs(D)**2) instead.')
-        magnitude = np.abs(S)
+        magnitude = np.abs(S) ** 2
     else:
         magnitude = S
 


bmcfee:
The semantics of `power_to_db` assume that the input is already squared in magnitude, so it should not be squared again.

I think the calculation you want here is `amplitude_to_db`, which already does the squaring as you've implemented here.

onkyo14taro:
Sorry, I had misunderstood the situation when `power_to_db` receives spectrogram whose dtype is complex (Maybe, the spectrogram whose dtype is complex is assumed to be `S * np.conj(S)`). Thank you for your reply.

PR #1059, opened 2020-01-14
added SRC samplerate dispatch to resample

### LINKED ISSUE #1047: Re-introduce libsamplerate support
#### Description

Now that the resample interface is fully flexible, we could easily introduce [`python-samplerate`](https://python-samplerate.readthedocs.io/en/latest/?badge=latest#) as an optional dependency, and add its implemented modes to our `res_type` dispatch:

- linear 
- src_best (sinc_best)
- src_fastest (sinc_fastest)
- src_medium (sinc_medium)
- zero_order_hold

Down the line, since libsamplerate is now BSD-licensed (and the python bindings are MIT-licensed), we could also put it back in the main dependency line and possibly deprecate resampy.  This is more a question of packaging and maintenance than of functionality, but it's something to consider.

bmcfee:
#### Reference Issue
Fixes #1047 

This PR re-introduces support for libsamplerate as a sample-rate conversion backend.  New `res_type` values are supported to map onto libsamplerate's modes.

Functionally, I don't think there's anything to do here.  We should perhaps have a conversation about packaging, dependencies, and possible deprecation of resampy though.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -455,6 +455,11 @@ def resample(y, orig_sr, target_sr, res_type='kaiser_best', fix=True, scale=Fals
 
             To use `scipy.signal.resample_poly`, set `res_type='polyphase'`.
 
+            To use `samplerate.resample`, use the following:
+                - `linear`: linear interpolation
+                - `zero_order_hold`: zero-order hold sampling
+                - `sinc_best`, `sinc_medium`, `sinc_fastest`: for high, medium, and low-quality sinc interpolation
+
         .. note::
             When using `res_type='polyphase'`, only integer sampling rates are
             supported.
@@ -487,6 +492,7 @@ def resample(y, orig_sr, target_sr, res_type='kaiser_best', fix=True, scale=Fals
     librosa.util.fix_length
     scipy.signal.resample
     resampy.resample
+    samplerate.resample
 
     Notes
     -----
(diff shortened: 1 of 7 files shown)


lostanlen:
looks like the tests are failing because libsamplerate is not properly installed on the VM?
`OSError: libsamplerate not found`

bmcfee:
Ah, because I forgot to wipe the build caches.

bmcfee:
Though it looks like appveyor is failing on matplotlib differences.  Oy.

bmcfee:
Strangely, now the linux builds are failing (but everything else is fine).  I'm very confused by this.

bmcfee:
I think this one's ready for CR.  Code is fine IMO, but I'd like a quick check on the docstring to make sure it's clear.  Any takers?

librosa/core/audio.py
```diff
@@ -455,6 +455,11 @@ def resample(y, orig_sr, target_sr, res_type='kaiser_best', fix=True, scale=Fals
 
             To use `scipy.signal.resample_poly`, set `res_type='polyphase'`.
 
+            To use `samplerate.resample`, set any of the following:
```
justinsalamon, line 458: it’d be helpful to add a comment about runtime performance - most ppl use the default resample to 22050 on load behavior, without knowing it’s potentially the slowest step in their chain

justinsalamon: Along the same vein, it might be worth considering using a faster resampling engine as the default

bmcfee: Thanks for looking it over!

I'll add some comments on speed, but I worry that we now have so many options that it could easily get out of hand.

Re changing the defaults: I think it's best to keep it as is. Going to 22k by default already angers the audiophiles enough, so I'd prefer to keep high quality by default here.

That said, I do think it warrants a blog post to discuss making things go fast.

justinsalamon: cool, yes a blogpost would be handy. For runtime, perhaps a table of algorithm|quality|runtime would be helpful/concise? Could be included in the docs directly or linked to from the docs.

Regarding resampling, is the current default engine categorically the best in terms of quality? E.g. sox resampling seems to be both high quality and fast? I agree default resampling should be high quality, was wondering if there's an alternative implementation that's HQ but faster.

bmcfee: It's the best in terms of what we can reliably ship without making the package dependency chain a total nightmare, yes. :upside_down_face: 

samplerate's `sinc_best` is better than resampy's due to having brute-forced a more efficient parametrization of their kaiser filter, but it's effectively the same method.  It's not the default because doing so would add external C library dependencies that make shipping very difficult.  The same applies to soxr, which is also great, but difficult to depend on.

PR #861, opened 2019-03-31
Test LPC with both float32 and float64

ajweiss:
#### Reference Issue

Adds tests for #856

#### What does this implement/fix? Explain your changes.

Adds explicit tests for single and double precision numbers for `librosa.lpc`

#### Any other comments?

I wasn't able to reproduce the issue by testing locally against Python 2.7 and 3.7.2.  (with numba 0.40.1 and 0.43.1 respectively).  Opening this PR because the tests are useful regardless and maybe the issue will surface in the CI.

Issue #725, opened 2018-06-13
Support for LPC


mmcauliffe:
In general, there's not a great way to extract LPC coefficients in Python.  The main way that I've seen referenced is talkbox (https://github.com/cournape/talkbox), but it's quite out of date at this point and doesn't support Python 3.  I've done some work for updating the talkbox code for python 3 (https://github.com/mmcauliffe/Conch-sounds/blob/master/conch/analysis/formants/lpc.py), but it's implemented in Python rather than anything faster (there's also this python implementation in a PR here: https://github.com/jsawruk/pymir/pull/8/files/1ac978c5120ed93689095535b432630792be6843?diff=split&short_path=04c6e90).

Is there any chance of implementing LPC in librosa?  The recommended route for migrating from talkbox for mfcc generation is to go to librosa, so it'd be nice if LPC generation could be done here as well.

r9y9:
What about the [SPTK](http://sp-tk.sourceforge.net/)'s implementation? I am maintaining a python wrapper: https://github.com/r9y9/pysptk and it supports both python 2 and 3. The core is implemented in C so it should be faster than naive python. lpc: http://pysptk.readthedocs.io/en/latest/generated/pysptk.sptk.lpc.html#pysptk.sptk.lpc.

bmcfee:
> Is there any chance of implementing LPC in librosa? The recommended route for migrating from talkbox for mfcc generation is to go to librosa, so it'd be nice if LPC generation could be done here as well.

That would be great to have, and I'm happy to help out if someone wants to take the lead on it.

Re: speed, I was going to suggest numba-jitting the code you linked, but it seems like you've already tried that (commented out)?  I haven't implemented lpc before, but I would expect that a numba implementation could be fast enough for regular use.  If we have a reference implementation that we trust, even if it's slow, that would be quite helpful in developing a faster implementation.

bmcfee:
Merged in #797

Issue #319, opened 2016-04-07
Move static test data to a submodule


faroit:
The test vectors significantly increase the repository size (See #240). I personally don't see the benefit of designing tests against static MATLAB generated test vectors. Many scientific python users now are not using MATLAB anymore and therefore identical bit-exact output is not that crucial anymore.
If you really want full MATLAB compatibility, dynamically generated vectors could better cover more test cases.

What about splitting the existing unit tests into functional tests (pure python) and MATLAB compatibility tests which could use oct2py (in case the matlab code runs on octave) or matlab_wrapper to run the tests without the intermediate vectors if available on local machines? For travis obviously only `oct2py` would work.

bmcfee:
> I personally don't see the benefit of designing tests against static MATLAB generated test vectors.

We have unit tests in place where it makes sense, but for certain functions, regression testing on pinned parameters is really the only way to be sure that something hasn't gone unexpectedly wrong.



> What about splitting the existing unit tests into functional tests (pure python) and MATLAB compatibility tests which could use oct2py (in case the matlab code runs on octave) or matlab_wrapper to run the tests without the intermediate vectors if available on local machines? For travis obviously only oct2py would work.

I'm open to the idea, but last time I tried something like this, octave didn't fully cover all of the things that we needed.  Maybe things are different now?

At any rate, the tests are coming due for an overhaul (perhaps a rewrite with py.test), but I'm treating this as a low-priority item.




> The test vectors significantly increase the repository size (See #240).

Incidentally, wouldn't the repo have to stay the same size after removing the files, for git to be able to revert to previous checkpoints?  Or would we have to dig into deep git/back to the future 2 magic to retroactively wipe out the large files?

faroit:
> I'm open to the idea, but last time I tried something like this, octave didn't fully cover all of the things  that we needed. Maybe things are different now?

I can run some tests next week and report about octave compatibility. 

For now I would recommend using Git LFS for the matlab vectors

> Incidentally, wouldn't the repo have to stay the same size after removing the files, for git to be able to revert to previous checkpoints? Or would we have to dig into deep git/back to the future 2 magic to retroactively wipe out the large files?

Yes, the history needs to be rewritten so maybe you want to make use of some of the available tools to do this automatically: looks like [this](https://github.com/rtyley/bfg-repo-cleaner/releases) would work

bmcfee:
Awesome, thanks for the pointers.

[...]

PR #1064, opened 2020-01-28
Test modernization

### LINKED ISSUE #783: Rewrite generator tests
#### Description
We've just about finished porting the test suite from nose to pytest #391 in #775, but almost all of the tests are still written using generators to sweep parameters.  This is a very nosy way to do things, and it's supported in pytest for now, but support is deprecated and will be removed in pytest 4.0.

Eventually, we will have to rewrite the test suite to use proper fixtures rather than generators.  I've started on this for a few test modules, but doing it properly will be a massive undertaking deserving of its own issue.

bmcfee:
#### Reference Issue
Fixes #783  and #1072 

This PR implements the long-overdue upgrade of our test suite to use proper fixtures instead of generators.  In doing so, we should be able to run on modern versions of pytest (>4).

#### Any other comments?

Some of the tests have been rewritten and/or simplified to separate out passing cases from failing cases.

Whenever possible, heavy fixtures (ie audio loaded from disk or spectrograms) will be shared at the module level.  This should significantly cut down our build times once all the smoke clears.

After consulting with the matplotlib devs about the current failures arising from freetype version discrepancies, I've relaxed the RMS threshold on image comparison tests (from 2 to 6).  This should keep things working for the foreseeable future.

It's definitely a work in progress, and the largest modules are yet to be done.  Here's the progress checklist:

- [x] test_beat.py
- [x] test_cache.py
- [x] test_constantq.py
- [x] test_core.py
- [x] test_decompose.py
- [x] test_display.py
- [x] test_dtw.py
- [x] test_effects.py
- [x] test_failures.py
- [x] test_features.py
- [x] test_filters.py
- [x] test_met_features.py
- [x] test_onset.py
- [x] test_segment.py
- [x] test_sequence.py
- [x] test_time_frequency.py
- [x] test_util.py


bmcfee:
Fair warning: code review for this will be terrrrrrrrible.

bmcfee:
Tagging @ajweiss -- In modernizing `test_core`, I noticed that we don't actually have the `core-lpcburg-*` regression fixtures checked into the repo! :fearful:

The way that the pytest generators worked in pytest3, this leads to the tests passing silently because the loop over file globs has zero iterations, and nothing ever fails.  This has me worried, since the only other test that covers lpc is pretty minimal and definitely does not hit all of our code paths.

On the other hand, I'm not keen on the idea of bloating the repo with even more regression fixtures, as we should be trying to eliminate these in favor of run-time comparisons instead.  Any thoughts?

bmcfee:
Note: the stream test is currently failing due to state leakage when when use an open file handle as the path fixture.  What needs to happen here is for the handle to be closed and re-opened at every instantiation of the test, but I don't see an easy way to do that right now.  Without this, the handle is closed at the end of one run, and the next attempt fails because it's trying to read from a closed handle.

ajweiss:
Hey!  I'll take a look this week.  Short answer, the generation code for
the fixtures is in there, but yeah it looks like I didn't check in the
binaries.  I did run it at dev time though for the original commit.  There
was an update that fixed an issue with the types that may not have had it
run though.  What to do long term, I'll think about it and get back to you!

bmcfee:
>  I'll take a look this week.

:+1:

From my end, I've just finished up with the features tests, so this PR is feature-complete.  The two issues to resolve are LPC and streaming, so hopefully we can just knock those out quickly and get this :shipit:

librosa/display.py
```diff
@@ -950,6 +950,9 @@ def __decorate_axis(axis, ax_type):
         axis.set_label_text('')
         axis.set_ticks([])
 
+    else:
+        raise ParameterError('Unsupported axis type: {}'.format(ax_type))
+
```
bmcfee: This was a missing error condition that upgrading the tests revealed (as an `xpass`). :tada:

librosa/sequence.py
```diff
@@ -1126,7 +1126,7 @@ def viterbi_discriminative(prob, transition, p_state=None, p_init=None, return_l
     if p_init is None:
         p_init = np.empty(n_states)
         p_init.fill(1./n_states)
-    elif np.any(p_init < 0) or not np.allclose(p_init.sum(), 1):
+    elif np.any(p_init < 0) or not np.allclose(p_init.sum(), 1) or p_init.shape != (n_states,):
         raise ParameterError('Invalid initial state distribution: '
                              'p_init={}'.format(p_init))
```
bmcfee: Another couple of error conditions that weren't checked before.

setup.py
```diff
@@ -57,7 +57,8 @@
                   'mock',
                   'pytest-mpl',
                   'pytest-cov',
-                  'pytest < 4'],
+                  'pytest',
+                  'contextlib2'],
```
bmcfee: contextlib2 is used to create dummy contexts that are used as dynamic failure catches in the new tests.  This is because we can no longer mark certain parametrizations as failure modes.

tests/test_display.py
```diff
@@ -534,7 +521,7 @@ def test_sharex_specshow_ms(S_abs, y, sr):
     return plt.gcf()
 
 
-@pytest.mark.mpl_image_compare(baseline_images=['sharex_waveplot_ms'], extensions=['png'])
+@pytest.mark.mpl_image_compare(baseline_images=['sharex_waveplot_ms'], extensions=['png'], tolerance=6)
```
bmcfee: increased RMS tolerance here is to handle deviations induced by different versions of the freetype library.  It's still small enough tolerance that actual failures should trigger, but now we can at least have minor changes in axis label kerning without triggering a failure.

bmcfee:
Streaming issue is now resolved, so I think it's just down to LPC.

Does anyone feel charitable enough to try reviewing this?  I did remove/reduce some test cases that seemed excessive, but otherwise it's almost entirely equivalent to the old test cases.  As long as our coverage is in good shape, I feel okay with not stressing CR too much here.

librosa/feature/spectral.py
```diff
@@ -873,7 +871,7 @@ def rms(y=None, S=None, frame_length=2048, hop_length=512,
         # Calculate power
         power = 2 * np.sum(x, axis=0, keepdims=True) / frame_length**2
     else:
-        raise ValueError('Either `y` or `S` must be input.')
+        raise ParameterError('Either `y` or `S` must be input.')
```
bmcfee: This should have been a ParameterError all along. See #1069

librosa/feature/spectral.py
```diff
@@ -838,8 +838,6 @@ def rms(y=None, S=None, frame_length=2048, hop_length=512,
     >>> plt.show()
 
     '''
-    if y is not None and S is not None:
-        raise ValueError('Either `y` or `S` should be input.')
```
bmcfee: This block was redundant with the final else block, so it's now removed.

librosa/segment.py
```diff
@@ -161,7 +161,7 @@ def cross_similarity(data, data_ref, k=None, metric='euclidean',
     data = np.atleast_2d(data)
 
     if data_ref.shape[0] != data.shape[0]:
-        raise ValueError("data_ref and data must have the same first dimension")
+        raise ParameterError("data_ref and data must have the same first dimension")
```
bmcfee: Should have been a ParameterError all along.

bmcfee:
Since this PR includes a rewrite of most of the tests anyway, I took the liberty of also blacking the test code per #1070. 

Also added a few more tests to increase coverage.

setup.py
```diff
@@ -57,7 +57,8 @@
                   'mock',
```
scjs: I think `mock` can be removed from the testing dependencies. Python >= 3.3 includes it in the standard library with `unittest`.

bmcfee:
@ajweiss nevermind, I got the lpcburg generator working in octave and checked in the files.  I think we're good to go on this now.

setup.py
bmcfee: Good catch, I missed that one previously.

bmcfee:
Note: this PR adds a goodly chunk of new .mat data to the test directory, which should have been there previously from #725 .  At some point, we'll probably want to purge this #319 and call directly via octave bridge, but i'm leaving it for now.  Once the remaining CI's finish, I think this is good to merge.

ajweiss:
Just catching up on this now.  The good news is, I made some follow up code changes in #861 and used the unit tests to verify correctness (the work was done on a computer that didn't have the LPC .mat data) and the fact that the regression test still passes now after being switched off for some time indicates that the unit test is pretty good at verifying correctness.

In terms of what to do long term, all the binary data seems a little bit overkill (I think I used existing tests as a guide in terms of building an exhaustive regression test) but it doesn't hurt if it doesn't cause problems with the main repo getting too large.  Dynamically generating the data from Octave is an interesting idea, but then that adds build dependencies on Octave which could get unwieldly.  That said, if there are enough reasons to pull Octave in (it can be used for most/all of the regressions) then it seems to make sense here as well.  This seems like the best solution assuming Octave results don't diverge too much from what we test against now.

Git LFS for the binary data works as well, but I worry that as the code that generates it ages, it may rot.

PR #1066, opened 2020-02-04
Make audioread work with open file objects

carlthome:
I'm currently working with MP3 ([MTG-Jamendo](https://github.com/MTG/mtg-jamendo-dataset)) so I rely on librosa's audioread integration. Got a bit confused that `librosa.load` accepts open file descriptors for soundfile but not for MP3. Until https://github.com/erikd/libsndfile/issues/258 is in, I propose that both librosa backends accept both string paths and open file objects. Thoughts?

Also noticed that there doesn't seem to be a unit test for `librosa.load` with open files. Should that be added or is it excessive? I'm on the fence, personally, since https://github.com/erikd/libsndfile/issues/258 will shake things up anyway.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -3,6 +3,7 @@
 """Core IO, DSP and utility functions."""
 
 import pathlib
+import tempfile
 import warnings
 
 import soundfile as sf
@@ -51,8 +52,7 @@ def load(path, sr=22050, mono=True, offset=0.0, duration=None,
         Any string file paths, or any object implementing Python's
         file interface (e.g. `pathlib.Path`) are supported as `path`.
 
-        If the codec is supported by `soundfile`, then `path` can also be
-        an open file descriptor (int).
+        `path` can also be an open file descriptor (int).
 
     sr   : number > 0 [scalar]
         target sampling rate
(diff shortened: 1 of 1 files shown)


bmcfee:
> I propose that both librosa backends accept both string paths and open file objects. Thoughts?

I see where you're coming from with this, and I agree that the api is a little quirky right now.  I would support this if audioread had direct support for open file descriptor decoding (as soundfile does). Of course, it doesn't (and can't, due to how it works at a basic level), so your proposed solution is probably the only way to get this kind of behavior.

I'm on the fence about whether this is something that we should merge though, for a couple of reasons:

1. The audioread dependency is (we hope) a temporary thing, and in general, I'd prefer to remove IO code rather than add more.
2. The use-case is relatively uncommon.  
3. The proposed solution hides a significant amount of overhead (flushing to disk and reading back). It could maybe be done a bit quicker with a named pipe instead of a tempfile, but it's going to be an awkward dance no matter what.  Awkward things are hard to test and maintain, so I prefer to avoid them. :grin: 

All that said, it's a legitimate problem and a reasonable workaround that you propose, but I don't think we should merge it.  What do you think about making it an example in the advanced gallery?  That way, people that really need this functionality can find a worked example, but we won't have to work too hard to maintain it.  Eventually it won't be necessary anymore, and we can just drop the example.



> Also noticed that there doesn't seem to be a unit test for `librosa.load` with open files. Should that be added or is it excessive?

Also a gray area here.  It probably should be there, but it's also functionality that should be tested at the soundfile side and that we can take for granted.  If we do want to expand the tests though, I suggest waiting until I'm done with #1064 (probably in the next few weeks, time allowing).

carlthome:
Thanks for the quick respone!

I agree, I'm also on the fence about adding more complexity for audioread in `load` when the general direction is to go for libsndfile. I looked into adding file object support in audioread but it involves updating [several class constructors](https://github.com/beetbox/audioread/blob/master/audioread/rawread.py#L62) and felt very mucky to verify. That's where this logic fits in best though IMO, going by conventions like in pandas etc.

Figured 3. wouldn't be so bad for MP3 but I agree in principle.

> What do you think about making it an example in the advanced gallery? 

Agree. Might even be enough to just leave this snippet here on GitHub for future reference?

```py
import tempfile

import librosa as lr
import tensorflow as tf

samplerate = 22050
filename = f"gs://my-bucket/my-audio.mp3"

with tf.io.gfile.GFile(filename, "rb") as f:
    with tempfile.NamedTemporaryFile() as ntf:
        ntf.write(f.read())
        waveform = lr.load(ntf.name, samplerate)[0]
```

aungpyae01012540-spec (approved):

PR #1068, opened 2020-02-12
Fix pitch shift doc

keunwoochoi:
It fixes the documentation of `pitch-shift`. I think current doc is confusing `step`, and `semitone`. I understood that `step` is defined in this function something that may or may not be equal to semitone.


lostanlen:
Yes, this is much better. LGTM
thanks!

bmcfee:
Thanks @keunwoochoi !

Issue #1070, opened 2020-02-19
Coding style updates


bmcfee:
#### Description

In implementing #1064, I noticed that our coding style is a bit inconsistent in places.  As much as I dislike automated style checking, it might be good to give it a once over with black using some sane parameters (eg, 120 line length because it's not 1987).

To be clear, I don't think we should automate this in CI, or require that new contributors always stick strictly to this style, but we are overdue IMO for a cleanup. (Expecially in tests!)

lostanlen:
> 120 line length

Really? Isn't that a bit long? I remember Raymond Hettinger advocating for "90-ish". It's also what black adheres too.

bmcfee:
> Really? Isn't that a bit long?

Ehh fine :)  i just want to make sure we don't go into that ridiculous 78-char limit nonsense.

bmcfee:
Fixed by #1193

PR #1077, opened 2020-02-23
[RFC, do not merge] run black on core/audio.py

lostanlen:
#### Reference Issue
Related to #1070 

This PR is the raw result of running `black librosa/core/audio.py` on the master branch.
It's meant to serve as a conversation starter around #1070
We can configure black a little bit more finely in the future depending on what we like.

### Any other comments?
* i like the new formatting of the function prototype, with every argument on a line of its own.

* double quotes vs single quotes: should we let black standardize this? right now it puts everything to double quotes.

* line 215 is a bit weird: `frame = frame[: s_end - n_prev]`. do we really need space after this colon?

* note that black adds whitespace around _every_ operator. Yet I tend to remove the whitespace to signify operator precedence. So black would destroy that visual cue.
`den = q*den - bwd_pred_error[-1]**2 - fwd_pred_error[0]**2`
becomes `den = q * den - bwd_pred_error[-1] ** 2 - fwd_pred_error[0] ** 2`

* compound return statements become parenthesized and multiline.
Before:
```python
    return np.pad((y_sign[tuple(slice_post)] != y_sign[tuple(slice_pre)]),
                  padding,
                  mode='constant',
                  constant_values=pad)
```

After:
```python
    return np.pad(
        (y_sign[tuple(slice_post)] != y_sign[tuple(slice_pre)]),
        padding,
        mode="constant",
        constant_values=pad,
    )
```

Before:
```python
        return np.digitize(x_comp,
                           np.linspace(-1, 1, num=int(1+mu), endpoint=True),
                           right=True) - int(mu + 1)//2
```
After:
```python
        return (
            np.digitize(
                x_comp, np.linspace(-1, 1, num=int(1 + mu), endpoint=True), right=True
            )
            - int(mu + 1) // 2
        )
```
It almost makes me think that we want to assign a variable before applying the `return` statement.

* On this example (`core/audio.py`), the overall line length remained almost the same. 1521 lines before and 1574 lines after. But the number of additions/deletions is quite high, mostly because of single quote to double quote conversion.

* Worth watching: https://www.youtube.com/watch?v=esZLCuWs_2Y

* Thoughts?

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -18,22 +18,40 @@ from .._cache import cache
 from .. import util
 from ..util.exceptions import ParameterError
 
-__all__ = ['load', 'stream', 'to_mono', 'resample',
-           'get_duration', 'get_samplerate',
-           'autocorrelate', 'lpc', 'zero_crossings',
-           'clicks', 'tone', 'chirp',
-           'mu_compress', 'mu_expand']
+__all__ = [
+    "load",
+    "stream",
+    "to_mono",
+    "resample",
+    "get_duration",
+    "get_samplerate",
+    "autocorrelate",
+    "lpc",
+    "zero_crossings",
+    "clicks",
+    "tone",
+    "chirp",
+    "mu_compress",
+    "mu_expand",
+]
 
 # Resampling bandwidths as percentage of Nyquist
-BW_BEST = resampy.filters.get_filter('kaiser_best')[2]
-BW_FASTEST = resampy.filters.get_filter('kaiser_fast')[2]
+BW_BEST = resampy.filters.get_filter("kaiser_best")[2]
+BW_FASTEST = resampy.filters.get_filter("kaiser_fast")[2]
 
 
 # -- CORE ROUTINES --#
 # Load should never be cached, since we cannot verify that the contents of
 # 'path' are unchanged across calls.
-def load(path, sr=22050, mono=True, offset=0.0, duration=None,
-         dtype=np.float32, res_type='kaiser_best'):
+def load(
+    path,
+    sr=22050,
+    mono=True,
+    offset=0.0,
+    duration=None,
+    dtype=np.float32,
+    res_type="kaiser_best",
+):
     """Load an audio file as a floating point time series.
 
     Audio will be automatically resampled to the given rate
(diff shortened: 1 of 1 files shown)


librosa/core/audio.py
```diff
@@ -345,20 +370,20 @@ def stream(path, block_length, frame_length, hop_length,
...
 
     if not (np.issubdtype(type(block_length), np.integer) and block_length > 0):
-        raise ParameterError('block_length={} must be a positive integer')
+        raise ParameterError("block_length={} must be a positive integer")
     if not (np.issubdtype(type(frame_length), np.integer) and frame_length > 0):
-        raise ParameterError('frame_length={} must be a positive integer')
+        raise ParameterError("frame_length={} must be a positive integer")
     if not (np.issubdtype(type(hop_length), np.integer) and hop_length > 0):
-        raise ParameterError('hop_length={} must be a positive integer')
+        raise ParameterError("hop_length={} must be a positive integer")
```
bmcfee: whoa, these raise lines are wrong!  there's no `.format(hop_length)` (etc) to interpolate the {} to actual values.

This is independent of the issue for this PR, but good to fix anyway.

bmcfee:
>     * i like the new formatting of the function prototype, with every argument on a line of its own.

I agree in general, but it does have the unfortunate side effect of breaking folds in vim:

image

This is probably something to be fixed on my side, but I wanted to point it out.


>     * double quotes vs single quotes: should we let black standardize this? right now it puts everything to double quotes.

I generally prefer single, but I'm okay with double.  It's mostly a question of which one requires more escaping (ie due to nested punctuation).

>     * line 215 is a bit weird: `frame = frame[: s_end - n_prev]`. do we really need space after this colon?

I don't think that we do, but it doesn't hurt.



>     * note that black adds whitespace around _every_ operator. Yet I tend to remove the whitespace to signify operator precedence. So black would destroy that visual cue.

I prefer your way, though it can get messy when compounded.

> It almost makes me think that we want to assign a variable before applying the `return` statement.

This is probably a good thing to do anyway.

librosa/core/audio.py
lostanlen: good catch ...

bmcfee:
#1193 implements code styling, so I think we can close this PR out.  (It seemed easier to do it that way than to rebase this one.)

PR #1078, opened 2020-03-03
fix wrong output of function stack_memory

### LINKED ISSUE #1061: librosa.feature.stack_memory returns WRONG VALUES
#### Description

Error when using `librosa.feature.stack_memory ` function, the returned array contains some values that doesn't exists in the input array, and the wrong result is random, the probability of failure increases with bigger `n_steps` parameter

#### Steps/Code to Reproduce

```python
import librosa
import numpy as np

# generate random array in range [0, 24)
x = np.remainder((np.random.random(10000)*100).astype(int), 24)
assert (x < 24).all(), f'this should not alarm'
stackedX = librosa.feature.stack_memory(
    x, n_steps=500, mode='constant')  # same error with mode type edge and reflect
print(f'{stackedX[stackedX >= 24]}')  # will print very big integers
assert (stackedX < 24).all(), f'this will alarm'

```

#### Expected Results

```
$ python test
(nothing happens)
```

#### Actual Results

```
$ python test
[2097163]
  File "testST.py", line 10, in <module>
    assert (stackedX < 24).all(), f'this will alarm'
AssertionError: this will alarm
```

#### Versions

```python
Linux-5.0.0-37-generic-x86_64-with-debian-buster-sid
Python 3.7.4 (default, Aug 13 2019, 20:35:49) 
[GCC 7.3.0]
NumPy 1.16.4
SciPy 1.3.0
librosa 0.7.2
>>> 
```

beantowel:
#### Reference Issue
Fixes #1061

Fixed wrong output of the function `stack_memory` when input array is large. Added the above test case in `test_features.py`

#### Any other comments?
Though this patch didn't use numba to rewrite this function more efficiently (which i tried a bit but it seems some functions is not compatible to jit). The patch fixed the issure #1061 and it's slightly faster.

### DIFF
--- a/librosa/feature/utils.py
+++ b/librosa/feature/utils.py
@@ -238,11 +238,7 @@ def stack_memory(data, n_steps=2, delay=1, **kwargs):
 
     data = np.pad(data, [(0, 0), padding], **kwargs)
 
-    history = data
-
-    # TODO: this could be more efficient
-    for i in range(1, n_steps):
-        history = np.vstack([np.roll(data, -i * delay, axis=1), history])
+    history = np.vstack([np.roll(data, -i * delay, axis=1) for i in range(n_steps)[::-1]])
 
     # Trim to original width
     if delay > 0:
--- a/tests/test_features.py
+++ b/tests/test_features.py
@@ -69,9 +69,9 @@ def test_delta_badwidthaxis(x, width, axis):
     librosa.feature.delta(x, width=width, axis=axis)
 
 
-@pytest.mark.parametrize("data", [np.random.randn(5), np.random.randn(5, 5)])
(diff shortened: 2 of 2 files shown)


bmcfee:
Thanks for taking care of this -- LGTM!

I still want to numba-fy this operation, but that can be a separate PR.  Do you have anything else to merge here, or shall we :shipit: ?

bmcfee:
@beantowel sorry to revive this thread, but we're preparing release notes.  How would you like to be addressed in the contributing authors list?

beantowel:
Thank you! beantowel would just be fine.

bmcfee:
ok, thanks!

PR #1083, opened 2020-03-07
fix peak_pick util func doc

changgunyee:
fix documentation of util.peak_pick func.
Since `n` is the length of `x` rows, it is a little bit confusing.

### DIFF
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -969,16 +969,16 @@ def peak_pick(x, pre_max, post_max, pre_avg, post_avg, delta, wait):
         input signal to peak picks from
 
     pre_max   : int >= 0 [scalar]
-        number of samples before `n` over which max is computed
+        number of samples before the current over which max will be computed
 
     post_max  : int >= 1 [scalar]
-        number of samples after `n` over which max is computed
+        number of samples after the current over which max will be computed
 
     pre_avg   : int >= 0 [scalar]
-        number of samples before `n` over which mean is computed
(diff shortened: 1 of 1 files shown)


lostanlen:
i'm not sure that this is truly needed. `n` is clearly defined above.

bmcfee:
Thanks @changgunyee for opening this PR.  I do agree with @lostanlen in this case, all terms are clearly defined in the context of the docstring, and I think it's clearer to refer explicitly to `n` here.

Maybe someone else a little less close to the code can chime in though?

changgunyee:
I thought there's two `n`. One in the description part for `x[n]` and the other in params for x's shape `shape=[n,]`. And it was confusing for me.

But it's like other people weren't, I will close this PR.

Thanks for CR.

lostanlen:
ah, yes. Good point. I would be in favor of changing the shape to a different letter, like `N` for example. But perhaps it's not necessary to define a math variable for it at all. We just need to convey the idea that it's a 1-D time series. It's best to remain consistent across the library.

changgunyee:
I read most of the documentation and it seems changing `n` in description part to `i` looks good as other function docstrings usually use `i` as index for loop and `n` for 1-D time series.

Is there another suggestion?

lostanlen:
looks good to me

changgunyee:
Thanks for review. I'll close this since this is not significant PR.

bmcfee:
Thanks for your effort on this anyway, @changgunyee !

PR #1094, opened 2020-04-02
STFT fails for large window sizes

fzalkow:
The STFT fails when using large window sizes. This is due to the computation of `n_columns`, which can be zero. `n_columns` is used as third argument for `range`, which must not be zero. To be honest, I don't fully get the idea of the computation of `n_columns`. I just added a quick fix to enforce that `n_columns` is at least one.

The following code example illustrates the issue:

```python
import numpy as np 
import librosa

x = np.empty(1000000)

for exponent in range(12, 17):
    N = 2 ** exponent
    H = N // 2
    print(N, H)
    X = librosa.stft(x, n_fft=N, hop_length=H, win_length=N)
```

Before this pull request, this code example raised an error: `ValueError: range() arg 3 must not be zero`

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -229,6 +229,7 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
     # how many columns can we fit within MAX_MEM_BLOCK?
     n_columns = int(util.MAX_MEM_BLOCK / (stft_matrix.shape[0] *
                                           stft_matrix.itemsize))
+    n_columns = max(n_columns, 1)
 
     for bl_s in range(0, stft_matrix.shape[1], n_columns):
         bl_t = min(bl_s + n_columns, stft_matrix.shape[1])
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -213,6 +213,15 @@ def test_stft(infile):
     assert np.allclose(D, DATA["D"].conj())
 
 
+def test_stft_winsizes():
+    x = np.empty(1000000)
+
+    for power in range(12, 17):
+        N = 2 ** power
+        H = N // 2
+        librosa.stft(x, n_fft=N, hop_length=H, win_length=N)
+
+
 # results for FFT bins containing multiple components will be unstable, as when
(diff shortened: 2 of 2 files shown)


bmcfee:
Thanks for catching this.

The general idea behind the `n_columns` calculation is to limit the amount of data (measured in bytes, not samples) to send through FFT.  This gives us an effective STFT block size, which doesn't change the computation at all, just the amount of columns that we do in parallel.  I think your thresholding fix is a good solution here.

I wonder if a similar problem pops up in our `nnls` solver -- can you look into that as well and see if the same problem (and fix) applies?

librosa/core/spectrum.py
```diff
@@ -229,6 +229,7 @@ def stft(y, n_fft=2048, hop_length=None, win_length=None, window='hann',
     # how many columns can we fit within MAX_MEM_BLOCK?
     n_columns = int(util.MAX_MEM_BLOCK / (stft_matrix.shape[0] *
                                           stft_matrix.itemsize))
+    n_columns = max(n_columns, 1)
```
bmcfee, line 232: LGTM.  Now might also be a good opportunity to switch this over to use integer division `//` and drop the explicit int cast to be more python3ic.

fzalkow:
Thanks for the explanation!

I applied the same fix for `nnls`. In principle, the same issue could happen there. However, I couldn't produce the problem in practice, because the problem only occurs if the matrix `A` is large in its second dimension. And I get a memory error in this case before the `n_columns` issue arises.

fzalkow:
The test ist much faster than a second. On my system:

```
%timeit test_stft_winsizes
15.2 ns ± 0.0796 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)
```

fzalkow:
Oops, sorry, I did the timing wrong... :sweat: Still, it's less than a second.

```
%timeit test_stft_winsizes()
133 ms ± 3.07 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

bmcfee:
Thanks @fzalkow !  Sorry for taking so long to merge this.  (It's been a wild month.)

PR #1107, opened 2020-05-01
update numba calls for version 0.50.0

pusalieth:
compatibility with version 0.50.0

--tested with SV2TTS

numba version 0.50.0 uses a different structure for import

#### Any other comments?
Tested the change with SV2TTS. Everything worked as normal.

### DIFF
--- a/librosa/util/decorators.py
+++ b/librosa/util/decorators.py
@@ -5,7 +5,7 @@
 
 import warnings
 from decorator import decorator
-from numba.decorators import jit as optional_jit
+from numba.core.decorators import jit as optional_jit
 
 __all__ = ['moved', 'deprecated', 'optional_jit']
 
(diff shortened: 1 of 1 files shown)


pusalieth:
Hold on this. There may be an issue after all.

pusalieth:
Okay, disregard. The program was using an older version of librosa, which used the output call. That's what was failing.

bmcfee:
Thanks for this!  LGTM, but do you know if we need to revise the minimum numba version requirement for this to work?  It looks like the numba refactor happened at 0.49, and our current minimum version (0.43) does not work with this change.  I'm okay with bumping to 0.49, but it would be good to figure out exactly what the minimum requirement is.

pusalieth:
I just tested the change against every version of numba. Looks like 0.49.0 is the fist to change the syntax. I change the code to seamlessly change the syntax for those using 0.49.0 or greater, and still keep functionality for those using 0.43.0. You should be able to merge it now, and it not affect anyone.

bmcfee:
Excellent, thanks!

PR #1112, opened 2020-05-06
[CR needed] fixed #1111 - revised stack_memory and tests

### LINKED ISSUE #1111: stack_memory unit tests fail randomly
#### Description
Stack_memory has heisenbugs in its unit test.  It's unclear whether this comes from randomness in the test fixture (probably) or a deeper bug in the implementation.

This may be a regression from #1078 , and may indicate that it's time to overhaul that entire function., as indicated in the comment thread on that PR.

bmcfee:
#### Reference Issue
Fixes #1111 

This PR rewrites `stack_memory` to use a numba-accelerated helper function for improved efficiency.

It also fixes a bug in the test fixtures for `stack_memory`.

#### Any other comments?

The new implementation is about 10x faster on realistic work-loads:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ addons:
         packages:
         - libsamplerate0
 
-    brew:
+    homebrew:
         packages:
         - libsamplerate
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,20 @@ jobs:
           language: generic
           osx_image: xcode11
 
+addons:
+    apt:
+        packages:
+        - libsamplerate0
+
+    brew:
+        packages:
+        - libsamplerate
+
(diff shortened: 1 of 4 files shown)


bmcfee:
Note: the implementation is still relatively memory-inefficient, as it invokes two copies: one for padding and one for the output array.  It at least improves on the old implementation by pre-allocating the entire output instead of repeatedly calling vstack.  I'm not sure it's worth optimizing further, since doing so would require implementing some implicit padding behavior.

bmcfee:
Also rolled in a couple of fixes to travis build caching with dependencies.

bmcfee:
This one's ready for a CR if anyone cares to volunteer.

PR #1118, opened 2020-05-16
Fix numba version deprecation warning

levkk:
Screenshot from 2020-05-16 10-15-33

### DIFF
--- a/librosa/util/decorators.py
+++ b/librosa/util/decorators.py
@@ -4,9 +4,10 @@
 '''Helpful tools for deprecation'''
 
 import warnings
+from packaging import version
 from decorator import decorator
 import numba
-if (numba.__version__ < '0.49.0'):
+if version.parse(numba.__version__) < version.parse('0.49.0'):
     from numba.decorators import jit as optional_jit
 else:
     from numba.core.decorators import jit as optional_jit
--- a/librosa/util/decorators.py
+++ b/librosa/util/decorators.py
@@ -4,10 +4,9 @@
 '''Helpful tools for deprecation'''
 
 import warnings
-from packaging import version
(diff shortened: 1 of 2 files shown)


bmcfee:
@levkk Thanks for this -- the initial commit was fine, but I'm confused about the reversion to the old string check (with the newer version).

If you roll back the last commit, I think this would be good to go.

Don't worry about the failing test on travis; it's due to a combination of issues that are out of scope for this one, and handled by a separate PR.

librosa/util/decorators.py
```diff
@@ -4,9 +4,10 @@
 '''Helpful tools for deprecation'''
 
 import warnings
+from packaging import version
```
bmcfee, line 7: Can you add `packaging` to the requirements list in setup.py?

Otherwise looks good!

bmcfee:
BTW, I've merged the fix for the travis build.  If you could rebase this PR to the current master, tests should run cleanly again.  Sorry about that!

setup.py
```diff
@@ -38,6 +38,7 @@
     install_requires=[
         'audioread >= 2.0.0',
         'numpy >= 1.15.0',
+        'packaging >= 20.3',
```
bmcfee, line 41: Thanks -- Is 20.3 the minimal version required here?  From a skim of the docs, it looks like anything above 17 would do the trick, but I don't have much experience with this package.

levkk: That's what I had in my virtualenv, totally not required I think. I can put 17.

levkk, line 41:
```suggestion
        'packaging >= 17',
```

bmcfee:
LGTM. Thanks again!

PR #1123, opened 2020-05-19
add python_requires to setup.py

stefansjs:
add python_requires to setup.py so that unsupported python interpreters don't install librosa by accident.

This adds the python_requires parameters to setup.py so that pip >= 9.0 doesn't install this package on incompatible python interpreters. I may or may not have discovered the hard way that librosa does not support python 3.5.

#### Any other comments?

According to the docs https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires, this requires setuptools >= 24.2.0 in order to build a package. The docs also say that users with older versions of pip will still be able to download and install the package, but IIUC pip 9.0 is actually the default interpreter for virtualenvs. My guess is that this isn't very many users, but it's not breaking any functionality for older pip versions anyways.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
         'numba >= 0.43.0',
         'soundfile >= 0.9.0',
     ],
-    python_requires='>=3.6, <=3.8',
+    python_requires='>=3.6',
     extras_require={
         'docs': ['numpydoc', 'sphinx!=1.3.1', 'sphinx_rtd_theme',
                  'matplotlib >= 2.0.0',
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,7 @@ setup(
         'numba >= 0.43.0',
         'soundfile >= 0.9.0',
     ],
+    python_requires='>=3.6, <=3.8',
     extras_require={
         'docs': ['numpydoc', 'sphinx!=1.3.1', 'sphinx_rtd_theme',
                  'matplotlib >= 2.0.0',


bmcfee:
Thanks for this.  Do you think we need the <= 3.8?

setup.py
```diff
@@ -46,6 +46,7 @@
         'numba >= 0.43.0',
         'soundfile >= 0.9.0',
     ],
+    python_requires='>=3.6, <=3.8',
```
bmcfee, line 49: <=3.8 fails on, say, 3.8.2.  (Two of our CI builds failed from this.)

Can we drop the upper bound?

Otherwise, LGTM.

stefansjs: I see. I thought it would take all 3.8 versions. That's my mistake.

This is really more of an opinion than anything. If you've tested on a specific set of python interpreters then you can think of python_requires as a statement about what versions are validated to work with librosa. OTOH, you can treat future breaking changes in the interpreter as exceptional and assume forwards compatibility. My preference is for the defensive approach, which is probably an artifact of living in dependency hell so often. But if you think most librosa users are in a more researchy environment then I can see the argument for assuming forwards compatibility.

Should I drop the upper bound in the PR? Or bump the 3.8 version higher (e.g 3.8.2)?

bmcfee: I'd prefer to drop the upper bound. Python releases more frequently than we do, and there's nothing on the 3.9 roadmap that should break compatibility for us anyway.

PR #1125, opened 2020-05-20
[CR needed] Updates to frame API

### LINKED ISSUE #916: Soften contiguity requirements for framing
#### Description

`util.frame` currently raises an exception if the input buffer is discontiguous.  This is required for efficient in-place view construction, but an exception might not be the right pattern here because contiguity can be a subtle thing from the user's perspective.

What do people think about relaxing this exception to a warning, which constructs a contiguous copy (if necessary) and provides instructions for how to force contiguity?  This way, things will still work, just less efficiently, and the user will have clearer directions for fixing it.

### LINKED ISSUE #965: Remove default length and hop parameters for util.frame
#### Description

This is a bit of a nitpick, but with the eternal yak shave of #916 and #944, we're opening up the possibility of applying `frame` to any sort of contiguous array, not just audio.  In that case, I don't think it makes sense to keep the default frame and hop parameters for the `frame` function.

This would be a minor API break.  It's not a huge deal because A) we always call it with explicit parameters, and B) `frame` is primarily an internally used function.  However, it is a public function, so we need to be a bit careful.

bmcfee:
#### Reference Issue
- [x] Fixes #965 
- [x] Fixes #916 

This PR implements the following API changes to `util.frame`:

- #965 : removes default values for `frame_length` and `hop_length`.  These made sense when frame was only used for audio buffers, but now that we're more general, we should drop the defaults to prevent unexpected behaviors.
- #916 Relax contiguity violations from `exception` to `warning`, and invoke a copy to get the necessary shape.

### DIFF
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 """Utility functions"""
 
+import warnings
 import scipy.ndimage
 import scipy.sparse
 
@@ -37,7 +38,11 @@ def frame(x, frame_length, hop_length, axis=-1):
 
     This implementation uses low-level stride manipulation to avoid
     making a copy of the data.  The resulting frame representation
-    is a new view of the input data.
+    is a new view of the same input data.
+
+    However, if the input data is not contiguous in memory, a warning
+    will be issued and the output will be a full copy, rather than
+    a view of the input data.
 
     For example, a one-dimensional input `x = [0, 1, 2, 3, 4, 5, 6]`
     can be framed with frame length 3 and hop length 2 in two ways.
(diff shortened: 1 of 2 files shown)


bmcfee:
This one's ready for CR, mainly looking for feedback on the updates to the documentation and warnings. Any takers?

lostanlen:
I like this. Regarding #916, i suppose that the old behavior (stricter) can still be achieved with a `with warning.catch_warnings()` block. Printing a warning and making a copy is less efficient, but is nice for interactive use. And in some intricate cases, there's really no other way than to make a copy.

bmcfee:
> i suppose that the old behavior (stricter) can still be achieved with a `with warning.catch_warnings()` block.

Indeed.  The thing that makes me a little nervous is that this is really the only way to test if a new memory allocation has been performed here, and that the behavior is now dynamic where it previously wasn't.  (You can't check the owndata flag of the framed array because the new version still doesn't own its data.)

That said, I think it's a good compromise, and the utility of more general framing far outweighs the slight reduction in predictability of behind-the-scenes work.

Issue #1130, opened 2020-05-27
Taking multi-channel seriously


bmcfee:
Most of librosa only supports monophonic audio.  But for many of the analyses we'd like to do, stereo or multi-channel support would be very useful and not all that difficult at this point.

This issue is meant to kick off discussion of how this will work, but I have some thoughts as outlined below.

# Conventions

In general, we should continue to support native mono `y.shape = (N,)` without artificial up-casting to explicit mono `(1, N)`.

For example, an stft on `(N,)` will still produce an output of shape `(# freqs, # frames)`.

However, an stft on `(1, N)` would produce output `(1, #freqs, #frames)`.  Now that framing is fully generalized to multichannel, this should not present any difficulties.  More generally, `(K, N)` would map to `(K, #freqs, #frames)`.

As a general rule, the trailing dimension will be (usually) treated as time-like (samples, frames, etc), and the leading dimension will be channels.

# How it will work

## Things that will generalize easily

- STFT and friends
- Linear filters (mel, mfcc, chroma_stft, and co)
- HPSS

## Things that will be a pain to generalize

- CQT and friends?
- Structure / recurrence
- General decomposition
- effects?
- inverse transforms (mfcc, mel, etc)

## Things that should stay monophonic

- Detectors (onset, beat, pitch)
- Display
- Sequence modeling (dtw, viterbi)

bmcfee:
Quick rejoinder on this:

For linear spectral projections, the following works* for generalizing single to multichannel in exactly the way we'd expect:

```python

# S is either (f, t) or (c, f, t), or generally (..., f, t)
# M is (m, f)
Mein = np.einsum('...ft,mf->...mt', S, M)  # Better with optimize=True

# Mein is (m, t) or (c, m, t) or generally (..., m, t)
```

\* by `works` I mean that it produces the correct values in the correct shape.  However, the memory layout may or may not adhere to any contiguity constraints.  Einsum does claim to let you force this with the `order=` parameter, but a few minutes of hacking didn't pay off for me yet.

\** Also, the summation path optimization could probably be cached/precomputed, rather than optimized every time.  Will take a bit of thinking on how to make that generally work in a flexible way.

bmcfee:
>  by `works` I mean that it produces the correct values in the correct shape. However, the memory layout may or may not adhere to any contiguity constraints. Einsum does claim to let you force this with the `order=` parameter, but a few minutes of hacking didn't pay off for me yet.

[...]

PR #1131, opened 2020-05-27
Tightened requirements on valid_audio

### LINKED ISSUE #1122: waveplot() segfaults when given input of shape (1, 35000)
#### Description
When waveplot() is given a numpy array of shape (1, 35000), the interpreter segfaults.

#### Steps/Code to Reproduce
```
$ python
Python 3.7.6 (default, Feb  8 2020, 13:20:58) 
[GCC 9.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from librosa.display import waveplot
/home/sliedes/.virtualenvs/torch/lib/python3.7/site-packages/librosa/util/decorators.py:9: NumbaDeprecationWarning: An import was requested from a module that has moved location.
Import requested from: 'numba.decorators', please update to use 'numba.core.decorators' or pin to Numba version 0.48.0. This alias will not be present in Numba version 0.50.0.
  from numba.decorators import jit as optional_jit
/home/sliedes/.virtualenvs/torch/lib/python3.7/site-packages/librosa/util/decorators.py:9: NumbaDeprecationWarning: An import was requested from a module that has moved location.
Import of 'jit' requested from: 'numba.decorators', please update to use 'numba.core.decorators' or pin to Numba version 0.48.0. This alias will not be present in Numba version 0.50.0.
  from numba.decorators import jit as optional_jit
>>> import numpy as np
>>> waveplot(np.zeros((1, 35000)))
Segmentation fault (core dumped)
```

#### Expected Results
waveplot fails gracefully, at least does not crash the interpreter.

#### Actual Results
Python interpreter segfaults.

#### Versions
Linux-5.6.7-arch1-1-x86_64-with-arch
Python 3.7.6 (default, Feb  8 2020, 13:20:58) 
[GCC 9.2.0]
NumPy 1.17.3
SciPy 1.2.0
librosa 0.7.2

bmcfee:
#### Reference Issue
Fixes #1122 

This PR fixes / prevents #1122 by tightening the constraints on `valid_audio`.

Monophonic data must be of shape `(N,)`, multichannel data cannot have shape `(1, N)`. (Only 2 and above are allowed.)

#### Any other comments?

This should be revised when we implement #1130

### DIFF
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -212,7 +212,7 @@ def valid_audio(y, mono=True):
     `(N,)` (number of samples).
 
     If `mono=False`, then `y` may be either monophonic, or have shape
-    `(2, N)` (stereo).
+    `(2, N)` (stereo) or `(K, N)` for `K>=2` for general multi-channel.
 
 
     Parameters
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -206,7 +206,13 @@ def frame(x, frame_length, hop_length, axis=-1):
 
 @cache(level=20)
 def valid_audio(y, mono=True):
-    '''Validate whether a variable contains valid, mono audio data.
+    '''Validate whether a variable contains valid audio data.
+
+    If `mono=True`, then `y` is only considered valid if it has shape
+    `(N,)` (number of samples).
+
+    If `mono=False`, then `y` may be either monophonic, or have shape
+    `(2, N)` (stereo).
 
 
     Parameters
@@ -215,7 +221,7 @@ def valid_audio(y, mono=True):
       The input data to validate
 
     mono : bool
-      Whether or not to force monophonic audio
+      Whether or not to require monophonic audio
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


bmcfee:
NB: Appveyor test fails are due to an old API key, and not indicative of real failures.

bmcfee:
Confirmed offline that this addresses the original waveplot issue #1122 .  I think this is good to go-- anyone care to look it over?

lostanlen:
unrelated, but: why is `valid_audio` cached? is it really faster?

bmcfee:
Any objections to merging this one?

lostanlen:
> It's O(N) in the "worst" case (valid audio) due to the isfinite check. It's cached because it may be called many times for the same audio in a processing pipeline.

Thanks for explaining. I was initially worried about the overhead of storing `y` in memory and comparing it to new inputs every time the function is called; but i had forgotten that librosa's cache was joblib-based (as opposed to memoized). So IIUC it caches by looking up a hash table, which is a small price to pay in comparison with `O(len(y))`.

LGTM

PR #1134, opened 2020-05-28
add B, C, D, Z weightings

beasteers:
#### Reference Issue

Implements #1007 

Currently only A weighting is implemented. This adds on B, C, D, and Z weighting as well.

It also adds helpers:
 - `frequency_weighting(frequencies, 'C')` - choose weighting from string (case insensitive). Use `None` or `'Z'` for unweighted.
 - `multi_frequency_weighting(frequencies, 'ACZ')` - multiple weightings from string/list - name might change tho

#### Notes
 - This is the source I used for the equations/constants. I was having trouble finding the actual source specification without going thru a paywall. If we find better sourced constants I'm happy to update them. (https://en.wikipedia.org/wiki/A-weighting#Function_realisation_of_some_common_weightings) I've compared against other plots online and it seems like they check out tho

 - There is a discrepancy between the constants currently in librosa and the ones on the wiki page. In `A_weighting` librosa uses 12200 whereas the wiki page uses 12194. The difference tops out around 0.0085db (checked up to 2e15) so the difference is completely insignificant, but I just wanted to document it. B and C use 12194. I was tempted to change A too, but maybe some silly person is relying on 3 decimals of precision for their dBs so I held off.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1590,7 +1590,7 @@ def db_to_amplitude(S_db, ref=1.0):
 def perceptual_weighting(S, frequencies, kind='A', **kwargs):
     '''Perceptual weighting of a power spectrogram:
 
-    `S_p[f] = A_weighting(f) + 10*log(S[f] / ref)`
+    `S_p[f] = frequency_weighting(f, 'A') + 10*log(S[f] / ref)`
 
     Parameters
     ----------
@@ -1599,6 +1599,10 @@ def perceptual_weighting(S, frequencies, kind='A', **kwargs):
 
     frequencies : np.ndarray [shape=(d,)]
         Center frequency for each row of `S`
+
+    kind : str
+        The frequency weighting curve to use.
+        e.g. `'A'`, `'B'`, `'C'`, `'D'`, `None or 'Z'`
 
     kwargs : additional keyword arguments
         Additional keyword arguments to `power_to_db`.
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1587,7 +1587,7 @@ def db_to_amplitude(S_db, ref=1.0):
 
 
 @cache(level=30)
-def perceptual_weighting(S, frequencies, **kwargs):
+def perceptual_weighting(S, frequencies, kind='A', **kwargs):
     '''Perceptual weighting of a power spectrogram:
 
     `S_p[f] = A_weighting(f) + 10*log(S[f] / ref)`
@@ -1654,7 +1654,8 @@ def perceptual_weighting(S, frequencies, **kwargs):
     >>> plt.show()
     '''
 
-    offset = time_frequency.A_weighting(frequencies).reshape((-1, 1))
+    offset = time_frequency.frequency_weighting(
+        frequencies, kind=kind).reshape((-1, 1))
 
     return offset + power_to_db(S, **kwargs)
 
--- a/librosa/core/time_frequency.py
+++ b/librosa/core/time_frequency.py
@@ -1474,6 +1474,7 @@ WEIGHTING_FUNCTIONS = {
     'C': C_weighting,
     'D': D_weighting,
     'Z': Z_weighting,
+    None: Z_weighting,
 }
 
 
--- a/librosa/core/time_frequency.py
+++ b/librosa/core/time_frequency.py
@@ -1346,7 +1346,7 @@ def B_weighting(f_sq):     # pylint: disable=invalid-name
     >>> plt.show()
 
     '''
-    const = np.array([12200, 20.6, 158.5]) ** 2.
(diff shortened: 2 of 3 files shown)


librosa/core/time_frequency.py
```diff
@@ -1257,33 +1284,299 @@ def A_weighting(frequencies, min_db=-80.0):     # pylint: disable=invalid-name
...
+
+    Returns
+    -------
+    weighting : scalar or np.ndarray [shape=(n,)]
+        `weighting[i]` is the weighting of `frequencies[i]`
+
+    See Also
+    --------
+    perceptual_weighting
+    frequency_weighting
```
bmcfee, line 1504: See also has a self-loop here

librosa/core/time_frequency.py
```diff
@@ -1257,33 +1284,299 @@ def A_weighting(frequencies, min_db=-80.0):     # pylint: disable=invalid-name
...
+    Returns
+    -------
+    weighting : scalar or np.ndarray [shape=(len(kinds), n)]
+        `weighting[i]` is the weighting of `frequencies[i]`
+
+    See Also
+    --------
+    perceptual_weighting
+    frequency_weighting
+    multi_frequency_weighting
```
bmcfee, line 1552: self-loop

librosa/core/time_frequency.py
```diff
@@ -1227,8 +1234,23 @@ def fourier_tempo_frequencies(sr=22050, win_length=384, hop_length=512):
...
+    def inner(frequencies, min_db=-80.0):
+        # Vectorize to make our lives easier
+        frequencies = np.asanyarray(frequencies)
+
+        # Call weighting function (with pre-computed squared frequency)
+        weights = func(frequencies**2.0)
+        if min_db is not None:
+            weights = np.maximum(min_db, weights)
+        return weights
+    return inner
```
bmcfee, line 1248: I see why you wrote it this way, but it would be more accessible if we just made this a normal subroutine instead of a decorator.  In general we try to keep python fanciness to a minimum :grin:

bmcfee:
Thanks @beasteers ! This looks good so far.  Aside from the CR, a couple of comments:

- Is the plan still to add a kwarg to `perceptual_weighting`?
- 12200 came from here: https://arxiv.org/pdf/1107.4969.pdf , which they pull from Smith'99.  However I'm having a tough time tracking down the exact number in that ref.  That said, if we can track down to a proper standard for this stuff, since the change is in milli-dBs, I'm okay with that.  Anyone relying on the 3rd decimal place in a dB calculation is asking for trouble anyway.

librosa/core/time_frequency.py
beasteers: good thing I restrained myself from setting the docstring with the decorator too 😝

So to clarify, you want me to just duplicate those lines directly into each weighting function and just get rid of the decorator? 

I just like how the weighting functions are the equation only, but however you want it since you're the maintainer

librosa/core/time_frequency.py
beasteers: good catch

librosa/core/spectrum.py
```diff
@@ -1587,7 +1587,7 @@ def db_to_amplitude(S_db, ref=1.0):
 
 
 @cache(level=30)
-def perceptual_weighting(S, frequencies, **kwargs):
+def perceptual_weighting(S, frequencies, kind='A', **kwargs):
     '''Perceptual weighting of a power spectrogram:
 
     `S_p[f] = A_weighting(f) + 10*log(S[f] / ref)`
```
bmcfee, line 1593: Probably should update this part too

beasteers: hmm update how? I added `kind='A', kind='C'`. Idk am I missing something?

tests/test_time_frequency.py
```diff
@@ -357,6 +357,79 @@ def test_A_weighting(min_db):
         assert not np.any(a_range < min_db)
 
 
+@pytest.mark.parametrize("min_db", [None, -40, -80])
```
bmcfee, line 360: These tests should not have prints in them

librosa/core/spectrum.py
bmcfee: :+1: thanks!

bmcfee: I think this didn't highlight properly -- what needs updating is the docstring, since it's no longer always the A_weighting.

tests/test_time_frequency.py
beasteers: oops my bad!

beasteers:
Don't think so! Good to go!

I was just porting some SONYC code and just wanted to get any weighting function definitions out of the code base. Glad it went smoothly!

bmcfee:
:shipit:

PR #1140, opened 2020-05-28
removed contiguity requirement for audio. fixes #1133

### LINKED ISSUE #1133: Should discontiguous audio be valid?
Since merging #1125 , we no longer have a strict requirement for the input to `util.frame` to be contiguous.  A discontiguous buffer will be copied to a proper arrangement to support framing.

Relatedly, `util.valid_audio` requires that the input buffer be F-contiguous to pass as valid.  This requirement was originally put in to ensure that framing can operate smoothly if the audio buffer is valid.

The question now is: should we relax the requirement in `valid_audio`?

*Why does this matter*: if we want to make a multi-dimensional STFT, it will need to frame a multi-channel input: `(2, N) -> (2, NFFT, F)`.  However, soundfile returns C-contiguous buffers, and the whole thing breaks.  If we relax the requirement in `valid_audio`, then the whole thing can proceed as expected, if not quite as efficiently as possible.

Related issue: #1130

bmcfee:
#### Reference Issue
Fixes #1133 

This PR removes the contiguity requirement for audio buffers.  As mentioned in the parent issue, this was originally put in place to guarantee that framing worked properly.

However, since merging #1125 , we no longer require contiguity (but it is encouraged), so this requirement does not seem necessary anymore.

#### Any other comments?

This PR initiates a long-term plan for #1130 .

### DIFF
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -236,8 +236,8 @@ def valid_audio(y, mono=True):
             - `y.dtype` is not floating-point
             - `mono == True` and `y.ndim` is not 1
             - `mono == False` and `y.ndim` is not 1 or 2
+            - `mono == False` and `y.ndim == 2` but `y.shape[0] == 1`
             - `np.isfinite(y).all()` is False
-            - `y.flags["F_CONTIGUOUS"]` is False
 
     Notes
     -----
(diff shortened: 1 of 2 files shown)


bmcfee:
Code-wise, I think this is good to go.  Probably good if we have some discussion around the implications of this change though.

Issue #1139, opened 2020-05-28
[Feature Request] cache removal


beasteers:
**Is your feature request related to a problem? Please describe.**
I'm looking into librosa cache for the first time and it looks really helpful, but I'm worried about there being no cache eviction at all which feels like it's asking for trouble.

I'm the type of person who would enable caching in my bashrc then forget about it, forget to call `librosa.cache.clear()` and then wonder why my disk space keeps filling up.

**Describe the solution you'd like**
Ideally, I'd like to be able to do something like this:

```python
import librosa
librosa.cache.enable('~/.librosa/cache', byte_limit=100*1024**2, ttl=60*60*24)
```
Though I see that with the current implementation, you can't enable caching after importing. So I guess this would be fine:
```python
import os
os.environ['LIBROSA_CACHE_DIR'] = '~/.librosa/cache'
os.environ['LIBROSA_CACHE_BYTE_LIMIT'] = str(100*1024**2)
os.environ['LIBROSA_CACHE_TTL'] = str(60*60*24)
import librosa
```

**Possible steps**
Flipping through the `joblib.Memory` implementation a bit, they have a parameter `byte_limit` already and they have a function `reduce_size` that will remove files that exceed `byte_limit`.

So the first course of action would be to add `LIBROSA_CACHE_BYTE_LIMIT` as an environment variable and find some place to call `.reduce_size()` periodically. My thought would be to call it at the end of a cached function if the time since the previous call exceeds some threshold (call reduce_size every 2 minutes) or something.

Then later for ttl, we could override the `get_cache_items_to_delete` behavior in joblib to check `datetime.now() - cache_item.last_access < os.getenv('LIBROSA_CACHE_TTL')`

I just wouldn't know how to fiddle with the `FunctionMaker` stuff.

**Describe alternatives you've considered**
I suppose one stopgap would be to automatically clear the librosa cache at the top/bottom of a script, but that feels a bit annoying and may not always be what you want to do.

bmcfee:
I agree this would be a great feature, but I think it's better implemented in joblib.  Cache eviction policies are tricky business though, and require serious thought.  https://github.com/joblib/joblib/issues/313 works through some of the difficulties of making this work.

beasteers:
>but I think it's better implemented in joblib.

Tbh I agree, and I'm surprised that it's not already part of the core. I was just looking at 4 years of inactivity on the issue and it seems like they just provided the `reduce_size` method and are expecting users to take it from there. 

[...]

PR #1142, opened 2020-05-29
allow bytes_limit in librosa cache

beasteers:
#### Reference Issue

Addresses #1139 

I figured it would be worth taking a second to sketch out a possible solution. 

What this does:
 - adds `LIBROSA_CACHE_BYTES_LIMIT` - the maximum cache size. If no bytes_limit is set, then the filesystem won't be checked.
 - adds `LIBROSA_CACHE_RESIZE_INTERVAL` - about how often we should check the cache size. I wasn't sure what a sensible default was so I set 5 minutes by default.
 - simplified `decorator.FunctionMaker(...)` to `functools.wraps`

If you'd rather close, that's fine. I would just feel better if we could set some cache limit.

### DIFF
--- a/librosa/_cache.py
+++ b/librosa/_cache.py
@@ -17,7 +17,6 @@ class CacheManager(object):
     preference for speed vs. storage usage.
     '''
 
-    last_resized_at = 0
     def __init__(self, *args, cache_resize_interval=None, level=10, **kwargs):
         # Initialize the memory object
         self.memory = Memory(*args, **kwargs)
--- a/librosa/_cache.py
+++ b/librosa/_cache.py
@@ -3,6 +3,8 @@
 """Function caching"""
 
 import os
+import time
+import functools
 from joblib import Memory
 
 
@@ -15,15 +17,16 @@ class CacheManager(object):
     preference for speed vs. storage usage.
     '''
 
-    def __init__(self, *args, **kwargs):
-
-        level = kwargs.pop('level', 10)
-
+    last_resized_at = 0
+    def __init__(self, *args, cache_resize_interval=None, level=10, **kwargs):
         # Initialize the memory object
         self.memory = Memory(*args, **kwargs)
         # The level parameter controls which data we cache
         # smaller numbers mean less caching
         self.level = level
+        # call reduce_size no more frequently than cache_resize_interval
+        self._throttled_reduce_size = throttle(
+            self.reduce_size, cache_resize_interval)
 
     def __call__(self, level):
         '''Example usage:
@@ -36,25 +39,17 @@ class CacheManager(object):
             '''Decorator function.  Adds an input/output cache to
             the specified function.'''
 
-            from decorator import FunctionMaker
-
-            def decorator_apply(dec, func):
-                """Decorate a function by preserving the signature even if dec
-                is not a signature-preserving decorator.
-
-                This recipe is derived from
-                http://micheles.googlecode.com/hg/decorator/documentation.html#id14
-                """
-
-                return FunctionMaker.create(
-                    func, 'return decorated(%(signature)s)',
-                    dict(decorated=dec(func)), __wrapped__=func)
-
             if self.memory.location is not None and self.level >= level:
-                return decorator_apply(self.memory.cache, function)
+                cached_function = self.memory.cache(function)
 
-            else:
-                return function
+                def inner(*a, **kw):
+                    '''Call function and maybe resize cache.'''
+                    _ = cached_function(*a, **kw)
+                    self._throttled_reduce_size()
+                    return _
+
+                return functools.wraps(function)(inner)
+            return function
         return wrapper
 
     def clear(self, *args, **kwargs):
@@ -73,9 +68,28 @@ class CacheManager(object):
         return self.memory.warn(*args, **kwargs)
 
 
+def throttle(func, interval):
+    '''Don't call a function if it has been called in the last `interval` seconds.'''
+    interval = interval or 0
+    @functools.wraps(func)
+    def inner(*a, _ignore_cache=False, **kw):
+        # rerun function
+        tnow = time.time()
+        if _ignore_cache or (tnow - inner.last_ran) > interval:
+            inner.last_result = func(*a, **kw)
+            inner.last_ran = tnow
+
+        return inner.last_result
+    inner.last_ran = 0
+    inner.last_result = None
+    return inner
+
 # Instantiate the cache from the environment
-cache = CacheManager(os.environ.get('LIBROSA_CACHE_DIR', None),
-                     mmap_mode=os.environ.get('LIBROSA_CACHE_MMAP', None),
-                     compress=os.environ.get('LIBROSA_CACHE_COMPRESS', False),
-                     verbose=int(os.environ.get('LIBROSA_CACHE_VERBOSE', 0)),
-                     level=int(os.environ.get('LIBROSA_CACHE_LEVEL', 10)))
+cache = CacheManager(
+    os.environ.get('LIBROSA_CACHE_DIR', None),
+    bytes_limit=os.environ.get('LIBROSA_CACHE_BYTES_LIMIT', None),
+    cache_resize_interval=os.environ.get('LIBROSA_CACHE_RESIZE_INTERVAL', 5*60),
+    mmap_mode=os.environ.get('LIBROSA_CACHE_MMAP', None),
+    compress=os.environ.get('LIBROSA_CACHE_COMPRESS', False),
+    verbose=int(os.environ.get('LIBROSA_CACHE_VERBOSE', 0)),
+    level=int(os.environ.get('LIBROSA_CACHE_LEVEL', 10)))


lostanlen:
@beasteers did you test this in a multithreaded setting? i'm a bit nervous at the idea of calling `os.environ.pop` and `os.environ.get` according to `time.time()`-based logic. 
Can you explain why nothing would go wrong?

beasteers:
> did you test this in a multithreaded setting?

I haven't, but what I've added is basically the same as doing this manually with joblib:

```python
# LIBROSA_CACHE_RESIZE_INTERVAL=10
import time
import librosa

resize_interval = os.getenv('LIBROSA_CACHE_RESIZE_INTERVAL')

t0 = time.time()
for f in [...]:
    y, sr = librosa.load(f)
    if resize_interval and time.time() - t0 > resize_interval:
        librosa.cache.reduce_size()
        t0 = time.time()
```
The user just doesn't have to track the time themselves. 

>how much is this piece of code exposed to the casual user? will it appear often in the middle of a stack trace?

It shouldn't appear any more than it was previously because it doesn't add any additional wrappers around the actual function call. The only time it would appear in the stack trace would be if you interrupted it (or some other exception occurred) in the `Memory.reduce_size()` call.

>with many nested layers of abstraction, though; so it's worth commenting it in great detail. 

I don't think it's too much more complicated than it was with the `FunctionMaker.create` stuff.

The additions are:
 - wrap `self.reduce_size` with a throttle decorator so that it won't run twice in some time interval
 - call `self.throttled_reduce_size` after every cache call (which will only really run after at minimum the specified time interval has passed).

But I'm happy to add more comments/explanation to the code.

beasteers:
Hi @bmcfee no worries! I kinda forgot about this lol.

1. yes it does. Heres a demo to show that functools wraps copies over the function name and signature.
```python
def cached(func):
    def inner(*a, **kw):
        return func(*a, **kw)
    return functools.wraps(func)(inner)

import functools
import inspect

# vanilla
def func(x, y):
    '''I am a docstring'''

print(func.__name__, inspect.signature(func))

# using functools.wraps
@cached
def func(x, y):
    '''I am a docstring'''
print(func.__name__, inspect.signature(func))
```
Which prints out:
```
func (x, y)
func (x, y)
```
And doing `func?` shows:
```python
Signature: func(x, y)
Docstring: I am a docstring
File:      /opt/Github/sonyc/sonycserver/<ipython-input-14-ff0e52bc63eb>
Type:      function
```

I can revisit this branch and add some tests to specifically check this tho!

2. Not sure I understand/if anything needs to be done for this

3. I'm open to taking another look at this to see if I can simplify the logic at all.

I was about to go into how users probably wouldn't realize that librosa was caching things, but I just went and looked and I guess caching is disabled by default lol 😅 I mean I guess we could provide both options? 

I think the main argument for having cache management be handleable at an environment level rather than a source level is that say one person writes some beautiful librosa code, and then another person goes to import and use that code and they have caching enabled in their `~/.bashrc`, they'd have to go in and edit the code to periodically purge the cache before they can use it (or remember it's in their rc and disable caching). But TBH I feel like caching is like one of the last things that I really want to be thinking about when I use librosa so I would probably forget to say "Oh hey it's probably time to purge my librosa cache" cuz also as a user, I don't think I would really have a sense of how often you'd need to do that

bmcfee:
> I can revisit this branch and add some tests to specifically check this tho!

Nah, that'd be overkill IMO.  I just wanted to make sure we don't lose any functionality here.


>     1. Not sure I understand/if anything needs to be done for this

I think it's fine.



>     2\. I'm open to taking another look at this to see if I can simplify the logic at all.

One thing that might be helpful is to expose the cache_reduce_size function up as a method of the cache object.  That way, users can trigger it directly if they want to.  Then, if the TTL is undefined (or set to 0?), we can bypass the throttling operation and automatic clear.



> I think the main argument for having cache management be handleable at an environment level rather than a source level is that say one person writes some beautiful librosa code, and then another person goes to import and use that code and they have caching enabled in their `~/.bashrc`, they'd have to go in and edit the code to periodically purge the cache before they can use it (or remember it's in their rc and disable caching).

That's part of it.  The main part for me was that it made it much simpler to implement.  Cache configuration needs to happen prior to import, or else we need some centralized object that can keep track of all cached functions in one place (yikes).  Configuring it through the environment eliminated the potential for someone to import librosa first, then configure the cache, and then be surprised when it doesn't work.



> But TBH I feel like caching is like one of the last things that I really want to be thinking about when I use librosa

True!  I left it flexible, but the way I use it most of the time is at a pretty low level (10) to avoid recomputing filterbanks all the time.  It's a bit like caching FFT plans in that way: retain the things that are commonly used and don't depend on the content of the inputs, just their shapes.

OTOH, sometimes I have relatively complex processes that would be too much of a pain to write efficiently.  This was the case with the seymour project, where a script would crawl over a db of thousands of tracks, exporting pretty much every feature and analysis we could do: CQTs, chroma features, beat timings, structure analyses, tempo estimates, etc etc etc.  There was a lot of redundancy in the various functions being called, but I wanted the individual feature calculations to be clean and modular.  The use pattern here was to cache at a high level (30 or 40), and purge after each track.  (Or, more realistically, every N tracks for small N.)

The in-between case is pretty rare in my experience, but maybe someone out there has a workflow where it makes sense.  (Re-analyzing the same track?)

beasteers:
> OTOH, sometimes I have relatively complex processes that would be too much of a pain to write efficiently. This was the case with the seymour project, ...

I see! This explanation was really helpful in understanding the realistic use case that you had defined it for, and I can definitely see how it could benefit from a `cache.reduce_size()` method. I guess my earlier impression was that it was basically a set-it-forget-it flag that you could turn on that sped-up librosa for repeated calls, but I see how strategic cache purging can certainly be beneficial. I guess this is mostly just a safety measure that someone can enable to be certain their disk won't balloon using librosa, even if they are doing some level of cache purging.



> The in-between case is pretty rare in my experience

I believe that! I'll give it another look (maybe this weekend) to see if I can clean this PR up a bit and add a `reduce_size` method. I'll also probably remove the throttle function and simplify that a bit so that it's less code that needs to be run when the feature's disabled. 

What are your preferences on the default values for these?
```python
LIBROSA_CACHE_BYTES_LIMIT=0
LIBROSA_CACHE_RESIZE_INTERVAL=20  # will be disabled if bytes_limit is falsey
```

> or else we need some centralized object that can keep track of all cached functions

not that I'm going to touch this at all, but it would be possible to do it with a simple closure, but you'd just have to check the cache params inside the function each time
```python
cached_function = self.memory.cache(function)
def inner(*a, **kw):
    if self.memory.location is not None and self.level >= level:
        return cached_function(*a, **kw)
    else:
        return function(*a, **kw)
```
(This is under the assumption that memory objects can call cache without a location - otherwise you'd need to create it inside the first if on the first call) - but even ignoring the location part, you could implement a dynamic cache level like this.

bmcfee:
> I see how strategic cache purging can certainly be beneficial.

What would actually be great is purging everything above a fixed level.  That way you could dump any track-dependent cache results, but keep the lower-level stuff like filterbanks.  It sure seems like a pain to implement that though, and not worth the overall effort.


> I guess this is mostly just a safety measure that someone can enable to be certain their disk won't balloon using librosa, even if they are doing some level of cache purging.

Yep exactly!

> I'll give it another look (maybe this weekend) to see if I can clean this PR up a bit and add a `reduce_size` method. I'll also probably remove the throttle function and simplify that a bit so that it's less code that needs to be run when the feature's disabled.

That'd be great.  The merge conflicts right now shouldn't be too bad to resolve.


> What are your preferences on the default values for these?

Bytes seems fine.

I'm thinking a bit more about the resize interval, and am wondering how stateful it is across runtimes.  For example, if you have a script that you fan out over a collection as separate processes (something dask might plausibly do, for example), would they each have separate throttle clocks?  If so, and if each script takes less time to run than the interval, what would happen?




> but you'd just have to check the cache params inside the function each time

Sure - I mostly wanted to avoid additional complexity and branching paths in the wrapper.

beasteers:
> What would actually be great is purging everything above a fixed level.

yeah I agree that'd be difficult, although, mainly you'd just have to manage a memory instance per level (which could be as simple as a defaultdict? idk). If you did that, you could have `librosa.cache.clear()` and `librosa.cache.clear(30)`

> I'm thinking a bit more about the resize interval, and am wondering how stateful it is across runtimes. 

Yeah that is a fair point - it would be managing `x` number of clocks, but I don't know how to mitigate that without introducing file locks and that seems waaaaaay out of the purview. In that case, I am absolutely fine with setting `LIBROSA_CACHE_RESIZE_INTERVAL=0` by default and have that completely disable the automatic resizing unless someone explicitly wants it. That way there won't be unexpected default functionality when someone distributes it.

> If so, and if each script takes less time to run than the interval, what would happen?

it runs reduce_size when it first initializes the object, so even if you set it to clear every 10 minutes and no script ran that long, it would still run on import each time the script runs.

bmcfee:
Oof.  Another eternity later...

I think we should close this PR out, for two reasons.  1) It's way out of sync (35K line changes :grimacing:), and 2) after giving it some thought, I think the most sensible thing is to let the user call `reduce_size` directly.  In joblib, the bytes_limit parameter of the memory object is now deprecated, and users are supposed to supply the limit directly to reduce_size.

I just gave it a shot, and this actually already works out of the box, eg:

```python
>>> librosa.cache.reduce_size("5M")
```

Unless I'm mistaken, this solves the problem for us and we can mark #1139 as wontfix.  @beasteers what do you think?

beasteers:
Yeah totally sounds good! I don't even remember what this was about lol but sounds like it's not an issue anymore

Issue #641, opened 2017-11-06
RFC: support for non-western systems


bmcfee:
#### Description
This issue popped up during poster discussions at ismir2017 (tagging @vrangasayee): librosa's support for non-western (eg, Indian) scales is currently non-existent, but it would be easy to add.

Things that would be useful to have and trivial to implement

- Unit conversion functions: `hz_to_*`, `note_to_*`, etc
- Visualization decorations, `specshow(y_axis='...')`

More fundamentally, it would be nice to have support built in for `chroma`, and if possible, down to `cqt` basis filters.  The latter might take some refactoring, but if it works out, we should also be able to support non-equal-temperament scales.  (Consider these as stretch goals.)

vrangasayee:
This would be really useful for a lot of non-western scales including Indian and other systems. Would be glad to collaborate to make this happen.

bmcfee:
@vrangasayee @stefan-balke great!  I'm a total newbie at this, but is [this wiki page](https://en.wikipedia.org/wiki/Svara) a reasonable starting point?  Is there a better standard reference for notation?

As I mentioned above, the first thing to do is get the unit converters specified.  Do you know if there's a standard notation for svara including octave height and fractional notes?  Like, does `Sa3+35` make sense the way that `C3+35` does?  If so, then the converters should be pretty trivial to code up with `note_to_svara` as the core converter, from which we can derive `hz_to_svara`, `midi_to_svara`, etc.  

After that, the viz ticker decorations are pretty trivial, and then we can worry about chroma-like features, and then extending into (say) Chinese systems.  (Side note: any recommendations for who to consult on implementing those?)

vrangasayee:
That looks reasonable. I have other texts that provide a better coverage
especially of the fractional notes. Let me try to scan and upload those (or
find online references).

vrangasayee:
Here's a authoritative text on the 22 sruthi which also covers the 12 and
16 sruthis as ratios.

stefan-balke:
Hm, I think your attachment got stuck somewhere. Can you share a dropbox link?

vrangasayee:
[srutis.pdf](https://github.com/librosa/librosa/files/1466980/srutis.pdf)
Uploaded to github. This should work.

bmcfee:
Thanks for that!  It's gonna take me a bit to digest, but a few questions immediately come to mind (apologies in advance if I'm being excessively ignorant):

[...]

PR #1150, opened 2020-06-04
Data loader / more example tracks

### LINKED ISSUE #952: RFC: more example tracks?

bmcfee:
#### Reference Issue
Fixes #952 

This PR replaces our single pre-packaged example audio clip by a collection maintained at https://github.com/librosa/librosa-data .

It uses `pooch` to handle remote file access and caching, and provides a couple of helpers to expose the available data to users.

Shortcuts `librosa.example` and `librosa.ex` are provided to make things easy to use.

The old example file is included in the new collection, and can be accessed by `librosa.ex('vibeace')`.  A deprecation shim is in place until 0.9 to preserve the old API.

#### Any other comments?

TBD:

- [x] cache the cache on appveyor
- [x] tests
- [x] improve documentation on how to control the cache location
- [x] audit the documentation to use more appropriate examples for each function
- [x] add more tracks to the mixtape (see #952 ) 
- ~[ ] migrate direct github downloads to an archived zenodo link~


bmcfee:
## How to use this

The old style:
```python
>>> fname = librosa.util.example_audio_file()
>>> y, sr = librosa.load(fname)
```
The new style:
```python
# Already downsampled to 22KHz mono
>>> y, sr = librosa.load(lbrosa.ex('vibeace'))

# Original quality
>>> y, sr = librosa.load(lbrosa.ex('vibeace', hq=True))
```

List the examples (so far):
```python
>>> librosa.util.list_examples()
AVAILABLE EXAMPLES
--------------------------------------------------------------------
brahms    	Brahms - Hungarian Dance #5
choice    	Admiral Bob - Choice (drum+bass)
nutcracker	Tchaikovsky - Dance of the Sugar Plum Fairy
vibeace   	Kevin MacLeod - Vibe Ace
waller    	Fats Waller - Ain't Misbehavin'
```

bmcfee:
I think this is basically feature complete and ready for cr.

Merge should eat until we finalize the track list and set up zenodo storage

bmcfee:
A couple more points of discussion here:

1. Should we set up the data files as a subtree to the `librosa-data` repo, so that they're always in sync (in dev anyway)?
2. Should we have the source url pull from github when `'dev' in librosa.__version__` and zenodo when on a stable release?

bmcfee:
> * Should we set up the data files as a subtree to the `librosa-data` repo, so that they're always in sync (in dev anyway)?
> 

I looked into this, and I think it will be more of a hassle than it's worth.  These data files shouldn't change too often.

> * Should we have the source url pull from github when `'dev' in librosa.__version__` and zenodo when on a stable release?

For this, I think we should use the `packaging` library's tools for [version parsing](https://packaging.pypa.io/en/latest/version/), and propose the following logic:

1. `v.is_prerelease and (v.pre is None)` ==> github   (example: `0.8.0.dev`)
2. `v.is_prerelease and (v.pre is not None)` ==> zenodo  (example: `0.8.0rc0`)
2. `not v.is_prerelease` ==> zenodo  (example: `0.8.0`)

bmcfee:
Relying on zenodo integration seems like a bit more of a hassle than I was hoping.

Instead, I've set up a direct URL for accessing the data at https://librosa.org/data/ ; if we ever want to migrate off github, the link should continue to work.

I think this is :shipit: once CI passes.  Anyone else want to chime in?

bmcfee:
Side note: I think we're good (for now) on example tracks.  The additional content that I'd like to add can be rolled into #641 when the time comes.

lostanlen:
@bmcfee i don't understand the nomenclature of track IDs. Is there a reason why the Hungarian Dance #5 gets labeled by its composer (`brahms`) but the Tchaikovsky piece by its title (`nutcracker`)? The list of tracks is very short, so i can live with the inconsistency; but i'm concerned that it would be worth to document why these tracks were chosen in the librosa "mixtape". Do they reflect a specific genre or instrumentation? Are they fit for a specific MIR task? At the moment, it's impossible to know without listening to all samples.

bmcfee:
> @bmcfee i don't understand the nomenclature of track IDs. Is there a reason why the Hungarian Dance #5 gets labeled by its composer (`brahms`) but the Tchaikovsky piece by its title (`nutcracker`)?

Yeah, because there isn't one :shrug: I was mainly aiming for salience and brevity.



> The list of tracks is very short, so i can live with the inconsistency; but i'm concerned that it would be worth to document why these tracks were chosen in the librosa "mixtape".

A fair point.  I'm not sure where would make sense to document this stuff though.  Maybe a blog post?  Or in the README for the data repo?

lostanlen:
Yes, I think that it would be good to add to the README a one-sentence description of the track, describing its content + the reason why it was added

Another thing which is frustrating is that there isn't a programmatic way to list examples. The `list_examples` function prints the list, but doesn't return it. I have only read the PR superficially, but it seems like the only way to list the example keys is to access a relatively obscure global variable: `librosa.files.__TRACKMAP.keys()`

bmcfee:
> Another thing which is frustrating is that there isn't a programmatic way to list examples. The `list_examples` function prints the list, but doesn't return it.

That's right.  I thought about adding this, but couldn't construct a plausible use-case for it.  The examples here are primarily meant to serve the documentation, and I want to encourage users to be deliberate about choosing example tracks rather than brute-forcing through all of them for some reason.  But maybe you have a different use-case in mind?

bmcfee:
> Yes, I think that it would be good to add to the README a one-sentence description of the track, describing its content + the reason why it was added. Even better if that description can be pulled directly from librosa. "Vibe Ace" and "Choice" aren't informative per se.

Hrm.  Maybe a page in the documentation would be appropriate then.  Off the top of my head, here's how I've been thinking about them:

- vibe ace : backward compatibility
- fishin : good for demonstrating structure / segmentation
- choice : hpss
- brahms : variable tempo (eg for tempogram and plp), soft onsets
- nutcracker : fixed tempo, harmony
- trumpet : monophonic

bmcfee:
> Maybe a page in the documentation would be appropriate then.

Done.

lostanlen:
Wonderful! Many thanks for addressing this so fast. Makes a lot of sense now.

bmcfee:
Ok, docstrings are now up to date for constantq.py, which finishes up the purging of `example_audio_file` from our example code.

Once CI finishes, I think this is good to go.

PR #1157, opened 2020-06-09
Fix hopsize issue for iirt

fzalkow:
We found a severe bug in the iirt function.

Three different sample rates are used to compute the bands for iirt. For each sample rate, a distinct hop size is used to compute the short-time mean-square power (STMSP). Without this bug fix, the hop sizes were rounded and then applied, which is problematic. The rounding error leads to increasing shifts that are different for each of the three sample rates. As a consequence, for longer waveforms, the three frequency bands are out of sync. We solved this by computing the start indices for the windows before rounding.

The following code example illustrates the issue:

    import numpy as np
    import librosa

    Fs = 22050
    x = librosa.clicks([10, 50, 90, 130, 170], sr=Fs, length=180*Fs)
    X = librosa.iirt(x)

    plt.figure(figsize=(10, 4))
    plt.imshow(np.log(X + 1e-7), aspect='auto', origin='lower', cmap='gray_r', extent=[0, 180, 24, 108])
    plt.colorbar()
    plt.xlabel('Time (seconds)')
    plt.ylabel('Pitch')

The result without the bug fix:
old

The result with the bug fix:
new

Thanks to @ahnonay, who found the issue! Also, ping to @stefan-balke who might be interested.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1321,9 +1321,6 @@ def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
     bands_power = []
 
     for cur_sr, cur_filter in zip(sample_rates, filterbank_ct):
-        factor = float(sr) / float(cur_sr)
-        win_length_STMSP = int(np.round(win_length / factor))
-        hop_length_STMSP = int(np.round(hop_length / factor))
 
         # filter the signal
         cur_sr_idx = np.flatnonzero(y_srs == cur_sr)[0]
@@ -1335,12 +1332,22 @@ def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
             cur_filter_output = scipy.signal.sosfiltfilt(cur_filter,
                                                          y_resampled[cur_sr_idx])
 
-        # frame the current filter output
-        cur_frames = util.frame(np.ascontiguousarray(cur_filter_output),
-                                frame_length=win_length_STMSP,
-                                hop_length=hop_length_STMSP)
+        factor = float(sr) / float(cur_sr)
+        hop_length_STMSP = hop_length / factor
+        win_length_STMSP_round = int(round(win_length / factor))
 
-        bands_power.append(factor * np.sum(cur_frames**2, axis=0)[:n_frames])
+        start_idx = np.arange(0, len(cur_filter_output)-win_length_STMSP_round, hop_length_STMSP)
+        if len(start_idx) < n_frames:
+            min_length = int(np.ceil(n_frames * hop_length_STMSP)) + win_length_STMSP_round
+            cur_filter_output = util.fix_length(cur_filter_output, min_length)
+            start_idx = np.arange(0, len(cur_filter_output)-win_length_STMSP_round, hop_length_STMSP)
+
+        start_idx = np.round(start_idx).astype(int)[:n_frames]
+        idx = np.tile(np.arange(win_length_STMSP_round),
+                      (len(start_idx), 1)) + start_idx[:, np.newaxis]
+
+        cur_band_power = factor * np.sum(cur_filter_output[idx] ** 2, axis=-1)
+        bands_power.append(cur_band_power)
 
     return np.asfortranarray(bands_power)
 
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks for catching this!  The fix makes sense to me, but could you add a unit test for it?  Could be as simple as looking for peaks in the output when provided a delayed impulse input.

librosa/core/spectrum.py
```diff
@@ -1335,12 +1332,22 @@ def iirt(y, sr=22050, win_length=2048, hop_length=None, center=True,
             cur_filter_output = scipy.signal.sosfiltfilt(cur_filter,
                                                          y_resampled[cur_sr_idx])
 
-        # frame the current filter output
-        cur_frames = util.frame(np.ascontiguousarray(cur_filter_output),
-                                frame_length=win_length_STMSP,
-                                hop_length=hop_length_STMSP)
+        factor = float(sr) / float(cur_sr)
+        hop_length_STMSP = hop_length / factor
+        win_length_STMSP_round = int(round(win_length / factor))
```
bmcfee, line 1337: Can these be written with integer division (`//`)?

The float casts for `factor` calculation are no longer necessary, since we don't support python2.

stefan-balke:
Another example of error propagation...damn! Thanks for catching this! Interesting that we were very close to the reference implementation though.

Are we getting closer now or was this bug also in the MATLAB implementation?

librosa/core/spectrum.py
fzalkow: I removed the float casts. I don't see an obvious application of integer division without changing the behavior. `win_length_STMSP_round` uses rounding, so we could convert it like `(win_length + factor // 2) // factor` (see [SO](https://stackoverflow.com/questions/3950372/round-with-integer-division)), but that is less readable IMO.

fzalkow:
I implemented a test. The downside is that it is quite slow to compute (for me, nearly a minute). But to really catch this issue, we have to compute iirt for a longer waveform, and that takes some time.

@stefan-balke: This bug did not exist in the Matlab implementation... I think we only thought that we had been close to the reference implementation because we only tested for short audio files, where the effect of the shift was minor...

bmcfee:
> The downside is that it is quite slow to compute (for me, nearly a minute).

Yikes!  Can't we bring that down by, eg, reducing the sampling rate?

librosa/core/spectrum.py
bmcfee: Yeah, that's fine for win length, but hop length should be integer division.

The current implementation hacks around this by casting the arange to int before doing anything with it, but that seems like it could introduce unpredictable bugs.  Better to make sure we have the right types up front.

fzalkow:
I strongly reduced the sample rate, and now the test lasts below two seconds. (I needed to change the filterbank parameters for that purpose. So, it is not a "standard" application of the function anymore, but it may still serve for test purposes.)

bmcfee:
> So, it is not a "standard" application of the function anymore, but it may still serve for test purposes

Excellent!  Confirmed locally that it fails on the main branch (before the fix is applied).

librosa/core/spectrum.py
fzalkow: The idea of the fix is that the hop size is passed as non-integer to arange. Doing so, we avoid the accumulated rounding errors. The discretization only happens at the very end, when indices are used. Why do you see the introduction of unpredictable bugs?

bmcfee: > The idea of the fix is that the hop size is passed as non-integer to arange.

Perhaps I misunderstood then.



> The discretization only happens at the very end, when indices are used. Why do you see the introduction of unpredictable bugs?

Force-casting after the fact can hide mistakes that we could otherwise catch, so as a general rule, I prefer to make the calculations explicit up front to avoid silent errors.  But if this behavior is intentional and part of the fix, I'm fine with it.  (Maybe add a line of documentation explaining what's going on though.)

fzalkow:
Sorry, one more commit. I was too hasty with throwing out castings. Floor division can yield a float when the arguments are float. Therefore, we still need to cast `n_frames` to int.

bmcfee:
Anything left on this one @fzalkow ?

fzalkow:
Nope, can be merged from my point of view. Thanks!

bmcfee:
Awesome, thanks!

Issue #1025, opened 2019-11-28
de-emphasis (↔ preemphasis)


keunwoochoi:
I feel like why not when we have `preemphasis`. A use case would be when people use librosa as a preprocessor of a system (speech analysis/enhancement for example) where they'd need to recover the preemphasis-ed input signals. Any thoughts?

bmcfee:
Yeah, this seems doable.

The preemph filter uses `b=[1, -coef], a = [1]`, which gives zeros, poles, and gain `z=[coef], p=[], k=1`.  The inverse filter would exchange poles and zeros, putting a pole at `coef` and producing coefficients `b'=a, a'=b`.  The inverse filter then has a pole at `0 < coef < 1` (for nontrivial filters) which is stable.

I tested this out and it seems to work well enough for default settings, at least after the filter warms up for a bit.  The hiccup here is getting the initial conditions right.

keunwoochoi:
> The hiccup here is getting the initial conditions right.

What do you mean by initial conditions?

bmcfee:
> What do you mean by initial conditions?

Pre-emphasis is a weighted first-order difference, which is undefined at the first input sample.  Our implementation uses `lfilter_zi` to initialize the filter to steady-state, which would need to be inverted to recover the initial `x[0]` (or something close to it).

keunwoochoi:
👍 thanks for the update and the answers!

bmcfee:
Reopening this issue because we haven't implemented it yet.

PR #1159, opened 2020-06-11
Remove pre-emphasis glitch

### LINKED ISSUE #1032: Preemphasis generates a glitch in the start of the output signal
The current implementation of the preemphasis generates a glitch in the output signal.

Steps to reproduce:
```python
a = np.array([9.12e-05, 9.15e-05, 9.13e-05, 6.10e-05, 3.05e-05, 3.07e-05, 6.10e-05])
b = librosa.effects.preemphasis(a)
print(b)

Out: 
array([-9.699088e-01,  3.036000e-06,  2.545000e-06, -2.756100e-05,
       -2.867000e-05,  1.115000e-06,  3.122100e-05])
```
Here `b[0]` is clearly off.

What can be done is to asign `b[1]` to `b[0]`. This hack would deal with the glitch while keeping the lengths of `b` and `a` the same.

bmcfee:
#### Reference Issue
Fixes #1032 

This PR implements the linear extrapolation initialization for preemphasis.

As a bonus, scalar inputs are now allowed for `zi` and they should be automatically upcast to match the filter shape.

### DIFF
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -594,7 +594,10 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
         At `coef=1`, the result is the first-order difference of the signal.
 
     zi : number
-        Initial filter state
+        Initial filter state.  When making successive calls to non-overlapping
+        frames, this can be set to the `zf` returned from the previous call.
+
+        By default `zi` is initialized as `2*y[0] - y[1]`.
 
     return_zf : boolean
         If `True`, return the final filter state.
@@ -642,7 +645,10 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
     a = np.asarray([1.0], dtype=y.dtype)
 
     if zi is None:
-        zi = scipy.signal.lfilter_zi(b, a)
+        # Initialize the filter to implement linear extrapolation
+        zi = 2*y[..., 0] - y[..., 1]
+
+    zi = np.atleast_1d(zi)
 
     y_out, z_f = scipy.signal.lfilter(b, a, y,
                                       zi=np.asarray(zi, dtype=y.dtype))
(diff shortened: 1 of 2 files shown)


librosa/effects.py
```diff
@@ -642,7 +645,10 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
     a = np.asarray([1.0], dtype=y.dtype)
 
     if zi is None:
-        zi = scipy.signal.lfilter_zi(b, a)
+        # Initialize the filter to implement linear extrapolation
+        zi = 2*y[..., 0] - y[..., 1]
```
lostanlen, line 649: I saw this line of code in a dream last night 😴
Weird in a beautiful way

lostanlen:
LGTM. I think that this linear extrapolation is neat. If the input has both high amplitude and high frequency, `zi` might be outside the expected range of audio; but I don't think that this will matter in practice. It's far more important to preserve the first derivative.

On a somewhat related note, it might be good to say that the default value of `zi=0.97` is for consistency with RASTAMAT's MFCC, which itself is consistent with HTK's MFCC.
Link to: http://www.ee.columbia.edu/~dpwe/resources/matlab/rastamat/mfccs.html ?

bmcfee:
Do you think we should also roll in #1025 here?  I have a notebook implementation where it works fine after a bit of warmup, but getting the initial conditions right is tricky.

lostanlen:
> Do you think we should also roll in #1025 here? I have a notebook implementation where it works fine after a bit of warmup, but getting the initial conditions right is tricky.

No, I don't see it as an urgent need. Pre-emphasis is a default in the HTK pipeline of MFCC, but is no longer customary in SOTA pipelines for time--frequency analysis; and for audio synthesis, even less so. The state of the debate on #1025 is "sure, why not", so that doesn't strike me as release-blocking.

bmcfee:
Ok, then I'll consider this one as :shipit: once CI finishes.

PR #625, opened 2017-09-22
Added numba strong requirement

bmcfee:
#### Reference Issue
Fixes #624 

#### What does this implement/fix? Explain your changes.

This PR updates the resampy requirement to 0.2, and replaces `optional_jit` by a direct import of `numba.jit`.

#### Any other comments?

The numba-free test environments have been removed.

`optional_jit` should be properly deprecated and removed in the 0.7 series.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/625)

PR #1164, opened 2020-06-15
Use Numba API location for jit import.

stuartarchibald:
#### Reference Issue

xref: https://github.com/numba/numba/issues/5865. Hopefully the change proposed herein would make it so that  changes in Numba's internals would not impact `librosa`.

Imports Numba's `jit` decorator from the Numba API location opposed to Numba's internals.

#### Any other comments?

There appears to be no use of `optional_jit` in the code base? Perhaps consider for deprecation/removal?

### DIFF
--- a/librosa/util/decorators.py
+++ b/librosa/util/decorators.py
@@ -4,13 +4,8 @@
 '''Helpful tools for deprecation'''
 
 import warnings
-from packaging import version
 from decorator import decorator
-import numba
-if version.parse(numba.__version__) < version.parse('0.49.0'):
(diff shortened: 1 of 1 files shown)


bmcfee:
Thanks for this PR.  Indeed, the `optional_jit` shim is a holdover and probably can be removed safely at this point.

Looking into the git history, I'm not sure why we ever used the internal API to begin with. :shrug:

bmcfee:
LOL https://github.com/librosa/librosa/pull/625 even says that `optional_jit` should have been deprecated and purged in 0.7, I guess we missed it!

PR #1166, opened 2020-06-16
return D_steps for backtracking by user

fzalkow:
I want to propose a new feature for DTW, which is helpful for subsequence DTW. Currently, only the optimal warping path is returned by the DTW function. However, it can also be useful to compute more than one path. For example, in the case of audio matching, one may be interested in multiple matching subsequences. For that purpose, one has to perform backtracking not only from the global minimum of the matching function but from several local minima.

To achieve the feature, ...
* ... the `dtw` function now has an optional keyword `return_D_steps` for returning the step size matrix.
* ... I wrote a function `dtw_backtracking` that is a wrapper for `__dtw_backtracking` to handle the step sizes.
* ... the `__dtw_backtracking` function now gets an argument `start_m`, which is the start position for backtracking.

The following code example illustrates the new feature:
```python
import numpy as np
import librosa
from scipy.spatial.distance import cdist

dim = 24
X_len = 10

X = np.random.random((dim, X_len))
Y = np.random.random((dim, 100))

Y[:, 10:10+X_len] = X
Y[:, 80:80+X_len] = X

C = cdist(X.T, Y.T, metric='cosine')
D, D_steps = librosa.sequence.dtw(C=C, backtrack=False, subseq=True, return_D_steps=True)

matching_curve = D[-1, :]
idx_sort = np.argsort(matching_curve)

wp1 = librosa.sequence.dtw_backtracking(D_steps, subseq=True, start_m=idx_sort[0])
wp2 = librosa.sequence.dtw_backtracking(D_steps, subseq=True, start_m=idx_sort[1])

plt.figure(figsize=(10, 4))
plt.imshow(C, aspect='auto', origin='lower', cmap='gray')
plt.ylabel('Sequence X')
plt.xlabel('Sequence Y')
plt.plot(wp1[:, 1], wp1[:, 0], 'r:o', label='Warping Path 1')
plt.plot(wp2[:, 1], wp2[:, 0], 'b:o', label='Warping Path 2')
plt.legend()
plt.colorbar()
```
dtw

### DIFF
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -107,17 +107,17 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
 
     Returns
     -------
-    D : np.ndarray [shape=(N,M)]
+    D : np.ndarray [shape=(N, M)]
         accumulated cost matrix.
-        D[N,M] is the total alignment cost.
+        D[N, M] is the total alignment cost.
         When doing subsequence DTW, D[N,:] indicates a matching function.
 
-    wp : np.ndarray [shape=(N,2)]
+    wp : np.ndarray [shape=(N, 2)]
         Warping path with index pairs.
-        Each row of the array contains an index pair n,m).
+        Each row of the array contains an index pair (n, m).
         Only returned when ``backtrack`` is True.
 
-    steps : np.ndarray [shape=(N,M)]
+    steps : np.ndarray [shape=(N, M)]
         Step matrix, containing the indices of the used steps from the cost
         accumulation step.
         Only returned when ``return_steps`` is True.
@@ -351,12 +351,12 @@ def __dtw_calc_accu_cost(C, D, steps, step_sizes_sigma,
 
     Returns
     -------
-    D : np.ndarray [shape=(N,M)]
+    D : np.ndarray [shape=(N, M)]
         accumulated cost matrix.
-        D[N,M] is the total alignment cost.
+        D[N, M] is the total alignment cost.
         When doing subsequence DTW, D[N,:] indicates a matching function.
 
-    steps : np.ndarray [shape=(N,M)]
+    steps : np.ndarray [shape=(N, M)]
         Step matrix, containing the indices of the used steps from the cost
         accumulation step.
 
@@ -414,7 +414,7 @@ def __dtw_backtracking(steps, step_sizes_sigma, subseq, start=None):  # pragma: 
     wp : list [shape=(N,)]
         Warping path with index pairs.
         Each list entry contains an index pair
-        (n,m) as a tuple
+        (n, m) as a tuple
 
     See Also
     --------
@@ -426,7 +426,7 @@ def __dtw_backtracking(steps, step_sizes_sigma, subseq, start=None):  # pragma: 
         cur_idx = (steps.shape[0] - 1, start)
 
     wp = []
-    # Set starting point D(N,M) and append it to the path
+    # Set starting point D(N, M) and append it to the path
     wp.append((cur_idx[0], cur_idx[1]))
 
     # Loop backwards.
@@ -478,7 +478,7 @@ def dtw_backtracking(steps, step_sizes_sigma=None, subseq=False, start=None):
     wp : list [shape=(N,)]
         Warping path with index pairs.
         Each list entry contains an index pair
-        (n,m) as a tuple
+        (n, m) as a tuple
 
     See Also
     --------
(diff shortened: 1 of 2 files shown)


librosa/sequence.py
```diff
@@ -112,6 +116,10 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
         Each row of the array contains an index pair n,m).
         Only returned when ``backtrack`` is True.
 
+    D_steps : np.ndarray [shape=(N,M)]
+        Step matrix, containg the indices of the used steps
+        Only returned when ``return_D_steps`` is True.
+
```
bmcfee, line 122: is there any reason not to call this just `steps`?

librosa/sequence.py
```diff
@@ -100,6 +101,9 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
         The Sakoe-Chiba band radius (1/2 of the width) will be
         ``int(radius*min(C.shape))``.
 
+    return_D_steps : binary
```
bmcfee, line 104: binary --> bool; docstring should explain what `D_steps` means.

librosa/sequence.py
```diff
@@ -367,7 +383,7 @@ def __dtw_calc_accu_cost(C, D, D_steps, step_sizes_sigma,
 
 
 @jit(nopython=True, cache=True)
-def __dtw_backtracking(D_steps, step_sizes_sigma, subseq):  # pragma: no cover
+def __dtw_backtracking(D_steps, step_sizes_sigma, subseq, start_m=None):  # pragma: no cover
```
bmcfee, line 386: why `start_m` and not just `start`?

librosa/sequence.py
```diff
@@ -397,9 +416,14 @@ def __dtw_backtracking(D_steps, step_sizes_sigma, subseq):  # pragma: no cover
     --------
     dtw
     '''
+    if start_m is None:
+        cur_idx = (D_steps.shape[0] - 1, D_steps.shape[1] - 1)
+    else:
+        assert subseq
```
bmcfee, line 422: assertions should only be used in tests, not library code.  If there's something wrong with the input value, raise an exception with an informative error message.

librosa/sequence.py
```diff
@@ -424,6 +448,54 @@ def __dtw_backtracking(D_steps, step_sizes_sigma, subseq):  # pragma: no cover
     return wp
 
 
+def dtw_backtracking(D_steps, step_sizes_sigma=None, subseq=False, start_m=None):
+    '''Backtrack optimal warping path.
+
+    Wrapper for ``__dtw_backtracking``.
```
bmcfee, line 454: External users should not need to know about `__dtw_backtracking`.

librosa/sequence.py
```diff
@@ -424,6 +448,54 @@ def __dtw_backtracking(D_steps, step_sizes_sigma, subseq):  # pragma: no cover
...
+        Saved indices of the used steps used in the calculation of D.
+
+    step_sizes_sigma : np.ndarray [shape=[n, 2]]
+        Specifies allowed step sizes as used by the dtw.
+
+    subseq : binary
+        Enable subsequence DTW, e.g., for retrieval tasks.
+
+    start_m : int
+        Start column index for backtraing (only allowed for ``subseq=True``)
```
bmcfee, line 472: same comments about variable naming, types, and descriptions from above.

librosa/sequence.py
fzalkow: The variable was named `D_steps` before, so I just kept the name. In our group, we usually call it `E` (in order of computation: cost matrix `C`, accumulated cost matrix `D`, step size matrix `E`). However, I can also rename it to `steps` if you prefer this.

librosa/sequence.py
fzalkow: My idea was that this is the start index for the column (instead of `M-1`). The start index for the row is still `N-1`. But I agree that this idea is not obvious and it may be better to just call it `start`.

fzalkow:
Thanks for the feedback. I adjusted everything, but the name for the step size matrix (see above). I can use `D_steps`, `steps`, or `E`, whatever you prefer...

librosa/sequence.py
```diff
@@ -87,19 +88,23 @@ def dtw(X=None, Y=None, C=None, metric='euclidean', step_sizes_sigma=None,
...
+    global_constraints : bool
         Applies global constraints to the cost matrix ``C`` (Sakoe-Chiba band).
 
     band_rad : float
         The Sakoe-Chiba band radius (1/2 of the width) will be
         ``int(radius*min(C.shape))``.
 
+    return_D_steps : bool
+        If true, the functions returns ``D_steps``, the step matrix,
+        containg the indices of the used steps from the cost accumulation step.
```
bmcfee, line 106: typo

librosa/sequence.py
bmcfee: I think `steps` is the clearest option here.  Single-letter names should be generally avoided.

bmcfee:
> I adjusted everything, but the name for the step size matrix (see above). I can use `D_steps`, `steps`, or `E`, whatever you prefer...

Thanks for the quick turnaround!

bmcfee:
Anything else to add here, or is this :shipit: ?

fzalkow:
Can be merged, from my point of view!

Issue #1074, opened 2020-02-20
Numpy FFT type promotion -> complex128


timlod:
#### Description
By default, librosa uses the numpy fft implementation (numpy.fft) when calling librosa.get_fftlib().
That implementation will always promote input from float32 such that the output is complex128 (2 64bit floats to make the complex number) [(Numpy fft type promotion)](https://numpy.org/devdocs/reference/routines.fft.html#type-promotion).
This can silently double memory consumption in pipelines that use methods based on fft.

librosa has `scipy` as a requirement, so it would be possible to default to `scipy.fft` instead, which respects the input types.

Currently this can be achieved by doing this before running any memory-constrained code:
```
import scipy.fft as scipyfft
librosa.set_fftlib(scipyfft)
```

#### Steps/Code to Reproduce

```
import librosa

S = np.float32(np.random.randn(1025))
S_cqt = librosa.cqt(S) # (uses fft internally)
print(S_cqt.dtype) # np.complex128
```

#### Expected Results
`print(S_cqt.dtype) # np.complex64`

#### Actual Results
`print(S_cqt.dtype) # np.complex128`

#### Versions
Librosa 0.7.2

bmcfee:
This is a fair point, thanks for raising it.

If I understand correctly, your suggestion is to make fftpack the default again?  (It was previously, but we reverted to numpy's implementation after doing some speed benchmarking.)

I'm not opposed to this, but it does warrant some discussion.

timlod:
I've not been using librosa extensively, so I don't know all the arguments for or against using either - but I did notice this particular issue and wanted to raise it, as Python carries on quietly without users realising certain behaviours. If you value consistency in types/memory consumption over some possible tradeoffs with regards to execution time, I would suggest it. In any case, I would add the fact that anything will be cast to 64bit resolution when using numpy's fft methods.

lostanlen:
Thank you for raising this, @timlod.
Related: #825 (closed by #832)

bmcfee:
Gave this a think, and here are some ideas:

1. STFT already has a `dtype` parameter to force expected behavior, regardless of the underlying FFT implementation.  (It could be smarter though, inferring the dtype from the precision of the input.)
2. OP's example code gets type promotion through CQT, not STFT.  This could be fixed by adding a similar `dtype=` parameter to CQT methods, and I'd support that in general.
[...]

PR #1171, opened 2020-06-19
[CR] Dtype inference and control in CQT

### LINKED ISSUE #1143: More intelligent dtype inference
Instigated from #1074 

Many of our functions (eg `stft`) have a `dtype=` parameter with a coded default value (eg `complex64`).

We could probably do a better job of adapting to input precision on the fly, without losing any expected behavior.  EG, if `stft` is given a `float32` input, its output should be `complex64`; if the input is `float64`, it could automatically shift up to `complex128` (unless overridden by the user).

Implementing this would probably require some helper methods to infer the correct precision for converting between real and complex, but I see no downside to doing this.

### LINKED ISSUE #1144: allow specified dtype in CQT
Also instigated from #1074 

The STFT function allows the user to specify a `dtype=` for the output; we could do the same in CQT. (And related methods: pseudo, hybrid, vqt, inverse, etc.)

bmcfee:
#### Reference Issue
- Fixes #1143 
- Fixes #1144 
- Addresses #1074 

This PR addresses dtype control in spectrogram calculations in a few steps:

1. Helper functions are added to `util` to infer minimal dtypes when converting between real and complex
2. `sparsify_rows` API has expanded to allow explicit dtype control (this supports CQT basis construction)
3. all CQT-related methods now have a `dtype=` parameter
4. all STFT-related methods (including GL) do automatic dtype inference

#### Any other comments?

Committing this now to get the CI on the way.

As mentioned in #1144, I'd also like to revamp the `__trim_stack` helper in cqt to be more memory efficient.  Rolling that into this PR because this is broadly concerned with memory efficiency and copying.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -954,7 +954,7 @@ def __trim_stack(cqt_resp, n_bins, dtype):
     # Copy per-octave data into output array
     n = 0
     for c_i in cqt_resp[::-1]:
-        cqt_out[n:min(n + c_i.shape[0], n_bins)] = c_i[:min(c_i.shape[0], n_bins - n)]
+        cqt_out[n:min(n + c_i.shape[0], n_bins)] = c_i[:min(c_i.shape[0], n_bins - n), :max_col]
         n += c_i.shape[0]
 
     return cqt_out
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -330,7 +330,8 @@ def hybrid_cqt(y, sr=22050, hop_length=512, fmin=None, n_bins=84,
                                    res_type=res_type,
                                    dtype=dtype)))
 
-    return __trim_stack(cqt_resp, n_bins, dtype)
+    # Propagate dtype from the last component
+    return __trim_stack(cqt_resp, n_bins, cqt_resp[-1].dtype)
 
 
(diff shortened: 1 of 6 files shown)


bmcfee:
trim_stack rewrite is done, and this is ready for CR.

I wound up not using numba acceleration because there wasn't much benefit, and the complexity of handling typed lists (list of ndarrays) didn't seem worth it.

I get about a 10ms improvement on nutcracker-HQ with 36bpo, but the bigger savings are in memory usage because an entire set of copies of the output are no longer needed.

bmcfee:
Ok, I think this one's good to go.

Quick benchmark for the new trim_stack:
**Before**
```python
In [3]: y, sr = librosa.load(librosa.ex('nutcracker', hq=True), sr=None)                      

In [4]: %timeit librosa.cqt(y=y, sr=sr, bins_per_octave=36, n_bins=216)                       
2.59 s ± 5.87 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
**After**
```python
In [5]: %timeit librosa.cqt(y=y, sr=sr, bins_per_octave=36, n_bins=216)                       
2.33 s ± 5.07 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Issue #1173, opened 2020-06-22
Mel filterbank


Thundor404:
**Describe the bug**
When building a Mel filter bank via librosa.filters.mel, **the results show that the first few banks have no equal maximum magnitude.** The problem is not conformant to general Mel design although it may have few influences.

**To Reproduce**
`import librosa
import librosa.display
import librosa.filters
import matplotlib.pyplot as plt
import numpy as np

sr = 22050
fftSize = 1024;

melFb = librosa.filters.mel(sr,fftSize ,20, norm=None)
melFb = melFb/np.max(melFb);
plt.subplot(3,1,3)

freq = np.linspace(0, sr, int(fftSize/2+1))
plt.plot(freq, melFb.T)
plt.ylabel('Mel filter')
plt.title('Mel filter bank')`

**Expected behavior**
My own design flow, convert Hz to Mel, divide Mel frequency, and then convert back to the normal frequency and build windows according to the aforementioned results. It should have the same magnitude for all filters. See in screenshots.

**Screenshots**
Mel filter bank via librosa
![Figure 2020-06-22 164005](https://user-images.githubusercontent.com/41377351/85269654-c6256180-b4aa-11ea-99b2-ba79d5a0e813.png)

normal Mel filter bank 
![mel](https://user-images.githubusercontent.com/41377351/85267877-5ca45380-b4a8-11ea-9ca3-edd19390a0c2.png)

**Software versions***
<!--
librosa: 0.7.2

audioread: 2.1.8
numpy: 1.18.1
scipy: 1.4.1
sklearn: 0.22.1
joblib: 0.15.1
decorator: 4.4.2
six: 1.15.0
soundfile: 0.10.3
resampy: 0.2.2
numba: 0.49.1

**Additional context**

bmcfee:
Yes, this is correct and intended behavior with the default normalization settings (Slaney-style).

See the discussion in #1030 and the related pr #1050 .  In 0.8, you'll have the ability to set `norm=np.inf` to get peak (l-infinity) normalized filters.

Tagging this issue for 0.8 to remind us to handle the corresponding deprecation.

PR #1175, opened 2020-06-22
finished deprecation of old mel normalization behaviors. #1050

bmcfee:
#### Reference Issue
#1173 #1030 #1050 

This PR resolves the deprecation cycle for Mel normalization behavior proposed in #1050 .

`norm=1` and `norm=np.inf` are now consistent with the rest of the library.

#### Any other comments?

I think this is good to go.


bmcfee:
no CR needed on this IMO, will merge after CI passes.

lostanlen:
tagging @gerazov and @onkyo14taro

bmcfee:
One last comment on this: I removed two regression test fixtures for mel normalization using the old parameter interpretations, since they are no longer relevant.

PR #1186, opened 2020-06-25
cmap returns binary colormaps for boolean data

### LINKED ISSUE #1185: display.cmap should return binary colormaps for binary data
`cmap()` infers an appropriate colormap based on the characteristics of the data.
If the `dtype` is boolean, it returns a `gray_r` colormap.
The colormap returned has 256 values by default, so colorbars show up as continuous.
However, we know that the data will be binary, so it should instead fix the number of values to 2 and provide a discrete map.

https://github.com/librosa/librosa/blob/6f609e6b664d5a17f6a5a60e438c46318957975a/librosa/display.py#L343

should be modified to call `get_cmap(cmap_bool, lut=2)`.

bmcfee:
#### Reference Issue
Fixes #1185 

This PR sets `lut=2` for getting the colormap of boolean data.

#### Any other comments?

The result of this can be seen with the following code snippet:
```python
import librosa
import librosa.display
c = librosa.feature.chroma_cqt(y=y, sr=sr)
R = librosa.segment.recurrence_matrix(c)
librosa.display.specshow(R)
plt.colorbar()
plt.show()
```
Before this change, produces:
image

with the proposed change, produces:
image

(in both cases, the colorbar ticks are out of control of cmap, and not affected by this PR).

As an aside, I also merged the quantile checks here, so the function should be ~2x as fast on non-binary data.

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -340,7 +340,7 @@ def cmap(data, robust=True, cmap_seq='magma', cmap_bool='gray_r', cmap_div='cool
     data = np.atleast_1d(data)
 
     if data.dtype == 'bool':
-        return get_cmap(cmap_bool)
+        return get_cmap(cmap_bool, lut=2)
 
     data = data[np.isfinite(data)]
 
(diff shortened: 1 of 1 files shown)


bmcfee:
No CR needed on this I think; will merge when CI passes.

Issue #1069, opened 2020-02-19
Standardize exception types


bmcfee:
#### Description
In implementing #1064 , I've noticed that our exception typing isn't totally consistent.  In some places, we throw `ValueError`, but in most other places, we throw `ParameterError`.

I've fixed this in a couple of places, but we should have a thorough audit and separate issue for this.

The rationale behind using our own exception class is to make it easy to isolate problems generated by violating the librosa API (which raises our own exception types) from those that occur deeper in the stack, eg within numpy or sklearn, or higher in the stack in case someone builds applications on top of us.

PR #1193, opened 2020-06-28
Audit for 0.8 release

### LINKED ISSUE #1024: 0.8 documentation audit
#### Description

The docstring examples use a mix of stateful and object-oriented API, which is not great style.  At some point, we should audit these for consistency and adhere to best practices and recommendations from the matplotlib developers.

This will be a huge and tedious undertaking, similar in scope to #783 , so I'm not attaching it to a specific major version yet.

bmcfee:
#### Reference Issue
Fixes #1024 

This PR will implement the final documentation audit for 0.8 release.

Aside from the usual stuff (versioning, release notes), we also have the following checklist:

- [x] Which versions get built
- [x] Logo rendering seems off for mobile devices, probably needs some svg editing
- [x] enforce numba<0.5 for historical builds, or do some monkey-patching at build time
- ~~[ ] banners for outdated versions~~
- [x] theme colors on mobile devices
- [x] exception audit #1069 
- [x] code style updates #1070

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -58,6 +58,12 @@ Contributors
 * Stef van der Struijk <https://github.com/NumesSanguis>
 * Shunsuke Hidaka <https://github.com/onkyo14taro>
 * Eric Semeniuc <https://github.com/esemeniuc>
+* Frank Cwitkowitz <https://github.com/cwitkowitz>
+* Bea Steers <https://github.com/beasteers>
+* beantowel <https://github.com/beantowel>
+* Thor Whalen <https://github.com/thorwhalen>
+* Stefan Sullivan <https://github.com/stefansjs>
+
 
 Some feature extraction code was based on <https://github.com/ronw/frontend> by Ron Weiss.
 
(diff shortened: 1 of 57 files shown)


bmcfee:
Some more comments here:

- Multi-version doc builds are pinned to sphinx-multiversion==0.2.3.  0.2.4 introduces some changes that may or may not be compatible with our requirements for historical example generation; still waiting on some resolution from the smv devs.
- Continuing the prior comments about which old versions to build, i think a 2-year window (current + last two majors) is fine.  Coincidentally, this maps onto what still "works" with the current dependency stack; 0.5 fails from joblib issues.  Eventually we will need to figure out a monkeypatch for the numba problem, but i'll punt on that while smv develops.
- Using aggressive caching (level=40 or higher), the entire doc build is now down to under an hour on my dev machine.  This might be close enough to consider doing CI or at least daily/weekly doc builds.
- A single version build with minimal caching (level=10) is pretty fast now: 406 seconds on my dev machine.

There are still a few build errors/warnings to chase down in the current documentation, but it's in much better shape now.

bmcfee:
Putting version warning banners in the docs might require hacking/forking the theme, smv, or both.

I'd prefer to avoid all of those things, so I guess the question is how much we really need warnings if someone's looking at an old version of the docs.

bmcfee:
I'm also rolling the cosmetic updates #1069 and #1070 into this.

bmcfee:
This is all good to go IMO; some of the upstream issues for sphinx-multiversion are out of our control right now.

I'll merge this when CI passes, and do the release tomrorow (2020-07-22).

PR #1211, opened 2020-07-22
Update changelog.rst

pcampr:
wrong link for librosa.yin in changelog.rst


bmcfee:
Ah, how embarrassing!  Thanks for catching it -- I guess copypasting and regexing issue numbers to make the changelog isn't the most robust of processes.

Issue #1214, opened 2020-08-01
Error message wording when n_fft is larger than length of the input


bshall:
***BEFORE POSTING A BUG REPORT*** Please look through [existing issues (both open and closed)](https://github.com/librosa/librosa/issues?q=is%3Aissue) to see if it's already been reported or fixed!

**Describe the bug**
The error/warning message when `n_fft` is larger than the length of the input seems to be the wrong way around. Currently, I get
`n_fft=2048 is too small for input signal of length=512`. Shouldn't it be `n_fft=2048 is too *large* for input signal of length=512`?

**To Reproduce**
```
import librosa
y = librosa.tone(220, length=512)
librosa.stft(y, center=False)
```

**Expected behavior**

**Screenshots**

**Software versions***
INSTALLED VERSIONS
------------------
python: 3.8.3 (default, Jul  2 2020, 16:21:59) 
[GCC 7.3.0]

librosa: 0.8.0

audioread: 2.1.8
numpy: 1.19.1
scipy: 1.5.2
sklearn: 0.23.1
joblib: 0.16.0
decorator: 4.4.2
soundfile: 0.10.3
resampy: 0.2.2
numba: 0.50.1

numpydoc: None
sphinx: None
sphinx_rtd_theme: None
sphinxcontrib.versioning: None
sphinx-gallery: None
pytest: None
pytest-mpl: None
pytest-cov: None
matplotlib: 3.3.0
presets: None

bmcfee:
Ah good catch. I think this is purely cosmetic (the behavior is correct, but the description is wrong). But we should fix it for the next release.

jamesb93:
Furthermore, what would be the appropriate way of dealing with such scenarios? Pad/stuff the frame with some value?

bmcfee:
@jamesb93 yes, either pad the frame or reduce the FFT size.  Note that frames are padded by default, in which case this is a warning (not an exception).

miltonG-isla:
Hello, we observed the same error. My student Kei may have found an easy fix. In spectrum.py lines 221 and 230, he noted the > sign should probably be a < sign.

bneslyn21:
Mil

> Hello, we observed the same error. My student Kei may have found an easy fix. In spectrum.py lines 221 and 230, he noted the > sign should probably be a < sign.

This doesn't work for me :(

bshall:
[...]

PR #1218, opened 2020-08-06
stft issues with large FFT size

fzalkow:
I wondered about some warnings raised by the `stft` function when `n_fft` is larger than the signal length. In theory, this should not be a problem when `win_length` is shorter because `n_fft` is achieved by zero-padding. The following example illustrates the issue:

```python
import librosa
import numpy as np

x = np.random.random(4000)
X = librosa.stft(x, hop_length=100, win_length=200, n_fft=4001)
```

This raises a warning.
```
UserWarning: n_fft=4001 is too small for input signal of length=4000
  n_fft, y.shape[-1]
```

I have various concerns, maybe based on a misunderstanding?

1. In general, should it not mean that the FFT size is too large (instead of too small)?
2. When `center=True`, it should pad zeros as necessary to fulfill the FFT size. In practice, I found the output to be fine for all FFT sizes. Isn't the warning unnecessary?
3. Currently, an error is raised if `center=False`. In theory, we could have an arbitrarily large FFT size if the window size fits the signal. As far as I see, this is only a problem in the specific librosa implementation, where we do windows of FFT size instead of using windows of the window size (with zero padding afterward). Of course, the librosa implementation is more efficient compared to an implementation with zero padding after windowing.

I don't know how to handle it? I want to start with the suggestion of removing the warning and re-phrasing the error message. However, I only see that the third point can be addressed by reimplementing the STFT for the case of `center=False` and `n_fft > y.shape[-1]`.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -218,18 +218,11 @@ def stft(
 
     # Pad the time series so that frames are centered
     if center:
-        if n_fft > y.shape[-1]:
-            warnings.warn(
-                "n_fft={} is too small for input signal of length={}".format(
-                    n_fft, y.shape[-1]
-                )
-            )
-
         y = np.pad(y, int(n_fft // 2), mode=pad_mode)
 
     elif n_fft > y.shape[-1]:
         raise ParameterError(
-            "n_fft={} is too small for input signal of length={}".format(
+            "n_fft={} is too large for input signal of length={}".format(
                 n_fft, y.shape[-1]
             )
         )
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -218,6 +218,13 @@ def stft(
 
     # Pad the time series so that frames are centered
     if center:
+        if n_fft > y.shape[-1]:
+            warnings.warn(
+                "n_fft={} is too small for input signal of length={}".format(
+                    n_fft, y.shape[-1]
+                )
+            )
+
         y = np.pad(y, int(n_fft // 2), mode=pad_mode)
 
     elif n_fft > y.shape[-1]:
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -269,6 +269,12 @@ def test_stft_toolong_left():
     librosa.stft(y, n_fft=2048, center=False)
 
 
+def test_stft_toolong_center():
+    y = np.zeros((128,))
+    with pytest.warns(UserWarning):
+        librosa.stft(y, n_fft=2048, center=True)
+
+
 def test_stft_winsizes():
     # Test for issue #1095
     x = np.zeros(1000000)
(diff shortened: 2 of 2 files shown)


lostanlen:
Related: https://github.com/librosa/librosa/issues/1214

(3) strikes me as an antipattern. If the signal is so short that it doesn't fit an entire STFT frame, then the whole pipeline is at risk. Solving the warning (either by modifying STFT parameters or by lengthening the signal) should be the responsibility of the user, not of librosa.

fzalkow:
Right, #1214 exactly covers my first point!

Concerning point 3: In my view, the length of an STFT frame is defined by the window length. So I agree that the signal has to fit an entire STFT frame. However, the FFT size can be larger than the window size and even larger than the signal length. If the FFT size is larger than the window size, the respective windowed frames need to be zero-padded.

If you do not disagree in principle, I can work on a more constructive PR to illustrate what I mean with this point.

bmcfee:
Coming back to this one, sorry for the lack of attention here!



>     1. In general, should it not mean that the FFT size is too large (instead of too small)?

Yes, I think that's a thinko in the error message.  As noted in #1214 , the behavior is correct but the message is backward.



>     2\. When `center=True`, it should pad zeros as necessary to fulfill the FFT size. In practice, I found the output to be fine for all FFT sizes. Isn't the warning unnecessary?

I think it's still helpful to have a warning here.  It doesn't actually prevent a user from computing STFTs on short signals, and warnings can be easily suppressed if the user is confident that this is what they want to be doing.  However, the warning could be helpful for detecting problematic examples in a feature extraction pipeline, and since the padding would otherwise obscure these short clips, it seems like the only way to achieve this.



>     3\. Currently, an error is raised if `center=False`. In theory, we could have an arbitrarily large FFT size if the window size fits the signal. As far as I see, this is only a problem in the specific librosa implementation, where we do windows of FFT size instead of using windows of the window size (with zero padding afterward).



> However, the FFT size can be larger than the window size and even larger than the signal length. If the FFT size is larger than the window size, the respective windowed frames need to be zero-padded.

Yeah, it's possible to extend the DFT length beyond the number of samples in the window.  Right now we do this by padding the window (if necessary) out to the frame length, but I guess you could also just set `n` explicitly in the call to `rfft` and bypass the padding step.  As I recall, we didn't do it that way originally because we were already padding the framed signal to get centering, so there wasn't much of a benefit to doing both. Maybe there's a benefit if we want to do left-aligned frequency over-sampling, but I think that's a pretty unusual use case.

bmcfee:
Hey @fzalkow , sorry to take so long to circle back on this. (I know you've been busy too!)

Would it be possible to do the following here:

1. Rebase to the current main, as we've changed quite a bit with respect to the testing infrastructure
2. Revert the removal of the userwarning and test.

Or would it be simpler to open a new PR to only correct the error in the warning message?

bmcfee (approved):

fzalkow:
Thanks for merging! :+1:

bmcfee:
Thanks for the fix! 🎉

PR #1228, opened 2020-08-23
Fix `soundfile` Error Catching Problem

chaonan99:
For the latest `soundfile`, error handling changed (used to throw `RuntimeError` only), as in [this pull request](https://github.com/bastibe/SoundFile/pull/272). This cause `librosa` failed to catch when other types of error are thrown. This PR fixes the issue by changing `RuntimeError` to `Exception` when it comes to `soundfile` in `librosa/core/audio.py`.


bmcfee:
Thanks for opening this!

Bare exception catching is not a good practice though.  We were catching a specific kind of error from soundfile, which is only raised when there's an unsupported codec.  Your proposed change would also trigger on filenotfound errors, which is not what we want to do here.

Can you revise the PR to catch the specific error that soundfile raises now?

bmcfee:
Thanks for the changes.  Having now looked at the linked PR though, I don't think the changes here implement what we want them to do.

The PR in question replaces the generic `RuntimeError` with a library-specific `SoundFileError`.  If we want to catch this, we'll need to bump the minimum version requirement, since this is new API from soundfile.  (I'm okay with that, but we might need to be a little careful about how exactly it's implemented.)

The changes you've suggested here do not cover this change though.  Instead, they seem related to the following:
https://github.com/bastibe/SoundFile/blame/master/README.rst#L74-L83 by catching ValueError or TypeError if the API is used incorrectly.  This is not the purpose of our exception catch though: the API is used correctly, and we're only trying to detect when a file has an unsupported codec.  Any other errors in usage should not be caught by us, and should therefore propagate up to the caller.

chaonan99:
Good point. I opened this PR because when I was using `librosa.load`, I got "ValueError: frames must be specified for non-seekable files" from `soundfile` causing `librosa.load` failed. But audioread seems to work just fine. So I ended up changing `RuntimeError` to `Exception` and the load method worked for me.

Having a closer look, I was running in a `ThreadPool` (`multiprocessing.pool.ThreadPool`). When I switched to run in a single thread, `RuntimeError` is thrown instead of `ValueError` from `soundfile`. It seems to me a `soundfile` bug that cannot throw the correct error in a multithread context. I'll try to file an issue in `soundfile`.

So from `librosa` perspective, is this something we would consider to handle?

bmcfee:
> Having a closer look, I was running in a `ThreadPool` (`multiprocessing.pool.ThreadPool`). When I switched to run in a single thread, `RuntimeError` is thrown instead of `ValueError` from `soundfile`. It seems to me a `soundfile` bug that cannot throw the correct error in a multithread context. I'll try to file an issue in `soundfile`.

Ok, yeah that sounds like a problem upstream of us and nothing to do with how we handle exceptions.



> So from `librosa` perspective, is this something we would consider to handle?

If soundfile is changing their exception API, we should keep up with it.  It does appear that the change you mentioned has been merged, but not yet released, so I don't think there's anything for us to do yet.  However, we will have problems when the next soundfile release drops, so we should prepare for it as best we can with the 0.8.1 point release.

bmcfee:
Circling back on this: it looks like the new release on soundfile (0.10.3) includes the new exceptions, so we can go ahead with this for the 0.8.1 release.  We'll have to bump the minimum required version accordingly.

bmcfee:
> Circling back on this: it looks like the new release on soundfile (0.10.3) includes the new exceptions, so we can go ahead with this for the 0.8.1 release. We'll have to bump the minimum required version accordingly.

Looks like I was mistaken -- 0.10.3 does not include the new exception class, but the next release will, as a subclass of `RuntimeError`: https://github.com/bastibe/python-soundfile/blob/744efb4b01abc72498a96b09115b42a4cabd85e4/soundfile.py#L1520 .  What this means is that the current exception catcher (`RuntimeError`) will continue to work for us into the future, but we will at some point have the opportunity to make our catch more specific so as to prevent it being triggered by other `RuntimeErrors`.  Maybe @bastibe can confirm that I have this right, or correct if I'm wrong?

As stated earlier in thread, I don't think we should modify the behavior here to also catch more exception classes, as that doesn't address the core issue.

I'm removing this PR from the 0.8.1 milestone, and we can revisit this issue when 0.10.4 (or whatever the next soundfile release is) comes out.

bastibe:
> What this means is that the current exception catcher (`RuntimeError`) will continue to work for us into the future, but we will at some point have the opportunity to make our catch more specific so as to prevent it being triggered by other `RuntimeErrors`. Maybe @bastibe can confirm that I have this right, or correct if I'm wrong?

That sounds correct to me. The thing is, soundfile is sort of in the same position as librosa: we are downstream from libsndfile, whose error reporting is not as rich as we are accustomed to in Python. Some file-not-found errors are nondescript "system errors", for example.

I am currently waiting for the next release of libsndfile, with MP3 support. Once that goes live, I will update soundfile, which will include the new exceptions.

bmcfee:
Thanks @bastibe !

bmcfee:
Doing a little gardening here - I think we need to close this PR out and start fresh.  Sorry to let this languish so long, but the release schedules for all the parts involved have not made this easy.

I'll open a new issue to document what exactly the status of this one is.

PR #1255, opened 2020-10-19
Return keys in list_examples

carlthome:
This PR introduces a minor change to `librosa.util.list_examples` such that it also returns the keys as a `List[str]`. This would be useful for finding all available files programmatically when testing DSP code.

1. Perhaps the current `list_examples` function is intentionally only printing and returning `None` though? If so, just let me know and I'll close this PR!
2. It might be sensible to make the print to stdout optional by introducing a function argument (e.g. `librosa.util.list_examples(display=True)`). As long as the current default is preserved, it shouldn't break anything.

### DIFF
--- a/librosa/util/files.py
+++ b/librosa/util/files.py
@@ -112,6 +112,11 @@ def list_examples():
 
     A brief description is provided in the second column.
 
+    Returns
+    -------
+    examples : List[str]
+        List of paths to the audio example files included with librosa
+
     See Also
     --------
     util.example
@@ -119,8 +124,10 @@ def list_examples():
     """
     print("AVAILABLE EXAMPLES")
     print("-" * 68)
-    for key in sorted(__TRACKMAP.keys()):
+    examples = sorted(__TRACKMAP.keys())
+    for key in examples:
         print("{:10}\t{}".format(key, __TRACKMAP[key]["desc"]))
+    return examples
 
 
 def example_info(key):


bmcfee:
Thanks @carlthome .  I see the utility of this, but I think we might want to think a little more about how it's done.

As you mention in (1), the function here really is intended only for interactive use.  Rather than extend it to return filenames, I would prefer to add a new function that returns paths separately.

The other issue I see is that the paths do not necessarily "exist" when listed.  We use pooch to download files on request, so you'll have to go through the librosa example loader to access things.

As a slightly more meta issue, I do think having a bit of friction at this part of the API isn't a bad thing.  One of the bonuses of moving to this example framework is that we've significantly reduced the file transfer footprint of the package.  If we make it easy to automatically download example, I worry a little about increasing our footprint again.

All that said, my general plan is to only grow the example list over time, so anything that's accessible now will stay that way.  For testing purposes, I think it's more useful to pick specific cases that demonstrate the behavior you want to test on, rather than fanning out over all examples.  This is why the examples are picked to satisfy different criteria (instrumentation, polyphony, length, structure, etc).  Is there a compelling use-case for testing on all examples that I'm not seeing?

bmcfee:
This PR's been idle for a while, and I think I'm coming down on the side of leaving it as is, primarily due to the friction argument above.

If people *really* want programmatic access to tracks for some reason (though I discourage this), it can be obtained by
```python
>>> track_keys = librosa.util.files.__TRACKMAP.keys()
```

Any objection to closing this out?

carlthome:
Agreed! And thanks for that one liner.

Sorry for losing track of this PR.

PR #1287, opened 2021-02-22
Add ffmpeg link for windows

cashlo:
Installing gstreamer and pygobject on windows is a lot of pain, it turns out installing ffmpeg is a lot easier. Adding a link for ffmpeg and I hope it will save someone some pain.


bmcfee:
Thanks for this, but I'm a bit puzzled.  The link we had (gstreamer.freedesktop.org) already included binaries, so why do we need a second link?

cashlo:
I just couldn't get pygobject to install on Windows, but for ffmpeg I downloaded the binaries and added them to path and that was it. Am I missing something? The new link is for ffmpeg.

bmcfee:
Ah, sorry about that -- I read too quickly.  This LGTM, thanks!

bmcfee (approved):

PR #1290, opened 2021-02-26
fixed #1283, format and tick locations for CQT plots

### LINKED ISSUE #1283: Maybe I found and fix a bug in librosa.display
Hi, Librosa Team:
As I mentioned in Issue  #1261 and Issue  #1243, the fig with a `cqt_note` y-axis didn't behave as I expect. So I read the code and matplotlib's document. Then I found some mistakes. After that, I use some code to correct them:

1. In `/librosa/display.py#L216` (the formatter of y-axis), the contributor wrote:https://github.com/librosa/librosa/blob/a53fa56bdb6695a994008d5b6ccd0100870a6036/librosa/display.py#L216, with a `int(x)` , in fact, this made the ticks away from where they should stay. My suggest is to delete the `int()` and use the float data instead.

2. In `/librosa/display.py#L1064`(the locator of y-axis), the contributor wrote:https://github.com/librosa/librosa/blob/a53fa56bdb6695a994008d5b6ccd0100870a6036/librosa/display.py#L1064, In my opinion, this doesn't make sense. I think we should use `FixedLocator(librosa.cqt_frequencies(85, fmin=librosa.note_to_hz('C1')))` instead, because `cqt_frequencies` is designed to compute the center frequencies of Constant-Q bins. See: https://github.com/librosa/librosa/blob/a53fa56bdb6695a994008d5b6ccd0100870a6036/librosa/core/convert.py#L1105-L1140

My full code:
```python
def set_axis(ax) -> None:
    """call AFTER you draw!!
    =========================
    Args:
        ax : Axes
    Returns:
        None
    """
    from matplotlib.ticker import LogLocator, FixedLocator
    from matplotlib.ticker import FuncFormatter

    def _major_fomatter(x, pos):
        """Fomatter for major ticks"""
        # print('MAJOR  ' + str(x) + '  ' + str(librosa.hz_to_note(x, octave=True, cents=True)))
        if x <= 0:
            return ""
        else:
            return librosa.hz_to_note(x, octave=True, cents=True)  # necessary for cents
            # return librosa.hz_to_note(x, octave=True, cents=False)

    def _minor_fomatter(x, pos):
        """Fomatter for minor ticks"""
        # print('MINOR  ' + str(x) + '  ' + str(librosa.hz_to_note(x, octave=True, cents=True)))
        if x <= 0:
            return ""
        vmin, vmax = ax.yaxis.get_view_interval()
        if vmax > 4 * max(1, vmin):
            return ""
        # return librosa.hz_to_note(x, octave=True, cents=True)  # debug
        return librosa.hz_to_note(x, octave=True, cents=False)
    """=Set Locator & Fomatter="""
    # log_C1 = np.log2(librosa.note_to_hz("C1"));
    # offset = 2.0 ** (log_C1 - np.floor(log_C1))
    offset = librosa.note_to_hz("C1") / 32
    ax.yaxis.set_major_locator(LogLocator(base=2.0, subs=(offset,)))
    ax.yaxis.set_minor_locator(
        FixedLocator(librosa.cqt_frequencies(
            85, fmin=librosa.note_to_hz('C1')))
    )
    ax.yaxis.set_major_formatter(_major_fomatter)
    ax.yaxis.set_minor_formatter(_minor_fomatter)
    return None
```

I didn't read the source code so carefully, and that's why I don't want to put my code into the librosa package by a PR.
If anyone have a good idea, feel free to make your own PR, just remember @unbadfish in you PR.

@bmcfee , I think I fix the "bug" I found

bmcfee:
#### Reference Issue
Fixes #1283 

Two bug-fixes here:

- Dropped an unnecessary integer case on frequency locations
- Corrected minor tick locator offsets in `cqt_note` and `cqt_hz` display modes.  `cqt_svara` was already correct.

#### Any other comments?

Thanks @unbadfish for catching these!


bmcfee:
Note: appveyor build failures are due to different versions of the `freetype` package (2.6 on windows vs 2.10 on linux builds).  These failures are not critical, and should not be a barrier to merging this PR.

bmcfee:
I think this is good to go, but if anyone would care to give it a once-over, I'd appreciate it.

lostanlen:
code looks 👍🏻 
have you looked at the images in the test folder? do they look 👍🏻 too ?

bmcfee:
>  have you looked at the images in the test folder? do they look 👍🏻 too ?

Yeah, no major changes.  I also tried to force the appveyor builds to use a more recent freetype, but I think we're stuck on it until the matplotlib build for windows also updates.  I think this one is good to go.

Thanks for the spot check!

PR #1293, opened 2021-03-10
1025 deemphasis

QEDan:
This implements `effects.py:deemphasis()`, which is the inverse operation of `effects.py:preemphasis()`. Thus, it resolves https://github.com/librosa/librosa/issues/1025. Tests are added that demonstrate recovery of the original signal prior to preemphasis.

#### Any other comments?
This is my first contribution. Please let me know how to improve it.

### DIFF
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -662,6 +662,7 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
 
     return y_out
 
+
 def deemphasis(y, coef=0.95, zi=None):
     """De-emphasize an audio signal with the inverse operation of preemphasis():
 
@@ -691,7 +692,8 @@ def deemphasis(y, coef=0.95, zi=None):
         Initial filter state. If inverting a previous preemphasis(), the same value should be used.
          When making successive calls to non-overlapping
 
-        By default ``zi`` is initialized as ``((2 - coef) * y[0] - y[1]) / (3 - coef)``. This
+        By default ``zi`` is initialized as
+        ``((2 - coef) * y[0] - y[1]) / (3 - coef)``. This
         value corresponds to the transformation of the default initialization of ``zi`` in ``preemphasis()``,
         ``2*y[0] - y[1]``.
 
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -645,6 +645,9 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
     >>> np.allclose(y_filt, np.concatenate([y_filt_1, y_filt_2]))
     True
 
+    See Also
+    --------
+    deemphasis
     """
     b = np.asarray([1.0, -coef], dtype=y.dtype)
     a = np.asarray([1.0], dtype=y.dtype)
@@ -667,7 +670,7 @@ def deemphasis(y, coef=0.95, zi=None):
     """De-emphasize an audio signal with the inverse operation of preemphasis():
 
     If y = preemphasis(x, coef=coef, zi=zi), the deemphasis is:
-        x[i] -> y[i] + coef * x[i-1]
+        x[i] = y[i] + coef * x[i-1]
         x = deemphasis(y, coef=coef, zi=zi)
 
 
@@ -699,7 +702,7 @@ def deemphasis(y, coef=0.95, zi=None):
     Returns
     -------
     y_out : np.ndarray
-        pre-emphasized signal
+        de-emphasized signal
 
     Examples
     --------
@@ -719,8 +722,6 @@ def deemphasis(y, coef=0.95, zi=None):
     a = np.asarray([1.0], dtype=y.dtype)
 
     if zi is None:
-        if coef == 3.0:
-            raise NotImplementedError("Unable to do deemphasis if coefficient is 3.")
         zi = ((2 - coef) * y[0] - y[1]) / (3 - coef)
     y[0] -= zi
 
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -666,9 +666,9 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
 def deemphasis(y, coef=0.95, zi=None):
     """De-emphasize an audio signal with the inverse operation of preemphasis():
 
-    If y~ = preemphasis(y, coef=coef, zi=zi)
-        y[i] -> y~[i] + coef * y[i-1]
-        y = deemphasis(y~, coef=coef, zi=zi)
+    If y = preemphasis(x, coef=coef, zi=zi), the deemphasis is:
+        x[i] -> y[i] + coef * x[i-1]
+        x = deemphasis(y, coef=coef, zi=zi)
 
 
     Parameters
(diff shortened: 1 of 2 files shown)


librosa/effects.py
```diff
@@ -660,3 +661,65 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
         return y_out, z_f
 
     return y_out
+
+
+def deemphasis(y, coef=0.95, zi=None):
+    """De-emphasize an audio signal with the inverse operation of preemphasis():
+
+    If y~ = preemphasis(y, coef=coef, zi=zi)
```
lostanlen, line 669: i don't recommend that you use the notation `y~` in your pseudo-code to mean "\tilde{y}" in LaTeX. Instead, call x the input of preemphasis and y the input of deemphasis.

y = preemphasis(x, coef=coef, zi=zi)
x = deemphasis(y, coef=coef, zi=zi)

Also, i would prefer an equal sign rather than an arrow (`->`)
x[t] = y[t] + coef * y[t-1]

so that your pseudo-code looks as close as possible to actually runnable Python code

librosa/effects.py
```diff
@@ -660,3 +661,65 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
...
+        At ``coef=1``, the result is the first-order difference of the signal.
+
+        The default (0.97) matches the pre-emphasis filter used in the HTK
+        implementation of MFCCs [#]_.
+
+        .. [#] http://htk.eng.cam.ac.uk/
+
+    zi : number
+        Initial filter state. If inverting a previous preemphasis(), the same value should be used.
+         When making successive calls to non-overlapping
```
lostanlen, line 693: incomplete sentence?

librosa/effects.py
```diff
@@ -660,3 +661,65 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
...
+        ``2*y[0] - y[1]``.
+
+    Returns
+    -------
+    y_out : np.ndarray
+        pre-emphasized signal
+
+    Examples
+    --------
+    Apply a standard pre-emphasis filter and invert it with deemphasis
```
lostanlen, line 707: de-emphasis (hyphen)

librosa/effects.py
```diff
@@ -660,3 +661,65 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
...
+        By default ``zi`` is initialized as
+        ``((2 - coef) * y[0] - y[1]) / (3 - coef)``. This
+        value corresponds to the transformation of the default initialization of ``zi`` in ``preemphasis()``,
+        ``2*y[0] - y[1]``.
+
+    Returns
+    -------
+    y_out : np.ndarray
+        pre-emphasized signal
+
```
lostanlen, line 704:
```
See Also
--------
preemphasis
```

lostanlen (changes requested):
welcome to librosa @QEDan :)
thank you for this. i have some minor comments regarding the documentation

librosa/effects.py
QEDan: Fixed in https://github.com/librosa/librosa/pull/1293/commits/5771637b181351a885e2fee266c0a03a397b4e09

librosa/effects.py
QEDan: Fixed in https://github.com/librosa/librosa/pull/1293/commits/5771637b181351a885e2fee266c0a03a397b4e09

librosa/effects.py
```diff
@@ -660,3 +661,68 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
         return y_out, z_f
 
     return y_out
+
+
+def deemphasis(y, coef=0.95, zi=None):
+    """De-emphasize an audio signal with the inverse operation of preemphasis():
+
+    If y = preemphasis(x, coef=coef, zi=zi), the deemphasis is:
+        x[i] -> y[i] + coef * x[i-1]
```
lostanlen, line 670: again, i would prefer an equal sign rather than an arrow here
(`->` is not an assignment operator in Python like it is in R or Mathematica ; and besides, i would argue that the arrow should point leftward in this case)

librosa/effects.py
```diff
@@ -660,3 +661,68 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
...
+
+        By default ``zi`` is initialized as
+        ``((2 - coef) * y[0] - y[1]) / (3 - coef)``. This
+        value corresponds to the transformation of the default initialization of ``zi`` in ``preemphasis()``,
+        ``2*x[0] - x[1]``.
+
+    Returns
+    -------
+    y_out : np.ndarray
+        pre-emphasized signal
```
lostanlen, line 702: de-emphasized signal

librosa/effects.py
```diff
@@ -660,3 +661,68 @@ def preemphasis(y, coef=0.97, zi=None, return_zf=False):
...
+    See Also
+    --------
+    preemphasis
+    """
+    b = np.asarray([1.0, -coef], dtype=y.dtype)
+    a = np.asarray([1.0], dtype=y.dtype)
+
+    if zi is None:
+        if coef == 3.0:
+            raise NotImplementedError("Unable to do deemphasis if coefficient is 3.")
```
lostanlen, line 723: i don't think we need to explicitly catch this exception, considering that we've already documented that the `coef` is expected to range between 0 and 1. Setting `coef` to `3.0` will naturally cause a `ZeroDivisionError`, which, i would argue, is exactly the right kind of error for this input.

@bmcfee feel free to object

lostanlen (commented):
getting there

librosa/effects.py
QEDan: Fixed in https://github.com/librosa/librosa/pull/1293/commits/c25e8c220b0e565473c16eeed82babed4ff1b1bd. Sorry I missed this in the first round.

librosa/effects.py
QEDan: Fixed in https://github.com/librosa/librosa/pull/1293/commits/c25e8c220b0e565473c16eeed82babed4ff1b1bd.

librosa/effects.py
QEDan: Removed in https://github.com/librosa/librosa/pull/1293/commits/c25e8c220b0e565473c16eeed82babed4ff1b1bd.

I was originally thinking that `coef=3` was a special case that should never come up, but could be handled in the future. Therefore, it was not yet implemented. But I took a second look at the math and I think the value of `zi` can't be deduced from the transformed signal alone if `coef=3`. Handling that case would require passing in more information, such as `x[0]` of the non-preemphasized signal. Anyway, as you said, it's reasonable to not handle that case because it's outside the range `0<coef<1`.

bmcfee:
Hey everyone, checking in on this issue: is there anything left to do here, or is it good to go?

I'm not concerned about the build fail on appveyor -- it's unrelated to this PR, and isn't something we can immediately fix anyway.

lostanlen:
as far as i am concerned this is ready for merging

bmcfee:
Great!  @QEDan anything else you want to add here, or shall I merge?

QEDan:
> Great! @QEDan anything else you want to add here, or shall I merge?

I'm happy if you're happy. Thanks!

bmcfee:
Done. Thanks again @QEDan !

lostanlen:
An important function for synthesizing audio from preemphasized MFCC. Many thanks @QEDan !

Issue #1307, opened 2021-04-13
Fix CI issues


bmcfee:
**Describe the bug**

We currently have build failures due to upstream issues in a few CI environments:

- Python 3.8 on Travis (linux), appears to be an issue with the sndfile package.  Not sure why this only seems to affect 3.8, but it's weird.
- Appveyor builds have failures on display tests.  This comes from a difference in the version of freetype used to render text compared to what we have for the baseline images, and is not actually a bug per se.

**Additional context**

For appveyor, we should either force a pin on the freetype version (assuming one is available), or disable the display tests.

PR #1308, opened 2021-04-13
trying to fix some CI issues

bmcfee:
#### Reference Issue
#1307 

Forcing sndfile installation on travis, and trying to update appveyor so that image tests work again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ addons:
     apt:
         packages:
         - libsamplerate0
+        - libsndfile1
 
     homebrew:
         packages:
--- a/.travis_dependencies.sh
+++ b/.travis_dependencies.sh
@@ -11,7 +11,7 @@ conda_create ()
     conda update -q conda
     conda config --add channels pypi
     conda info -a
-    deps='pip numpy scipy coverage scikit-learn!=0.19.0 matplotlib numba libsndfile'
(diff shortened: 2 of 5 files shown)


bmcfee:
Python 3.8 issue stems from libsndfile not being packaged on mainline conda for python 3.8 (?!).  It's not a python package, so I don't see why this is the case, but perhaps we can work around it by forcing from conda-forge.

bmcfee:
The appveyor issue is also tricky.

Although we force installation of `freetype>=2.10` (and this indeed works), the matplotlib build seems to be linked against freetype 2.6.  This is ultimately the source of the image comparison failures, as the two versions differ by just enough to through off the RMS calculations.

Since fixing this would probably require rebuilding matplotlib on our appveyor instance (which seems costly and not worth it), I propose that we simply bypass the image comparison tests (or mark as allowed to fail) if matplotlib's freetype version is below 2.10.  This is probably a good fix anyway, as it makes explicit the requirements for running the tests relative to the environment in which i created the original fixtures.

bmcfee:
Forcing conda-forge on travis sort of works -- libsndfile is now installed in the environment.  However, the soundfile package can't find it, and the tests still fail.  I'm pretty flummoxed at this point.

bmcfee:
Update: installing libsndfile1 via apt fixes this.  It's not a great fix, in that the sndfile packaging is still kinda broken, but at least it gets that issue out of our way.

Will merge once all outstanding builds finish.

Issue #1176, opened 2020-06-22
Simplify CQT tests


bmcfee:
At present, our test suite takes somewhere in the neighborhood of 15 minutes to run (on Travis).  A substantial portion of that is due to the CQT tests, which brute-force through a very large parameter grid.

Rather than trying all combinations of parameters, we should refactor the tests to focus on the behavior of individual parameters (or small grids of parameters where interactions are inevitable).

---

For reference, here's the runtime on my dev machine for the full test suite:

```
1 failed, 12963 passed, 1 skipped, 466 xfailed, 10394 warnings in 849.78s (0:14:09)
```

and for just the cqt module:

```
2528 passed, 16 xfailed, 8593 warnings in 519.08s (0:08:39)
```

bmcfee:
Let's make some sense of this by first mapping out what the current test sweeps are.  To keep it simple, I'm not counting all of our specific test cases, only the broadest tests with huge parameter grids.

<details>
<summary>`test_cqt` sweeps 9 parameters (8 non-trivial), 768 tests:</summary>

  - `fmin` (x2)
  - `n_bins` (x4)
  - `bins_per_octave` (x2)
  - `tuning` (x3)
  - `filter_scale` (x2)
  - `norm` (x2)
  - `res_type` (x2)
  - `sparsity`  (x1)
  - `hop_length` (x2)
 </details>

<details>
<summary>`test_vqt` sweeps 10 (4 non-trivial), 24 tests</summary>

  - `fmin` (x2)
  - `n_bins` (x2)
  - `gamma` (x3)
  - `bins_per_octave` (x2)
  - `tuning` (x1)
  - `filter_scale (x1)
  - `norm` (x1)
  - `res_type` (x1)
  - `sparsity` (x1)
  - `hop_length` (x1)
</details>

<details>
<summary>`test_hybrid_cqt` sweeps 10 (7 non-trivial), 672 tests</summary>

  - `sr` (x1)
  - `hop_length` (x1)
  - `sparsity` (x1)
  - `fmin` (x2)
  - `n_bins` (x7)
  - `bins_per_octave` (x2)
  - `tuning` (x3)
  - `resolution` (x2)
  - `norm` (x2)
  - `res_type` (x2)
</details>

<details>  
<summary>`test_icqt` sweeps 6, 96 tests </summary>

  - `over_sample` (x2)
  - `scale` (x2)
  - `hop_length` (x2)
  - `length` (x2)
  - `res_type` (x3)
[...]

PR #1317, opened 2021-05-06
CI on github actions

### LINKED ISSUE #1316: CI migration: travis -> github actions
With Travis's new model, we're running out of OSS credits at a pretty fast clip.  We can migrate the CI over to github actions following the guidelines [here](https://docs.github.com/en/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions).

bmcfee:
#### Reference Issue
Fixes #1316 

This PR initiates the migration from travis to github actions for CI testing.

#### Any other comments?

This is a work in progress, and it will probably take several tries before we get it right.

I may also roll in windows testing and remove appveyor if it looks like it would be easier all around.


bmcfee:
Workflow is basically working now, but we're hitting a bunch of problems with unresolvable conda environments, as well as some matplotlib regressions.  Probably we just need to re-render the image fixtures with mpl 3.4.1, but i'm not sure what's going on with the environments.

Also to-do: caching.

bmcfee:
> we're hitting a bunch of problems with unresolvable conda environments

These turned out to be due to mixing conda and conda-forge builds of numpy and scipy packages.  Shifting everything in the test env over to conda-forge seems to have resolved the issue.

bmcfee:
Coveralls' github action is kinda broken, in that the official action doesn't quite work with python, and the python add-on doesn't work outside of linux.  As a result, getting the action to run on the correct subset of the build matrix was proving to be kind of a pain.

I switched things over to run on codecov, which does appear to work pretty much out of the box.  Functionality isn't 100% yet because we don't have a codecov run from the main branch, but I'm happy to stick with it.

bmcfee:
This all appears to be basically working at this point.  We don't have build caching working, but I think that can be added later on.  For reference, the environment setup takes about 3 minutes right now, while the test suite itself takes about 16 minutes.

Other outstanding issues #1176 ought to bring down the test times substantially, at which point we can revisit build caching.  (Amdahl's law FTW!)

Last things to do here are clean up old CI configs and update the badges.

PR #1322, opened 2021-05-09
Make the default hop_length=None for mel_to_audio

### LINKED ISSUE #1318: hop_length defaults to 512 instead of n_fft//4 in feature.inverse.mel_to_audio
When inverting a Mel spectrogram to audio using feature.inverse.mel_to_audio, the hop_length defaults to 512 samples regardless of the specified n_fft.  The function doc, however, indicates that it will default to n_fft // 4, which seems like a better choice.

As far as I can see, changing the default value of hop_length to None will allow griffinlim to choose the correct default.
https://github.com/librosa/librosa/blob/4d92350c5bb661feb196f2659dc8ff3fa60fa29b/librosa/feature/inverse.py#L96

dpwe:
Fixes #1318

The docstring indicates that the default hop_length for mel_to_audio will be n_fft//4, which is what the underlying griffinlim() will do with hop_length=None. 

However, the function definition previously had a fixed default of 512, which led to unexpected results for non-default n_ffts.

### DIFF
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -93,7 +93,7 @@ def mel_to_audio(
     M,
     sr=22050,
     n_fft=2048,
-    hop_length=512,
+    hop_length=None,
     win_length=None,
     window="hann",
     center=True,


bmcfee (approved):
LGTM, thanks!

PR #1326, opened 2021-05-15
Adaptive waveplot; waveshow

### LINKED ISSUE #1207: Adaptive waveplot

`display.waveplot` works by downsampling the input waveform to get an amplitude envelope, rather than the raw samples.

This is great when we want a big-picture view of a long signal, but it's pretty lousy when we're looking at a signal close up.  The `max_points` parameter sets a threshold of sample counts, below which the raw signal is displayed.

The problem is that there's no easy way to switch between the two modes based on the actual display limits of plot in question.  For example, it'd be nice to be able to just `waveplot` some data, set the axis limits later, and have the plot decide dynamically whether to show the aggregated envelope or the raw samples.   Even better, for interactive backends (ipympl, qt), a user could zoom in or out, and the plot should update automatically.

The [matplotlib example gallery](https://matplotlib.org/3.3.0/gallery/event_handling/resample.html?highlight=resample) shows how to do this kind of thing using a callback for whenever the display limits change.  The example there isn't perfect: it recomputes the downsampled signal on any limit change, including panning (not just zooming), so it's likely pretty slow.  But I think we could adapt it to be more efficient.

bmcfee:
#### Reference Issue
Implements #1207 

This PR fixes #1207 by introducing a new function `librosa.display.waveshow`, as documented in that thread.

#### Any other comments?

This PR also deprecates `librosa.display.waveplot` to be removed in the 0.9 release.  Waveshow's functionality generally subsumes that of waveplot, but there wasn't a nice way to maintain API compatibility for a more seamless transition.


bmcfee:
Note: I have not yet implemented tests.

bmcfee:
Tests and docs are updated, this is ready for review.

Note: callback support with backend=Agg (as used in tests and doc builds) doesn't seem to work, so we have to trigger the plot update manually (for now).  Maybe there's a clever solution to this, but even setting `interactive` mode in pyplot didn't seem to fix this.

librosa/display.py
```diff
@@ -1220,8 +1323,197 @@ def __coord_time(n, sr=22050, hop_length=512, **_kwargs):
...
+    marker : string
+        Marker symbol to use for sample values. (default: no markers)
+
+    where : string, {'pre', 'mid', 'post'}
+        This setting determines how both waveform and envelope plots interpolate
+        between observations.
+
+        See `matplotlib.pyplot.step` for details.
+
+        Default: 'post'
```
lostanlen, line 1415: what is the reason why we default to `post` if plt.step defaults to `pre`?

librosa/display.py
```diff
@@ -1220,8 +1323,197 @@ def __coord_time(n, sr=22050, hop_length=512, **_kwargs):
...
+
+
+    ax : matplotlib.axes.Axes or None
+        Axes to plot on instead of the default `plt.gca()`.
+
+    offset : float
+        Horizontal offset (in seconds) to start the waveform plot
+
+    marker : string
+        Marker symbol to use for sample values. (default: no markers)
```
lostanlen, line 1407: maybe refer to matplotlib.markers for more details?

librosa/display.py
```diff
@@ -1220,8 +1323,197 @@ def __coord_time(n, sr=22050, hop_length=512, **_kwargs):
...
+    ----------
+    y : np.ndarray [shape=(n,) or (2,n)]
+        audio time series (mono or stereo)
+
+    sr : number > 0 [scalar]
+        sampling rate of ``y``
+
+    max_points : postive integer
+        Maximum number of samples to draw.  When the plot covers a time extent
+        smaller than `max_points / sr` (default: 1/2 second), samples are drawn.
```
lostanlen, line 1374: "all samples are drawn"?

librosa/display.py
```diff
@@ -1220,8 +1323,197 @@ def __coord_time(n, sr=22050, hop_length=512, **_kwargs):
...
+
+
+    Examples
+    --------
+    Plot a monophonic waveform with an envelope view
+
+    >>> import matplotlib.pyplot as plt
+    >>> y, sr = librosa.load(librosa.ex('choice'), duration=10)
+    >>> fig, ax = plt.subplots(nrows=3, sharex=True)
+    >>> librosa.display.waveshow(y, sr=sr, ax=ax[0])
```
lostanlen, line 1447: maybe demo a marker? i would vote for `marker="o"`or `marker="s"`

librosa/display.py
```diff
@@ -1220,8 +1323,197 @@ def __coord_time(n, sr=22050, hop_length=512, **_kwargs):
...
+
+    When using interactive rendering (e.g., in a Jupyter notebook or IPython
+    console), the plot will automatically update as the view-port is changed, either
+    through widget controls or programmatic updates.
+
+
+    Parameters
+    ----------
+    y : np.ndarray [shape=(n,) or (2,n)]
+        audio time series (mono or stereo)
```
lostanlen, line 1367: very important to explain here what happens if y is stereo because that's non-trivial

librosa/display.py
```diff
@@ -1220,8 +1323,197 @@ def __coord_time(n, sr=22050, hop_length=512, **_kwargs):
...
+
+    sr : number > 0 [scalar]
+        sampling rate of ``y``
```
lostanlen, line 1370: in Hertz (you never know these days :D )

librosa/display.py
```diff
@@ -1220,8 +1323,197 @@ def __coord_time(n, sr=22050, hop_length=512, **_kwargs):
...
+    label : string [optional]
+        The label string applied to this plot.
+        Note that the label
+
+    kwargs
+        Additional keyword arguments to `matplotlib.pyplot.fill_between` and
+        `matplotlib.pyplot.step`.
+
+        Note that only those arguments which are common to both functions will be
+        supported.
```
lostanlen, line 1426: i looked at the kwargs of step and fill_between and here are the kwargs in common (at least the ones most likely to be overriden by the user)
alpha, animated, color, label, linestyle, linewidth, snap

maybe worth documenting here for convenience?

librosa/display.py
lostanlen: also worth noting that the marker doesn't need to be a string. it can also be a Marker object (which seems necessary if you want the marker to be a different color than the wave line?)

lostanlen (commented):
amazing
here are my comments re: docs. will make a usability test later

librosa/display.py
bmcfee: This is covered in the main docstring above

librosa/display.py
bmcfee: No, just the samples within the viewport.

librosa/display.py
bmcfee: There is a slight hiccup here that the sphinx extension doesn't quite handle callbacks gracefully, so getting it to draw samples int he docstring examples might require a bit of hackery.  But we could certainly do it.

librosa/display.py
lostanlen: ah, ok, i understand the sentence now. the when should read as an "iff"

librosa/display.py
bmcfee: Great question.  I actually have no idea why step defaults to `pre` when it's obviously wrong :grin: 

Here's how it looks with post:
image

and here's how it looks with pre:
image

librosa/display.py
bmcfee: I'd say even Hz isn't specific enough :)  but sure.

librosa/display.py
bmcfee: I'd rather not, in case it changes in the future and we get out of sync with mpl docs.

librosa/display.py
lostanlen: ... let's keep `post`

bmcfee:
Thanks for the CR!  Doc comments addressed.

bmcfee:
Note: I threw in an extra unit test to cover bits that were missed in #1312.  This was showing up as a coverage regression in this PR because blacking display.py caused a modification to an uncovered line.

bmcfee:
@lostanlen what do you think about rolling #1329 into this PR?  I know it's unrelated scope, but it would almost certainly involve regenerating fixtures for the new display tests here, so we may as well just do it once rather than twice.

lostanlen:
@bmcfee excellent idea!!

bmcfee:
Ok, I think this is goodish.

One usability quirk that I encountered this morning.  The resolution of the envelope artist is now adaptive, in that inputs of different length will be downsampled differently to ensure that the resulting artist has at most `max_points` positions (like the samples plot).  This is mostly fine, until you plot a long waveform and then zoom in.  Long signals will have more aggressive downsampling than short signals, which makes for a blocky display at intermediate (ie still envelope-based) zoom levels.

---

For example, here's the nutcracker example fully zoomed out:
image

Zoomed in to a 2-second window:
image

1-second:
image

And then 1/2 second (samples take over)
image


For comparison purposes, here's our (shorter) trumpet example at a 1-second window:
image

---

This makes me wonder if it's worth having an optional second parameter for controlling the envelope resolution separately of the sample resolution?  Or maybe I'm overthinking it?

lostanlen:
I can reproduce these displays on my machine, @bmcfee 
(Note to self and others: don't forget the `%matplotlib notebook` magic command to enable zoom)

bmcfee:
>  (Note to self and others: don't forget the `%matplotlib notebook` magic command to enable zoom)

`%matplotlib widget` these days :grin: 

@lostanlen what are your thoughts about the envelope resolution issue in the last post?

lostanlen:
> %matplotlib widget these days 😁

OK i'm taking good note of this! i didn't know about `%matplotlib widget`. It is clearly better than `notebook` because it refreshes the figure automatically after resizing :) which `notebook` doesn't (at least not on my machine). But `%matplotlib widget` requires to install `ipympl` before launching the notebook


> @lostanlen what are your thoughts about the envelope resolution issue in the last post?

I think it's fine! It's nothing that a change of `max_points` cannot solve.
And besides, i have the _impression_ that the appearance is much more fluid (less pixelated) now that i'm using `%matplotlib widget` (but maybe that's a wishful-thinking bias

lostanlen:
Frustrating: zooming on an extreme portion of the envelope occasionally does not show the corresponding waveform.
I'm no FPS gamer, but i expected to be able to aim at that peak properly!

https://user-images.githubusercontent.com/3943142/118949307-5ee18d00-b959-11eb-9c11-cca623b08ddd.mp4

bmcfee:
> @bmcfee could it be that the envelope estimation and downsampling misrepresents the position of actual maxima?

Yes, there are two reasons why this can happen:

1. The envelope is calculated from `|y|`, not `y`, so the peak you're seeing at that time might actually be out of phase.  This is correct behavior IMO, and not something that would be easily remedied.
2. The envelopes are left-aligned, not centered, so the peak you're seeing could be displaced in time.  This is primarily an efficiency hack, as otherwise we'd invoke pad-copy to compute the envelope.  Note that centering frames would not fix this though, as peaks can come from anywhere within the frame.

So... I'm inclined to mark this as correct/intended behavior.

Back on the sampling rate question:

> I think it's fine! It's nothing that a change of `max_points` cannot solve.

This is true, but it does make things a little clumsy at times.  I can live with it for now, but perhaps in the future we might want to decouple these parameters.

> And besides, i have the _impression_ that the appearance is much more fluid (less pixelated) now that i'm using `%matplotlib widget` (but maybe that's a wishful-thinking bias

That may be the case, depending on which browser you're using!  I've noticed that widget on firefox is snappier than on chrome, but YMMV.

PR #1335, opened 2021-05-22
Support notation in STFT plots 

### LINKED ISSUE #1210: Support notation in STFT plots
`specshow` allows STFT to be shown in linear or logarithmic scales, with Hz formatting on the frequency axis.

If a user wants a pitch-based decoration (notes or svara), they currently have to use a CQT spectrogram rather than an STFT.  There's no good technical reason for that though, as they all render the data in natural coordinates.

Add axis modes to specshow:

- `fft_note`
- `fft_svara`

which work like their CQT counterparts, but applied to STFT data.

A savvy user could probably hack this by manually constructing the axis decoration objects, but we should make it easier for them.

P-bibs:
#### Reference Issue
Fixes #1210

This PR adds two new axis modes for STFT Plots: note and svara. The changes are mostly copies of the relevant code from the CQT counterparts for the same axis modes, but adjusted to work with the STFT plot.

#### Any other comments?
Still requires documentation (I'll handle this) and testing (@bmcfee has said he can do this)

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -889,6 +889,7 @@ def __mesh_coords(ax_type, coords, n, **kwargs):
         "linear": __coord_fft_hz,
         "fft": __coord_fft_hz,
         "fft_note": __coord_fft_hz,
+        "fft_svara": __coord_fft_hz,
         "hz": __coord_fft_hz,
         "log": __coord_fft_hz,
         "mel": __coord_mel_hz,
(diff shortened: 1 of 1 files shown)


P-bibs:
Note: I can also rebase this after adding documentation to make the merge cleaner if desired.

librosa/display.py
```diff
@@ -39,7 +39,7 @@
 import numpy as np
 from matplotlib.cm import get_cmap
 from matplotlib.axes import Axes
-from matplotlib.ticker import Formatter, ScalarFormatter
+from matplotlib.ticker import AutoLocator, Formatter, MultipleLocator, ScalarFormatter
```
bmcfee, line 42: Seems like these extra imports weren't used -- can we remove them?

librosa/display.py
```diff
@@ -964,6 +966,12 @@ def __scale_axes(axes, ax_type, which):
         mode = "log"
         kwargs[base] = 2
 
+    elif ax_type == "fft_note":
```
bmcfee, line 969: This case is identical to `log` (line 959 above) -- should we just merge them like we have for the cqt types?  This would also catch the presently missing `fft_svara` scaling.

librosa/display.py
```diff
@@ -1089,12 +1097,36 @@ def __decorate_axis(axis, ax_type, key="C:maj", Sa=None, mela=None, thaat=None):
...
+        )
+        axis.set_label_text("Svara")
+
     elif ax_type in ["mel", "log"]:
         axis.set_major_formatter(ScalarFormatter())
         axis.set_major_locator(SymmetricalLogLocator(axis.get_transform()))
         axis.set_label_text("Hz")
 
-    elif ax_type in ["linear", "hz"]:
+    elif ax_type in ["linear", "hz", "fft"]:
```
bmcfee, line 1129: these parts all seem good.

bmcfee (changes requested):
This looks basically good, but we can simplify a bit.  Thanks for getting this moving!

bmcfee:
> Note: I can also rebase this after adding documentation to make the merge cleaner if desired.

Thanks, doesn't seem necessary though as we don't have any conflicts.

bmcfee (approved):

bmcfee:
Made a couple of edits in the PR suggestions; we have a redundant block here, but I can fix it later on when adding tests.  Once this passes CI, I'm happy to merge.

Issue #1327, opened 2021-05-16
0.8.1 release audit


bmcfee:
- [x] migrate packaging spec to pyproject.toml [helpful example post](http://ivory.idyll.org/blog/2021-transition-to-pyproject.toml-example.html)
- [x] lint and style
- [x] update version string
- [x] update changelog
- [x] update authors.md
- [x] add developer documentation for building docs and package releases
- [x] verify doc build / historical doc build

bmcfee:
Instead of adding docs for packaging, I think we should finally get around to setting up github actions for this.

https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

PR #1337, opened 2021-05-23
0.8.1 release prep

bmcfee:
#### Reference Issue
Implements #1327


bmcfee:
I added py36 tests, which we somehow lost track of when migrating to gh actions.  Tests are currently failing for py36 because the conda-forge package for matplotlib is still 3.3, but our test fixtures were generated in 3.4.  The two are different enough to trigger failures in image comparison tests (excessive rmse).

lostanlen (approved):
LGTM
don't forget to set the day (`2021-05-??`) in the changelog :)

bmcfee:
I think this is good to go: docs and historical docs are building now, and the release audit is complete.

I've added a github action for pushing packages to pypi on release tag, but it's keyed to the main branch so we'll have to merge this PR to see if it's working properly.

The only question for me is whether we want to include #1338 in this release.

PR #1340, opened 2021-05-24
support np.integer wherever we support int

### LINKED ISSUE #1275: Would support on more numpy integer types be a good idea?
When playing with librosa especially with other audo libraries and numpy functions, I always meet erros from time to time by type errors where for example, `n_bins` requires `int` (as reflected by `isinstance()` in source code), but other integer types are not acceptable, for example `np.int64`. Would it be a good idea to support more integer types instead of asking users to transfer a numpy integer to a "pythonic" integer? I know it is a small thing but it can avoid some unexpected behaviors and additional bothering for users if done properly. 

I can implement it and do tests if @bmcfee you think it's a good idea.

bmcfee:
#### Reference Issue
Fixes #1275 

Wherever we check for `isinstance(foo, int)` we should also allow `np.integer`.

#### Any other comments?

Late addition to the 0.8.1 release.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -877,7 +877,7 @@ def lpc(y, order):
     >>> ax.set_title('LP Model Forward Prediction')
 
     """
-    if not isinstance(order, int) or order < 1:
+    if not isinstance(order, (int, np.integer)) or order < 1:
         raise ParameterError("order must be an integer > 0")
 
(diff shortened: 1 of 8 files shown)


bmcfee:
All modified functions are verified to work with `np.int32(n)` as inputs, which previously failed.  I think this is good to go.

PR #1341, opened 2021-05-24
Safe and optional onset envelope normalization

### LINKED ISSUE #1291: Add a normalize flag for librosa.onset.onset_detect
`librosa.onset.onset_detect` normalizes the onset_envelope by default, which means that I get different results based on how much of a clip I use. Although I understand the purpose of this, in my case I don't want this normalized.

Add a flag, e.g. `normalize` to `librosa.onset.onset_detect`, defaulting to True for backwards compatability.

```
def onset_detect(y=None, sr=22050, onset_envelope=None, hop_length=512,
                 backtrack=False, energy=None,
                 units='frames', normalize=True, **kwargs):
    # First, get the frame->beat strength profile if we don't already have one
    if onset_envelope is None:
        if y is None:
            raise ParameterError('y or onset_envelope must be provided')

        onset_envelope = onset_strength(y=y, sr=sr, hop_length=hop_length)

    # Shift onset envelope up to be non-negative
    # (a common normalization step to make the threshold more consistent)
    onset_envelope -= onset_envelope.min()

    # Do we have any onsets to grab?
    if not onset_envelope.any():
        return np.array([], dtype=np.int)

    # Add check for normalize flag
    if normalize:
        # Normalize onset strength function to [0, 1] range
        onset_envelope /= onset_envelope.max()

    ...

    return onsets
```

I currently reimplement `onset_detect` without the normalize step.

**Additional context**
I'd be happy to make a PR for this.

bmcfee:
#### Reference Issue
Fixes #1291 

This PR implements three fixes:

1. Normalization of onset envelopes wtihin `onset_detect` is now optional (new feature)
2. When normalization is applied, it is no longer modifies the envelope in-place (bug fix)
3. When no onsets are detected, the dtype of the output array is consistent with the case when they are.  Note that the dtype depends on the `units=` parameter, so frames and samples would use int and time would use float. (bug fix)

#### Any other comments?

I think this should be :shipit: once CI passes.

### DIFF
--- a/librosa/onset.py
+++ b/librosa/onset.py
@@ -33,6 +33,7 @@ def onset_detect(
     backtrack=False,
     energy=None,
     units="frames",
+    normalize=True,
     **kwargs,
 ):
     """Locate note onset events by picking peaks in an onset strength envelope.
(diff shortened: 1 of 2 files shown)


bmcfee:
@lostanlen thoughts on this one? Should be uncontroversial, but another set of eyes is always helpful.

lostanlen (approved):

PR #1347, opened 2021-06-24
GitHub Action to lint Python code

### LINKED ISSUE #1348: Typos discovered by codespell
$ `codespell --ignore-words-list="ba,trough,ue" --skip="*demo.ipynb"`  # See #1347
```
./docs/changelog.rst:1151: compatibilty ==> compatibility
./docs/examples/plot_vocal_separation.py:69: minimium ==> minimum
./librosa/filters.py:927: bandwith ==> bandwidth
./librosa/display.py:627: postive ==> positive
./librosa/display.py:1433: postive ==> positive
./librosa/util/utils.py:726: mininum ==> minimum
./librosa/core/notation.py:3: utilties ==> utilities
./librosa/core/notation.py:127: indicies ==> indices
./librosa/core/notation.py:158: indicies ==> indices
./librosa/core/spectrum.py:1291: bandwith ==> bandwidth
```
***BEFORE POSTING A BUG REPORT*** Please look through [existing issues (both open and closed)](https://github.com/librosa/librosa/issues?q=is%3Aissue) to see if it's already been reported or fixed!

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Software versions***

**Additional context**

cclauss:
#### Reference Issue

Fixes #1348

#### Any other comments?
Output: https://github.com/cclauss/librosa/actions

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,0 +1,21 @@
+name: lint_python
+on: [pull_request, push]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install bandit black codespell flake8 isort mypy pytest pyupgrade safety
+      - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
+      - run: black --check . || true
+      - run: codespell || true  # --ignore-words-list="" --skip=""
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88 --show-source --statistics
+      - run: isort --check-only --profile black . || true
+      - run: pip install -r requirements.txt || true
+      - run: mypy --ignore-missing-imports . || true
+      - run: pytest . || true
+      - run: pytest --doctest-modules . || true
+      - run: shopt -s globstar && pyupgrade --py36-plus **/*.py || true
+      - run: safety check
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -9,9 +9,9 @@ jobs:
       - run: pip install bandit black codespell flake8 isort mypy pytest pyupgrade safety
       - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
       - run: black --check . || true
-      - run: codespell --ignore-words-list="ba" --skip="*demo.ipynb" || true
+      - run: codespell --ignore-words-list="ba,trough,ue" --skip="*demo.ipynb" || true
       - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88 --show-source --statistics
+      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=120 --show-source --statistics
       - run: isort --check-only --profile black . || true
       - run: pip install -r requirements.txt || true
       - run: mypy --ignore-missing-imports . || true
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -11,7 +11,7 @@ jobs:
       - run: black --check . || true
       - run: codespell --ignore-words-list="ba,trough,ue" --skip="*demo.ipynb" || true
       - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=120 --show-source --statistics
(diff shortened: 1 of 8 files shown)


bmcfee:
Thanks for this!  I hadn't heard of codespell before, seems like a nice addition.

I'm a bit confused by some of the scripting in the action though, specifically the use of `|| true` to bypass any failures.  As it turns out, most of the lines in the action actually do fail, but this is obscured by the shortcut operator.  I don't think most of these lines are necessary though (pytest, doctest, mypy, pyupgrade, etc) and it would be better to remove them IMO.  Since this PR is nominally for linting, I'm not sure what the rest of these steps are intended to do.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,21 @@
...
+on: [pull_request, push]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install bandit black codespell flake8 isort mypy pytest pyupgrade safety
+      - run: bandit --recursive --skip B101,B110 .
+      - run: black --check . || true
```
bmcfee, line 11: This check fails because it includes sphinx-gallery examples.  These examples have specific formatting requirements, and I don't think should be subject to code style constraints from black.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,21 @@
...
+      - run: codespell --ignore-words-list="ba,trough,ue" --skip="*demo.ipynb" || true
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: flake8 . --count --exit-zero --max-complexity=28 --max-line-length=204 --show-source --statistics
+      - run: isort --check-only --profile black . || true
```
bmcfee, line 15: This check fails, primarily on tests.  However, the test modules do imports in a specific order so we can disable librosa's cache module at runtime.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,21 @@
...
+      - run: pip install -r requirements.txt || true
```
bmcfee, line 16: We don't use requirements.txt, so this fails.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,21 @@
...
+      - run: mypy --ignore-missing-imports . || true
```
bmcfee, line 17: What does this line do?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,21 @@
...
+      - run: pytest . || true
```
bmcfee, line 18: This line fails because pip install above failed.  However, we have extensive test runners elsewhere, and it should not be needed here for a linting action.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,21 @@
...
+      - run: pytest --doctest-modules . || true
```
bmcfee, line 19: We don't use doctests.  This is mainly because it's a bit too strict to be practical when many of our example outputs derive from either randomness or numerically sensitive floating point operations.  As a result, example code is often "correct" but gives slightly different output values to what is encoded in the docstrings.  I don't think it's worth pursuing to try to make these fully replicable and stably tested.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,21 @@
...
+      - run: shopt -s globstar && pyupgrade --py36-plus **/*.py || true
```
bmcfee, line 20: What exactly does this do?  We already have tests on 3.6-3.10.

cclauss:
The idea behind `|| true` is to see the results of failing test while a allowing the tests that follow to run.  This allows reviewer to suggest modifictions to make tests pass or eliminate tests which do not make sense for the project.

The line `run: pip install bandit black codespell flake8 isort mypy pytest pyupgrade safety` is a signal to the revier to read up on any of these packages that are new to them.  `mypy` is a project started by the creater of Python the type hints.  `pyupgrade` actually modifies code to modernize code and remove legacy syntax.  I start with `pytest` in please just to see if unittests and/or doctests can be found and if they pass.

In the review process, we can eliminate any lines / tools that do not make sense for the project.  We can also attempt to fix failing tests and remove the corresponding `|| true`.

bmcfee:
Thanks for the clarifications, and for simplifying the action.

I'm generally happy with how it stands now, though I still think the checks should only be applied to the librosa package and not the entire source tree.  That is, `flake8 librosa` instead of `flake8 .`.

I'm also happy to roll in fixes for #1348 to this PR so we can remove the failure suppression without needing a separate PR.

bmcfee (approved):

bmcfee:
Thanks again @cclauss !  This LGTM.  Anything else we should consider here, or is it good to merge?

cclauss:
Good to merge.

PR #1351, opened 2021-07-06
The big Multi-Channel PR

bmcfee:
#### Reference Issue
This PR begins the work of #1130, extending librosa to support multi-channel wherever possible.

#### What does this implement/fix? Explain your changes.

This branch will serve as the target for smaller pull requests implementing multi-channel support throughout the package.

As a first step, I've implemented multi-channel stft.  It's numerically equivalent to our previous implementation, and comparably efficient for single-channel inputs.
For multi-channel inputs, it's not as efficient as I think is possible, but optimization will take a bit of sleuthing.

#### Any other comments?

This PR will not be merged until #1130 is complete.

PR #1365, opened 2021-08-17
Multichannel beat

iranroman:
#### Reference Issue

Implements #1130

work in progress for multichannel beat

### DIFF
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -372,7 +372,9 @@ def tempo(
         max_idx = np.argmax(bpms < max_tempo)
         logprior[:max_idx] = -np.inf
     # explicit axis expansion
-    logprior = np.expand_dims(logprior, axis=(-1, *range(tg.ndim-2)))
+    shape = [1 for _ in tg.shape]
+    shape[-2] = len(logprior)
+    logprior = logprior.reshape(shape)
 
     # Get the maximum, weighted by the prior
     # Using log1p here for numerical stability
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -356,10 +356,10 @@ def tempo(
 
     # Eventually, we want this to work for time-varying tempo
     if aggregate is not None:
-        tg = aggregate(tg, axis=1, keepdims=True)
+        tg = aggregate(tg, axis=-1, keepdims=True)
 
     # Get the BPM values for each bin, skipping the 0-lag bin
-    bpms = core.tempo_frequencies(tg.shape[0], hop_length=hop_length, sr=sr)
+    bpms = core.tempo_frequencies(tg.shape[-2], hop_length=hop_length, sr=sr)
 
     # Weight the autocorrelation by a log-normal distribution
     if prior is None:
@@ -374,7 +374,7 @@ def tempo(
 
     # Get the maximum, weighted by the prior
     # Using log1p here for numerical stability
-    best_period = np.argmax(np.log1p(1e6 * tg) + logprior[:, np.newaxis], axis=0)
+    best_period = np.argmax(np.log1p(1e6 * tg) + logprior[:, np.newaxis], axis=-2)
 
     return bpms[best_period]
 
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -543,29 +543,34 @@ def plp(
     )
 
     if tempo_min is not None:
-        ftgram[tempo_frequencies < tempo_min] = 0
+        ftgram[..., tempo_frequencies < tempo_min, :] = 0
     if tempo_max is not None:
-        ftgram[tempo_frequencies > tempo_max] = 0
+        ftgram[..., tempo_frequencies > tempo_max, :] = 0
+
+    # reshape lengths to match dimension properly
+    shape = [1 for _ in ftgram.shape]
+    shape[-2] = len(tempo_frequencies)
+    tempo_frequencies = tempo_frequencies.reshape(shape)
 
     # Step 3: Discard everything below the peak
     ftmag = np.log1p(1e6 * np.abs(ftgram))
     if prior is not None:
-        ftmag += prior.logpdf(tempo_frequencies)[:, np.newaxis]
+        ftmag += prior.logpdf(tempo_frequencies)
 
-    peak_values = ftmag.max(axis=0, keepdims=True)
+    peak_values = ftmag.max(axis=-2, keepdims=True)
     ftgram[ftmag < peak_values] = 0
 
     # Normalize to keep only phase information
-    ftgram /= util.tiny(ftgram) ** 0.5 + np.abs(ftgram.max(axis=0, keepdims=True))
+    ftgram /= util.tiny(ftgram) ** 0.5 + np.abs(ftgram.max(axis=-2, keepdims=True))
 
     # Step 5: invert the Fourier tempogram to get the pulse
-    pulse = core.istft(ftgram, hop_length=1, length=len(onset_envelope))
+    pulse = core.istft(ftgram, hop_length=1, length=onset_envelope.shape[-1])
 
     # Step 6: retain only the positive part of the pulse cycle
-    np.clip(pulse, 0, None, pulse)
+    pulse = np.clip(pulse, 0, None, pulse)
 
     # Return the normalized pulse
-    return util.normalize(pulse)
+    return util.normalize(pulse, axis=-1)
 
 
 def __beat_tracker(onset_envelope, bpm, fft_res, tightness, trim):
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -371,12 +371,14 @@ def tempo(
     if max_tempo is not None:
         max_idx = np.argmax(bpms < max_tempo)
         logprior[:max_idx] = -np.inf
+    # explicit axis expansion
+    logprior = np.expand_dims(logprior, axis=(-1, *range(tg.ndim-2)))
 
     # Get the maximum, weighted by the prior
     # Using log1p here for numerical stability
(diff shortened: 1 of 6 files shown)


iranroman:
Sharing the changes I've done towards the multichannel beat module. So far I've written the onset strength function and corresponding test. Looking forward to your comments and advice.

bmcfee:
Can you open this pr against the multichannel branch instead? We haven't merged to main yet, so this pr is showing a bunch of unrelated changes.

bmcfee:
Ah, it turns out that I could change the PR target branch from the web interface (but not the app), so disregard the previous comment.

That said, you have conflicts, so please resolve those first.

iranroman:
> That said, you have conflicts, so please resolve those first.

Thanks for the heads up. It was just a missing `optimize=True` that I did not have in my temporary multichannel melspectrogram implementation.

librosa/onset.py
```diff
@@ -345,7 +345,7 @@ def onset_strength(
         **kwargs,
     )
 
-    return odf_all[0]
+    return np.squeeze(odf_all)
```
bmcfee, line 348: This seems dangerous to me.  `np.squeeze` removes all singleton dimensions, where the old code was only removing one dimension.

Would `return odf_all[..., 0, :]` work instead?

librosa/onset.py
```diff
@@ -602,22 +602,27 @@ def onset_strength_multi(
...
     pad_width = lag
     if center:
         # Counter-act framing effects. Shift the onsets by n_fft / hop_length
         pad_width += n_fft // (2 * hop_length)
 
-    onset_env = np.pad(onset_env, ([0, 0], [int(pad_width), 0]), mode="constant")
+    onset_env_ndims = onset_env.ndim
+    onset_env = np.pad(onset_env, 
+                    [[0,0] if i != (onset_env_ndims-1) else [int(pad_width),0] 
+                        for i in range(onset_env_ndims)],
```
bmcfee, line 616: I'd prefer to expand this list comprehension out into something more readable:
```python
padding = [(0,0) for _ in onset_env.shape]
padding[-1] = (int(pad_width), 0)
onset_env = np.pad(onset_env, padding, mode='constant')
```

We use this pattern elsewhere, and it's a bit more readable IMO.

tests/test_multichannel.py
```diff
@@ -59,6 +92,25 @@ def test_stft_multi(y_multi):
...
+
+    D0 = librosa.onset.onset_strength(y[0])
+    D1 = librosa.onset.onset_strength(y[1])
+
+    # Check each channel
+    assert np.allclose(D[0], D0, atol=1e-1, rtol=1e-1) # test in dB units
+    assert np.allclose(D[1], D1, atol=1e-1, rtol=1e-1) # test in dB units
+
+    # Check that they're not both the same
+    assert not np.allclose(D0, D1, atol=1e-1, rtol=1e-1) # test in dB units
```
bmcfee, line 111: These tolerances are way loose.  How far off are the results actually?  (Also I don't think they're in dB)

iranroman: mmhmm, then there's something I do not understand here: 
https://github.com/librosa/librosa/blob/fad5538330312635a7ac91a9b9bb92d078b16593/librosa/onset.py#L571

Here are some general stats on the comparison being made during the test.

```
np.abs(D[0] - D0)
min: 0.0
max: 0.14240837
mean: 0.008507246

np.abs(D[1] - D1)
min: 0.0
max: 1.4305115e-06
mean: 7.587927e-08


np.abs(D0 - D1)
min: 0.0
max: 0.44862747
mean: 0.100642346
```

bmcfee: Ah, right.  This is another instance of the `power_to_db` calculation being sensitive to all channels simultaneously, as noted here: https://github.com/librosa/librosa/pull/1351#issuecomment-894711524

TLDR: it truncates to 80dB below peak, but the peak finder operates across all dimensions.  (I think this is correct behavior, and what we want most of the time.)

The fix here, I think, would be to use spectrogram input directly instead of implicitly computing a melspec by providing the time-domain input.

bmcfee (changes requested):

librosa/core/spectrum.py
```diff
@@ -245,6 +245,7 @@ def stft(
 
     if dtype is None:
         dtype = util.dtype_r2c(y.dtype)
+        print(dtype)
```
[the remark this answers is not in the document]
iranroman: hehehe, oops (busted!)

librosa/beat.py
```diff
@@ -374,7 +374,7 @@ def tempo(
 
     # Get the maximum, weighted by the prior
     # Using log1p here for numerical stability
-    best_period = np.argmax(np.log1p(1e6 * tg) + logprior[:, np.newaxis], axis=0)
+    best_period = np.argmax(np.log1p(1e6 * tg) + logprior[:, np.newaxis], axis=-2)
 
     return bpms[best_period]
```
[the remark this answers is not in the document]
iranroman: Done, but not clear to me whether you mean `best_period` may be a multi-element vector (instead of just a list with one scalar), or something with more dimensions (i.e. matrix or ndarray).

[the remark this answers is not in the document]
bmcfee: > or something with more dimensions (i.e. matrix or ndarray).

Yes: if you do `np.argmax(X, axis=n)`, the result will have shape equal to `X.shape` except dimension `n` is removed.  

In our case, we used to have `tg.ndim=2`, so argmaxing results in a 1d array, so `best_period` has shape `(n_frames,)`.  When we expand out to multichannel, we'll be collapsing out the tempo dimension, but retaining channels and frames.

My comment here comes from the concern that using a multidimensional index array can sometimes do unexpected things (see [here](https://numpy.org/doc/stable/reference/arrays.indexing.html#advanced-indexing)).  However, in this case we're okay because `bpms` is always 1d, so we don't get into any of the weird cases.

librosa/beat.py
```diff
@@ -371,12 +371,14 @@ def tempo(
     if max_tempo is not None:
         max_idx = np.argmax(bpms < max_tempo)
         logprior[:max_idx] = -np.inf
+    # explicit axis expansion
+    logprior = np.expand_dims(logprior, axis=(-1, *range(tg.ndim-2)))
```
bmcfee, line 375: Fancy :monocle_face:

But this relies on deprecated functionality in expand_dims when `tg.ndim-2 > logprior.ndim`.  See the deprecation note [here](https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html).

librosa/feature/rhythm.py
```diff
@@ -139,6 +139,7 @@ def tempogram(
         raise ParameterError("win_length must be a positive integer")
 
     ac_window = get_window(window, win_length, fftbins=True)
+    ac_window = ac_window.reshape((-1,1)) # explicit broadcast
```
bmcfee, line 143: This needs to expand out to the full shape of the framed onset envelope, not just one leading dimension.

tests/test_multichannel.py
```diff
@@ -59,6 +92,119 @@ def test_stft_multi(y_multi):
...
+
+    # Check each channel
+    assert np.allclose(D[0], D0)
+    assert np.allclose(D[1], D1)
+
+    # Check that they're not both the same
+    assert not np.allclose(D0, D1)
+
+
+def test_tempogram(y_multi):
```
bmcfee, line 116: this test should just use `s_multi` instead of y_multi`.

If `onset_envelope` is provided, then `y` will not be used in the tempogram calculations.

tests/test_multichannel.py
```diff
@@ -59,6 +92,119 @@ def test_stft_multi(y_multi):
...
+
+    # Check each channel
+    assert np.allclose(t[0], t0)
+    assert np.allclose(t[1], t1)
+
+    # Check that they're not both the same
+    assert not np.allclose(t0, t1)
+
+
+def test_fourier_tempogram(y_multi):
```
bmcfee, line 139: same comment as above, we should use the `s_multi` fixture here

bmcfee:
Thanks @iranroman !  This all looks good now, modulo a couple of small comments on the tests.

Only other remaining question is if this PR will include plp or not?  I'm fine with either way, just curious.

iranroman:
Thanks for the feedback and guidance @bmcfee. 

> Only other remaining question is if this PR will include plp or not? 

I'll try to get plp done today (as well as the test changes)

tests/test_multichannel.py
```diff
@@ -201,7 +203,64 @@ def test_tempo_multi(y_multi):
 
     # Check that they're not both the same
     assert not np.allclose(t0, t1)
-    print(t,t0,t1)
+
+
```
bmcfee, line 207: PLP Implementation looks good!

Some notes on the test:

1. It looks like `ctx` is always `dnr` (do not raise) here, in which case we should be able to skip it entirely.
2. One hop and window is sufficient here, no use in burning cycles for things we already test elsewhere
3. `use_onset` parameter is not used in the test.  I don't think it's critical that we test multi-channel with and without onset envelope construction since the code path is identical and we test that elsewhere.  May as well remove that parameter.

iranroman: Thanks!
    
    1. I was trying other parameter combinations but I removed some, and now there's only `dnr()`
    2. alright
    3.  same story as 1

Pushing a commit with fixes now.

bmcfee (approved):
Thanks!  This all looks great, happy to merge once CI finishes. Unless there's anything else to add in this PR?

iranroman:
I think for `beat.py` this is it. You said on #1351 that we are only implementing `tempo` and `plp`. So those are done.

PR #1372, opened 2021-08-30
get_duration function now compatible with int16 and other dtypes

tcelox:


bmcfee (approved):
LGTM, thanks!  This might conflict a bit with #1371 and related changes, but I'll deal with that after merging.

bmcfee:
Coverage check is failing because the PR removed a covered line in a module with less than 100% coverage.  I'm not worried about this, and am happy to merge.

PR #1374, opened 2021-09-05
multichannel nnls

iranroman:
Reference Issue

Implements #1130

What does this implement/fix? Explain your changes.

work in progress for multichannel beat

### DIFF
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -80,7 +80,7 @@ def mel_to_stft(M, sr=22050, n_fft=2048, power=2.0, **kwargs):
     """
 
     # Construct a mel basis with dtype matching the input data
-    mel_basis = filters.mel(sr, n_fft, n_mels=M.shape[0], dtype=M.dtype, **kwargs)
+    mel_basis = filters.mel(sr, n_fft, n_mels=M.shape[-2], dtype=M.dtype, **kwargs)
 
     # Find the non-negative least squares solution, and apply
     # the inverse exponent.
@@ -237,9 +237,12 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm="ortho", ref=1.0, lifter=0):
     scipy.fftpack.dct
     """
     if lifter > 0:
-        n_mfcc = mfcc.shape[0]
+        n_mfcc = mfcc.shape[-2]
         idx = np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype)
-        lifter_sine = 1 + lifter * 0.5 * np.sin(np.pi * idx / lifter)[:, np.newaxis]
+        shape = [1 for _ in mfcc.shape]
+        shape[-2] = len(idx)
+        idx = idx.reshape(shape)
+        lifter_sine = 1 + lifter * 0.5 * np.sin(np.pi * idx / lifter)
 
         # raise a UserWarning if lifter array includes critical values
         if np.any(np.abs(lifter_sine) < np.finfo(lifter_sine.dtype).eps):
@@ -254,7 +257,7 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm="ortho", ref=1.0, lifter=0):
     elif lifter != 0:
         raise ParameterError("MFCC to mel lifter must be a non-negative number.")
 
-    logmel = scipy.fftpack.idct(mfcc, axis=0, type=dct_type, norm=norm, n=n_mels)
+    logmel = scipy.fftpack.idct(mfcc, axis=-2, type=dct_type, norm=norm, n=n_mels)
     return db_to_power(logmel, ref=ref)
 
 
--- a/librosa/util/_nnls.py
+++ b/librosa/util/_nnls.py
@@ -143,7 +143,7 @@ def nnls(A, B, **kwargs):
     if B.ndim == 1:
         return scipy.optimize.nnls(A, B)[0]
 
-    n_columns = MAX_MEM_BLOCK // (A.shape[-1] * A.itemsize)
+    n_columns = MAX_MEM_BLOCK // (np.prod(B.shape[:-1]) * A.itemsize)
     n_columns = max(n_columns, 1)
 
     # Process in blocks:
--- a/librosa/util/_nnls.py
+++ b/librosa/util/_nnls.py
@@ -149,7 +149,7 @@ def nnls(A, B, **kwargs):
     # Process in blocks:
     if B.shape[-1] <= n_columns:
         return _nnls_lbfgs_block(A, B, **kwargs).astype(A.dtype)
-    
+
     x = np.einsum('fm,...mt->...ft', np.linalg.pinv(A), B, optimize=True)
     np.clip(x, 0, None, out=x)
     x_init = x
@@ -159,5 +159,4 @@ def nnls(A, B, **kwargs):
         x[..., bl_s:bl_t] = _nnls_lbfgs_block(
             A, B[..., bl_s:bl_t], x_init=x_init[..., bl_s:bl_t], **kwargs
         )
-
     return x
--- a/librosa/util/_nnls.py
+++ b/librosa/util/_nnls.py
@@ -12,6 +12,7 @@
 import numpy as np
 import scipy.optimize
 from .utils import MAX_MEM_BLOCK
+from .exceptions import ParameterError
 
 
 __all__ = ["nnls"]
@@ -28,7 +29,7 @@ def _nnls_obj(x, shape, A, B):
     diff = np.einsum('mf,...ft->...mt',A,x, optimize=True) - B
 
     # Compute the objective value
-    value = 0.5 * np.sum(diff ** 2, axis=(-1,-2))
+    value = 0.5 * np.sum(diff ** 2)
 
     # And the gradient
     grad = np.einsum('fm,...mt->...ft',A.T,diff, optimize=True)
@@ -79,6 +80,23 @@ def _nnls_lbfgs_block(A, B, x_init=None, **kwargs):
     )
     # reshape the solution
     return x.reshape(shape)
+
+
+def _process_in_blocks(A, B, n_columns, **kwargs):
+    # Process in blocks:
+    if B.shape[-1] <= n_columns:
+        return _nnls_lbfgs_block(A, B, **kwargs).astype(A.dtype)
+    
+    x = np.einsum('fm,...mt->...ft', np.linalg.pinv(A), B, optimize=True)
+    np.clip(x, 0, None, out=x)
+    x_init = x
+    
+    for bl_s in range(0, x.shape[-1], n_columns):
+        bl_t = min(bl_s + n_columns, B.shape[-1])
+        x[..., bl_s:bl_t] = _nnls_lbfgs_block(
+            A, B[..., bl_s:bl_t], x_init=x_init[..., bl_s:bl_t], **kwargs
+        )
+    return x
 
 
 def nnls(A, B, **kwargs):
@@ -146,17 +164,13 @@ def nnls(A, B, **kwargs):
     n_columns = MAX_MEM_BLOCK // (A.shape[-1] * A.itemsize)
     n_columns = max(n_columns, 1)
 
-    # Process in blocks:
-    if B.shape[-1] <= n_columns:
-        return _nnls_lbfgs_block(A, B, **kwargs).astype(A.dtype)
-
-    x = np.einsum('fm,...mt->...ft', np.linalg.pinv(A), B, optimize=True)
-    np.clip(x, 0, None, out=x)
-    x_init = x
-
-    for bl_s in range(0, x.shape[-1], n_columns):
-        bl_t = min(bl_s + n_columns, B.shape[-1])
-        x[..., bl_s:bl_t] = _nnls_lbfgs_block(
-            A, B[..., bl_s:bl_t], x_init=x_init[..., bl_s:bl_t], **kwargs
-        )
-    return x
+    # if single channel
+    if B.ndim == 2:
+        x = _process_in_blocks(A,B,n_columns,**kwargs)    
+    elif B.ndim == 3:
+        x = []
+        for iB in B:
+            x.append(_process_in_blocks(A,iB,n_columns,**kwargs))
+    else:
+        raise ParameterError("Target matrix B should have ndim <= 3 but has B.ndim: {:d}".format(B.ndim))
+    return np.array(x)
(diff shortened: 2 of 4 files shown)


iranroman:
sharing my work for nnls. The main issue is that the optimization flattens the basis and thus multichannel optimization depends on flattening ordering. I need to think more about how to make this perfect in the context of `lbfgsb`

iranroman:
Yeah, the default `order='C'` flattening is resulting in strange behavior in the multi-channel test. 

Alternatively, if `_nnls_obj` returns `return value, grad.flatten(order='F')`, my multichannel test passes within a `1e-6` error tolerance for all assertions. 

The investigation continues.

iranroman:
We could also simply add a "channel" `for` loop outside the `for bl_s` loop in `nnls`. This would be the simplest (but hackiest) solution. Otherwise I'll have study what is `scipy` doing with flattening order in its `lbfgsb` optimization.

librosa/util/_nnls.py
```diff
@@ -25,13 +25,13 @@ def _nnls_obj(x, shape, A, B):
     x = x.reshape(shape)
 
     # Compute the difference matrix
-    diff = np.dot(A, x) - B
+    diff = np.einsum('mf,...ft->...mt',A,x, optimize=True) - B
 
     # Compute the objective value
-    value = 0.5 * np.sum(diff ** 2)
+    value = 0.5 * np.sum(diff ** 2, axis=(-1,-2))
```
bmcfee, line 31: do we need target axes here?  value should be scalar.

bmcfee:
> Yeah, the default `order='C'` flattening is resulting in strange behavior in the multi-channel test.
> 
> Alternatively, if `_nnls_obj` returns `return value, grad.flatten(order='F')`, my multichannel test passes within a `1e-6` error tolerance for all assertions.

Yeah, that makes sense.  I hit a similar snag recently in this commit https://github.com/librosa/librosa/pull/1351/commits/c23850a28d90e71528d150ba4e62a952242ce3a1  There's a reusable pattern here that you might want to look at (transpose/reshape), but it seems like you already have it working reasonably enough.

I think order=F is correct here, as it will unwrap column-wise rather than row-wise.

tests/test_multichannel.py
```diff
@@ -902,3 +902,30 @@ def test_click_multi():
...
+    # channel 1
+    M1 = librosa.feature.melspectrogram(S=S[1], sr=sr, power=1)
+    S1_recover = librosa.util.nnls(mel_basis, M1)
+
+    # Check each channel
+    assert np.allclose(S_recover[0], S0_recover, atol = 0.5, rtol = 0.5)
+    assert np.allclose(S_recover[1], S1_recover, atol = 0.5, rtol = 0.5)
+
+    # Check that they're not both the same
+    assert not np.allclose(S0_recover, S1_recover, atol = 0.5, rtol = 0.5)
```
bmcfee, line 931: These tolerances should be much lower; there's no reason we shouldn't have strict numerical precision on this operation.

Side note: it might be better if you construct M from mel_basis and S rather than going through melspectrogram here, just to cut down on the number of moving parts.

librosa/util/_nnls.py
iranroman: I thought that keeping the sum for each channel independently would help to make my test pass. In any case, the results are the same. I guess the optimizer does a summation on its own?

tests/test_multichannel.py
iranroman: Thanks for the comment. I have followed your suggestion in the side note, but the results are still the same.

iranroman: The ravel/flatten associated with `lbfgsb` coefficient optimization is messing with the precision.

iranroman:
> I think order=F is correct here, as it will unwrap column-wise rather than row-wise.

Changing `_nnls_obj` to return `return value, grad.flatten(order='F')` does solve the precision issues for my multichannel test. However, the original `nnls` tests in `tests/test_util.py` now fail.

iranroman:
Settled for the looping over channels trick after addressing and trying everything else discussed in this PR and still seeing precision issues.

librosa/util/_nnls.py
```diff
@@ -25,13 +26,13 @@ def _nnls_obj(x, shape, A, B):
     x = x.reshape(shape)
```
bmcfee, line 26: I think you need order=F here to make sure that flatten/unflatten is done consistently

librosa/util/_nnls.py
```diff
@@ -25,13 +26,13 @@ def _nnls_obj(x, shape, A, B):
...
 
     # Compute the objective value
     value = 0.5 * np.sum(diff ** 2)
 
     # And the gradient
-    grad = np.dot(A.T, diff)
+    grad = np.einsum('fm,...mt->...ft',A.T,diff, optimize=True)
 
     # Flatten the gradient
     return value, grad.flatten()
```
bmcfee, line 38: This should be `grad.flatten(order='F')`

librosa/util/_nnls.py
```diff
@@ -63,11 +64,11 @@ def _nnls_lbfgs_block(A, B, x_init=None, **kwargs):
     # If we don't have an initial point, start at the projected
     # least squares solution
     if x_init is None:
-        x_init = np.linalg.lstsq(A, B, rcond=None)[0]
+        x_init = np.einsum('fm,...mt->...ft', np.linalg.pinv(A), B, optimize=True)
         np.clip(x_init, 0, None, out=x_init)
 
     # Adapt the hessian approximation to the dimension of the problem
-    kwargs.setdefault("m", A.shape[1])
+    kwargs.setdefault("m", A.shape[-1])
```
bmcfee, line 71: We don't need to use -1 here. It's okay to restrict to A being a 2d array (matrix) here.  Only B needs to support multichannel.

librosa/util/_nnls.py
```diff
@@ -146,17 +164,13 @@ def nnls(A, B, **kwargs):
...
+    # if single channel
+    if B.ndim == 2:
+        x = _process_in_blocks(A,B,n_columns,**kwargs)    
+    elif B.ndim == 3:
+        x = []
+        for iB in B:
+            x.append(_process_in_blocks(A,iB,n_columns,**kwargs))
+    else:
+        raise ParameterError("Target matrix B should have ndim <= 3 but has B.ndim: {:d}".format(B.ndim))
+    return np.array(x)
```
bmcfee, line 176: Please revert this change.  We should not restrict to ndim<3; if it doesn't work using the reshape operations from before, then something is wrong.

librosa/util/_nnls.py
iranroman: Yeah, I changed `order=F` all over and the original `test_util.py` tests passed, but the same issue persisted with my test.

librosa/util/_nnls.py
```diff
@@ -25,13 +26,13 @@ def _nnls_obj(x, shape, A, B):
...
     # Compute the difference matrix
-    diff = np.dot(A, x) - B
+    diff = np.einsum('mf,...ft->...mt',A,x, optimize=True) - B
 
     # Compute the objective value
     value = 0.5 * np.sum(diff ** 2)
 
     # And the gradient
-    grad = np.dot(A.T, diff)
+    grad = np.einsum('fm,...mt->...ft',A.T,diff, optimize=True)
```
bmcfee, line 35: we could even skip this transpose here, since einsum is awesome:
```python
    grad = np.einsum('mf,...mt->...ft',A,diff, optimize=True)
```

bmcfee:
Interesting that the tests are passing except on the py37 environment.  I don't think this has anything to do with python 3.7 per se, but seeing as the tests are deterministic, it's a little troubling.

One other thing that occurred to me here: `nnls` works by chopping the input `B` into blocks, similar to stft.  This was originally designed to balance a cpu-memory trade-off, which is where this line comes from:

https://github.com/librosa/librosa/blob/9732ae0775d359e24172bd396e98b44b7cddf19a/librosa/util/_nnls.py#L146-L147

In the current PR, this is unchanged, but note that it does not account for the number of channels in the input signal, so the memory balance is no longer fixed.  In STFT, I modified the block size to work as follows:

https://github.com/librosa/librosa/blob/349b95d39399ec4f6fd05962a4d0a03a33aa2ae1/librosa/core/spectrum.py#L252-L253

Note that it scales up by the product of leading channels in the input, so the number of columns per block will decrease proportionally for multichannel inputs.

Why does this matter for LBFGS?  Remember that it's performing a high-dimensional convex optimization problem, and using (approximate) second order information to accelerate convergence.  This second-order information depends on all input dimensions simultaneously, so increasing the number of dimensions (channels and columns per block) might have some effect on the convergence of the method, as well as its memory consumption.

Similar reasoning may apply here:
https://github.com/librosa/librosa/blob/349b95d39399ec4f6fd05962a4d0a03a33aa2ae1/librosa/util/_nnls.py#L69-L70
which we might want to scale up as `m=np.prod(B.shape[:-1])`.

iranroman:
Thanks for your feedback. I am pushing a version with the change `n_columns = util.MAX_MEM_BLOCK // (np.prod(B.shape[:-1]) * A.itemsize) `. Hopefully this will make the test pass in python 3.7

Changing the other line `kwargs.setdefault("m", np.prod(B.shape[:-1])) ` was causing `test_util.py` to (marginally) not pass.

bmcfee (approved):

bmcfee:
Excellent, thanks @iranroman !  

I think this mostly covers the feature inversion functions as well, except for a couple of axis targets, eg here: https://github.com/librosa/librosa/blob/349b95d39399ec4f6fd05962a4d0a03a33aa2ae1/librosa/feature/inverse.py#L240-L257

Do you want to roll those changes into this PR, or should I merge this and handle those separately?

iranroman:
I've quickly addressed those axes targets. Maybe we should include some multichannel tests before calling this a done deal?

bmcfee:
> I've quickly addressed those axes targets. Maybe we should include some multichannel tests before calling this a done deal?

That'd be great.  :grin:

librosa/feature/inverse.py
```diff
@@ -237,9 +237,12 @@ def mfcc_to_mel(mfcc, n_mels=128, dct_type=2, norm="ortho", ref=1.0, lifter=0):
...
     """
     if lifter > 0:
-        n_mfcc = mfcc.shape[0]
+        n_mfcc = mfcc.shape[-2]
         idx = np.arange(1, 1 + n_mfcc, dtype=mfcc.dtype)
-        lifter_sine = 1 + lifter * 0.5 * np.sin(np.pi * idx / lifter)[:, np.newaxis]
+        shape = [1 for _ in mfcc.shape]
+        shape[-2] = len(idx)
+        idx = idx.reshape(shape)
+        lifter_sine = 1 + lifter * 0.5 * np.sin(np.pi * idx / lifter)
```
bmcfee, line 245: This bit can now be done with `util.expand_to(idx, ndim=mfcc.ndim, axes=-2)`

iranroman:
oh boy! isn't this `scipy.optimize.fmin_l_bfgs_b` thing interesting?

tests/test_multichannel.py
```diff
@@ -902,3 +903,90 @@ def test_click_multi():
...
+# -- feature inversion tests
+@pytest.mark.parametrize("power", [1, 2])
+@pytest.mark.parametrize("n_fft", [1024, 2048])
+def test_mel_to_stft_multi(power,  n_fft):
+    srand()
+
+    # Make a random mel spectrum, 4 frames
+    mel_basis = librosa.filters.mel(22050, n_fft, n_mels=128)
+
+    stft_orig = np.random.randn(2, n_fft // 2 + 1, 4) ** power
```
bmcfee, line 944: This seems like it shouldn't work when power=1; randn can produce negative values, but mel_to_stft can't.  It's not a huge deal here, but it might put us in a weird part of the solution space that could cause trouble.

bmcfee:
The test failures here are pretty puzzling, i gotta admit.  The windows build failure seems unrelated to nnls, since it's triggering from a memory allocation failure (?!).

For the nnls test failures, the debug log shows numerically identical results at the beginning and end of the arrays, so there must be errors elsewhere in the data that are large enough to fail the allclose.  It would be helpful to show those, which can be done by adding a second param to the assert:

https://github.com/librosa/librosa/pull/1374/files#diff-d61f35d9354076e4b40f7e1483accd4948f0f63e431e8812c83e20b87083e501R928

becomes 
```python
    assert np.allclose(S_recover[0], S0_recover), np.max(np.abs(S_recover[0] - S0_recover))
```

Making that change would at least help us determine if this is a real cause for concern, or if we can just roll back the tolerances a little.

tests/test_multichannel.py
iranroman: I just adapted the `mel_to_stft` test in `test_features.py`. We might want to change that too then?

bmcfee: Nah, we can leave it as is for consistency / minimal surprise.  I was just flagging it as a possible thing to take note of.

bmcfee:
As a quick followup, I did clone this PR and run it locally on my dev machine, and everything passes fine.  I suspect that the allclose failures we're seeing on CI are down to differences in the underlying BLAS putting us just outside the numerical tolerance threshold.  This should be easy to confirm with the suggested addition to the assertion statements.

iranroman:
I added the function to assert the distance between multichannel and mono recovered signals, and as we can see the numerical error is in the order of `e-5`. I'll add that margin to the test so that it now passes in all (but the window's build)

bmcfee:
Okay, it looks like it's indeed just barely outside of tolerance on some platforms:

```
E       AssertionError: 9.536743e-06
```

If we bump atol to 1e-5, it should be fine here.

Alternately, if we use float64 instead of float32 in the test data, it would probably also work. My preference is for bumping the tolerance though, so that the tests are a bit more explicit in what we're allowing.

Windows ci is still choking on memory. Maybe we can reduce the size of the test data here?  Use a shorter excerpt?

iranroman:
The Window's error is kinda wild `Unable to allocate 6.30 GiB for an array with shape (845689575,) and data type float64`. I'll make the signal even shorter.

iranroman:
Alright. Seems like we are almost there. Not sure what's causing the codecov errors are though.

bmcfee:
> Not sure what's causing the codecov errors are though.

https://codecov.io/gh/librosa/librosa/compare/349b95d39399ec4f6fd05962a4d0a03a33aa2ae1...09f9776d90ccb10b87f045aba8681f091293fa21/src/librosa/util/_nnls.py?before=librosa/util/_nnls.py

It looks like we're no longer hitting the block iteration because we're slicing too small.

bmcfee (approved):
Lgtm, thanks! Anything else here, or should I merge?

iranroman:
I think that's it. Thanks for guiding me through this!

lostanlen:
woah, i'm only reading this now ...

`np.einsum('fm,...mt->...ft', np.linalg.pinv(A), B, optimize=True)` is epic!

congrats @iranroman 🚀

Issue #1353, opened 2021-07-14
(1, n_samples) is not accepted as a mono waveform


dpwe:
**Description**
Librosa typically operates on one-dimensional, one-channel signals whose shape is `(n_samples,)`.  In some cases, stereo signals are accepted in "time-major" format, e.g. of shape `(2, n_samples)`.  But the analogous single-channel signal, whose shape is `(1, n_samples)`, is rejected -- at least by `numpy.resample()`

**To Reproduce**

```
import librosa
import numpy as np

# Conventional mono
mono_signal = np.random.randn(1000)
resampled_mono = librosa.resample(mono_signal, 1000, 2000)

# Conventional stereo
stereo_signal = np.random.randn(2, 1000)
resampled_stereo = librosa.resample(stereo_signal, 1000, 2000)

# Stereo-like single channel
onechan_signal = np.random.randn(1, 1000)
resampled_onechan = librosa.resample(onechan_signal, 1000, 2000)

```

**Expected behavior**
`resampled_mono == resampled_onechan[0]`

**Actual behavior**
```
---------------------------------------------------------------------------
ParameterError                            Traceback (most recent call last)
<ipython-input-7-3776517cab8b> in <module>()
      1 onechan_signal = np.random.randn(1, 1000)
----> 2 resampled_onechan = librosa.resample(onechan_signal, 1000, 2000)
      3 print(resampled_onechan.shape)

1 frames
/usr/local/lib/python3.7/dist-packages/librosa/util/utils.py in valid_audio(y, mono)
    304     elif y.ndim == 2 and y.shape[0] < 2:
    305         raise ParameterError(
--> 306             "Mono data must have shape (samples,). " "Received shape={}".format(y.shape)
    307         )
    308 

ParameterError: Mono data must have shape (samples,). Received shape=(1, 1000)
```

**Software versions**
```
Linux-5.4.104+-x86_64-with-Ubuntu-18.04-bionic
Python 3.7.10 (default, May  3 2021, 02:48:31) 
[GCC 7.5.0]
NumPy 1.19.5
SciPy 1.4.1
librosa 0.8.1
INSTALLED VERSIONS
------------------
python: 3.7.10 (default, May  3 2021, 02:48:31) 
[GCC 7.5.0]

librosa: 0.8.1

audioread: 2.1.9
numpy: 1.19.5
scipy: 1.4.1
sklearn: 0.22.2.post1
joblib: 1.0.1
decorator: 4.4.2
[...]

PR #1382, opened 2021-10-13
[CR needed] Use zero-padding by default most of the time

bmcfee:
#### Reference Issue

Related:
- #1353 
- #1381 

This PR changes the default padding mode for most spectral transforms (stft, cqt, and friends) to `constant` (i.e., zero-padding) instead of reflection padding.  The only place where reflection padding is retained is in HPSS, where it is specifically useful for handling the boundary conditions on median filters.

This is a pretty minor change, but it does result in slightly cleaner features at the boundaries (most of time time), and is probably a better assumption for things like CQT.

Deep in the technical weeds, this change was motivated by the discussion in #1353 , where short or empty signals cause problems for reflection padding (the old default), which in turn results in some counterintuitive behavior when we were generalizing the multichannel functionality.

#### Any other comments?

The only practical drawback I'm seeing so far is that onset detection now fires on the first frame if the signal starts at non-zero values.  Maybe this is expected behavior, but it's certainly different.


bmcfee:
Any objections to merging this one?  @lostanlen ?

PR #1384, opened 2021-10-20
Fix package name typo in doc generation dependencies

ajweiss:
#### Reference Issue

N/A

Fixes a typo in setup.cfg so that `pip install -e .[docs]` will install dependencies for documentation generation.

#### Any other comments?

Out of curiosity, are these generated manually or is there automation that generates them on release?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,7 +67,7 @@ docs =
     matplotlib >= 2.0.0, < 3.3
     sphinx-multiversion >= 0.2.3
     sphinx-gallery >= 0.7
-    spinxcontrib-svg2pdfconverter
+    sphinxcontrib-svg2pdfconverter
     presets
 tests =
     matplotlib >= 3.0


ajweiss:
> Ah how embarrassing! Thanks for catching this.

It's a little annoyingly gotcha-ish, but I was just trying to give the documentation generation process a little attention since you were talking about making it a replacement for the example notebooks.  (ie, turning documentation generation into public/human facing interface of the project for developers looking to add examples)

I actually do have some questions regarding preferred development and runtime environment (anaconda vs. vanilla pip, which at least on Intel means different math libraries which actually caused a problem with the LPC stuff back in the day) but I'm not sure where would be a good place to start the discussion.  I wasted a fair bit of time yesterday getting a dev environment setup again for this stuff and if you think it's useful, would like to at least try to squeeze something useful out of that time by maybe improving some of the contributor/developer documentation to reduce barrier of entry for newcomers.

PR #1404, opened 2021-11-14
Removes unused imports in setup.py & spectral.py

alexmalins:
#### Reference Issue
N/A

This PR fixes three imports in the code that are no longer used.

In setup.py:
- `import sys` no longer needed as of [this commit](https://github.com/librosa/librosa/commit/0c500a454a5e4855f3bb660a43edf3179c9e5666#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7) ("removing 2.7 cruft")
- `from setuptools import find_packages` no longer needed as of PR #1337's switch to using `packages = find:` in setup.cfg

In librosa/feature/spectral.py:
- `from ..core.audio import to_mono` not used since PR #1351 and changes to the `rms()` function ([this commit](https://github.com/librosa/librosa/commit/eb603e7a91598d1e72d3cdeada0ade21a33f9c0c#))

#### Any other comments?
Discovered these while looking for an example of how to split off setup.py into pyproject.toml & setup.cfg (thanks!).

I used pylint to confirm the unused imports. The tests still ran successfully on my local machine after removing them.

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -12,7 +12,7 @@ from .. import filters
 from ..util.exceptions import ParameterError
 
 from ..core.convert import fft_frequencies
-from ..core.audio import zero_crossings, to_mono
+from ..core.audio import zero_crossings
 from ..core.spectrum import power_to_db, _spectrogram
 from ..core.constantq import cqt, hybrid_cqt
 from ..core.pitch import estimate_tuning
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,5 @@
-import sys
 from importlib.machinery import SourceFileLoader
-from setuptools import setup, find_packages
+from setuptools import setup
 
 
 version = SourceFileLoader('librosa.version',


bmcfee:
Thanks for this! Changes look good to me, so I'm happy to merge after ci passes. Unless there's anything else to add here?

bmcfee (approved):

alexmalins:
Thanks! Nope, nothing else to add other than maybe a suggestion for the project to consider incorporating pylint checking? It was catching other unused variables and arguments that I didn't have the time to look into, along with various other recommendations. Nothing major, but I find its checking very useful on other projects.

bmcfee:
> Nope, nothing else to add other than maybe a suggestion for the project to consider incorporating pylint checking?

Thanks.  I usually do a pylint check as part of release auditing, but I guess we could consider rolling it into the CI now.  I've avoided that in the past because most of the things that pylint complains about are specific choices we've made (eg variable names) for specific reasons.  I know that this can be configured, but I haven't made the time to set that up properly.

Issue #18, opened 2013-07-31
Make filterbanks generic


craffel:
As of now the mel fb is triangular, chroma fb is Gaussian.  But, sometimes people use triangular fbs for computing chroma, and so on.  I propose that instead of melfb and chromafb we have triangularfb and gaussianfb (or something) which take as an argument the center frequencies of each filter, so that we could create a triangular-binned chroma fb, for example.  We could also create rectangularfb, etc...

lostanlen:
As part of the HAMR hackathon, I'm addressing this issue today.
I'll try to make a spec this morning on which we could discuss, and then write the code in the afternoon.

### My take on the triangles vs. Gabors discrepancy
The main argument in pro of triangular filter banks is that it is simple to design filter banks satisfying energy conservation (Littlewood-Paley criterion) with them, whatever be the chosen scale of center frequencies : linear, log, or mel. Yet they have a poor temporal localization (essentially sinc squared) owing to the discontinuity of the first derivative in the Fourier domain. So my impression is that they should only be given as default for the mel scale, whereas we can do much better for linear scale and log scale in terms of time-frequency localization while still satisfying the Littlewood-Paley criterion. That gives asymmetric triangular filters, by the way. This argument is in accordance with making the new API non breaking with respect to the current one, i.e. triangular-based MFCC and Gabor-based chromas.

### Vocabulary
I'm not in favor of calling the Gabor filters "gaussians". Granted, they are Gaussians in the Fourier domain, but calling them gaussians clashes dangerously with `scipy.signal.gaussian`, which is a baseband Gaussian window.

### Syntax
I concur with the keyword syntax
```
f = filters.chroma( ... , shape = filter.triangle)
```
proposed by @bmcfee.
IIUC filter.triangle is a function, like filter.gabor would.
This means that each of them should have its own API as well.

bmcfee:
@lostanlen this all sounds great.  The one thing I'd like to sort out is how to deal with parametrizing windows in linear or log space.  I think it should be possible to have a consistent interface for this that acts as expected, but i haven't thought deeply about it yet.

lostanlen:
So @bmcfee sat together for a while and worked on a strategy for this. The plan is now to write a backend function to design a filter in the Fourier domain, with four arguments:
* vector of linearly spaced frequencies involved in STFT, in Hertz
* center frequency in Hertz
* bandwith in Hertz, in terms of full width at half maximum (FWHM)
[...]

Issue #1396, opened 2021-10-29
Better error messages for CQT


BurningKarl:
**Is your feature request related to a problem? Please describe.**
I'm currently building a real-time audio visualization using an RGB LED strip and a Raspberry Pi. In this process, I figured out that `np.fft.fft` is not very good at turning an audio snippet into frequencies, because usually the signal is not periodic. Then, I found this library and its STFT and CQT implementations, which are designed for audio analysis. Thank you for providing this open-source library!

Back to the problem: I experimented a bit with the arguments to CQT using the following code snippet

```python
import librosa
import np

FRAMERATE = 44100
FFT_SIZE = 2 ** 13
num_octaves = 1
leds_per_octave = 10

signal = np.zeros(FFT_SIZE, dtype=np.float64)
amplitudes = np.abs(
    librosa.cqt(
        signal,
        sr=FRAMERATE,
        hop_length=FFT_SIZE * 2,  # The output should only have one frame
        fmin=min_frequency,
        n_bins=num_octaves * leds_per_octave,
        bins_per_octave=leds_per_octave,
        sparsity=0.1,
    )
).reshape(-1)
```
for different values of `min_frequency`. My understanding is that CQT will resample the audio signal for each octave and then use an STFT to get the power (I hope this is the right word) for the requested frequencies in that octave. Now, naturally this will have two limits. The minimum requested frequency `fmin` can't be too low because the length of the audio snippet `signal` is limited, but also the maximum requested frequency (determined by `fmin` and the number of octaves) can't be too high because the frame rate of the signal is limited.

Using `fmin=librosa.note_to_hz("C2")` the first kind of problem occurs and I get a warning:
```
/home/pi/Documents/LightOrgan/venv/lib/python3.7/site-packages/librosa/core/spectrum.py:224:
UserWarning: n_fft=256 is too small for input signal of length=128
```
First, I'm pretty sure this should say "is too large". Second, this is not very helpful and does not relate to the underlying problem (as I understand it) at all. I did neither provide `n_fft=256` nor a signal of `length=128`.

Using `fmin=librosa.not_to_hz("C11")` the maximum requested frequency is too high and I get an error:
```
  File "/home/pi/Documents/LightOrgan/venv/lib/python3.7/site-packages/librosa/filters.py", line 646, in constant_q_lengths
    raise ParameterError("Filter pass-band lies beyond Nyquist")
librosa.util.exceptions.ParameterError: Filter pass-band lies beyond Nyquist
```
[...]

PR #1405, opened 2021-11-15
[CR] generalizing cqt/vqt filterbank construction

### LINKED ISSUE #1073: [Discussion] CQT, VQT, energy preservation, and frequency band alignment
I wanted to make a separate issue to consolidate the discussion around a number of related points that have popped up in the implementation of VQT #1018, inverse CQT #165, and so on.  @lostanlen and I have discussed these things in various places (often offline), but it would be helpful to have a more permanent record.

# Frequency band definitions

In the CQT implementation, we follow the SK2010 definition of the frequency bands, which are essentially left-aligned: `[f, a * f]` (for some constant `a`).  This works well enough, but it makes some calculations awkward, eg in the VQT.  Later on in the CQT toolbox, the definition shifted to a centered representation, eg `[f / sqrt(a), f * sqrt(a)]` (again for some constant `a`), so that the  frequency of each filter is centered within the band.

Some questions:

- Should we add support for centered frequency bands?  I don't think it would change much, except possibly some boundary effects at the top end of octaves.
- Should we make centering the default?  Should we even bother with left-aligned bands anymore?

# Energy preservation

Our CQT implementation has historically had a bunch of headaches around normalization and energy preservation.  Of course we can't exactly preserve energy with a lossy frequency representation, but our current cqt/icqt round trip gets pretty close.  However, other feature representations (notably Mel) are not so forgiving, and it would be nice if we could strive for consistency here.

bmcfee:
#### Reference Issue
Fixes #1073, #18, #1396

This PR will generalize the VQT filterbank construction to allow for non-uniform spacing of filter frequencies.

Additionally, the bandwidth calculation for VQT (and CQT) filters is now determined by the separation between the midpoints to neighboring frequencies `f[k+1], f[k-1]` rather than comparison to the next frequency up.  This is a very small change for CQT (ET)-based analyses, but will give us a bit more flexibility when spacing is non-uniform.

Before I dig further into the implementation, I wonder if we should refactor the interface a bit here?  Or perhaps create a parallel interface entirely?

The CQT filter constructor (and length helper) is quite simple, and I don't see much benefit in extending the API to support arbitrary frequencies.  (This would be very non-constant-Q indeed!)  Instead, we could create a new interface for wavelet construction that could be used directly by VQT, and accept an array of frequencies as parameters rather than BPO, fmin, nbins, etc.  (At some point, we may want to deprecate `filters.constant_q_*` if the functionality is fully supplanted.)  It's already a bit of a stretch having `gamma=` in the CQT constructor, so I think this kind of change might be overdue.  But I wanted to get thoughts from others (ie @lostanlen ) before going ahead with it.

#### Any other comments?

Once this is merged, we can look at implementing just-intonation VQT and related methods, as described in #1402 .

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -462,17 +462,26 @@ def pseudo_cqt(
     # Apply tuning correction
     fmin = fmin * 2.0 ** (tuning / bins_per_octave)
 
-    fft_basis, n_fft, _ = __cqt_filter_fft(
+    freqs = cqt_frequencies(
+        fmin=fmin, n_bins=n_bins, bins_per_octave=bins_per_octave
+    )
+
+    alpha = __bpo_to_alpha(bins_per_octave)
+
+    lengths, _ = filters.wavelet_lengths(
+        freqs, sr=sr, window=window, filter_scale=filter_scale, alpha=alpha
+    )
+
+    fft_basis, n_fft, _ = __vqt_filter_fft(
         sr,
-        fmin,
-        n_bins,
-        bins_per_octave,
+        freqs,
         filter_scale,
         norm,
         sparsity,
         hop_length=hop_length,
         window=window,
         dtype=dtype,
+        alpha=alpha
     )
 
     fft_basis = np.abs(fft_basis)
@@ -492,13 +501,6 @@ def pseudo_cqt(
     if scale:
         C /= np.sqrt(n_fft)
     else:
-        freqs = cqt_frequencies(
-            fmin=fmin, n_bins=n_bins, bins_per_octave=bins_per_octave
-        )
-        alpha = __bpo_to_alpha(bins_per_octave)
-        lengths, _ = filters.wavelet_lengths(
-            freqs, sr=sr, window=window, filter_scale=filter_scale, alpha=alpha
-        )
         # reshape lengths to match dimension properly
         lengths = util.expand_to(lengths, ndim=C.ndim, axes=-2)
 
@@ -1007,54 +1009,6 @@ def vqt(
 
 
 @cache(level=10)
-def __cqt_filter_fft(
-    sr,
-    fmin,
-    n_bins,
-    bins_per_octave,
-    filter_scale,
-    norm,
-    sparsity,
-    hop_length=None,
-    window="hann",
-    gamma=0.0,
-    dtype=np.complex64,
-):
-    """Generate the frequency domain constant-Q filter basis."""
-
-    freqs = cqt_frequencies(fmin=fmin, n_bins=n_bins, bins_per_octave=bins_per_octave)
-    alpha = __bpo_to_alpha(bins_per_octave)
-    basis, lengths = filters.wavelet(
-        freqs,
-        sr=sr,
-        filter_scale=filter_scale,
-        norm=norm,
-        pad_fft=True,
-        window=window,
-        gamma=gamma,
-        alpha=alpha,
-    )
-
-    # Filters are padded up to the nearest integral power of 2
-    n_fft = basis.shape[1]
-
-    if hop_length is not None and n_fft < 2.0 ** (1 + np.ceil(np.log2(hop_length))):
-
-        n_fft = int(2.0 ** (1 + np.ceil(np.log2(hop_length))))
-
-    # re-normalize bases with respect to the FFT window length
-    basis *= lengths[:, np.newaxis] / float(n_fft)
-
-    # FFT and retain only the non-negative frequencies
-    fft = get_fftlib()
-    fft_basis = fft.fft(basis, n=n_fft, axis=1)[:, : (n_fft // 2) + 1]
-
-    # sparsify the basis
-    fft_basis = util.sparsify_rows(fft_basis, quantile=sparsity, dtype=dtype)
-
-    return fft_basis, n_fft, lengths
-
-
 def __vqt_filter_fft(
     sr,
     freqs,
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -626,14 +626,14 @@ def icqt(
     n_bins = C.shape[-2]
 
     # truncate the cqt to max frames if helpful
+    freqs = cqt_frequencies(
+        fmin=fmin, n_bins=n_bins, bins_per_octave=bins_per_octave
+    )
+    alpha = __bpo_to_alpha(bins_per_octave)
+    lengths, _ = filters.wavelet_lengths(
+        freqs, sr=sr, window=window, filter_scale=filter_scale, alpha=alpha
+    )
     if length is not None:
-        freqs = cqt_frequencies(
-            fmin=fmin, n_bins=n_bins, bins_per_octave=bins_per_octave
-        )
-        alpha = __bpo_to_alpha(bins_per_octave)
-        lengths, _ = filters.wavelet_lengths(
-            freqs, sr=sr, window=window, filter_scale=filter_scale, alpha=alpha
-        )
         n_frames = int(np.ceil((length + max(lengths)) / hop_length))
         C = C[..., :n_frames]
 
@@ -666,17 +666,17 @@ def icqt(
         # Slice out the current octave
         sl = slice(bins_per_octave * i, bins_per_octave * i + n_filters)
         C_oct = C[..., sl, :]
+        freq_oct = freqs[sl]
 
-        fft_basis, n_fft, lengths = __cqt_filter_fft(
+        fft_basis, n_fft, lengths = __vqt_filter_fft(
             my_sr,
-            fmin * 2.0 ** i,
-            n_filters,
-            bins_per_octave,
+            freq_oct,
             filter_scale,
             norm,
             sparsity,
             window=window,
             dtype=dtype,
+            alpha=alpha,
         )
 
         # Re-scale the filters to compensate for downsampling
@@ -885,6 +885,7 @@ def vqt(
 
     # First thing, get the freqs of the top octave
     freqs = cqt_frequencies(n_bins=n_bins, fmin=fmin, bins_per_octave=bins_per_octave)
+    
     freqs_top = freqs[-bins_per_octave:]
 
     fmin_t = np.min(freqs_top)
@@ -918,20 +919,22 @@ def vqt(
     vqt_resp = []
 
     # Skip this block for now
+    oct_start = 0
     if auto_resample and res_type != "kaiser_fast":
 
         # Do the top octave before resampling to allow for fast resampling
-        fft_basis, n_fft, _ = __cqt_filter_fft(
+        freqs_top = freqs[-n_filters:]
+
+        fft_basis, n_fft, _ = __vqt_filter_fft(
             sr,
-            fmin_t,
-            n_filters,
-            bins_per_octave,
+            freqs_top,
             filter_scale,
             norm,
             sparsity,
             window=window,
             gamma=gamma,
             dtype=dtype,
+            alpha=alpha,
         )
 
         # Compute the VQT filter response and append it to the stack
@@ -939,31 +942,38 @@ def vqt(
             __cqt_response(y, n_fft, hop_length, fft_basis, pad_mode, dtype=dtype)
         )
 
-        fmin_t /= 2
-        fmax_t /= 2
-        n_octaves -= 1
+        oct_start = 1
 
         res_type = "kaiser_fast"
 
     # Iterate down the octaves
     my_y, my_sr, my_hop = y, sr, hop_length
 
-    for i in range(n_octaves):
-        fft_basis, n_fft, _ = __cqt_filter_fft(
+    for i in range(oct_start, n_octaves):
+
+        # Slice out the current octave of filters
+        if i == 0:
+            sl = slice(-n_filters, None)
+        else:
+            sl = slice(-n_filters*(i+1), -n_filters*i)
+
+        # This may be incorrect with early downsampling
+        freqs_oct = freqs[sl]
+
+        fft_basis, n_fft, _ = __vqt_filter_fft(
             my_sr,
-            fmin_t * 2.0 ** -i,
-            n_filters,
-            bins_per_octave,
+            freqs_oct,
             filter_scale,
             norm,
             sparsity,
             window=window,
             gamma=gamma,
             dtype=dtype,
+            alpha=alpha
         )
 
         # Re-scale the filters to compensate for downsampling
-        fft_basis[:] *= np.sqrt(2 ** i)
+        fft_basis[:] *= np.sqrt(2 ** (i-oct_start))
 
         # Compute the vqt filter response and append to the stack
         vqt_resp.append(
@@ -1030,6 +1040,50 @@ def __cqt_filter_fft(
 
     if hop_length is not None and n_fft < 2.0 ** (1 + np.ceil(np.log2(hop_length))):
 
+        n_fft = int(2.0 ** (1 + np.ceil(np.log2(hop_length))))
+
+    # re-normalize bases with respect to the FFT window length
+    basis *= lengths[:, np.newaxis] / float(n_fft)
+
+    # FFT and retain only the non-negative frequencies
+    fft = get_fftlib()
+    fft_basis = fft.fft(basis, n=n_fft, axis=1)[:, : (n_fft // 2) + 1]
+
+    # sparsify the basis
+    fft_basis = util.sparsify_rows(fft_basis, quantile=sparsity, dtype=dtype)
+
+    return fft_basis, n_fft, lengths
+
+
+def __vqt_filter_fft(
+    sr,
+    freqs,
+    filter_scale,
+    norm,
+    sparsity,
+    hop_length=None,
+    window="hann",
+    gamma=0.0,
+    dtype=np.complex64,
+    alpha=None,
+):
+    """Generate the frequency domain variable-Q filter basis."""
+
+    basis, lengths = filters.wavelet(
+        freqs,
+        sr=sr,
+        filter_scale=filter_scale,
+        norm=norm,
+        pad_fft=True,
+        window=window,
+        gamma=gamma,
+        alpha=alpha,
+    )
+
+    # Filters are padded up to the nearest integral power of 2
+    n_fft = basis.shape[1]
+
+    if hop_length is not None and n_fft < 2.0 ** (1 + np.ceil(np.log2(hop_length))):
         n_fft = int(2.0 ** (1 + np.ceil(np.log2(hop_length))))
 
     # re-normalize bases with respect to the FFT window length
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -875,7 +875,7 @@ def vqt(
     n_filters = min(bins_per_octave, n_bins)
 
     # Relative difference in frequency between any two consecutive bands
-    alpha = 2.0 ** (1.0 / bins_per_octave) - 1
+    alpha = 2.0 ** (0.5 / bins_per_octave) - 2.0**(-0.5 / bins_per_octave)
 
     if fmin is None:
         # C1 by default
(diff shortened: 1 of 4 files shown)


bmcfee:
Ran into a snag when porting over CQT to use the new wavelet constructor.

The alpha derivation in #1073 requires at least 3 frequencies to be defined: `f[k-1], f[k], f[k+1]`.  Some of our unit tests (eg in pseudo-cqt) use only a single frequency in some cases, which causes and IndexError.  What should we do?

Seems to me that alpha is not well-defined here, so an error is appropriate.

bmcfee:
Gave this a bit of thought, I think we can add a special case to the wavelet constructor where `alpha` is explicitly included and only used in the case where `len(freqs)==1`.  This does mean that the call site needs to know something about alpha, which i'm not too happy about.  On the other hand, this information is knowable in the cases where it comes up since it will always result from slicing a large frequency set into smaller pieces.

bmcfee:
Alpha backup hack is in place.  We'll need to relax that when exposing a non-ET wavelet interface, but I'll leave that for the PR that properly addresses #1402 .

At this point, I think this one is ready for CR.

lostanlen (commented):
I have an issue with the docstring of VQT. Quoting:


> If 'gamma=None', gamma will be calculated such that filter bandwidths are equal to a
        constant fraction of the equivalent rectangular bandwidths (ERB). This is accomplished
        by solving for the gamma which gives::
> `B_k = alpha * f_k + gamma = C * ERB(f_k)`

If i understand correctly, the actual bandwdidth is not `B_k` but `B_k / filter_scale`.

librosa/filters.py
```diff
@@ -634,23 +648,308 @@ def constant_q_lengths(
...
+
+        If two or more frequencies are provided, this parameter is ignored.
+
+    Returns
+    -------
+    lengths : np.ndarray
+        The length of each filter.
+
+    f_cutoff : float
+        The lowest frequency at which all filters' main lobes have decayed by
```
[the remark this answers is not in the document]
bmcfee: Lowest is correct here: it's the minimum frequency such that the condition is satisfied by all filters.  I couldn't think of a better way to word that.

librosa/core/constantq.py
```diff
@@ -680,33 +680,29 @@ def icqt(
...
-        y_oct = audio.resample(y_oct, 1, sr // my_sr, res_type=res_type, scale=not scale, fix=False)
+        y_oct = audio.resample(
+            y_oct, 1, sr // my_sr, res_type=res_type, scale=not scale, fix=False
+        )
 
         if y is None:
             y = y_oct
         else:
-            y[..., :y_oct.shape[-1]] += y_oct
+            y[..., : y_oct.shape[-1]] += y_oct
```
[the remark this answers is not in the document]
bmcfee: Yeah, black did it. :confused:

bmcfee:
> If i understand correctly, the actual bandwidth is not `B_k` but `B_k / filter_scale`.

Ah, good point.

librosa/filters.py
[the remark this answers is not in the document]
lostanlen: OK, my bad. 

Maybe still add "This second output serves in vqt to ensure that all wavelet bands remain below the Nyquist frequency." as a second sentence ?

lostanlen:
I like this but i wish we defined `alpha` in the `__bpo_to_alpha` function, and how we came up with that formula.

Considering that the equation of alpha is 

```
B_k = (alpha * f_k + gamma) / filter_scale
```


it's not entirely obvious why it should be equal to `2**(0.5/bpo) - 2**(0.5/bpo)`

bmcfee:
> should we expect `f_k + B_k` and `f_{k+1} - B_{k+1}` to be equal or something like that?

Yup, that's the idea.  The bandwidths are intended to be disjoint and cover the frequency range so that each frequency in [fmin, fmax] hits at least 3dB of some filter.

lostanlen:
Well then :

```
f_{k+1} - B_{k+1}  - (f_k + B_k) = 
f_k * 2**(1/bpo) - f_k * 2**(1.5/bpo) + f_k * 2**(0.5/bpo) - (f_k + 2**(0.5/bpo) f_k - 2**(-0.5/bpo) f_k)
```

Setting `r=2**(0.5/bpo)` we have

```
(f_{k+1} - B_{k+1}  - (f_k + B_k)) / f_k = r**2 - r**3 + r - 1 - r - 1/r
```

Only the `r` term cancels. We obtain:

```
(f_{k+1} - B_{k+1}  - (f_k + B_k)) / f_k = r**2 - r**3 - 1 - 1/r = (r**2 - 1) * (1-1/r) ≠ 0
```

bmcfee:
> Only the `r` term cancels. We obtain:

Yep, gotcha.  They don't intersect exactly, though they approach each other as BPO increases.  For bpo = [12, 24, 36, 48, 60] we get a difference:
```
In [6]: (r**2 - 1)*(1-1/r)
Out[6]: array([1.693e-03, 4.201e-04, 1.863e-04, 1.046e-04, 6.692e-05])
```

lostanlen:
RIght. But i'm more worried about the low-BPO regime. 

I think the best solution would be:

```
alpha = (1 - 2**(1/bpo)) / (1 + 2**(1/bpo))
```

So that: `1 + alpha = 2**(1/bpo) (1 - alpha)`

lostanlen:
Some copy-pasted comments from an offline discussion with @bmcfee 

```python
In [1]: import numpy as np

In [2]: bpo = np.array([12, 24, 36, 48, 60])

In [3]: r = 2**(1/ bpo)

In [4]: r-1
Out[4]: array([0.05946309, 0.02930224, 0.01944064, 0.01454533, 0.01161944])

In [5]: r-1/r
Out[5]: array([0.11558878, 0.0577703 , 0.03851056, 0.02888214, 0.02310542])

In [6]: r**0.5 - r**-0.5
Out[6]: array([0.0577703 , 0.02888214, 0.01925439, 0.01444069, 0.01155252])

In [7]: (r-1)/(r+1)
Out[7]: array([0.02887311, 0.01443956, 0.00962675, 0.00722016, 0.00577616])
```

hm ... we're yet one factor of two below what we had 0__0

this is actually a long-standing design choice in wavelets 
assuming you want one per octave, do you want the bandwidth to be [2/3, 4/3]  ?  or [0.707, 1.414] ?
former --> symmetric, covers exactly one octave. center frequency at the arithmetic midpoint
latter --> asymmetric, covers exactly one octave. center frequency at the geometric midpoint


> i mean it's comforting that we're not the first to run into this, but i'm not sure i see how it helps.  our windows are essentially symmetric no matter what, and our frequency centers are already determined by a geometric series.  seems like it's impossible

that's not a problem. 
[2/3, 4/3] and [4/3, 8/3] are juxtaposed, in geometric progression (center freqs = 1, 2), and symmetric

if you're unhappy about (r-1) / (r+1)  being too small, make it (r**k - 1) / (r**k + 1)  for some k>1
this way your filter won't be meeting the next wavelet but the k 'th next wavelet
(usually in scattering.m  we take k=2 )


> it's not so much that i'm unhappy about it, but that's another parameter that we'd need to control right?
> and then what do you do when frequencies are unevenly spaced?

`f_{n±k}` instead of `f_{n±1}`


```python
>>> bpo = np.array([1, 3, 6, 12, 24, 36, 48, 60])
>>> r = 2**(1/bpo)
>>> (r**2-1)/(r**2+1)
> array([0.6       , 0.22702358, 0.11501333, 0.05769811, 0.02887311,
>        0.01925171, 0.01443956, 0.01155194])
>>> r**0.5 - r**-0.5
> array([0.70710678, 0.23156333, 0.11558878, 0.0577703 , 0.02888214,
>        0.01925439, 0.01444069, 0.01155252])
> 
```

top: my formula but going two filters upward / downward instead of one
bottom: your formula

the difference is only significant for bpo<12
however my formula satisfies `f_n + alpha f_n = f_{n+2} - alpha f_{n+2}`
and vice-versa  `f_n - alpha f_n = f_{n-2} + alpha f_{n-2}`


> back to bandwidth: i'm not seeing how to implement this for unevenly spaced filters so that you get an alpha for each frequency.  looks to me like we have 2n equations in n unknowns: f[n]*(1+a[n]) = f[n+2] *(1 - a[n+2]) and f[n]*(1-a[n]) = f[n-2] *(1 + a[n-2])

yes, with unevenly spaced filters it's impossible. you can imagine that f[n+2]/f[n] is very large in front of 1 while f[n]/f[n-2] is close to 1.

to make it work rigorously and with energy conservation, it would be necessary to move away from Gabor wavelets and replace them by wavelets whose design accommodates a low cutoff and a high cutoff.
spline wavelets come to mind

chapter 7 of the wavelet tour lists some options (although it focuses on the bpo=1 case, the design can be adapted to any sequence of f_k)
https://www.di.ens.fr/~mallat/papiers/WaveletChap7.pdf 
see in particular the Meyer splines between Eqs. 7.84 and 7.89 
but this has (to my knowledge) never been done for unevenly spaced filters. in fact, i don't know any unevenly spaced filterbank that isn't piecewise constant or triangular

all this may quickly turn uncontrollable and bug-ridden. plus, it's a massive breaking change to put splines in librosa instead of Gabors
another, more reasonable option is to imagine that the bpo is local and of the order of
`bpo[n] = 4/log2(f[n+2]/f[n-2])`
for every frequency f[n], this will return a certain bpo number


In the special case of an evenly spaced CQT or VQT, we have bpo[n] = 4/log2(2**(2/B)/2**(-2/B))=4/(4/B) = B , i.e., a constant (as we'd expect)
so our definition of alpha could become


```python
alpha[n] = (2**(2*bpo[n]) - 1) / (2**(2*bpo[n]) + 1)
```

if the bins are evenly spaced in the log-frequency domain, the above guarantees an exact juxtaposition of passbands n-2, n, n+2, etc.
otherwise, the juxtaposition is no longer perfect. but assuming that bpo[n] ~ constant , the coverage should be good enough to avoid bringing the Littlewood-Paley sum too high or too low.
this has the advantage of being backwards compatible in the sense that we can keep Gabor wavelets with symmetric frequency profiles. no need to redesign the wavelet filterbank for the JustIT
considering that our bpo[n]  doesn't vary by huge amounts from n  to n±2  i think that this is the best compromise

bmcfee:
Ok, I've revised the implementation in light of the above.  Bandwidth plot appears to still have good coverage, even with uneven spacing:
image

I'm not 100% satisfied with it yet though, as we now require >=4 bins to compute alpha from local BPO estimates, and it seems a bit hacky the way I've implemented it for now.  There ought to be a more elegant way to handle the edge cases, maybe it'll be clearer after a few more :coffee:

librosa/core/constantq.py
```diff
@@ -1441,4 +1441,5 @@ def __bpo_to_alpha(bins_per_octave):
     alpha : number > 0
     """
 
-    return 2.0 ** (0.5 / bins_per_octave) - 2.0 ** (-0.5 / bins_per_octave)
+    r = 2**(1/bins_per_octave)
+    return (r**2 - 1) / (r**2 + 1)
```
lostanlen, line 1445: nice

librosa/filters.py
```diff
@@ -759,25 +762,18 @@ def wavelet_lengths(
...
-        alpha[0] = (freq_roots[1] - freq_low) / freq_roots[0]
-
-        freq_high = freqs[-1] / freq_roots[-2]
-        alpha[-1] = (freq_high - freq_roots[-2]) / freq_roots[-1]
+    # We need at least 4 frequencies to infer alpha
+    if len(freqs) > 4:
+        # Approximate the local octave resolution
+        bpo = np.empty(len(freqs))
+        logf = np.log2(freqs)
+        bpo[0] = 2/(logf[2] - logf[0])
```
lostanlen, line 770: any way we could pass `bins_per_octave` to this function so as to use octave equivalence for local BPO estimation?

```
bpo[2:-2] = 4/(logf[4:] - logf[:-4])
bpo[:2] = bpo[bins_per_octave:(bins_per_octave+2)]
bpo[-2:] = bpo[-(bins_per_octave+2):-bins_per_octave]
```

where `bins_per_octave` is the user-provided constant

lostanlen: if you're unhappy with the `len(freqs)>4` requirement, another way to estimate the local `bpo` is:

```
bpo[1:-1] = 2/(logf[2:] - logf[:-2])
```

now we have a `len(freqs)>2` requirement

bmcfee: > any way we could pass bins_per_octave to this function so as to use octave equivalence for local BPO estimation?

Eh... we could, but it seems like it's inching toward API bloat.  We already pass through a backup alpha parameter to cover this case, and I don't think we need both.  Maybe BPO is better than alpha?

> if you're unhappy with the len(freqs)>4 requirement, another way to estimate the local bpo is:

I like that; it does seem a bit more direct to me.  But does it satisfy the conditions you had before?

lostanlen: if we work with evenly spaced filters, then `bpo` will be a constant array, so whether we estimate it from second-order differences or fourth-order differences (or indeed, differences of any order) will not change the result: it will remain correct.

> But does it satisfy the conditions you had before?

yes.

lostanlen:
```
logf[n] = logf0 + n/B
logf[n+k] - logf[n-k] = (n+k)/B - (n-k)/B = 2k/B
2k/(logf[n+k] - logf[n-k]) = B
```

set `k=1` for second-order differences. `k=2` for fourth-order differences. the result is always `B`.
in light of this, i think that `k=1` should suffice

bmcfee: Yeah, I'd agree with that.  Even if the filter spacing is not uniform, I expect that it would be preferable to have a more local estimate here anyway given the intended use in controlling bandwidth.

lostanlen: > Eh... we could, but it seems like it's inching toward API bloat.

As i said offline, another option is to "discover" octave equivalence by seeking `(n, B)` such that `logf[n]+logf[n+B]=1` and work from there. No extra kwarg needed

lostanlen: here's how to discover `bins_per_octave`

```python
bins_per_octave = 1 + np.argmax(np.real(np.fft.ifft(np.abs(np.fft.fft(np.modf(logf)[0])))**2)[1:])
```

bmcfee:
Ok, I think we're basically in good shape with this now.  Possibly the comments / documentation could still be improved, but the functionality appears to be there.

librosa/filters.py
```diff
@@ -634,23 +648,302 @@ def constant_q_lengths(
...
+
+    gamma : number >= 0 [scalar, optional]
+        Bandwidth offset for determining filter lengths, as used in
+        Variable-Q transforms.
+
+        Bandwidth for the k'th filter is determined by::
+
+            B[k] = alpha[k] * freqs[k] + gamma
+
+        where ``alpha[k]`` is difference between the geometric means of
```
lostanlen, line 701:
```
``alpha[k]`` is twice the relative difference between ``freqs[k+1]`` and ``freqs[k-1]``:

    alpha[k] = (freqs[k+1]-freqs[k-1]) / (freqs[k+1]+freqs[k-1])

If ``freqs`` follows a geometric progression (as in CQT and VQT), the vector ``alpha`` is constant and such that

    (1 + alpha) * freqs[k-1] = (1 - alpha) * freqs[k+1]

Furthermore, if ``gamma=0`` (default), ``alpha`` is such that even-``k`` and odd-``k`` filters are interleaved:

    freqs[k-1] + B[k-1] = freqs[k+1] - B[k-1]
```

lostanlen:
> Ok, I think we're basically in good shape with this now. Possibly the comments / documentation could still be improved, but the functionality appears to be there.

Thanks. I have suggested an explanation for `__bpo_to_alpha`

lostanlen:
I believe that this looks good and will close #1073 (and probably #1396 although i haven't tracked that other one closely).
However, does this PR really close #18? From what i've understood, a central feature request in #18 was to accommodate triangular filters in CQT/VQT.
(this also has implications for a JI transform. i'm under the impression that the difference in sparsity between JI and ET will be greater with triangular filters than with Gabor filters)

what's the current status of that feature request?

bmcfee:
> However, does this PR really close #18? From what i've understood, a central feature request in #18 was to accommodate triangular filters in CQT/VQT.

Originally, I think this was mainly intended to just provide a unified backend for mel and chroma bases.  It was only much latter that the cqt/vqt connection became apparent.



> It would be more efficient to define them directly in the frequency domain (as we do with `filters.mel`)

That is a good point.  Right now we do all filter construction in the time domain and then expect the caller to convert to the frequency domain.  I suppose we could also provide a purely frequency-domain interface for wavelet construction, but it seems like a lot of work at this point.


My $2e-2 is that this, at present, provides a half solution to #18 - it can construct arbitrary filters, but does not really provide an interface to use them yet.

lostanlen:
> Originally, I think this was mainly intended to just provide a unified backend for mel and chroma bases. It was only much latter that the cqt/vqt connection became apparent.

Yes, that's true.

You're right that it's a half-solution to #18, in the sense that it allows to construct mel-frequency filterbanks with any wavelet basis. (Hann for example)

The other half to #18 is to implement `triangularfb` (here i am re-using the name proposed by @craffel). For this we may take the `filter.mel` implementation and just remove the first line:

```python
    # 'Center freqs' of mel bands - uniformly spaced between limits
    mel_f = mel_frequencies(n_mels + 2, fmin=fmin, fmax=fmax, htk=htk)

    fdiff = np.diff(mel_f)
    ramps = np.subtract.outer(mel_f, fftfreqs)

    for i in range(n_mels):
        # lower and upper slopes for all bins
        lower = -ramps[i] / fdiff[i]
        upper = ramps[i + 2] / fdiff[i + 1]

        # .. then intersect them with each other and zero
        weights[i] = np.maximum(0, np.minimum(lower, upper))

    if norm == "slaney":
        # Slaney-style mel is scaled to be approx constant energy per channel
        enorm = 2.0 / (mel_f[2 : n_mels + 2] - mel_f[:n_mels])
        weights *= enorm[:, np.newaxis]
    else:
        weights = util.normalize(weights, norm=norm, axis=-1)
```


>  I suppose we could also provide a purely frequency-domain interface for wavelet construction, but it seems like a lot of work at this point.

I would say that this is out of scope. and yes, would be a lot of work
(this is what Kymatio does by the way)

lostanlen:
But meanwhile, i'm for merging this and considering #1073 and #1396 as closed.

bmcfee:
I might want to pull in a little refactoring of the vqt guts to make this a bit cleaner. There won't be any external changes here, but it will reduce a little of the redundant code in cqt_filter_fft. I have this working in a notebook, will try to pull it in tomorrow.

bmcfee:
As promised, I've pushed up the minor refactoring of the VQT implementation.  This should make the code a little easier to follow.

The key changes here are:

1. Basis frequencies are constructed once and sliced out for each octave, rather than being constructed independently for each octave.  This makes it much closer to how the `icqt` function operates, and is more readable IMO.
2. Slight change / simplification to early downsampling.  Rather than reducing the frequencies and number of octaves, we now just override the starting point of the octave loop.  This makes the state more transparent.
3. Added a new function `__vqt_filter_fft` that basically does what `__cqt_filter_fft` does but using the new wavelet constructor and accepting frequencies instead of frequency-generation parameters.

Remaining question here:

- Should we completely eliminate `__cqt_filter_fft`?  The only place it's used now is in `pseudo_cqt`.  It would be easy enough to revise that as well, but I wanted to make sure that tests are passing first before completing the refactor.

lostanlen:
The only change i'm seeing is that the `@cache` goes from `10` to `20`. What are the consequences?

bmcfee:
> The only change i'm seeing is that the `@cache` goes from `10` to `20`. What are the consequences?

Where are you seeing that?  It should be 10 for anything that's independent of track content.

lostanlen:
ah sorry, i misinterpreted. i though that the code from `__cqt_filter_fft` (cache=20) was being copied into `pseudo_cqt` (cache=10) but in fact it is replaced by a call to `__vqt_filter_fft` which is also cache=20

bmcfee:
> i though that the code from `__cqt_filter_fft` (cache=20) was being copied into `pseudo_cqt` (cache=10) but in fact it is replaced by a call to `__vqt_filter_fft` which is also cache=20

Is something going haywire with github's code folding?  It should look as follows:

image

bmcfee:
Windows build failure is due to pushing coverage results to codecov; tests passed as expected.

bmcfee:
Anything left to do here?

PR #1406, opened 2021-11-18
Move responsibility for fetching version number to setup.cfg

alexmalins:
#### Reference Issue

N/A

This PR streamlines the way setuptools fetches the librosa version number. It moves the responsibility from `setup.py` to `setup.cfg`. The benefit is the import and version fetch statements in `setup.py` can be removed, thus reducing the librosa code base that needs to be maintained. The responsibility can be offloaded to setuptools, [which handles version fetching since v46.4.0](https://packaging.python.org/guides/single-sourcing-package-version/).

#### Any other comments?
The current `pyproject.toml` requires setuptools >= 48, so there is no downside to this change AFAIK. It built and tested successfully on my local machine.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,6 +10,7 @@ filterwarnings =
 
 [metadata]
 name = librosa
+version = attr: librosa.version.version
 description = Python module for audio and music processing
 long_description = file: README.md
 long_description_content_type = text/markdown; charset=UTF-8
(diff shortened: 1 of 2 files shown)


bmcfee:
Nice; I didn't realize that was possible!  I guess this functionality was added to setuptools mid-2020, but hadn't made it into whatever example docs i was cribbing from when i last revised the packaging configs.  Anyway this LGTM, thanks!

alexmalins:
Thanks! I like it - having a three line setup.py is dope 🔥

bmcfee (approved):

bmcfee:
:shipit: thanks!

Issue #1161, opened 2020-06-12
Inverse VQT, griffinlim-vqt


bmcfee:
**Is your feature request related to a problem? Please describe.**

#1018 implements variable-Q transforms.  This issue exists to factor out the related methods for inverse-VQT and VQT phase recovery.  These are not critical to implementing VQT, and can be punted to a later release.

lostanlen:
I agree. I don't mean to sound pessimistic, but there's a chance that our VQT implementation is slightly buggy and has weird corner cases. Let's wait till those bugs get reported and fixed before worrying about feature inversion.

bmcfee:
This is probably worth investigating again since merging #1405.

bmcfee:
Ok, I think i've tracked down the source of the scaling error that I've been running into with vqt inversion, and it's pretty subtle.

I noticed when implemented #1405 that the length normalization in the forward VQT here:
https://github.com/librosa/librosa/blob/8bdfc032d1b362a4c29becbee8dc89f082e5fb34/librosa/core/constantq.py#L992-L1006

needs to recompute the filter lengths at the end because the sampling rate may have changed by early downsampling:
https://github.com/librosa/librosa/blob/8bdfc032d1b362a4c29becbee8dc89f082e5fb34/librosa/core/constantq.py#L917-L919

In the CQT case (gamma=0), this doesn't matter because the resampling factor cancels out, so lengths/sr is the same.  However, when `gamma != 0`, it becomes a problem.  Making this work correctly will require inferring the effective sampling rate of the forward vqt inside the inverse vqt.  This seems a little dicey to me, but it's approximately doable.

bmcfee:
Compensating for early downsampling did not make much of a difference here.  I have something that more or less works, but it's still not giving a better inverse than using `gamma=0` even when a nonzero gamma was used in the forward transform.

bmcfee:
The plot thickens.

It looks like our inverse cqt consistently over-amplifies by 0.5 dB.

```python
>>> sr = 44100
>>> y, sr = librosa.load(librosa.ex('nutcracker'), sr=sr, duration=10, offset=5)

# VQT with gamma=0 ==> cqt
>>> V = librosa.vqt(y=y, sr=sr, gamma=0, hop_length=128, sparsity=0, bins_per_octave=60, n_bins=5*12*8, scale=False, res_type='soxr_hq')

# Inverse CQT
[...]

Issue #1208, opened 2020-07-21
CQT's hop length


andrekassis:
It is my understanding that the constant-Q transform doesn't have a fixed hop size as it varies for different frequencies. However,  LibRosa's CQT takes an argument "hop_length" which I'm finding very confusing (the documentation is not helpful). What is this argument and how is it consistent with the definition of the transform? Also, using the Matlab implementation of the transform produces different results than those return by librosa. why is that?

bmcfee:
I think your understanding may not be correct.  The hop length is fixed as it determines the center position of the frames; what changes is the frame lengths (as a function of frequency). 

And yes, the Matlab version is different, for many reasons. The main one being the implementation of resampling, though I'm sure there are others. (IIRC they also do some non uniform frequency spacing tricks to cover all the way up to Nyquist, which we don't do.)

andrekassis:
Thank you. That was helpful

PR #1407, opened 2021-11-28
[CR] Fixing cqt filterbank normalization

bmcfee:
#### Reference Issue

Observed in #1161 

This PR fixes a scaling error in the inverse CQT, where we previously weren't correctly normalizing by the basis filters' frequency-domain power.

This PR also fixes a bug introduced in #1381 in the forward CQT/VQT, where filters were scaled to compensate for octave resampling when they should not have been.

#### Any other comments?

This does not fix the inverse VQT problem, but it gets us closer.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -627,20 +627,24 @@ def icqt(
     # Get the top octave of frequencies
     n_bins = C.shape[-2]
 
+    n_octaves = int(np.ceil(float(n_bins) / bins_per_octave))
+
     # truncate the cqt to max frames if helpful
     freqs = cqt_frequencies(
         fmin=fmin, n_bins=n_bins, bins_per_octave=bins_per_octave
     )
     alpha = __bpo_to_alpha(bins_per_octave)
-    lengths, _ = filters.wavelet_lengths(
+
+    lengths, f_cutoff = filters.wavelet_lengths(
         freqs, sr=sr, window=window, filter_scale=filter_scale, alpha=alpha
     )
+
+    # Trim the CQT to only what's necessary for reconstruction
     if length is not None:
         n_frames = int(np.ceil((length + max(lengths)) / hop_length))
         C = C[..., :n_frames]
 
-    # How many octaves do we have?
-    n_octaves = int(np.ceil(float(n_bins) / bins_per_octave))
+    C_scale = np.sqrt(lengths)
 
     # This shape array will be used for broadcasting the basis scale
     # we'll have to adapt this per octave within the loop
@@ -667,12 +671,10 @@ def icqt(
 
         # Slice out the current octave
         sl = slice(bins_per_octave * i, bins_per_octave * i + n_filters)
-        C_oct = C[..., sl, :]
-        freq_oct = freqs[sl]
 
-        fft_basis, n_fft, lengths = __vqt_filter_fft(
+        fft_basis, n_fft, _ = __vqt_filter_fft(
             my_sr,
-            freq_oct,
+            freqs[sl],
             filter_scale,
             norm,
             sparsity,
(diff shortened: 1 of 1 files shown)


bmcfee:
I just noticed that when we have a relatively prime hop length (so that downsampling is impossible), we end up with some octave amplification errors.  I'm not sure if this is coming from the forward or inverse transform yet, but I'll look into it.  Relevant: #1208

bmcfee:
> I'm not sure if this is coming from the forward or inverse transform yet, but I'll look into it.

It was the forward transform.  We had

```python
fft_basis[:] *= np.sqrt(2 ** (i-oct_start))
```
which is meant to compensate for resampling at each octave's basis functions.  This is fine when each octave counts a downsampling operation, but more generally, it needs to be
```python
fft_basis[:] *= sr / my_sr
```
where `my_sr` is the sampling rate for the current octave (which may be the same as `sr`).

This error does not effect previous releases, as it was only possible starting from #1381 .

librosa/core/constantq.py
```diff
@@ -627,20 +625,22 @@ def icqt(
     # Get the top octave of frequencies
     n_bins = C.shape[-2]
 
+    n_octaves = int(np.ceil(float(n_bins) / bins_per_octave))
```
lostanlen, line 628: not sure why you had to move up this definition of `n_octaves`, but OK

librosa/core/constantq.py
```diff
@@ -681,24 +679,35 @@ def icqt(
...
+
+        # Compensate for length normalization in the forward transform
+        freq_power *= n_fft / lengths[sl]
 
         # Inverse-project the basis for each octave
-        D_oct = np.einsum("fc,c,...ct->...ft", inv_basis, C_scale, C_oct, optimize=True)
+        if scale:
+            # scale=True ==> re-scale by sqrt(lengths)
+            D_oct = np.einsum(
+                "fc,c,c,...ct->...ft",
```
lostanlen, line 695: brilliant

librosa/core/constantq.py
```diff
@@ -971,11 +979,11 @@ def vqt(
             window=window,
             gamma=gamma,
             dtype=dtype,
-            alpha=alpha
+            alpha=alpha,
         )
 
         # Re-scale the filters to compensate for downsampling
-        fft_basis[:] *= np.sqrt(2 ** (i-oct_start))
+        fft_basis[:] *= np.sqrt(sr / my_sr)
```
lostanlen, line 986: this change is transparent iff `hop_length` is a power of two but fixes the #1381 bug when it isn't— do i read this right?

librosa/core/constantq.py
bmcfee: Mostly just for legibility

librosa/core/constantq.py
bmcfee: Yeah exactly. We had accidentally been carrying the assumption that each octave is at a different sampling rate and using the implicitly calculated ratio to compensate. Here we just use the explicit ratio of rates, which will always do the right thing.

lostanlen (approved):

gdoras:
I am a bit confused by this normalisation. If I understand correctly, we are rescaling during [resampling](https://github.com/librosa/librosa/blob/main/librosa/core/audio.py#L619):
    
    if scale:
        y_hat /= np.sqrt(ratio)

where ``ratio=float(target_sr) / orig_sr``, i.e. ``ratio=1/2`` in this case, so we end up multiplying the output by ``np.sqrt(2)`` after each downsampling.

To [comensate](https://github.com/librosa/librosa/blob/main/librosa/core/constantq.py#L986), we are also multiplying the basis by ``np.sqrt(2)`` after each downsamling:

    fft_basis[:] *= np.sqrt(sr / my_sr)

Aren't we multiplying by ``2.0`` after each downsampling ? The compensation should not be  ``fft_basis[:] /= np.sqrt(sr / my_sr)`` instead ?

bmcfee:
> Aren't we then multiplying by `2.0` after each downsampling ?

Basically yes: this is to account for the halving of length (# of samples) induced by resampling.  I don't think the compensation you propose is correct, since it would reduce the norm of the basis rather than increase it, and we definitely want it to increase here because it should be effectively covering a larger extent of time.  (Remember that this is all working with an unnormalized forward DFT, so larger N = larger magnitudes.)

More generally, the scaling and normalization API is probably a bit more complicated than it needs to be.  This is mainly due to many years of extension to the original API which aimed for compatibility with a previous python implementation.  I rather doubt that anyone uses `scale=False` or `norm!=1` these days in practice, but we've never had any specific cause to remove these features either.

gdoras:
Thanks for the explanation Brian.

PR #1414, opened 2021-12-08
[CR] specshow coordinate centering

### LINKED ISSUE #1357: librosa.display.specshow does not display the correct time scale
When trying to display a chromagram using librosa.display.specshow, the time scale is not correct. The time scale is correct when the hop_length is its default value (512). However, it is incorrect whenever the hop_length is changed. (The input length increases as I increase the hop_length).

# Code I used to generate and plot the chromagram:
```
# 512 is the default, when this number is altered, the time scale is wrong
HOP_SIZE = 512

y, sr = librosa.load(mp3)
chroma_feature = librosa.feature.chroma_stft(y=y, sr = sr, hop_length = HOP_SIZE)

fig, ax = plt.subplots()
img = librosa.display.specshow(chroma_feature, y_axis='chroma', x_axis='time', ax=ax, sr = sr, hop_length = HOP_SIZE)
fig.colorbar(img, ax=ax)
plt.show()
```

bmcfee:
#### Reference Issue
Fixes #1357 

This PR implements centered coordinates on specshow, made possible in mpl 3.3.

#### Any other comments?

This change required rebuilding the image comparison fixtures.  This is because pytest-mpl uses `classic` theme by default, which changes the `autolimit_mode` from `data` to `round_numbers` and resulted in strange padding mode.  I'm now forcing it to `default`.  This might be a bad idea, as `default` could change going forward (while other styles probably won't), but it does at least get us working for now.

CR here should take a careful look at the updated image fixtures.


bmcfee:
To follow up on the original issue, here's a chromagram from the brahms example at standard hop length:

image

and at absurdly high hop length:

image

Compared to the current main branch:

image

and

image

bmcfee:
Took a quick pass through the documentation to make sure that everything still works properly.  A couple of snags came up:

1. The stack_memory example still "works" (After a bit of correction), but now that chroma bins are centered properly, the decorative boundaries look weird.  Probably this example plot just needs to be redone.
2. The laplacian segmentation demo fails to illustrate segment labels now.  This is because if we only supply a single column and use bin centers, it has no way to infer the width of the bins.  Using specshow for this was always a hack, and we can probably do better.

bmcfee:
Fixed the stack memory example to be a bit clearer.  (Still not great IMO, but better)

Before:

image


After:

image

bmcfee:
Ok, I think I'm happy with where this one is at.

I dug into the mpl styles, and it seems like "classic" is the only one that uses round_numbers mode.  It seems very unlikely that anyone will accidentally stumble into specshow plots with outsized margin padding due to our switch over to centered coordinates.

Everything else seems to be functioning just like before, though the code is a bit simpler now.  Another set of eyes on this would be appreciated, but I think it's basically done.

bmcfee:
Meanwhile, I'm going to spend a bit of time today verifying that the updated version requirements (mpl 3.3) still work with our target documentation builds for the next release.

bmcfee:
Confirmed that historical doc builds still work with revised requirements.

lostanlen (approved):
LGTM !

Let's remember to publicize the new specshow default (`shading='auto'` instead of `shading='flash'`) in the next release notes.

https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.pcolormesh.html

```
'auto': Choose 'flat' if dimensions of X and Y are one larger than C. Choose 'nearest' if dimensions are the same.
```

and of course the new MPL version requirement

PR #1418, opened 2021-12-13
[CR] add soundfile object handling to load and stream

### LINKED ISSUE #1349: It would be nice if librosa.load(path, sr) would support a format-kwarg.
I am currently dealing with python inmemory file objects (`io.BytesIO` etc.). They don't have a file extension which is why `librosa.load()` cannot recognise them properly as paths and I get: `RuntimeError: Error opening <_io.BytesIO object at 0x7f1ed40ff090>: Format not recognised.`

`librosa.load()` uses a `SoundFile` object internally which in fact already supports format as a `kwarg`. Please propagate that `kwarg` to the function call so that a user can define the extension of the file manually (=> `librosa.load(path, sr, format)`)

At the moment, I am using `pydub.AudioSegments` for that. But it is not as clean as that one `librosa` call would be.

bmcfee:
#### Reference Issue
Fixes #1349 (indirectly)

This PR adds support for existing soundfile objects in load and stream.  This makes it possible to work with raw objects, and provides a way for users to completely control the decoding process within our input wrappers.  This will obviate the need for exposing more API through `load` so that #1349 is no longer an issue.

#### Any other comments?

This is WIP; load works as expected, but stream will take a bit of refactoring.

I guess we should also extend this to `get_duration` and `get_samplerate`.

### DIFF
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -25,6 +25,7 @@ dependencies:
   - contextlib2
   - coverage
   - ffmpeg
+  - tomli<2.0
   - pip:
     - soxr
     - samplerate
--- a/.github/environment-minimal.yml
+++ b/.github/environment-minimal.yml
@@ -25,6 +25,7 @@ dependencies:
   - contextlib2
   - coverage
   - ffmpeg
+  - tomli<2.0
   - pip:
     - soxr
     - samplerate
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -731,10 +731,11 @@ def get_samplerate(path):
 
     Parameters
     ----------
-    path : string, int, or file-like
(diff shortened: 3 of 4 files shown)


bmcfee:
Python 3.6 test failures are due to the new release of tomli=2.0, which apparently does not support python 3.6.  Argh.

bmcfee:
I think this is good to :shipit: .  The change to `stream` could use an extra set of eyes - maybe @carlthome could chime in?  I've replaced the `seek(0)` by explicitly seeking the soundfile object to the start position of the stream.  Everything behaves as expected, but I want to make sure I'm not missing some weird edge case.

bmcfee:
After looking through #942 , I think I've convinced myself that the changes in this PR are correct and there's no difference in behavior.  Unless there are objections, I'll merge this PR later today.

Issue #1415, opened 2021-12-09
Add audio output to examples


bmcfee:
PR #1410 added an example to the gallery demonstrating Audio widget playback.  I think some of the other advanced examples could benefit from audio playback as well:

- HPSS
- Vocal separation
- Maybe new examples?

This would be pretty easy to do, and would make the examples much more compelling.  I don't want to go overboard with it though, since audio embeds are relatively expensive.

PR #1423, opened 2021-12-18
Added new examples

### LINKED ISSUE #1411: Audio examples: non-musical excerpts and other concepts

All of our example data are musical recordings, but librosa is often used for general audio and not only music.  It seems like we ought to include some non-musical example data.  I'm mainly thinking of this with an eye toward expanding our example gallery.

Additionally, we could expand the musical data to cover more phenomena than are currently included.  #952 documents the original track selection process.  We had discussed including makam and raga tracks, but never settled on specific choices.

Here's a brief wish list of things that would complement our current example selection:

1. Speaking voices
    a. Could be helpful for demonstrating parameters restricted to human vocal range
2. Bird song and/or other bioacoustic data
    a. Could be good for showing how to modify default parameters for non-human data
3. Environmental sound / acoustic scenes
4. Solo piano
    a. Might be nice for a nmf demo
4. Solo drums
    a. For onset detection and beat tracking demos
5. Makam
    a. For continuous pitch estimation
6. Raga
    a. same, but also for demonstating Indian notation
7. Something multichannel
    a. Now that we support multichannel properly, it would be nice to demo
8. A track with an odd time signature

The same constraints as listed in #952 apply here: must be freely shareable, non-offensive, and relatively short (to minimize data footprint).

bmcfee:
#### Reference Issue
Fixes #1411 

This PR adds a handful of new example clips.  Two bioacoustic (`robin` and `humpback`), three speech (`libri[123]`), and two 3/4 time songs (`pistacho` and `sweetwaltz`).

#### Any other comments?

I'll also roll in some updates to the advanced examples per #1415 here.

What I haven't  yet decided is whether to include the Carnatic Flute example demoed in #1411, or a multi-channel environmental sound example (due to codec issues with ogg vorbis).  Any commentary here would be much appreciated.


bmcfee:
I'm also rolling some documentation fixes into this PR, since the technical changes are pretty minute.

---
## Docstring example changes

### lpc: trumpet -> libri1

New LPC plot looks like this:
image

Compared to the old one here: 
image

No qualitative change, but I think speech is a better demonstrative case for this method.


### harmonic interpolation: choice -> sweetwaltz

New interp plot for tempo looks like this:
image

Compared to the old one:
image

The tempo is more prominent in the sweetwaltz example, and the switch to an odd-meter track makes the plot a bit more interesting.

### pcen: trumpet -> robin

New plots:
image
image

Compared to the old ones:

image
image

Obviously the default analysis parameters aren't the best for the bird song example, but I think it still gives a better visual demonstration of the various pcen options than the trumpet example.  Notably, the trumpet had no appreciable vibrato or scooping, so the max_size parameter was hard to see.

### mfcc: trumpet -> libri1

image
vs
image

Not entirely convinced of this switch yet; might make more sense if we also plot the melspec above the MFCCs so it's easier to visually parse.

### decompose: choice -> sweetwaltz

image
image

vs

image
image


---

Any other suggestions / candidates for using the new examples?

bmcfee:
One more update to the NMF example, now using subplot mosaic to lay things out a bit better:
image

bmcfee:
Revised pcen-stream example using humpback:
image

Note: this switches from mean to max-aggregation along pcen channels, just to have an output that looks a little more useful.  I also tried setting this up as an animation, but it turns out that animating quadmesh plots (spectrograms) is kind of a pain in matplotlib.  The resulting code was way more complex than I'd like for this kind of thing.

bmcfee:
I think this is basically ready to merge.  I'm still on the fence about the Carnatic flute example. It would be nice to have, but think our current examples essentially cover the same bases (pitch tracking, notation).  If we did want to build out a more elaborate advanced example focusing specifically on Indian music -- which we should probably do, but I don't have the bandwidth or expertise for it -- then it would make sense to include.

bmcfee:
Any objections to merging?

PR #1427, opened 2021-12-27
Tiny fix for docstring of `reassigned_spectrogram`

bluenote10:

The example is currently using `power_to_db` on the returned magnitudes. Since the docstring doesn't mention that the magnitudes are squared I'd assume that it should be `amplitude_to_db` instead, right?

#### Any other comments?

Just a tiny amendment to avoid confusion.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -999,7 +999,7 @@ def reassigned_spectrogram(
     ...     1e-6 * np.random.randn(2*sr)
     >>> freqs, times, mags = librosa.reassigned_spectrogram(y=y, sr=sr,
     ...                                                     n_fft=n_fft)
-    >>> mags_db = librosa.power_to_db(mags, ref=np.max)
+    >>> mags_db = librosa.amplitude_to_db(mags, ref=np.max)
 
     >>> fig, ax = plt.subplots(nrows=2, sharex=True, sharey=True)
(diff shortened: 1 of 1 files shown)


bmcfee:
Quite correct - thanks for catching that!

librosa/core/spectrum.py
```diff
@@ -999,8 +999,7 @@ def reassigned_spectrogram(
     ...     1e-6 * np.random.randn(2*sr)
     >>> freqs, times, mags = librosa.reassigned_spectrogram(y=y, sr=sr,
     ...                                                     n_fft=n_fft)
-    >>> mags_db = librosa.power_to_db(mags, ref=np.max)
-
+    >>> mags_db = librosa.amplitude_to_db(mags, amin=amin)
```
bmcfee, line 1002: I had to resolve a merge conflict here, but it looks like you've also changed the behavior of the db conversion.  Could you switch that back to `ref=np.max` instead of `amin=amin`?

bluenote10: That was unintentional indeed. I had literally just copy/pasted the different function name, but my mistake was to branch off of `master` instead of `main` out of habit. Should be fixed by rebasing now.

bmcfee (approved):

bmcfee:
Thanks!

PR #1430, opened 2021-12-31
Remove unnecessary dynamic __all__ modifications

### LINKED ISSUE #1429: Remove dynamic __all__ to facilitate static analysis

I noticed that librosa's IDE support is a bit lacking, because pylance doesn't understand librosa's module structure very well. For instance, while this works

image

the top-level symbols don't work:

image

In practice it is just a bit of a pity to have no auto-completion on the plain `librosa` module, and no jump to documentation / implementation.

I'm mainly opening the issue for discussion of possible solutions.

On first glance, the way librosa re-exports symbols looks sensible to me. The fact that the dynamic wildcard exports were working in one case but not in the other suggested it was a pylance issue. I've raised an upstream issue https://github.com/microsoft/pylance-release/issues/2200, but apparently it is expected behavior, and the suggestion was to fix it on librosa side.

I'm wondering if we can come up with a solution that doesn't involve massive boilerplate / anti-DRY. Any ideas?

bluenote10:
#### Reference Issue

Fixes #1429 

image

This PR makes librosa conformant with the official [library interface standard](https://github.com/python/typing/blob/master/docs/source/libraries.rst#library-interface).

#### Any other comments?

It turned out that that the number of dynamic `__all__` usages is actually quite manageable. Also wildcard usage isn't an issue, i.e., there is no need to also list all symbols everywhere in the imports. So overall, the change would be relatively minor actually.

### DIFF
--- a/librosa/core/__init__.py
+++ b/librosa/core/__init__.py
@@ -12,4 +12,12 @@ from .fft import *  # pylint: disable=wildcard-import
 from .notation import *  # pylint: disable=wildcard-import
 
 
-__all__ = [_ for _ in dir() if not _.startswith("_")]
+__all__ = []
+__all__ += convert.__all__
+__all__ += audio.__all__
+__all__ += spectrum.__all__
+__all__ += pitch.__all__
+__all__ += constantq.__all__
+__all__ += harmonic.__all__
+__all__ += fft.__all__
+__all__ += notation.__all__
--- a/librosa/core/__init__.py
+++ b/librosa/core/__init__.py
@@ -10,14 +10,3 @@ from .constantq import *  # pylint: disable=wildcard-import
 from .harmonic import *  # pylint: disable=wildcard-import
 from .fft import *  # pylint: disable=wildcard-import
 from .notation import *  # pylint: disable=wildcard-import
-
-
-__all__ = []
-__all__ += convert.__all__
-__all__ += audio.__all__
-__all__ += spectrum.__all__
-__all__ += pitch.__all__
-__all__ += constantq.__all__
-__all__ += harmonic.__all__
-__all__ += fft.__all__
-__all__ += notation.__all__
--- a/librosa/util/__init__.py
+++ b/librosa/util/__init__.py
@@ -82,4 +82,9 @@ from ._nnls import *  # pylint: disable=wildcard-import
 from . import decorators
 from . import exceptions
 
-__all__ = [_ for _ in dir() if not _.startswith("_")]
+__all__ = ["decorators", "exceptions"]
+__all__ = utils.__all__
+__all__ = files.__all__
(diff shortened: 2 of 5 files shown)


librosa/core/__init__.py
```diff
@@ -12,4 +12,12 @@
...
-__all__ = [_ for _ in dir() if not _.startswith("_")]
+__all__ = []
+__all__ += convert.__all__
+__all__ += audio.__all__
+__all__ += spectrum.__all__
+__all__ += pitch.__all__
+__all__ += constantq.__all__
+__all__ += harmonic.__all__
+__all__ += fft.__all__
+__all__ += notation.__all__
```
bmcfee, line 23: I don't think this is necessary.

`__all__` will be automatically updated by the imports at the top.  What line 15 did was remove all "private" imported objects (those starting with a `_`).  Your proposed change here does not do this, and without that functionality, we may as well just use the automatic updates implied by import, right?

librosa/util/__init__.py
```diff
@@ -82,4 +82,9 @@
 from . import decorators
 from . import exceptions
 
-__all__ = [_ for _ in dir() if not _.startswith("_")]
+__all__ = ["decorators", "exceptions"]
+__all__ = utils.__all__
+__all__ = files.__all__
+__all__ = matching.__all__
+__all__ = deprecation.__all__
+__all__ = _nnls.__all__
```
bmcfee, line 90: These bits should have `+=`, if you're going to do it this way, not `=`.  But again, I don't think we need to do all of this if we're not going to block privates.

librosa/util/deprecation.py
```diff
@@ -5,6 +5,7 @@
 import inspect
 import warnings
 
+__all__ = ["Deprecated", "rename_kw"]
```
bmcfee, line 8: No filtration was done here, so we don't need to override `__all__` at all.

bmcfee:
After doing a bit more digging, I think we can actually simplify this quite a bit.

According to pep8, the `from submod import *` pattern will bypass anything objects with leading underscores, and it will be automatically prevented from polluting `__all__`.

There are a few private functions which start with `__`, and those should be renamed to use a leading single underscore.  It won't affect the API in any way, but it will be a bit more internally consistent.

Making those changes, we can ditch all attempted cleverness in `__all__` mangling.

bluenote10:
Sorry for the delay!

Yes fully agreed, most likely we don't need really need to do anything at all with `__all__` :wink:

I've now removed all explicit usages, hope that is what you meant. If not, also feel free to close the PR if it is easier to apply the tiny change with the next maintenance commit or so.

bmcfee (approved):
LGTM, tested locally and everything seems fine. Thanks!

Issue #80, opened 2014-03-29
Use decorators to cache function output


craffel:
See this example:
https://wiki.python.org/moin/PythonDecoratorLibrary#Memoize
np.ndarrays are not hashable, but there may be a way to get a suitable hash from the ndarray object directly.

PR #1431, opened 2022-01-05
[CR] Required kwargs

### LINKED ISSUE #1114: Keyword-only arguments
#### Description

Since we're now a python3-only package, what do people think about moving to [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/) wherever possible?

Some arguments off the top of my head:

### Pro
- Forces users to be explicit about their parameters
- Improves code readability / reduces bugs from implicit parameter ordering
- Allows us to change the ordering of keyword arguments at any point in the future without fear of breaking things for users

### Con
- Existing code using positional parameter values will break
- Code will be necessarily more verbose
- Soft support for legacy python will definitely break. (Right now, things mostly work on 2.7, but we make no promises of support.)

---
If we do this, it will need to be at a major revision point.  0.8 seems like a logical place to do it (sooner being better), but I could be convinced otherwise.

bmcfee:
#### Reference Issue
Fixes #1114 

This PR implement keyword-only arguments throughout the package, as discussed in #1114.  The majority, but not all parameters will be keyword-only going forward.

Progress tracker (loosely in topological order from the import DAG):

- [x] _cache
- [x] util.exceptions
- [x] util.deprecation
- [x] util.decorators
- [x] util.files
- [x] util.utils
- [x] util.matching
- [x] feature.utils
- [x] core.fft
- [x] core.notation
- [x] core.convert
- [x] core.audio
- [x] core.harmonic
- [x] filters
- [x] segment
- [x] core.spectrum
- [x] core.constantq
- [x] core.pitch
- [x] decompose
- [x] feature.inverse
- [x] feature.spectral
- [x] effects
- [x] onset
- [x] feature.rhythm
- [x] beat
- [x] sequence
- [x] display

---
Docstring audits
- [x] beat.py
- [x] _cache.py
- [x] core/audio.py
- [x] core/constantq.py
- [x] core/convert.py
- [x] core/fft.py
- [x] core/harmonic.py
- [x] core/notation.py
- [x] core/pitch.py
- [x] core/spectrum.py
- [x] decompose.py
- [x] display.py
- [x] effects.py
- [x] feature/inverse.py
- [x] feature/rhythm.py
- [x] feature/spectral.py
- [x] feature/utils.py
- [x] filters.py
- [x] onset.py
- [x] segment.py
- [x] sequence.py
- [x] util/decorators.py
- [x] util/deprecation.py
- [x] util/exceptions.py
- [x] util/files.py
- [x] util/matching.py
- [x] util/_nnls.py
- [x] util/utils.py
- [x] version.py

#### Any other comments?

I'm starting the PR early to get full-package CI tests.  At the end, we'll also need to sweep back through for a docstring test.  I'll use comments below to keep track of contentious points.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -181,7 +181,7 @@ def load(
         y = to_mono(y)
 
     if sr is not None:
-        y = resample(y, sr_native, sr, res_type=res_type)
+        y = resample(y, orig_sr=sr_native, target_sr=sr, res_type=res_type)
 
     else:
         sr = sr_native
@@ -473,7 +473,7 @@ def to_mono(y):
 
 @cache(level=20)
 def resample(
-    y, orig_sr, target_sr, *, res_type="kaiser_best", fix=True, scale=False, **kwargs
+    y, *, orig_sr, target_sr, res_type="kaiser_best", fix=True, scale=False, **kwargs
 ):
     """Resample a time series from orig_sr to target_sr
 
@@ -566,7 +566,7 @@ def resample(
     Downsample from 22 KHz to 8 KHz
 
     >>> y, sr = librosa.load(librosa.ex('trumpet'), sr=22050)
-    >>> y_8k = librosa.resample(y, sr, 8000)
+    >>> y_8k = librosa.resample(y, orig_sr=sr, target_sr=8000)
     >>> y.shape, y_8k.shape
     ((117601,), (42668,))
     """
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -466,7 +466,7 @@ def to_mono(y):
     util.valid_audio(y, mono=False)
 
     if y.ndim > 1:
-        y = np.mean(y, axis=tuple(range(y.ndim-1)))
+        y = np.mean(y, axis=tuple(range(y.ndim - 1)))
 
     return y
 
@@ -1154,7 +1154,7 @@ def zero_crossings(
 
 
 def clicks(
-    *, 
+    *,
     times=None,
     frames=None,
     sr=22050,
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -381,7 +381,7 @@ def stream(
     ...                         frame_length=2048,
     ...                         hop_length=2048)
     >>> for y_block in stream:
-    ...     m_block = librosa.feature.melspectrogram(y_block, sr=sr,
+    ...     m_block = librosa.feature.melspectrogram(y=y_block, sr=sr,
     ...                                              n_fft=2048,
     ...                                              hop_length=2048,
     ...                                              center=False)
@@ -806,7 +806,7 @@ def autocorrelate(y, *, max_size=None, axis=-1):
 
     >>> import matplotlib.pyplot as plt
     >>> odf = librosa.onset.onset_strength(y=y, sr=sr, hop_length=512)
-    >>> ac = librosa.autocorrelate(odf, max_size=4* sr / 512)
+    >>> ac = librosa.autocorrelate(odf, max_size=4 * sr // 512)
     >>> fig, ax = plt.subplots()
     >>> ax.plot(ac)
     >>> ax.set(title='Auto-correlation', xlabel='Lag (frames)')
@@ -889,14 +889,14 @@ def lpc(y, *, order, axis=-1):
     Compute LP coefficients of y at order 16 on entire series
 
     >>> y, sr = librosa.load(librosa.ex('libri1'))
-    >>> librosa.lpc(y, 16)
+    >>> librosa.lpc(y, order=16)
 
     Compute LP coefficients, and plot LP estimate of original series
 
     >>> import matplotlib.pyplot as plt
     >>> import scipy
     >>> y, sr = librosa.load(librosa.ex('libri1'), duration=0.020)
-    >>> a = librosa.lpc(y, 2)
+    >>> a = librosa.lpc(y, order=2)
     >>> b = np.hstack([[0], -1 * a[1:]])
     >>> y_hat = scipy.signal.lfilter(b, [1], y)
     >>> fig, ax = plt.subplots()
@@ -1431,15 +1431,15 @@ def chirp(*, fmin, fmax, sr=22050, length=None, duration=None, linear=False, phi
     --------
     Generate a exponential chirp from A2 to A8
 
-    >>> exponential_chirp = librosa.chirp(110, 110*64, duration=1)
+    >>> exponential_chirp = librosa.chirp(fmin=110, fmax=110*64, duration=1)
 
     Or generate the same signal using ``length``
 
-    >>> exponential_chirp = librosa.chirp(110, 110*64, sr=22050, length=22050)
+    >>> exponential_chirp = librosa.chirp(fmin=110, fmax=110*64, sr=22050, length=22050)
 
     Or generate a linear chirp instead
 
-    >>> linear_chirp = librosa.chirp(110, 110*64, duration=1, linear=True)
+    >>> linear_chirp = librosa.chirp(fmin=110, fmax=110*64, duration=1, linear=True)
 
     Display spectrogram for both exponential and linear chirps.
 
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -45,6 +45,7 @@ BW_FASTEST = resampy.filters.get_filter("kaiser_fast")[2]
 # 'path' are unchanged across calls.
 def load(
     path,
+    *,
     sr=22050,
     mono=True,
     offset=0.0,
@@ -245,6 +246,7 @@ def __audioread_load(path, offset, duration, dtype):
 
 def stream(
     path,
+    *,
     block_length,
     frame_length,
     hop_length,
@@ -471,7 +473,7 @@ def to_mono(y):
 
 @cache(level=20)
 def resample(
-    y, orig_sr, target_sr, res_type="kaiser_best", fix=True, scale=False, **kwargs
+    y, orig_sr, target_sr, *, res_type="kaiser_best", fix=True, scale=False, **kwargs
 ):
     """Resample a time series from orig_sr to target_sr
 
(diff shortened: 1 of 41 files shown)


bmcfee:
A couple of contentious points so far (utils submodule):

- `util.softmask(X, X_ref, ...)`.  Question is whether to make `X_ref` a named parameter or not.  It's ambiguous because by my reasoning in #1114, `X` is a non-informative name here and therefore should be unnamed.  But this bleeds into `X_ref`, which is partially informative and should therefore be named.  Possible compromise: rename `X_ref` to `ref`?
- `util.pad_center(data, size, ...)` should `size` be keyword-only?  `data` is uninformative, but `size` does mean something specific and the two arguments are strictly necessary.
- Relatedly, `sync(data, idx, ...)` ; idx is a bit more obscure of a name here, but the issue is otherwise quite similar to `pad_center`.
- `match_events(events_from, events_to, ...)` (ditto `intervals`): it's a bit of a mouthful here, and parameter names are redundant to the function name.  How about shortening to `match_events(*, from, to, ...)`?

bmcfee:
> * `util.softmask(X, X_ref, ...)`.

thought this one over a bit more, and I think I prefer `util.softmask(X, X_ref, *, ...)` - that is, not making either input a named parameter.  The reasoning here is that this is essentially a non-commutative binary operation `f: a,b -> a**p / (a**p + b**p)` (for some power `p`), but neither input really should be thought of as a "reference" like we do in dB calculations.

bmcfee:
> * How about shortening to `match_events(*, from, to, ...)`?

Oh right, `from` is a reserved word in python. :laughing: 

How about `source` and `dest`?

bmcfee:
Running into some interesting snags here with the interaction between the cache object's decorator and kwarg-only functions.

```python
../../miniconda/envs/py39/lib/python3.9/site-packages/_pytest/python.py:578: in _importtestmodule
    mod = import_path(self.fspath, mode=importmode)
../../miniconda/envs/py39/lib/python3.9/site-packages/_pytest/pathlib.py:524: in import_path
    importlib.import_module(module_name)
../../miniconda/envs/py39/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:986: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:680: in _load_unlocked
    ???
../../miniconda/envs/py39/lib/python3.9/site-packages/_pytest/assertion/rewrite.py:170: in exec_module
    exec(co, module.__dict__)
tests/test_features.py:10: in <module>
    import librosa
librosa/__init__.py:211: in <module>
    from . import core
librosa/core/__init__.py:5: in <module>
    from .convert import *  # pylint: disable=wildcard-import
librosa/core/convert.py:7: in <module>
    from . import notation
librosa/core/notation.py:247: in <module>
    def mela_to_svara(mela, *, abbr=True, unicode=True):
librosa/_cache.py:58: in wrapper
    return decorator_apply(self.memory.cache, function)
librosa/_cache.py:50: in decorator_apply
    return FunctionMaker.create(
../../miniconda/envs/py39/lib/python3.9/site-packages/decorator.py:196: in create
    return self.make(body, evaldict, addsource, **attrs)
../../miniconda/envs/py39/lib/python3.9/site-packages/decorator.py:159: in make
    code = compile(src, filename, 'single')
E     File "<decorator-gen-0>", line 2
E       return decorated(mela, *, abbr=None, unicode=None)
E                               ^
E   SyntaxError: iterable argument unpacking follows keyword argument unpacking
```

It's possible that our decorator_apply function needs to be rewritten to accommodate kwonly parameters?

bmcfee:
Following up on the decorator problem above:

- The decorator wrapping business originally came up in #80, when we wanted to preserve the docstring and function signature of cached functions.  This is critical for interactive help browsing, as well as autodoc functionality in sphinx.  We can't just ditch it.
- I've raised the issue over in the `decorator` package.  It seems related to at least one other known issue, and may be of low priority.
- The author of `makefun` chimed in with a relatively easy fix: `makefun.wraps` does everything we want here, and I've verified that the proposed fix works with kwarg-only functions.  It's pretty easy, the cache object's call method simplifies to:
```python
def __call__(self, level):
        """Example usage:

        @cache(level=2)
        def semi_important_function(some_arguments):
            ...
        """

        def wrapper(function):
            """Decorator function.  Adds an input/output cache to
            the specified function."""

            from makefun import wraps
            
            if self.memory.location is not None and self.level >= level:
                return wraps(function)(self.memory.cache(function))
            else:
                return function
        return wrapper
```
(we can bump the import out to the module, of course).

Now the question: do we want to add makefun as a dependency here?  It wouldn't replace our dependency on the `decorator` package, which we use for .. well, decorators.  It's also relatively young (2019), but it seems solid and pretty minimal.

bmcfee:
47d0803 includes the makefun recipe for cache decoration.  We can roll it back if we decide against it, but I want to keep CI tests working to make progress on the rest of this PR.  That said, I am leaning toward the makefun recipe going forward.

bmcfee:
A couple more cases to think about:

- `segment.cross_similarity(data, data_ref, ...)` - similar deal to softmask.  I think neither of the first two args should be kwarged.
- `segment.agglomerative(data, k, ...)` - in this case, the first two arguments are both necessary and of distinct types (array, int).  I think it's not necessary to kwarg these, or cases like these.  We can in some other cases, like `core.phase_vocoder(D, rate)` where the name is explicitly meaningful, but we can handle it case-by-case.

librosa/core/audio.py
```diff
@@ -471,7 +473,7 @@ def to_mono(y):
 
 @cache(level=20)
 def resample(
-    y, orig_sr, target_sr, res_type="kaiser_best", fix=True, scale=False, **kwargs
+    y, orig_sr, target_sr, *, res_type="kaiser_best", fix=True, scale=False, **kwargs
```
bmcfee, line 476: Not sure if this one should kwarg the sampling rates.  Normally I would because they're both values that could be of the same type and therefore easily confused.  I don't particularly like these parameter names though.  Maybe something like:

```python
def resample(y, *, sr, new_sr, ...)
```
would be more clear?

bmcfee:
Docs build clean now.  Remaining todos here:

- [ ] resolve kwarging for matching functions
- [ ] resolve kwarging for resample
- [ ] final decision on `makefun`

bmcfee:
Turns out we didn't need makefun at all; I've reverted the change and fixed cache accordingly.

bmcfee:
One more thought: I think we should rename `h_range=` in salience/interp_harmonics to just `harmonics=`.

It's a bit more verbose, but is unambiguously a better name.  The only question is whether this should go through a deprecation-rename cycle or not, since it was previously a required positional argument.  I suspect that we can get away with simply renaming it directly while making it a required named argument and it will be fine; code using positional argument here is going to break anyway, so we may as well fix the name in one shot.

bmcfee:
Some more thoughts:

- I think it's okay to leave resample as is, with unnamed positional arguments.  Naming schemes I can think of don't seem to me to add any clarity:
  - `sr=, new_sr=` - probably the best option, but it's conceptually backward from how `sr` is used in `load` to specify the output sampling rate.  Could lead to trouble with `presets`.
  - `orig_sr=, target_sr=` - simultaneously too terse and too verbose.
  - `old_sr, new_sr` - "old" and "new" only make sense if the resampled signal is meant to replace the input signal.  Often it is not, eg in the case of cqt/wavelets.
- I'm still ambivalent about changing the matching functions.  It would be best if we could use `from=, to=` here, but we can't.  `source=,dest=` (as in graph matching algorithms) don't seem sufficiently clearer IMO.  The functions actually are inherently confusing, and I often have to reread the docs to remember how exactly to use them, ie which acts as the "query" and which is the "result".  (Those terms aren't any better either.)  Any suggestions here would be much appreciated.

lostanlen:
> How about source and dest?

`source` and `destination` as in https://numpy.org/doc/stable/reference/generated/numpy.moveaxis.html
or `src` and `dst` as in https://docs.python.org/3/library/shutil.html

> util.pad_center(data, size, ...) should size be keyword-only? 

Yes. Rename to `target_size` ?

> Relatedly, sync(data, idx, ...) ; idx is a bit more obscure of a name here, but the issue is otherwise quite similar to pad_center.

Rename `idx` to `boundaries` ?

> I think we should rename h_range= in salience/interp_harmonics to just harmonics=. It's a bit more verbose, but is unambiguously a better name.

Yes. Clarity wins in the age of tab-completion

bmcfee:
> `source` and `destination` as in https://numpy.org/doc/stable/reference/generated/numpy.moveaxis.html
> or `src` and `dst` as in https://docs.python.org/3/library/shutil.html

So in those cases, something is clearly moving "from" a source "to" a destination and it makes sense.  For the matching functions, it's less clear in my opinion.  Really what's happening is that each `events_from` is being assigned a match out of `events_to`, so one could make the case that matchings are drawn "from" the "to" set.  It's grammatically a bit weird, and I don't want to lock us into names that would be more confusing.



> Rename `idx` to `boundaries` ?

Interesting.  I'd worry that `boundaries` could be misleading since it also accepts a list of slice objects and not just "boundaries" which in segmentation land (ie mir_eval) would mean one frame index per interval.



> Yes. Clarity wins in the age of tab-completion

Agreed.

bmcfee:
Following up on this:

- I've implemented `h_range -> harmonics`.
- The closest function to our matching functions that I can think of is `np.searchsorted`, which names the two inputs `a` and `v`.  Curiously, the ordering there is reversed from how we do it: the first argument is the sorted array to draw from, the second argument contains the values to insert.  I don't think we should change the order here, but how about moving to more abstract names (`a` and `b` or something like)?  The idea here being to remove any suggestive variable naming which might lead to confusion, thereby forcing people to read the docstring to be certain of the behavior.  No matter what, I think I'm convinced that these arguments should stay positional and not keyword.
- resample: I think I'm convinced that `sr=, new_sr=` is the best option and will implement it unless there are objections.
- sync, idx: since `idx` can contain either frame indices or intervals, I don't think there's a good catch-all name for this, and we should leave it as is (relatively abstract).

Final thoughts @lostanlen ?  Do things otherwise look reasonable in this PR?

bmcfee:
> resample: I think I'm convinced that `sr=, new_sr=` is the best option and will implement it unless there are objections.

On second thought, it will be less friction to stick to `orig_sr=, target_sr=`.

bmcfee:
Unless there are objections, I'd like to merge this one today.

lostanlen:
Yes 👍🏻

Issue #1350, opened 2021-07-02
[DeprecationWarning] np aliases of builtin types deprecated in numpy 1.20


guillemcortes:
**Describe the bug**
Numpy deprecated their aliases of built-in types in NumPy 1.20 (https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations) and that triggers a warning every time one of these aliases is used. I am using librosa 0.6.2, but I've seen that aliases are also used in the 0.8.2 version (the latest as of now)

**To Reproduce**
```
from librosa.util import utils
import numpy as np
arr = np.random.rand(3,2)
utils.normalize(arr)
```

**Expected behavior**
```
/usr/local/lib/python3.7/dist-packages/librosa/util/utils.py:724: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the NumPy scalar type, use `np.float64` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
  mag = np.abs(S).astype(np.float)
```

**Solution**
As the same warning suggests, change all aliases for the built-in types.

**Disclaimer**
Please, note that this is my first issue reported ever, and I'm still familiarizing myself with it. I would happily fork the code and change these aliases, but I don't know if it's better to have a short discussion about it first.

**Software versions***
import platform; print(platform.platform())
`Linux-4.19.0-16-amd64-x86_64-with-debian-10.2`
import sys; print("Python", sys.version)
`Python 3.7.3 (default, Jul 25 2020, 13:03:44)`
`[GCC 8.3.0]`
import numpy; print("NumPy", numpy.__version__)
`NumPy 1.20.1`
import scipy; print("SciPy", scipy.__version__)
`SciPy 1.4.1`
import librosa; print("librosa", librosa.__version__)
`librosa 0.6.2`

librosa.show_versions()
```
INSTALLED VERSIONS
------------------
python: 3.7.3 (default, Jul 25 2020, 13:03:44)
[GCC 8.3.0]

librosa: 0.6.2

audioread: 2.1.9
numpy: 1.20.1
scipy: 1.4.1
sklearn: 0.24.1
joblib: 1.0.1
decorator: 4.4.2
six: 1.12.0
resampy: 0.2.2

numpydoc: None
sphinx: None
sphinx_rtd_theme: None
[...]

PR #1436, opened 2022-01-20
Warning stacklevels

### LINKED ISSUE #1432: Add stacklevel=2 to deprecation warnings

As noted in the recently released [IPython 8 docs](https://ipython.readthedocs.io/en/8.0.0/whatsnew/version8.html#deprecation-and-removal), deprecation warnings should trace to the caller, not the callee.

This is implemented by setting the stacklevel to 2 when issuing a warning.

No downside to this.  We might have to do a little extra checking to make sure it behaves properly when sent from a decorator though.

bmcfee:
#### Reference Issue
Fixes #1432 

This PR adds stacklevels to warnings issued by librosa.

#### Any other comments?

As noted in #1432, this is actually a bit more delicate than I initially thought.  Some basic rules here:

1. deprecations should always come outside of caching
2. warnings generated by cached functions are tricky to manage.  I haven't resolved this completely yet.
3. Stacklevel should be 2 for functions that are called directly by users, but can be larger for internal functions.  This gets messy quite rapidly.

Additionally, I've rolled in some minor cleanups to better deal with upstream deprecations that were previously swept under the rug, see #1350 .

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -171,7 +171,8 @@ def load(
     except RuntimeError as exc:
         # If soundfile failed, try audioread instead
         if isinstance(path, (str, pathlib.PurePath)):
-            warnings.warn("PySoundFile failed. Trying audioread instead.")
+            warnings.warn("PySoundFile failed. Trying audioread instead.",
+                          stacklevel=2)
             y, sr_native = __audioread_load(path, offset, duration, dtype)
         else:
             raise (exc)
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -1381,7 +1381,8 @@ def griffinlim_cqt(
     if momentum > 1:
         warnings.warn(
             "Griffin-Lim with momentum={} > 1 can be unstable. "
-            "Proceed with caution!".format(momentum)
+            "Proceed with caution!".format(momentum),
+            stacklevel=2
         )
     elif momentum < 0:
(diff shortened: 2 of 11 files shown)


bmcfee:
I've taken a first pass at adding stacklevels to all of our warnings.  For the most part, this is unambiguous, but there are a few cases where the level could be either 2 (user calls directly) or 3 (user calls indirectly).  For example, `mfcc_to_mel` could be called directly by a user, or indirectly through `mfcc_to_audio`.

In these cases, I've erred on the side of the lower number.  It's not always going to be exactly correct, but I'm not sure what else we can do here?

A massively over-engineered solution would parse the entire callstack for the first level outside of the librosa module and use that.  Maybe this would also solve our problem with warnings from inside cached functions.

bmcfee:
After a bit of consultation, it does seem like sticking to stacklevel=2 is the simplest and most effective solution here.

The only remaining problem to deal with is handling warnings from decorated functions. Cache levels make this difficult to predict, at it adds a runtime dependency on the actual stack level of the function. I think we might just have to declare defeat on those functions and live with the consequences.

bmcfee:
I think the simplest option here is probably the correct one: leave the stacklevel at 2 and not worry about caching or other decorators. It's not ideal, but it's not worse than what we currently do either.

Unless there are objections, I'll merge this pr by Monday.

PR #1447, opened 2022-02-09
github action for documentation builds

### LINKED ISSUE #1443: Github action for documentation builds
At present (0.9.0), our release flow is fairly well automated.  On tag, DOIs are minted, and packages are automatically pushed up to pypi.  (With luck, conda-forge detects the new pypi packages and updates as well.)

The one part of the release process that still requires manual intervention is building the documentation site.  As I've stated elsewhere, this is complex because we want to maintain historical documentation for 3 major revisions (or 3 years), and each doc build has numerous computationally expensive examples that produce figures.  This rules out using readthedocs, as our examples take longer to build than is allowed there.  (Memory requirements are huge as well. :frowning_face: )

Relatedly, the conda environment that I maintain to build historical docs is rather quirky and brittle.  There is often a very specific configuration of dependency versions that satisfy our target versions' requirements.  Keeping this on my private development machine is okay for the short term, but hardly a good solution long term.

Ideally, we could migrate this up to a github action.  This could be done easily enough by exporting my local conda environment and setting up a workflow to run sphinx-multiversion.

The only parts that I'm not sure about are:

1. How to set up an appropriate trigger?  Should it be any time we merge to main and/or tag?  Or should it be manual?
2. How to ensure that the action does not time out?  I'm not too familiar with the action configs for this sort of thing, but it wouldn't surprise me if it's not configurable.
3. How would we actually deploy the build?  At present, I rsync the build over to the librosa-doc repository, do some manual cleanup for dead files, and then push the result.  It's not obvious to me how this could be automated, and I suspect that getting the permissions right across repositories could be a hassle.

Anyone with experience in these issues, please comment and help us out!

bmcfee:
#### Reference Issue
Fixes #1443 

This PR creates a new workflow / action for documentation builds.

#### Progress tracker

- [x] conda environment for doc builds
- [x] workflow to run sphinx-mv
- [x] deploy to testing doc repo doc-playground
- [x] auto-generate "latest" link
- [x] decide how to handle `main` or git-hash versioning
- [x] revert version matching to full (3yr) historical build
- [x] deploy to live doc repo
- [x] switch triggers to main/merge and tag


bmcfee:
This PR also relaxes the sphinx-rtd-theme requirement to 1.x.

bmcfee:
Got this working almost the way we want it.  The only snag is that the deployment action we're using does not propagate symbolic links, so the `latest` link does not show up.  I've opened an issue for this, and will pick this back up when it's resolved.

bmcfee:
A few updates on this one:

- I think we should trigger on merge->main and tag.  The former to maintain an active doc site for development, the latter so that docs automatically deploy on release.  I don't think we should trigger on manual dispatch.
- I've renamed the versions in the sidebar to be more descriptive for users: "Releases" and "Development versions" rather than "Tags" and "Branches".  I expect that only "main" will appear in the dev versions section, but am leaving open the possibility of an extended dev branch needing a dedicated doc build (e.g. if we make huge changes from 1.x to 2.x).
- I've fixed the banners so that old versions issue a warning and link to the latest; dev versions issue a slightly different warning.
- The playground site for testing this deployment is up: https://librosa.org/doc-playground/0.9.0/ 
- I've opened a PR to the git-deploy action so that we can get the `latest` link to propagate automatically.

bmcfee:
Ok, https://librosa.org/doc-playground/ is now behaving exactly as I want it to with symlink to latest, recent releases, and a dev branch build.  I think this is basically good to ship once I figure out how to set up the right triggers.

PR #1453, opened 2022-02-18
Migrate test data to submodule

### LINKED ISSUE #319: Move static test data to a submodule

bmcfee:
#### Reference Issue
Fixes #319 

This PR moves our large, static test fixture set into its own submodule repository.  The eventual goal here is to bring down the main repo size without sacrificing our fixture suite.

- [x] create submodule repo: librosa/librosa-test-data
- [x] update CI workflow
- [x] update instructions for contributors
-  BFG the old test data folder --> cannot be done in branch

#### Any other comments?

There's probably a discussion to be had about maintaining the old fixture data, in case we want to add new examples. (I doubt that we will, but it's something to consider.)

I've also taken the liberty of adding a concurrency quick-cancel to the CI jobs here so that a new push will cancel currently running jobs.


bmcfee:
Note: the BFG step is probably safest to do after merging this one.

bmcfee:
I think this is basically done, except for the history purge which I'll manage separately.

It would be nice if someone else could take a look at this and verify that the instructions work as expected, etc.  Maybe @faroit (raised the original issue, ages ago...) has thoughts?

bmcfee:
It occurs to me that one other source of repo bloat here is our old doc site in the gh-pages branch.  The current doc site weighs about 550MB (~45MB per version + 300MB .git), but we only started that repo in 2020.  Prior to that, doc builds were synchronized to the main repo's `gh-pages` branch (now just a shallow redirect), and that probably accounts for quite a bit of .git bloat.

Clearing the old doc site from our history will take a bit more work, but it can be done outside this PR.

bmcfee:
Aha!  Purging the old example notebook gets us down into a good spot:

```
 →  dust -r .
  59M └─┬ .                                                         │██████████████████ │ 100%
  54M   ├─┬ .git                                                    │ █████████████████ │  90%
```

For reference, this command https://stackoverflow.com/a/57560917 is super useful for tracking down piggy files.

bmcfee:
Ok, I'm happy with it now :grin: 

```
 →  dust -r .
  25M └─┬ .                                                         │██████████████████ │ 100%
  19M   ├─┬ .git                                                    │    ██████████████ │  75%
```

bmcfee:
Aside: I'll handle most of the BFG'ing of non-test data (ie old docs) outside of this PR.

bmcfee:
I've been test-driving the updated contributor instructions on separate copies of the repo, and while I think the instructions are basically correct, there might be a bit of a snag when it comes to testing this out from the current PR.  This is because switching to a branch where a submodule replaces a directory that exists in the main branch, possibly because the new branch also removes the old files, leads to a somewhat weird state.  Switching into the new branch is fine, but switching back to main afterwards causes issues because the submodule includes one more file (README.md) that wasn't in the main branch, and this is apparently confusing git.  I'm pretty sure this is a temporary thing though, and once we merge then everything will be fine.

It also seems to require an extra `git submodule init` or `git pull --recurse-submodules` on my machine (compared to our updated CI).  This might be down to my local git config, I'm not 100% sure.  That said, I've revised the contributor docs to account for this, so I think it's okay to merge.  We can always amend things later if this isn't completely correct.

bmcfee:
I think this is all working as expected, and I'd like to merge and finish up the cleanup unless there are objections.

The only potential hiccup in this is that devs that already have a clone/fork will need to download a fresh clone (to see the reduced repo size).  I'm not sure what the best medium for communicating that message will be, so any suggestions would be appreciated

stefan-balke:
LGTM and thanks for doing this. Can remember this was on the list forever!
For communication:
One could put it in the docs as a sidetone, also in the release notes and when announcing it on Twitter. In the end it doesn't harm people if they still "waste" some memory and in the end, active devs will need the test data...

bmcfee:
> in the release notes and when announcing it on Twitter

Yeah, that's probably the best way to call attention to it. I've also updated the contributing docs here, so I think we're all set.

bmcfee:
Following up to confirm that after merging into main, the recursive clone instructions indeed work as expected.

PR #1477, opened 2022-04-19
Allow preconstructed audioread objects in load

### LINKED ISSUE #1450: Option to force audioread to use ffmpeg
When loading an mp3, librosa calls audioread. On my system audioread then calls gstreamer instead of ffmpeg, but gstreamer does a poor job decoding the file. It decodes much better with ffmpeg. When I call audioread.available_backends(), the output is:

```
[<class 'audioread.rawread.RawAudioFile'>, <class 'audioread.gstdec.GstAudioFile'>, <class 'audioread.ffdec.FFmpegAudioFile'>]
```
So I'd like an option that forces audioread to prefer ffmpeg. Or even better, always just force it to use ffmpeg if it is available. The audioread audio_open method has a backends argument you can use for that. Or a change could be made in audioread so it checks for ffmpeg before checking for gstreamer. That's probably the simplest solution.

bmcfee:
#### Reference Issue
Fixes #1450 

This PR makes it possible to use a pre-constructed audioread object for loading.  This can be used, as pointed out in #1450, to force the use of a specific decoder rather than relying on audioread to select one automatically.  The interface is otherwise the same as we have for pre-opened Soundfile objects.

#### Any other comments?

This PR bumps the minimum supported version of audioread up from 2.1.5 to ~~2.1.7~~ 2.1.9, which introduced the `audioread.available_backends()` function.  ~~2.1.7 came out in 2019~~ 2.1.9 came out in 2020, so I don't expect any major issues here.

(*EDIT*: technically only 2.1.7 is necessary, but it does not exist on conda-forge.  Since .8 and .9 are minor bugfix patches ontop of that, I've taken the liberty of rounding up here.)

Since libsndfile 1.1 is out now, I've softened the language around mp3 support to save us some confusion going forward.  The updates haven't quite propagated through conda-forge yet, but once they do, things should start working automatically without any additional changes from users.

The one thing I'm a little dicey on here is that I've added a TypeError catch to detect Soundfile failure.  This is triggered when an audioread object is given as input to soundfile, whereas we previously only needed to catch RuntimeErrors. This is a little sloppy, IMO, but it was the minimal viable change that I could sort out to get the intended effect.  A better solution would involve rewriting the dispatch logic around which decoders are called from load.

All that said, my goal is still to remove audioread support eventually, so I don't feel too strongly about this.  Unless it introduces really incorrect behavior, I'm inclined to keep it as is out of simplicity and inertia.


bmcfee:
Tests appear fine, but failing because codecov action deprecation.  Rolling in an update to codecov-action@v3 to see if that fixes it.

bmcfee:
Well that was a CI yak shave if ever there was one.  But I think this is good to go now.

Maybe @stefan-balke would be up for taking a quick look at this, having previously contributed on IO-related issues?

bmcfee:
After pondering this one a bit more, I think it's running aground of #1228, and I'm less happy with how i've implemented the dispatching here.  I see two options:

1) Punt this feature until soundfile updates so that we can catch the SoundFileError exception instead of guessing at failures with value/typeerrors.  This would probably move this feature from 0.9.2 to the 0.10.x series since we'd be significantly bumping a dependency version.
2) Refactor the load function to be more explicit about detecting pre-built loader objects.

(2) seems like it's probably the safer bet, but also it's more work.  I'll give it some more thought.

bmcfee:
I went ahead with option 2 and refactored the load function.  We're back to the simple exception catch now: if the user provides an audioread object, it does not attempt soundfile at all.

I expect that this PR changes things enough that #1228 will need to be completely rewritten.

bmcfee:
I think this is good to merge - the dangling coverage drop doesn't worry me much as it's just an exception-reraise that will likely disappear when we resolve #1228

PR #1485, opened 2022-05-21
Support for "h" and "m" in TimeFormatter

### LINKED ISSUE #1364: specshow's xticklabels do not print hour/minutes for short signals with offset
**Describe the bug**
I am using `librosa.specshow` with `x_axis="time"` to display spectrograms. I think i've found a bug in the xticklabels.
If the signal is less than one minute long (e.g., Brahms), it all goes well.
If it is more than one minute long and i'm showing all of it, it also goes well.
But if the signal is 90 seconds long and i want to show the last 40 seconds of it, librosa does not print xticklabels in `mm:ss` format but in scientific notation. (see below)

**To Reproduce**

Example:
```python
import librosa

def specshow_with_offset(y, sr, offset):
    S = librosa.power_to_db(np.abs(librosa.stft(y)))
    times = librosa.times_like(S, sr=sr)
    librosa.display.specshow(
        S, sr=sr, x_coords=times+offset, x_axis="time",
        y_axis="hz")
    
plt.figure(figsize=(6, 6))
y, sr = librosa.load(librosa.ex("brahms"))
for fig_id, offset in enumerate([0, 60, 120]):
    plt.subplot(3, 1, 1+fig_id)
    specshow_with_offset(y, sr, offset)
```

**Expected behavior**
Use `mm:ss` or `hh:mm:ss` notations depending on the values taken by `x_coords` rather than depending on the duration of the input.

**Screenshots**
Screenshot 2021-08-17 at 17 23 56

**Software versions***
```
macOS-10.15.7-x86_64-i386-64bit
Python 3.8.10 (default, May 19 2021, 11:01:55) 
[Clang 10.0.0 ]
NumPy 1.21.0
SciPy 1.4.1
librosa 0.8.1
INSTALLED VERSIONS
------------------
python: 3.8.10 (default, May 19 2021, 11:01:55) 
[Clang 10.0.0 ]

librosa: 0.8.1

audioread: 2.1.9
numpy: 1.21.0
scipy: 1.4.1
sklearn: 0.22.1
joblib: 1.0.1
decorator: 5.0.9
soundfile: 0.9.0
resampy: 0.2.2
numba: 0.48.0

```

**Additional context**
Related: #737 #760

lostanlen:
#### Reference Issue
Fixes #1364: "specshow's xticklabels do not print hour/minutes for short signals with offset"

This PR allows to pass `x_axis=` (or `y_axis=`) `"h"` (resp. `"m"`) to `specshow` (or `waveshow`, or anything using `TimeFormatter`) to display time units in hours (resp. minutes) instead of seconds.

#### Any other comments?
* I have updated images `test_time_unit.png` and `test_time_unit_lag.png` to reflect this new feature.

Before:
baseline

After:
test_time_unit

* Given `x_axis="s"`, i suggest `Time (seconds)` as an `xlabel` instead of `Time (s)`. I have updated `__decorate_axis` for this. This is because i find the previous label `Time (s)` difficult to read, especially when the text is small. Likewise for other units: `Time (minutes)` instead of `Time (m)`, etc.

* `lag_h` and `lag_m` are also supported

* Note that this PR leads to a net reduction in LOC count :) by 5 lines


bmcfee:
This generally looks good at a glance, but I'll be on vacation for the next week and unable to give a proper review until I get back.

bmcfee:
This LGTM.  I'd like to hold off on merging until we get the 0.9.2 release out though.

lostanlen:
> This LGTM. I'd like to hold off on merging until we get the 0.9.2 release out though.

OK so, proposed plan:

(1) release 0.9.2
(2) merge this #1485
(3) support strftime formats in specshow (`if "%" in x_axis:` branch)
(3bis, optional: release 0.9.3)
(4) bump requirement: matplotlib 3.6
(5) support two new types in x_coords and y_coords: numpy array of datetime64 and timedelta64
(6) timedelta_like(X, sr, hop_length)
(7) release 0.10
(8) write a blog post (https://librosa.org/blog/) entitled `Specshow time, folks!`
(9) profit

bmcfee (approved):

bmcfee:
I think this is good to go, but is there anything here that needs to be updated or integrated with the now merged changes in #1501 ?

lostanlen:
you're right, i should integrate these new units in #1501 code

bmcfee:
I think this needs to be rebased to main also.

librosa/display.py
```diff
@@ -149,27 +152,22 @@ def __call__(self, x, pos=None):
             value = x
             sign = ""
 
-        if self.unit == "s":
+        if self.unit == "h" or (self.unit==None and (vmax - vmin > 3600)):
```
bmcfee, line 155: Equality testing against `None` is frowned upon; should be `self.unit is None`

librosa/display.py
```diff
@@ -1098,30 +1106,21 @@ def __decorate_axis(
         axis.set_major_locator(MaxNLocator(prune=None, steps=[1, 1.5, 5, 6, 10]))
         axis.set_label_text("Time")
 
-    elif ax_type == "s":
-        axis.set_major_formatter(TimeFormatter(unit="s", lag=False))
+    elif isinstance(ax_type, str) and ax_type in time_units.keys():
```
bmcfee, line 1109: No need for `.keys()` here, can just say `ax_type in time_units`.

I'm also not sure about the type check here.  All the keys are strings anyway, so any ax_type which fails a string type check will also fail the membership test.

The only reason to do a type check here would be to short-circuit fail the membership test, which could be useful if `ax_type` is ever a non-hashable object.  Can that situation arise?

bmcfee (changes requested):
A few minor pythonic changes requested, but otherwise looks good!

bmcfee (approved):

PR #1493, opened 2022-06-02
bug-fix for multichannel splitting

### LINKED ISSUE #1489: Numpy array truthiness error during effects.split
When loading a file and trying to run librosa.effects.split() on it, I get this error:

```
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/librosa/effects.py", line 574, in split
    if non_silent[0]:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

**To Reproduce**
```
import librosa
import numpy as np

y, sr = librosa.load("path/to/file.mp3", sr=44100, mono=False)
intervals = librosa.effects.split(y, top_db=22, ref=np.max, frame_length=44100, hop_length=44100)
```

**Expected behavior**
The split effect returning an array of non-silent intervals.

**Software versions***
```
INSTALLED VERSIONS
------------------
python: 3.8.12 (default, Oct 13 2021, 06:42:42) 
[Clang 13.0.0 (clang-1300.0.29.3)]

librosa: 0.9.1

audioread: 2.1.9
numpy: 1.22.4
scipy: 1.8.1
sklearn: 1.1.1
joblib: 1.1.0
decorator: 5.1.1
soundfile: 0.10.3
resampy: 0.2.2
numba: 0.55.2

numpydoc: None
sphinx: None
sphinx_rtd_theme: None
sphinxcontrib.versioning: None
sphinx-gallery: None
pytest: None
pytest-mpl: None
pytest-cov: None
matplotlib: None
presets: None
```

**Additional context**
This is a file I haven't touched in a while, so I apologize if it is something that is covered in a changelog somewhere. However, I was unable find any similar issues.

bmcfee:
#### Reference Issue
Fixes #1489

This PR adds a squeeze to the non-silence detection helper function in the effects module.  This ensures that the output here is always one-dimensional.

The squeeze operation here only applies to leading axes, and explicitly skips the trailing (frame) axis.

#### Any other comments?

I think this should be good to merge after CI passes.  Eyeballs are welcome as always.

### DIFF
--- a/tests/test_effects.py
+++ b/tests/test_effects.py
@@ -252,6 +252,19 @@ def test_trim_empty():
     assert idx[1] == 0
 
 
+def test_trim_multi(y_multi):
+    # Test for https://github.com/librosa/librosa/issues/1489
+    y, sr = y_multi
+    librosa.effects.trim(y=y)
+
+
+def test_split_multi(y_multi):
+    # Test for https://github.com/librosa/librosa/issues/1489
+    y, sr = y_multi
+
+    librosa.effects.split(y=y)
+
+
 @pytest.fixture(
     scope="module",
     params=[0, 1, 2, 3],
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -445,6 +445,9 @@ def _signal_to_frame_nonsilent(
     # Aggregate everything but the time dimension
     if db.ndim > 1:
         db = np.apply_over_axes(aggregate, db, range(db.ndim - 1))
+        # Squeeze out leading singleton dimensions here
+        # We always want to keep the trailing dimension though
+        db = np.squeeze(db, axis=tuple(range(db.ndim - 1)))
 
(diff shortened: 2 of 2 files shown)


tests/test_effects.py
```diff
@@ -252,6 +252,19 @@ def test_trim_empty():
...
+    # Test for https://github.com/librosa/librosa/issues/1489
+    y, sr = y_multi
+    librosa.effects.trim(y=y)
+
+
+def test_split_multi(y_multi):
+    # Test for https://github.com/librosa/librosa/issues/1489
+    y, sr = y_multi
+
+    librosa.effects.split(y=y)
```
bmcfee, line 265: It is confirmed that this test fails on main.  Trim had no problem, and we do already have a multichannel trim test.

Issue #1462, opened 2022-03-10
More options for bandwidth estimation in similarity/recurrence matrix construction


bmcfee:
**Is your feature request related to a problem? Please describe.**

At present, the recurrence_matrix function provides a rudimentary bandwidth estimator, which uses the median k'th nearest neighbor distance.  A user can also provide an explicit value for bandwidth, but doing anything else turns out to be a bit of a pain.

**Describe the solution you'd like**

We could expand out the functionality for the bandwidth parameter to support a few different options.  Median-to-k is one option, another would be the pairwise adaptive method used by `Bo Wang, Jiayan Jiang, Wei Wang, Zhi-Hua Zhou, and Zhuowen Tu, “Unsupervised metric fusion by cross diffusion,” in Computer Vision and Pattern Recognition (CVPR), 2012 IEEE Conference on. IEEE, 2012, pp. 2997–3004.` (and later in the paper I coauthored with @ctralie), which essentially averages the distance D(i, j), the mean distance among i's nearest neighbors, and the mean distances among j's nearest neighbors.

One hiccup here, compared to the reference paper(s), is that we should ideally support arbitrary distance functions and not just euclidean distance.  This doesn't matter too much, except where squares of distances are concerned, and how that might interact with averaging.  (Squaring or not squaring distances has no effect on median calculation, since squaring is a monotonic-up function, but it will slightly change other aggregators like mean.  Just making a note of it here.)

**Describe alternatives you've considered**
We can always do this kind of thing by hand, but it's tedious and should be easier.

**Additional context**

We could probably come up with a few other recipes to include here.  I'm thinking of identifying them by name/string.

Additionally, we might want to consider adding a parameter `full=False` (by default) which controls whether the recurrence matrix is a k-nearest-neighbor matrix or a full matrix.  Right now we hack it by setting `k=N`, but this messes with the use of `k` in bandwidth estimation.  Setting `full=True` would essentially bypass the knn step and build the full affinity / kernel / distance matrix.

I'm including this idea in this issue mainly because it's relevant to bandwidth estimation, but we could also handle it separately.

Tagging @tomxi here because he's expressed interest in working on implementing the bandwidth issue.

tomxi:
Documenting some of the discussions that happened off-line:
We want to support more options for bandwidth estimation. Currently, bandwidth estimation supports only a scalar input or `None` in which case it would estimate a scalar bandwidth (Median-to-K).

A couple of other bandwidths would make sense besides Median-to-K: 
[...]

PR #1495, opened 2022-06-07
added more options for bandwidth estimation

tomxi:
#### Reference Issue

Implements #1462

This PR adds more options for bandwidth estimation for recurrence matrix calculation.
Without breaking changes to the current API, there are now more string based options to do some more advanced bandwidth estimation.

#### Any other comments?
Not sure how to write meaningful unit tests for the more fancy bandwidth options...
Probably would need some input on doc string as well.


librosa/segment.py
```diff
@@ -347,12 +348,39 @@ def recurrence_matrix(
         ``exp( - distance(i, j) / bandwidth)`` where ``bandwidth`` is
         as specified below.
 
-    bandwidth : None or float > 0
+    bandwidth : None, float > 0, or str in {'med_k_scaler', 'mean_k', 'gmean_k',
+    'mean_k_avg', 'gmean_k_avg', 'mean_k_avg_and_pair'}
```
bmcfee, line 352: The linting CI is currently failing because this line is not properly indented

librosa/segment.py
```diff
@@ -1138,3 +1176,77 @@ def path_enhance(
...
+        bandwidth = bw_mode
+        if bandwidth <= 0:
+            raise ParameterError(
+                "Invalid scalar bandwidth={}. " "Must be strictly positive.".format(bandwidth)
+            )
+        return bandwidth
+
+    if bw_mode is None:
+        bw_mode = 'med_k_scaler'
+
```
bmcfee, line 1194: should be `scalar` not `scaler`

librosa/segment.py
```diff
@@ -488,16 +525,17 @@ def recurrence_matrix(
     for diag in range(-width + 1, width):
         rec.setdiag(0, diag)
 
-    # Retain only the top-k links per point
-    for i in range(t):
-        # Get the links from point i
-        links = rec[i].nonzero()[1]
+    if not full:
```
bmcfee, line 528: If we're in full mode, does it make sense to hollow out the diagonal in line 525?

librosa/segment.py
```diff
@@ -1138,3 +1176,77 @@ def path_enhance(
         np.clip(R_smooth, 0, None, out=R_smooth)
 
     return R_smooth
+
+
+def __affinity_bandwidth(rec, bw_mode, k):
+
+    # the api allows users to specify a scalar bandwidth directly, besides the string based options.
+    if isinstance(bw_mode, (int, float)):
```
bmcfee, line 1184: This check doesn't seem quite right to me, as we might also want to support a numeric vector input (one bandwidth per point).  We should figure out the right way to identify numeric values including numpy dtypes.

bmcfee (changes requested):
First pass review here

bmcfee:
For getting test coverage up, I think the best thing to do is construct a small example (say 10 datapoints) and pre-compute the bandwidth using each method (and for different values of k as needed), then just regression-test them.

librosa/segment.py
tomxi: Thanks. Will fix all instance of this typo.

librosa/segment.py
tomxi: I think you are right / it wouldn't make sense to hollow out the diagonal. That would mean `width` will also be ignored when in full mode. If the user wants to hollow out the diagonal it's pretty easy for them to do post-hoc as well, although this would have an effect on the bandwidth calculation...

librosa/segment.py
```diff
@@ -1138,3 +1176,77 @@ def path_enhance(
...
+        # Compute k nearest neighbors' distance and sort ascending
+        knn_dist_row = np.sort(rec[i, links].toarray()[0])[:k]
+
+        # make sure self-links have a distance of 0 and not -1 in self mode
+        knn_dist_row[knn_dist_row < 0] = 0
+        knn_dists.append(knn_dist_row)
+
+    # take the last element of each list for the distance to kth neighbor
+    dist_to_k = np.asarray([dists[-1] for dists in knn_dists])
+    avg_dist_to_first_ks = np.asarray([np.mean(dists) for dists in knn_dists])
```
tomxi, line 1219: In the case that `sym == True` knocks out some links, should I treat them as 0 distance in my average distance to k? or not include them in the average calculation? The current code compute the means of the links that are left if there are less than `k` links left after taking only mutual k nearest neighbors. The alternative would be `np.sum(dists) / k` in place of `np.mean(dists)`

librosa/segment.py
tomxi: Yea... I got lazy here and only supported scalar bandwidth besides the string-based options. So for numeric vector input, we only want to support arrays with shapes the same as `rec.shape` right?

librosa/segment.py
```diff
@@ -1138,3 +1176,77 @@ def path_enhance(
...
+
+    if bw_mode == 'med_k_scaler':
+        return np.nanmedian(dist_to_k)
+
+    if bw_mode in ['mean_k', 'gmean_k']:
+        # building bandwidth components (sigma) using sparse matrix structures and indices
```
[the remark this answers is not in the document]
tomxi: What's the best way to do this? I'm thinking calling `rec = rec.tocsr()` at the beginning of this helper function, does that solve it?

bmcfee (changes requested):
This all LGTM so far!

As discussed offline, only remaining functionality here is to port these features over to cross_similarity.

bmcfee:
Since merging #1564 , a conflict popped up with this PR.  I've resolved that conflict for you, but you'll likely need to go in and add type annotations to the new helper functions to get lint / type annotation tests to pass.

librosa/segment.py
```diff
@@ -1220,3 +1262,94 @@ def path_enhance(
         np.clip(R_smooth, 0, None, out=R_smooth)  # type: ignore
 
     return np.asanyarray(R_smooth)
+
+
+def __affinity_bandwidth(rec, bw_mode, k):
```
bmcfee, line 1267:
```suggestion
def __affinity_bandwidth(rec: scipy.sparse.csr_matrix, bw_mode: Union[np.ndarray, int, float, str], k: int) -> Union[float, np.ndarray]:
```

librosa/segment.py
```diff
@@ -1220,3 +1262,94 @@ def path_enhance(
...
+
+        # Compute k nearest neighbors' distance and sort ascending
+        knn_dist_row = np.sort(rec[i, links].toarray()[0])[:k]
+        knn_dists.append(knn_dist_row)
+
+    # take the last element of each list for the distance to kth neighbor
+    
+    
+    dist_to_k = np.asarray([dists[-1] for dists in knn_dists])
+    avg_dist_to_first_ks = np.asarray([np.mean(dists) for dists in knn_dists])
```
bmcfee, line 1322: This seems like it could be faster by collecting knn_dists into an ndarray first, and then using `np.mean(knn_dists, axis=-1)`

tomxi: I think I did it this way because the lists `dists` in `knn_dists` can be of different length, and won't collect nicely into an ndarray... Not exactly sure when different lengths `dists` list would pop up, but I'll leave this potential efficiency issue for another PR if that's ok.

bmcfee:
The CI failures here are again due to codecov's rate limiting, and not a big deal.

The lint failures should be fixed though - I suggest installing mypy locally so you can test it more quickly than waiting for each push to fail.

bmcfee (approved):

bmcfee:
Thanks @tomxi ! This all looks great to me.  I'll give the doc build a quick check, but otherwise have no further comments.  Is there anything else you think needs attention here, or is it ready to go?

tomxi:
Thanks for the review. Nothing on my end, I think it's ready.

bmcfee:
:rocket: :shipit:

PR #1501, opened 2022-06-10
Square plots when different axis types are compatible

### LINKED ISSUE #1499: specshow auto-aspect should match on data, not decoration

PR #1312 implemented auto-aspect pinning for the case where we have a square image to show (e.g., a self-similarity matrix).  The logic requires that the x and y coordinates are equivalent, and that `x_axis == y_axis`:

https://github.com/librosa/librosa/blob/7ac022a8496126e95b46a7d56ad880a328359bda/librosa/display.py#L1278-L1282

This works great much of the time, but is I think overly strict when we might want to have similarly-typed axes with different decorations, eg `time` vs `time_s`.  In that situation, the axes matching fails, and the aspect is not set square.

I could see this being useful in the context of #1364 

I think we can soften the comparison logic a bit, so that any "time-like" axis types are considered compatible; ditto for frequency, ditto for tempo.  Probably the easiest way to implement this is to create a module-level `set` object with all tuples of compatible axis types:

```python
AXIS_COMPAT = set([('time', 's'), ('time', ms'), ...])

...

axes_same_and_not_none = (x_axis, y_axis) in AXIS_COMPAT
...
```

kianelbo:
#### Reference Issue
Fixes #1499 

This PR softens the comparison logic to determine whether plots could be squared. In addition to having exact axis types, different types in the same category, such as 'time' and 'ms', are also considered compatible and can have equal-scaling.

#### Any other comments?
I'm not sure if every two types in the same category are compatible, e.g., `linear` and `cqt_note`.

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -618,16 +618,20 @@ def __envelope(x, hop):
     return x_frame.max(axis=1)
 
 
-_categorical_axis_types = ['chroma', 'chroma_h', 'tonnetz', 'frames']
-_time_axis_types = ['linear', 'fft', 'hz', 'log,' 'fft_note', 'fft_svara', 'mel', 'cqt_hz', 'cqt_note', 'cqt_svara']
-_freq_axis_types = ['time', 's', 'ms', 'lag', 'lag_s', 'lag_ms']
-_tempo_axis_types = ['tempo', 'fourier_tempo']
+_chroma_ax_types = ('chroma', 'chroma_h', 'chroma_c',)
+_cqt_ax_types = ('cqt_hz', 'cqt_note', 'cqt_svara',)
+_freq_ax_types = ('linear', 'fft', 'hz', 'fft_note', 'fft_svara',)
+_time_ax_types = ('time', 's', 'ms',)
+_lag_ax_types = ('lag', 'lag_s', 'lag_ms',)
+_misc_ax_types = ('tempo', 'fourier_tempo', 'mel', 'log', 'tonnetz', 'frames',)
 
 _AXIS_COMPAT = set(
-    [(t, t) for t in _categorical_axis_types] +
-    [t for t in product(_time_axis_types, _time_axis_types)] +
-    [t for t in product(_freq_axis_types, _freq_axis_types)] +
-    [t for t in product(_tempo_axis_types, _tempo_axis_types)]
+    [(t, t) for t in _misc_ax_types] +
+    [t for t in product(_chroma_ax_types, _chroma_ax_types)] +
+    [t for t in product(_cqt_ax_types, _cqt_ax_types)] +
+    [t for t in product(_freq_ax_types, _freq_ax_types)] +
+    [t for t in product(_time_ax_types, _time_ax_types)] +
+    [t for t in product(_lag_ax_types, _lag_ax_types)]
 )
 
 
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -35,6 +35,7 @@ Miscellaneous
 
 """
 
+from itertools import product
 import warnings
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks for getting this going!  It looks like a good start, but there are a couple of subtleties to work out.  I'll note them in the code review.

The main other thing that we'll need here is for you to update the baseline image for the revised image comparison test.  This actually raises the issue that we haven't exactly documented how to do this in a useful way, so I'll add a separate issue for that.  The short version is that once you have your pytest properly setup, to run:

```
pytest --mpl-generate-path=tmp  -v tests/test_display.py -k test_[WHICHEVER_TEST_CHANGED]
```
ie in this case, it would be `-k test_auto_aspect`.  This will generate a new baseline image stored under the `tmp/` folder, which can be copied over to `tests/baseline_images/test_display/` (assuming it looks correct) and added to the PR.

librosa/display.py
```diff
@@ -617,6 +618,19 @@ def __envelope(x, hop):
     return x_frame.max(axis=1)
 
 
+_categorical_axis_types = ['chroma', 'chroma_h', 'tonnetz', 'frames']
```
bmcfee, line 621: Not all categorical axes should be considered "equal".  Probably only chroma and chroma_h make sense here.

frames is a time-like axis, but really is only compatible with itself.

librosa/display.py
```diff
@@ -617,6 +618,19 @@ def __envelope(x, hop):
...
+_time_axis_types = ['linear', 'fft', 'hz', 'log,' 'fft_note', 'fft_svara', 'mel', 'cqt_hz', 'cqt_note', 'cqt_svara']
```
bmcfee, line 622: These are frequency-like axes, not time-like.  I think the equivalence class is too broad here though, and should be reduced to the following:

- linear, fft, hz, fft_note, fft_svara
- log
- mel
- cqt_hz, cqt_note, cqt_svara

Since singletons (log and mel) are silly to include, we could shortcut this in the check by saying `x_axis == y_axis or (x_axis, y_axis) in _AXIS_COMPAT`.  This way we won't have to clutter the object too much.

librosa/display.py
```diff
@@ -617,6 +618,19 @@ def __envelope(x, hop):
...
+_freq_axis_types = ['time', 's', 'ms', 'lag', 'lag_s', 'lag_ms']
```
bmcfee, line 623: Time and lag should (I think?) be treated separately.

librosa/display.py
```diff
@@ -617,6 +618,19 @@ def __envelope(x, hop):
...
+_tempo_axis_types = ['tempo', 'fourier_tempo']
```
bmcfee, line 624: I'm not sure if we should treat tempo and fourier_tempo as equivalent.  (Honest question.)  The hiccup here is that `tempo` can include "infinite" bpm while fourier_tempo cannot, and this might do some screwy things if we try to match them.

The more I think about this, I'm starting to wonder if there would be value in doing an allclose check on the actual coordinates used, rather than matching on the shape of the array being displayed.

bmcfee (changes requested):
Good start, but there are some things noted in the review comments to narrow down.

tests/test_display.py
```diff
@@ -841,15 +842,15 @@ def test_same_axes(x_axis, y_axis, xlim, ylim, out):
 
 def test_auto_aspect():
```
kianelbo, line 843: > ie in this case, it would be -k test_auto_aspect. This will generate a new baseline image stored under the tmp/ folder, which can be copied over to tests/baseline_images/test_display/ (assuming it looks correct) and added to the PR.

This test is not decorated with `mpl_image_compare`, so I thought maybe image comparison is not needed. The input data is also very mock-ish (`np.zeros((12, 12))`) and does not produce a useful output.

bmcfee: oh fair enough!  Upon a closer inspection, this test works entirely on the axis properties and not on the rendered image.  Please disregard my earlier comment.

librosa/display.py
kianelbo: Thanks for the review. I also think dividing the types into smaller groups makes more sense.
I don't get what should be checked with allclose though. Is checking xlim and ylim inadequate?

bmcfee: > I don't get what should be checked with allclose though. Is checking xlim and ylim inadequate?

I'm not sure.  Limit-matching is obviously necessary, but it might not be sufficient for imposing an equal aspect ratio.  We can have time-vs-time over the same limits but with different spacing along one axis (while still having the same array shape).  This is admittedly a very unusual use-case, but I just want to make sure we think through the implications here.

kianelbo: Do you mean doing something like `np.allclose(x_coords, y_coords)` instead of `xlim == ylim`?

bmcfee: That's what I was thinking, yes.  Like I said, I'm not 100% sure that it's necessary.  Maybe @lostanlen has some thoughts on this?

bmcfee (approved):
I think this looks good, sorry for the lag.  I'm curious for @lostanlen 's opinion though.

bmcfee (approved):

bmcfee:
Thanks @kianelbo ! LGTM, will merge once CI finishes. :shipit:

PR #1504, opened 2022-06-13
Speed up `magphase`

futurulus:

Dividing the spectrum by the magnitude is equivalent to `exp(1j * angle(D))`, but about 4x faster:

```
$ python -m timeit -s 'import librosa; y, sr = librosa.load(librosa.ex("trumpet")); D = librosa.stft(y)' 'librosa.magphase_old(D)'
100 loops, best of 3: 13.5 msec per loop
$ python -m timeit -s 'import librosa; y, sr = librosa.load(librosa.ex("trumpet")); D = librosa.stft(y)' 'librosa.magphase_new(D)'
100 loops, best of 3: 3.58 msec per loop
```

[More thorough benchmark here.](https://github.com/futurulus/librosa/blob/faster-magphase-notebook/docs/magphase.ipynb) The proposed calculation is slightly slower for very small arrays, but is faster already on a 100-element array, and continues to be ~3x faster up through at least 10M elements. 

#### Any other comments?

Some potential speedup (and simplicity) is sacrificed to prevent the new implementation from producing NaNs where the old one didn't. The new implementation is also ~5x closer to a perfect reconstruction of the original complex spectrum when the magnitude and phase are multiplied back together, but the original was already accurate enough that I'm not sure this will matter in practice.

This idea was prompted by the result of profiling `librosa.decompose.hpss`; the current implementation spends about 10% of its time in `magphase`.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1168,8 +1168,17 @@ def magphase(D, *, power=1):
     """
 
     mag = np.abs(D)
+
+    # Prevent NaNs and return magnitude 0, phase 1+0j for zero
+    zeros_to_ones = mag == 0
+    mag_nonzero = mag + zeros_to_ones
+    # Compute real and imaginary separately, because complex division can
+    # produce NaNs when denormalized numbers are involved (< ~2e-39 for
+    # complex64, ~5e-309 for complex128)
+    phase = D.imag / mag_nonzero * 1.0j
+    phase += D.real / mag_nonzero + zeros_to_ones
+
     mag **= power
-    phase = np.exp(1.0j * np.angle(D))
 
     return mag, phase
 
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1169,14 +1169,18 @@ def magphase(D, *, power=1):
 
     mag = np.abs(D)
 
+    if not np.iscomplexobj(D):
+        return mag, np.sign(D, dtype=util.dtype_r2c(D.dtype)) + (D == 0)
+
     # Prevent NaNs and return magnitude 0, phase 1+0j for zero
     zeros_to_ones = mag == 0
     mag_nonzero = mag + zeros_to_ones
     # Compute real and imaginary separately, because complex division can
     # produce NaNs when denormalized numbers are involved (< ~2e-39 for
     # complex64, ~5e-309 for complex128)
-    phase = D.imag / mag_nonzero * 1.0j
-    phase += D.real / mag_nonzero + zeros_to_ones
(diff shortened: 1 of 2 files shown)


librosa/core/spectrum.py
```diff
@@ -1168,8 +1168,17 @@ def magphase(D, *, power=1):
...
     mag = np.abs(D)
+
+    # Prevent NaNs and return magnitude 0, phase 1+0j for zero
+    zeros_to_ones = mag == 0
+    mag_nonzero = mag + zeros_to_ones
+    # Compute real and imaginary separately, because complex division can
+    # produce NaNs when denormalized numbers are involved (< ~2e-39 for
+    # complex64, ~5e-309 for complex128)
+    phase = D.imag / mag_nonzero * 1.0j
+    phase += D.real / mag_nonzero + zeros_to_ones
```
bmcfee, line 1179: I wonder if this could be faster by working directly with the components of the phase array to avoid an extra multiply?

```python
phase = np.empty_like(D)
phase.imag = D.imag / mag_nonzero
phase.real = D.real / mag_nonzero + zeros_to_ones
```

bmcfee (commented):
Thanks for this!  It generally looks great.  I have a couple of ideas for how to squeeze this a little more: one is noted below by working directly with the components of the output array.

The other would be to bypass this nonzero array construction and implement this in numba as a ufunc.  This would be a bit more effort, and might not provide as much speedup, but it would probably reduce the (temporary) memory footprint.  We can either discuss that further or not in this PR, but I wanted to make a note of it here as another option to pursue.

futurulus (commented):
Thanks for the quick review! I certainly won't discourage discussion of the numba/ufunc implementation here, but I haven't used numba directly before, so I'd be a bit out of my depth implementing it.

bmcfee:
> but I haven't used numba directly before, so I'd be a bit out of my depth implementing it.

Ok, I'm happy to leave it out of scope for this PR and come back to it later when I have time to prototype it.  (Maybe at the scipy sprint next month.)

bmcfee:
Just to follow up, I took a crack at a numba ufunc implementation for just the phase part.  It didn't improve over the above implementation.  Maybe a pure numba implementation of the whole magphase operation might be better, but for now i'm happy to drop this idea.

futurulus:
Updated to set the components individually (and added another test)

bmcfee (approved):
LGTM!

bmcfee:
This one seems good to merge, unless there's anything else to add?

lostanlen (approved):

bmcfee:
:shipit: Thanks @futurulus !

PR #1508, opened 2022-06-20
Fix typo in warning in inverse.py

LorenzNickel:


bmcfee:
Thanks for this - I wonder why our codespell action didn't catch it though?

bmcfee (approved):

lostanlen (approved):

bmcfee:
> I wonder why our codespell action didn't catch it though?

After thinking more on this (i.e. post-:coffee:) I believe it's because this is a string value and not code.  codespell does not appear to inspect string contents.

CI failures here appear to be due to a temporary outage with codecov, nothing to worry about.

Issue #599, opened 2017-07-21
Automate benchmarking


bmcfee:
This issue comes up from time to time -- it would be nice to have a dedicated page, either in the documentation or self-contained elsewhere, for storing benchmarks of the most common operations.

I think this shouldn't be hard to do with a tool like [Benchmarker](https://pythonhosted.org/Benchmarker/), but the interesting things would be to have pages of results for different librosa versions, with and without numba, and comparisons to other libraries (madmom, essentia, etc.) on the common functionality.

This is probably best done as a new project under the librosa org, and could myabe be automated using sphinx-gallery.

bmcfee:
Coming back to this: [asv](https://asv.readthedocs.io/en/stable/) seems to be the tool of choice these days.  It makes it easy to compare things like time, memory, peak memory, and arbitrary numbers (eg reconstruction error).

lostanlen:
Probably relevant: https://github.com/KinWaiCheuk/nnAudio
and Figure 11 of https://arxiv.org/abs/1912.12055

PR #1514, opened 2022-06-24
[CR needed] STFT pre-allocation

### LINKED ISSUE #871: STFT/ISTFT pre-allocate output buffers
#### Description

STFT allocates an output buffer, but sometimes you might want to have it compute directly into an existing buffer.  For example, in griffin-lim, the method alternates stft/istft for each iterate, which is then discarded.  It would be better if we could give it an `out=` variable, which it would use instead of allocating a new buffer; this way, we could cut down on redundant memory allocations.

If the shape and/or dtype doesn't line up, then it should throw an exception.

bmcfee:
#### Reference Issue
Fixes #871 

This PR adds parameter `out=` to `stft` and `istft` methods, allowing re-use of previously allocated output buffers.

Additionally, it implements some more careful handling of edge padding (when using centered mode) to reduce unnecessary copying.  The result should be a substantial speedup when processing long signals.

#### Any other comments?

This is currently WIP.  To-dos:

- [x] complete tests for stft pre-allocation
- [x] add istft and tests
- [x] update other parts of the library to make use of pre-allocation where possible, eg griffinlim
- [x] add an example to the gallery illustrating the utility, or maybe just extend the pcen example

That said, I'd like to get some early feedback on the implementation, mainly from the perspective of readability.

### DIFF
--- a/docs/examples/plot_pcen_stream.py
+++ b/docs/examples/plot_pcen_stream.py
@@ -11,6 +11,10 @@ to do dynamic per-channel energy normalization on a spectrogram incrementally.
 
 This is useful when processing long audio files that are too large to load all at
 once, or when streaming data from a recording device.
+
+It also illustrates how to use a pre-allocated output buffer for block-wise
+short-time Fourier transforms.  This provides a minor speed boost and reduction
+in memory usage when processing audio streams.
 """
 
 ##################################################
@@ -59,10 +63,15 @@ pcen_blocks = []
 # Initialize the PCEN filter delays to steady state
 zi = None
 
+# Create a handle for storing the block STFT outputs
+# After the first block has been processed, we can re-use
+# this buffer instead of allocating a new one for each block.
+D = None
+
 for y_block in stream:
     # Compute the STFT (without padding, so center=False)
     D = librosa.stft(y_block, n_fft=n_fft, hop_length=hop_length,
-                     center=False)
+                     center=False, out=D)
 
     # Compute PCEN on the magnitude spectrum, using initial delays
     # returned from our previous call (if any)
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -367,6 +367,7 @@ def istft(
     center=True,
     dtype=None,
     length=None,
+    out=None,
 ):
     """
     Inverse short-time Fourier transform (ISTFT).
@@ -425,6 +426,11 @@ def istft(
         If provided, the output ``y`` is zero-padded or clipped to exactly
         ``length`` samples.
 
+    out : np.ndarray, optional
+        Optional pre-allocated output array.
+        FIXME: more
+
+
     Returns
     -------
     y : np.ndarray [shape=(..., n)]
@@ -481,7 +487,7 @@ def istft(
     # For efficiency, trim STFT frames according to signal length if available
     if length:
         if center:
-            padded_length = length + int(n_fft)
+            padded_length = length + 2 * (n_fft//2)
         else:
             padded_length = length
         n_frames = min(stft_matrix.shape[-1], int(np.ceil(padded_length / hop_length)))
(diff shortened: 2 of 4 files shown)


bmcfee:
And we've hit a pretty serious snag that I hadn't considered.  The optimization I've implemented to only pad at the edges breaks padding semantics for `wrap` mode.  This isn't a huge deal in the big picture, but in the short term it does break some of our unit tests (eg on reassigned frequencies).

Now, I'm not entirely sure why we're using wrap padding in these tests - probably this is to get around edge transients for the stationary test signals, but it's a pretty crude hack.

However, given that this does change semantics in a pretty subtle way, I don't think we can include this optimization in a minor point release.  We do have the option of rolling back to just have pre-allocated output arrays without the copy-pad optimization, but my preference would be to keep this intact and bump it to a later release (eg 0.10).

bmcfee:
I've added exceptions for now-unsupported padding modes in stft.

I wound up disabling the `center=True` test for reassigned spectrograms for now.  We used `pad_mode='wrap'` not out of actual necessity, but just to minimize the influence of edge transients on our test case.  Ultimately this was a bit of a hack, and I don't think it diminishes much to simplify this test and only check `center=False` (no padding).

The remaining to-do's in the issue are still TBD though.

lostanlen:
@bmcfee apologies for that vagrant commit 4303fa8 🤦🏻  i meant to push it to #1485 instead

bmcfee:
All good - had to rebase this anyway, so i just rolled back that commit at the same time.

bmcfee:
Prototype of griffin-lim using pre-allocation for both forward and inverse STFT's gets us a minor boost.  Example is 90 seconds of random noise at float32.

Here, `librosa.griffinlim` is not using pre-allocation, but is using the optimized padding from this PR.  `griffinlim` is using pre-allocation.  Results are numerically identical:

```python
>>> sr = 22050
>>> y = np.random.randn(22050 * 90).astype(np.float32)
>>> %timeit librosa.griffinlim(S, random_state=0, n_iter=64)
16.4 s ± 1.64 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
>>> %memit librosa.griffinlim(S, random_state=0, n_iter=64)
peak memory: 518.35 MiB, increment: 151.34 MiB
>>> %timeit griffinlim(S, random_state=0, n_iter=64)
15.1 s ± 370 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
>>> %memit griffinlim(S, random_state=0, n_iter=64)
peak memory: 518.42 MiB, increment: 151.40 MiB
```
Peak memory hasn't changed much, which I think just means the gc was doing a good job previously of cleaning up after itself.

At this point, I think we're still paying primarily in temp storage from the intermediate reconstruction here:
https://github.com/librosa/librosa/blob/9f46e987488a98ad383aef2875e95ae970c057e5/librosa/core/spectrum.py#L2434
we could probably absorb this directly into the `angles` array with an in-place multiply and cut down on storage a bit more.

bmcfee:
I had to add a couple of additional tweaks to GL here, mainly to preserve numerical precision in angles dtype.

Previously we forced complex64 even if the input was higher precision.  When moving from a temporary multiply (`istft(S * angles, ...)`) to an in-place multiply (`angles *= S`), this was causing disagreements on double-precision inputs (single-precision retained numerical equivalence).  This PR now has GL adapting the complex dtype to the precision of the input (`S`).

I'm still not seeing a ton of reliable speed improvements here, though we should automatically inherit the previously reported speedups in stft.

bmcfee:
> I'm still not seeing a ton of reliable speed improvements here, though we should automatically inherit the previously reported speedups in stft.

Quick follow-up: I am seeing reliable memory improvements though, so I think it's an optimization worth taking.

0.9.2, 10-second noise at 22050:

```python
>>> %memit librosa.griffinlim(S, random_state=0, n_iter=64)  # 0.9.2 branch
peak memory: 267.21 MiB, increment: 14.00 MiB
>>> %memit griffinlim(S, random_state=0, n_iter=64)  # PR 1514
peak memory: 198.05 MiB, increment: 14.25 MiB
```

bmcfee:
Ok, I think this one is feature-complete.  It does have some fairly substantial changes though.  While everything here "should be" numerically equivalent to the 0.9.2 branch (notwithstanding the previous comment about numerical precision in griffin-lim phase estimates), I think this PR does warrant some careful scrutiny by someone apart from myself.

bmcfee:
Some additional benchmarking data, relevant to #599.  I've been playing around with asv to get a more long-term sense of efficiency gains and regressions.  Here are some benchmark results for time and peak memory (on my laptop) for the current main branch vs this PR when running STFT on data of different durations (in seconds), different hop lengths, and centering on/off:

main branch (before preallocation optimizations):
```
[  0.00%] · For librosa commit 25538adb <abs2>:
[  0.00%] ·· Benchmarking conda-py3.9
[ 50.00%] ··· Running (benchmarks.STFTSuite.time_stft--).
[ 75.00%] ··· benchmarks.STFTSuite.peakmem_stft                                                                                                                                             ok
[ 75.00%] ··· ============ =========== =========== ============ ============ ============ =============
              --                                        center / duration                              
              ------------ ----------------------------------------------------------------------------
               hop_length   True / 10   True / 60   True / 240   False / 10   False / 60   False / 240 
              ============ =========== =========== ============ ============ ============ =============
                  256          154M        193M        336M         153M         188M          315M    
                  512          150M        172M        251M         149M         167M          230M    
                  1024         148M        162M        209M         148M         156M          188M    
              ============ =========== =========== ============ ============ ============ =============

[100.00%] ··· benchmarks.STFTSuite.time_stft                                                                                                                                                ok
[100.00%] ··· ============ ============= ============ ============ ============= ============ =============
              --                                          center / duration                                
              ------------ --------------------------------------------------------------------------------
               hop_length    True / 10    True / 60    True / 240    False / 10   False / 60   False / 240 
              ============ ============= ============ ============ ============= ============ =============
                  256       10.6±0.07ms   71.7±20ms     491±10ms     10.8±0.3ms   67.2±20ms      453±10ms  
                  512        5.81±0.2ms   31.3±0.4ms    253±20ms     5.47±0.1ms   31.2±0.8ms     205±7ms   
                  1024       3.65±0.4ms   18.5±0.7ms    132±6ms     3.09±0.05ms   17.1±0.5ms    77.5±20ms  
              ============ ============= ============ ============ ============= ============ =============
```
PR 1514:
```
[  0.00%] · For librosa commit c95314eb <stft-preallocate>:
[  0.00%] ·· Benchmarking conda-py3.9
[ 50.00%] ··· Running (benchmarks.STFTSuite.time_stft--).
[ 75.00%] ··· benchmarks.STFTSuite.peakmem_stft                                                                                                                                             ok
[ 75.00%] ··· ============ =========== =========== ============ ============ ============ =============
              --                                        center / duration                              
              ------------ ----------------------------------------------------------------------------
               hop_length   True / 10   True / 60   True / 240   False / 10   False / 60   False / 240 
              ============ =========== =========== ============ ============ ============ =============
                  256          153M        188M        315M         153M         188M          315M    
                  512          149M        167M        230M         149M         167M          230M    
                  1024         148M        156M        188M         148M         156M          188M    
              ============ =========== =========== ============ ============ ============ =============

[100.00%] ··· benchmarks.STFTSuite.time_stft                                                                                                                                                ok
[100.00%] ··· ============ ============= ============ ============ ============= ============ =============
              --                                          center / duration                                
              ------------ --------------------------------------------------------------------------------
               hop_length    True / 10    True / 60    True / 240    False / 10   False / 60   False / 240 
              ============ ============= ============ ============ ============= ============ =============
                  256        11.6±0.6ms   74.1±0.5ms    291±1ms      11.3±0.3ms   74.0±0.3ms     292±2ms   
                  512       5.94±0.02ms   33.1±0.8ms   149±0.9ms    5.72±0.09ms   31.8±0.8ms     149±1ms   
                  1024      3.32±0.04ms   17.6±0.6ms   77.8±0.8ms    3.24±0.1ms   17.7±0.7ms    77.4±0.9ms 
              ============ ============= ============ ============ ============= ============ =============
```

At some point, I'll have this up to run continuously on a server and publish results on a website.  But for now, it at least gives a rough sense of time and memory improvements for this PR.

bmcfee:
Thinking this over, I'm wondering if we should make this "view trimming" mode the default behavior, or just always implement it.  I see very little downside to allowing over-sized inputs, and the benefits to ease-of-use seem substantial.

Issue #1526, opened 2022-07-11
Improve autocorrelation implementation


bmcfee:
**Is your feature request related to a problem? Please describe.**

Our autocorrelation implementation could be a bit more efficient.  For reference, here's how the code currently looks:

https://github.com/librosa/librosa/blob/a3095313a2ba9e37e24f6180c4decc77cd6a2b0d/librosa/core/audio.py#L846-L868

**Describe the solution you'd like**

If `y` is real-valued (as it usually is), we can optimize this a bit by using `rfft` instead of `fft`.  The only catch here is that we need to be careful with maintaining shapes, but that shouldn't be a big deal.

I propose that we rearrange this a little so that the complex check happens first (not last), and just having two branches for real or complex input handling.

Additionally, once #745 is merged, we can use `abs2` to optimize the power spectrum calculation.

**Describe alternatives you've considered**
This isn't a high-priority item, so we could leave it as is.  But it would be a nice and easy optimization to have, and autocorrelation does pop up elsewhere in the package (eg for rhythm analysis).

bmcfee:
This issue should be blocked until #745 is resolved.

bmcfee:
We had a couple of scipy sprinters looking into this one, but they didn't comment here and I don't have their contact information.  (If this is you, please chime in!)

If there's no action on this issue in the next couple of weeks, we can reassign it to someone else.

bmcfee:
Started digging into this and it's a bit subtler than I anticipated.  I'm pretty sure we can bring the memory footprint down here with a combination of clever fft usage and some shifting / phase hacking.  Will report back when I have something working.

bmcfee:
Quick followup: it turns out that we can't rely on `ifft(..., n=max_size)` to handle the array allocation for us.  The phrasing of the documentation on this point is a little ambiguous, and probably I should have known this already, but the `n=` parameter truncates the input here instead of the output.  So I think we're stuck with the over-allocation.

That said, we can still use rfft/irfft to cut this down a little.

PR #1529, opened 2022-07-14
Accelerated abs2 and phasor

### LINKED ISSUE #745: Vectorized utility functions: abs2, unit phasor, others?
#### Description
It would be useful to have an efficient function for getting magnitude-squared (power) from real or complex variables.  This came up in offline discussion of #742 , where `abs` is doing unnecessary work if we actually want power anyway.

This could be efficiently accomplished by using `numba.vectorize` to make a ufunc out of `x * x.conj()`.

bmcfee:
#### Reference Issue
Fixes #745 

This PR implements numba-accelerated utility functions for squared absolute value (abs2) and (unit) phasor construction.

#### Any other comments?

The following functions have been updated to use the new accelerated utilities:

- `core.autocorrelate`
- `core.phase_vocoder`
- `core.griffinlim`
- `core.griffinlim_cqt`
- `core.icqt`
- `feature.rms`
- `filters.wavelet`
- `filters.constantq`

Results appear to be numerically identical in all cases.

As long as the documentation and tests seem clear enough, I think this should be good to merge.  Doing so will unblock #1526.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -845,7 +845,7 @@ def autocorrelate(y, *, max_size=None, axis=-1):
     # Compute the power spectrum along the chosen axis
     # Pad out the signal to support full-length auto-correlation.
     fft = get_fftlib()
-    powspec = np.abs(fft.fft(y, n=2 * y.shape[axis] + 1, axis=axis)) ** 2
+    powspec = util.abs2(fft.fft(y, n=2 * y.shape[axis] + 1, axis=axis))
 
     # Convert back to time domain
     autocorr = fft.ifft(powspec, axis=axis)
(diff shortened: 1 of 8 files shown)


bmcfee:
The functions as implemented here are, with a few non-functional changes, equivalent to the prototypes reported in #745.  Results are numerically equivalent and much more efficiently computed on large inputs.

bmcfee:
Unintended functionality I've observed in phasor: angles and mag can be different shapes, as long as they're compatible for multiplication (eg by broadcasting). I think this is probably a good thing, could be useful, and feels very numpythonic.

On the other hand, it's not really obvious (though i did document it in an example). Since this is a util function mainly for our own use, I'm okay with leaving it as is. Just wanted to highlight it here so we're all aware of it.

bmcfee:
I think this is fine to merge, but a quick spot check on the docs would be appreciated.

PR #1536, opened 2022-07-16
adds keyword argument definitions to docstrings

### LINKED ISSUE #1525: Clarify kwargs in feature functions
Some of our feature functions (melspectrogram, mfcc, etc) use kwargs to pass through parameters to core functions (stft, filter construction etc).  This isn't terribly well documented, and can be a little difficult for new users to navigate and understand what the options are.

At this point, it's probably worth replacing some of these kwargs, eg
https://github.com/librosa/librosa/blob/25538adb3aed3485a06e60b6dad88be3d540f0c2/librosa/feature/spectral.py#L1767-L1769
by
```python 
def mfcc(
    *, y=None, sr=22050, S=None, n_mfcc=20, dct_type=2, norm="ortho", lifter=0,
    n_mels=128, n_fft=2048, ...
):
```

Alternatively, if we don't want to make the API totally redundant, we could at least provide more documentation in the kwarg'ed functions' docstrings to better explain what options are available.

Embot2000:
#### Reference Issue

Fixes #1525

This adds clarity to the docstrings for feature functions using kwargs 

#### Any other comments?
Changes only pertain to docstrings (also one small add to gitignore to ignore vscode settings)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ coverage.xml
 
 #IDE
 .idea/
+.vscode
--- a/librosa/feature/inverse.py
+++ b/librosa/feature/inverse.py
@@ -30,7 +30,6 @@ def mel_to_stft(M, *, sr=22050, n_fft=2048, power=2.0, **kwargs):
         number of FFT components in the resulting STFT
     power : float > 0 [scalar]
         Exponent for the magnitude melspectrogram
-    
     **kwargs : additional keyword arguments for Mel filter bank parameters
     n_mels : int > 0 [scalar]
         number of Mel bands to generate
@@ -154,7 +153,6 @@ def mel_to_audio(
         samples.
     dtype : np.dtype
         Real numeric type for the time-domain signal.  Default is 32-bit float.
-
     **kwargs : additional keyword arguments for Mel filter bank parameters
     n_mels : int > 0 [scalar]
         number of Mel bands to generate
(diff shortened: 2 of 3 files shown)


bmcfee (commented):
This all looks great, content-wise, but we're failing the numpydoc formatting check in the linter (velin).  I think this is because the parameters are nested underneath the kwargs and numpydoc doesn't like it.  I'm not sure what the best solution is here, but I'll try to find someone who knows.

Embot2000:
I did look around for "best practices" for this and I didn't find much.. except a stack overflow thread (https://stackoverflow.com/questions/1137161/what-is-the-correct-way-to-document-a-kwargs-parameter).

I'm open to formatting this however you think is best!

bmcfee:
Ok, I just chatted with some numpydoc folks, and the recommended answer right now is just to tab the parameters over inline with the regular parameter list.  It's a little clumsy, but it should work.  Maybe something like:

```
**kwargs : additional parameters listed below

n_fft : ...
```

Embot2000:
OK sorry it took me a while, I had to run some errands. I removed the indents as suggested. Let me know if you need me to change anything else!

bmcfee:
Thanks!  It looks like velin is failing now from missing newlines between parameter blocks:

```
    **kwargs : additional keyword arguments for Mel filter bank parameters
        See `librosa.filters.mel` for details
    n_mels : int > 0 [scalar]
        number of Mel bands to generate
    fmin : float >= 0 [scalar]
        lowest frequency (in Hz)
```
should be
```
    **kwargs : additional keyword arguments for Mel filter bank parameters
        See `librosa.filters.mel` for details
    
    n_mels : int > 0 [scalar]
        number of Mel bands to generate
    
    fmin : float >= 0 [scalar]
        lowest frequency (in Hz)
```

Otherwise everything looks great.  If you want to test-drive the linter locally, you can `pip install velin` and run it like we do in the lint CI:

https://github.com/librosa/librosa/blob/3a5d71ef55c01fcfb8626d3b624cde898bad16d0/.github/workflows/lint_python.yml#L13

Embot2000:
Ha ok thanks! I use flake8 normally but obviously it didn't catch these so I'll give velin a shot. I will get these fixed and submitted tomorrow!

bmcfee:
Cool, thanks! We've adjourned for the night after a Eurovision rerun binge here.

Flake8 checks code, velin checks docstring formatting, so it makes sense that this slipped through.

Embot2000:
It looks like velin is looking for consistency within the docstring for each individual function because some of these (ie `mel_to_audio`) do not have newlines between each parameter but others (ie `mfcc_to_audio`) do. 

Would you like me to standardize them while I'm here and make them all the same? If so, do you have a preference?

bmcfee:
> Would you like me to standardize them while I'm here and make them all the same? If so, do you have a preference?

Ah, strange!  I didn't realize it did that.  I don't think we need to audit everything, but for  at lest the ones being changed in this PR, my (slight) preference is to have a newline between each parameter.

bmcfee (approved):
LGTM!  Happy to merge once CI finishes.  Thanks @Embot2000 !

PR #1539, opened 2022-07-17
Refactor AdaptiveWaveplot, disconnection support

### LINKED ISSUE #1532: librosa.display.waveshow memory leak?
**Describe the bug**
When I try to run `librosa.display.waveshow` on a long wav file (hour long), 44.1kHz, and plot the waveform, it works in Colab. However, try running the same cell a few times and the memory just keeps growing until the session crashes. This is problematic because I'd like to be able to run this waveform generating function in a loop over many hour-long files. 

**To Reproduce**

Example:
Here's a colab link (assuming you have access to 1 hour-long wav files: https://colab.research.google.com/drive/1HIXqFM4NIw6qflcS4Ss5yAMvJD_XNmta )
```
import librosa.display
import librosa

file_path = '/content/drive/MyDrive/wav_files/denoised_hour_long_wav_file.wav'
y, sr = librosa.load(file_path, mono=False, sr=None)

orig_path ='/content/drive/MyDrive/wav_files/hour_long_wav_file.wav'
y_2, sr_2 = librosa.load(orig_path, mono=False, sr=None)


fig = plt.figure(figsize=(20,4), dpi=100)
for i in range(10):
  #Running this in a loop to show how it will crash after roughly 3-5 times of doing this.
  librosa.display.waveshow(y_2, sr=sr_2, alpha=0.4)
  librosa.display.waveshow(y, sr=sr, color='r', alpha=0.5)
fig.tight_layout()
fig.show()
```
Try running this code in a cell multiple times. If you isolate just these 2 lines:

```
librosa.display.waveshow(y_2, sr=sr_2, alpha=0.4)
librosa.display.waveshow(y, sr=sr, color='r', alpha=0.5)
```

And run those in their own cell multiple times (or in a while loop, or for loop if you have an hour long file), you'll notice the memory growing until it crashes and restarts the session.

**Expected behavior**
I generate waveforms without the memory filling up and crashing Colab.

**Software versions***

Please run the following Python code snippet and paste the output below.
```
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import librosa; print("librosa", librosa.__version__)

librosa.show_versions()

Linux-5.4.188+-x86_64-with-Ubuntu-18.04-bionic
Python 3.7.13 (default, Apr 24 2022, 01:04:09) 
[GCC 7.5.0]
NumPy 1.21.6
SciPy 1.7.3
librosa 0.9.2
INSTALLED VERSIONS
------------------
python: 3.7.13 (default, Apr 24 2022, 01:04:09) 
[GCC 7.5.0]

librosa: 0.9.2

audioread: 2.1.9
numpy: 1.21.6
scipy: 1.7.3
sklearn: 1.0.2
joblib: 1.1.0
decorator: 4.4.2
soundfile: 0.10.3
resampy: 0.3.1
numba: 0.51.2
pooch: v1.6.0
packaging: 21.3

numpydoc: None
sphinx: 1.8.6
sphinx_rtd_theme: None
sphinx_multiversion: None
sphinx_gallery: None
mir_eval: None
ipython: None
sphinxcontrib-svg2pdfconverter: None
pytest: 3.6.4
pytest-mpl: None
pytest-cov: None
matplotlib: 3.2.2
samplerate: None
soxr: None
contextlib2: installed, no version number available
presets: None

```

bmcfee:
#### Reference Issue
Fixes #1532 

This PR refactors the `AdaptiveWaveplot` object a bit to make connection and disconnection from callbacks easier.  This doesn't "automatically" resolve #1532, but it does make things a bit cleaner.

You can now do the following to disable the adaptivity:
```python
adaptor = librosa.display.waveshow(...)

adaptor.disconnect()
```

And if you want to release memory and references, you can do:
```python
adaptor = librosa.display.waveshow(...)
del adaptor
```

This also allows adaptor objects to be reconnected to different axes objects and signals.

A deconstructor method is added to facilitate clean disconnection.

#### Any other comments?

The functionality is all here now, though I'm not sure what kind of documentation would be appropriate to explain things in a way that would avoid #1532.  Any feedback from @youssefabdelm would be appreciated.

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -531,7 +531,6 @@ class AdaptiveWaveplot:
         See Also
         --------
         disconnect
-        matplotlib.axes.callbacks.connect
         """
         # Disconnect any existing callback first
         self.disconnect()
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -162,11 +162,11 @@ class TimeFormatter(Formatter):
             s = "{:d}:{:02d}".format(int(value / 60.0), int(np.mod(value, 60)))
         elif self.unit == "s":
             s = "{:.3g}".format(value)
-        elif self.unit==None and (vmax - vmin >= 1):
+        elif self.unit == None and (vmax - vmin >= 1):
             s = "{:.2g}".format(value)
         elif self.unit == "ms":
             s = "{:.3g}".format(value * 1000)
-        elif self.unit==None and (vmax - vmin < 1):
+        elif self.unit == None and (vmax - vmin < 1):
             s = "{:.3f}".format(value)
 
         return "{:s}{:s}".format(sign, s)
@@ -509,6 +509,44 @@ class AdaptiveWaveplot:
         self.envelope = envelope
         self.sr = sr
         self.max_samples = max_samples
+        self.cid = None
+        self.ax = None
+
+    def connect(self, ax, signal="xlim_changed"):
+        """Connect the adaptor to a signal on an axes object.
+
+        Note that if the adaptor has already been connected to an axes object,
+        that connect is first broken and then replaced by a new callback.
+
+        Parameters
+        ----------
+        ax : matplotlib.axes.Axes
+            The axes to connect this adaptor to
+
+        signal : string, {'xlim_changed', 'ylim_changed'}
+            The signal to connect the update to
+
+        See Also
+        --------
+        disconnect
+        matplotlib.axes.callbacks.connect
+        """
+        # Disconnect any existing callback first
+        if self.ax:
+            self.ax.callbacks.disconnect(self.cid)
+
+        self.ax = ax
+        self.cid = ax.callbacks.connect(signal, self.update)
+
+    def disconnect(self):
+        """Disconnect the adaptor's update callback.
+
+        See Also
+        --------
+        connect
+        """
+        if self.ax:
+            self.ax.callbacks.disconnect(self.cid)
 
     def update(self, ax):
         """Update the matplotlib display according to the current viewport limits.
@@ -517,7 +555,7 @@ class AdaptiveWaveplot:
 
         Parameters
         ----------
-        ax : matplotlib axes object
+        ax : matplotlib.axes.Axes
             The axes object to update
         """
         lims = ax.viewLim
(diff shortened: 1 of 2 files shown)


librosa/display.py
```diff
@@ -509,6 +509,56 @@ def __init__(self, times, y, steps, envelope, sr=22050, max_samples=11025):
...
+        ----------
+        ax : matplotlib.axes.Axes
+            The axes to connect with this adaptor's `update`
+        signal : string, {"xlim_changed", "ylim_changed"}
+            The signal to connect
+
+        See Also
+        --------
+        disconnect
+        matplotlib.axes.callbacks.connect
```
lostanlen, line 534: i am not finding the documentation for this

closest i found is
https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.add_callback.html

librosa/display.py
```diff
@@ -509,6 +509,56 @@ def __init__(self, times, y, steps, envelope, sr=22050, max_samples=11025):
...
+        self.ax = ax
+        self.cid = ax.callbacks.connect(signal, self.update)
+
+    def disconnect(self, *, strict=False):
+        """Disconnect the adaptor's update callback.
+
+        Parameters
+        ----------
+        strict : bool
+            If `True`, remove references to the connected axes.
```
lostanlen, line 549: maybe explain the consequences of removing these references?

librosa/display.py
bmcfee: Ah I see; it should be the connect method of `CallbackRegistry`.

librosa/display.py
bmcfee: There shouldn't really be any user-facing consequences.  This is just a safety measure to eliminate circular references when you call `__del__`.

librosa/display.py
bmcfee: I'll actually just remove this see-also, since it's not necessary or helpful in the abstract.

librosa/display.py
lostanlen: cool. i see it now. maybe say that?

lostanlen (approved):

PR #1543, opened 2022-07-23
Fixes the PR reference number in the Changelog

cr2007:
The PR number in the change log for my PR was the wrong one.
The earlier one was 1487 (by @LorenzNickel )

This PR fixes the PR number mentioned in the changelog.
As I was reading the changelog for the [latest release](https://librosa.org/doc/0.9.2/changelog.html), I noticed that the PR reference number for my changes was the wrong one, and directed to a different PR.

This PR fixes that minor error, although it could be present in other releases too.

#### Any other comments?
This issue could be present in other releases/PR's listed too


bmcfee:
Thanks for catching that - sorry for the typo!

The fix will appear in the dev doc build when this merges, but the changelogs for releases will not change since they're tied to git tags.

cr2007:
Alright, as long as it's updated on the main website changelogs
image
Like the one showed above

bmcfee:
Yes - once the doc build completes (it usually takes about 40 minutes, give or take) it will appear updated at https://librosa.org/doc/main/changelog.html

Issue #1419, opened 2021-12-14
Drop official python 3.6 support in 0.9?


bmcfee:
Python 3.6 is nearing [end-of-life](https://endoflife.date/python), and we're almost certainly not going to get 0.9 out before then.

This raises the question of whether we should bump the minimum supported python to 3.7.  This isn't so much an issue for librosa itself - we generally don't use [new-fangled features](https://docs.python.org/3/whatsnew/3.7.html) - but it is an issue for our dependency stack and testing environments.

The issue arose recently due to the release of tomli 2.0, which is an upstream dependency for our testing environment (via test coverage), and led to breaking builds.  For now I've pinned the maximum tomli version (`<2.0`) in the in-progress PR #1418.  However, I'd prefer to be more explicit about minimum supported versions.

This is an easy enough change to make, but it might cause some problems.  According to our current [package statistics](https://pypistats.org/packages/librosa), somewhere between 1/3 and 1/2 of download traffic comes from 3.6 environments.  That of course will decrease going forward, but it's still a large portion of our user base that might be inconvenienced.

--- 

As an alternative, we could potentially keep 3.6 support for the 0.9 release (to ease transition), but remove it from the general build matrix.  We'd still keep the "minimal" environment with the pinned tomli -- up until the next major release after 0.9, anyway.

What do people think?  

- Option 1: remove 3.6 support entirely (even though we don't really *need* to)
- Option 2: reduce 3.6 testing, but keep support

bmcfee:
For a bit of context about how these decisions happen elsewhere, https://numpy.org/neps/nep-0029-deprecation_policy.html

PR #1551, opened 2022-08-11
Drop use of contextlib2

s-t-e-v-e-n-k:
As of Python 3.7, contextlib in the standard library now supports all
functionality that this library requires. As a consequence, remove the
external requirement, and switch to it. Drive by removing some now
unneeded __future__ imports.

Please note: This may fail CI for Python 3.6, but 3.6 has been EOL now for eight months.

### DIFF
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -22,7 +22,6 @@ dependencies:
   - pytest-mpl
   - pytest-cov
   - pytest
-  - contextlib2
   - coverage
   - ffmpeg
   - pip:
--- a/.github/environment-minimal.yml
+++ b/.github/environment-minimal.yml
@@ -22,7 +22,6 @@ dependencies:
   - pytest-mpl
   - pytest-cov
   - pytest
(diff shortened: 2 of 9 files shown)


bmcfee:
Thanks for this.  It's on my radar for when we drop 3.6.

> Please note: This may fail CI for Python 3.6, but 3.6 has been EOL now for eight months.

We decided in #1419 to retain 3.6 support for at least one more release cycle.  No decisions have been made yet for the 0.10 release, but we do still see a lot of 3.6 traffic on pypistats.

bmcfee:
It might also be worth mentioning that our use of contextlib(2) is extremely minimal, and is entirely confined to the unit tests.  Specifically, it's used to provide a dummy context manager for tests that conditionally pass or fail depending on certain configurations of input parameters.  We could probably just rewrite those tests and drop this requirement entirely.

bmcfee:
:wave: I'm retargeting this PR against a new branch to handle environment updates.

PR #1560, opened 2022-08-23
Correct window bandwidth calculation and documentation

### LINKED ISSUE #1527: Where does the formula in `window_bandwidth` come from?
Here's our formula for `window_bandwidth`

```python
        win = get_window(window, n)
        WINDOW_BANDWIDTHS[key] = n * np.sum(win ** 2) / np.sum(np.abs(win)) ** 2
```

This formula is (proportional to) the ratio between the square of the L2 norm and the square of the L1 norm.
I see how this relates to the (inverse) "peakiness" of the window vector, but i'd like some mathematical context around it.

Some explanation in the docstring as to why we use this formula + where it comes from. Maybe a bibliographical /textbook reference.

* Knowledge of the First Kind
* Knowledge of the Third Kind

https://www.gutenberg.org/files/3800/3800-h/3800-h.htm

**Additional context**
Found while discussing #745

bmcfee:
#### Reference Issue
Fixes #1527 

This PR fixes a very small error in `filters.window_bandwidth`, and expands the documentation.

Only window functions with negative values are affected.

#### Any other comments?

This should be good to merge after CI.

### DIFF
--- a/librosa/filters.py
+++ b/librosa/filters.py
@@ -1080,7 +1080,16 @@ def cq_to_chroma(
 
 @cache(level=10)
 def window_bandwidth(window, n=1000):
-    """Get the equivalent noise bandwidth of a window function.
+    """Get the equivalent noise bandwidth (ENBW) of a window function.
+
+    The ENBW of a window is defined by [#]_ (equation 11) as the normalized
+    ratio of the sum of squares to the square of sums::
+
+        enbw = n * sum(window**2) / sum(window)**2
+
+    .. [#] Harris, F. J.
+        "On the use of windows for harmonic analysis with the discrete Fourier transform."
+        Proceedings of the IEEE, 66(1), 51-83.  1978.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


librosa/filters.py
```diff
@@ -1115,7 +1124,7 @@ def window_bandwidth(window, n=1000):
 
     if key not in WINDOW_BANDWIDTHS:
         win = get_window(window, n)
-        WINDOW_BANDWIDTHS[key] = n * np.sum(win ** 2) / np.sum(np.abs(win)) ** 2
+        WINDOW_BANDWIDTHS[key] = n * np.sum(win ** 2) / (np.sum(win) ** 2 + util.tiny(win))
```
bmcfee, line 1127: Just a note that I'm adding a tiny here to stabilize the division.  This shouldn't affect anything substantively —I'd be worried if a window sums to an underflow value!—but it makes me feel better.

lostanlen:
thanks. does the Harris paper have a proof that this formula indeed yields "the width of an equivalent ideal rectangular spectral response that will pass the same noise power as the window (filter) under test" ?

bmcfee:
> does the Harris paper have a proof that this formula indeed yields

It's not quite as streamlined as the other link that I provided in #1527 discussion, but yes:
image

lostanlen (commented):
Thank you

PR #1561, opened 2022-08-23
High-dimensional resampling, axis targeting

### LINKED ISSUE #1559: Support ndim>2 in all resamplers
Noted in #1556 , our support for resampling of high-dimensional (ndim>2) signals is not exactly complete.  Some converters support this out of the box (resampy, scipy) and others do not (samplerate, soxr).

There are two options here:

1. Send PRs to the upstream packages to extend their support for high-dimensional signals.
2. Implement vectorization shims on our side so that the upstream packages don't have to change.  This would use `np.vectorize` to automatically convert high-D resamples into a collection of low-D resamples.

I'm not sure that option (1) is preferable, as it could complicate our dependency version requirements, and it may not even be in scope for those packages in the first place.

Option 2 could have some complications due to memory layout, dtypes, etc; we should prototype this out to see if it's feasible before committing

**Additional context**

Related issues: #1556 #1558

bmcfee:
#### Reference Issue
Fixes #1559 

This PR extends multichannel support to all resampling backends.

It also extends the resample API to support a target axis.

#### Any other comments?

This is ready for CR.

### DIFF
--- a/tests/test_multichannel.py
+++ b/tests/test_multichannel.py
@@ -1033,7 +1033,7 @@ def test_resample_highdim(x, res_type):
 
 @pytest.mark.parametrize('res_type', ('scipy', 'polyphase', 'sinc_fastest', 'kaiser_fast', 'soxr_qq'))
 @pytest.mark.parametrize('x, axis', [(np.zeros((2, 2, 2, 22050)), -1), (np.zeros((22050, 2, 3)), 0)])
-def test_resample_highdim(x, axis, res_type):
+def test_resample_highdim_axis(x, axis, res_type):
     # Resample along the target axis
     y = librosa.resample(x, orig_sr=22050, target_sr=11025, axis=axis, res_type=res_type)
 
--- a/tests/test_multichannel.py
+++ b/tests/test_multichannel.py
@@ -1008,3 +1008,35 @@ def test_trim_multichannel(y_multi):
     # intersection of the individual intervals
     assert ival[0] == max(ival0[0], ival1[0])
     assert ival[1] == min(ival0[1], ival1[1])
+
+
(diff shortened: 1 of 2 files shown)


lostanlen (commented):
might be worth revisiting the definition of y


```python
y : np.ndarray [shape=(..., n)]
audio time series. Multi-channel is supported.
```

bmcfee:
> might be worth revisiting the definition of y

good catch, done.

tests/test_multichannel.py
```diff
@@ -1008,3 +1008,35 @@ def test_trim_multichannel(y_multi):
...
+@pytest.mark.parametrize('res_type', ('scipy', 'polyphase', 'sinc_fastest', 'kaiser_fast', 'soxr_qq'))
+@pytest.mark.parametrize('x', [np.zeros((2, 2, 2, 22050))])
+def test_resample_highdim(x, res_type):
+    # Just run these to verify that it doesn't crash with ndim>2
+    y = librosa.resample(x, orig_sr=22050, target_sr=11025, res_type=res_type)
+
+
+@pytest.mark.parametrize('res_type', ('scipy', 'polyphase', 'sinc_fastest', 'kaiser_fast', 'soxr_qq'))
+@pytest.mark.parametrize('x, axis', [(np.zeros((2, 2, 2, 22050)), -1), (np.zeros((22050, 2, 3)), 0)])
+def test_resample_highdim(x, axis, res_type):
```
lostanlen, line 1036: consider renaming this `def` to avoid collision with the other `test_resample_highdim`

bmcfee: lol this is what happens when i jam out tests 5 minutes before leaving the office...

lostanlen (approved):
thanks

PR #1563, opened 2022-08-30
Explain `cmap` in `specshow` docs

### LINKED ISSUE #1524: Bypass automatic colormap selection

`librosa.display.specshow` allows the user to set the colormap for the rendered image in the usual way, eg `cmap='magma'`.  If no colormap is supplied, the helper function `librosa.display.cmap` tries to guess an appropriate colormap based on the contents of the data to be drawn (eg, sequential, binary, or diverging).

As noted in #738 , it would sometimes be helpful to be able to disable this behavior and fall back on whatever the default colormap of the matplotlib environment is.  This is especially handy when using a style / context manager to handle these decisions, where it may not be explicit or obvious to the user what colormap is being chosen.

At present, the best workaround here is to explicitly set the cmap, but that leads to inflexible code that does not adapt to the matplotlib environment.

The current logic for cmap inference is here:
https://github.com/librosa/librosa/blob/9f46e987488a98ad383aef2875e95ae970c057e5/librosa/display.py#L868

I propose to replace this by the following:

```python
if 'cmap' in kwargs and kwargs['cmap'] is None:
    # Remove cmap from the pcolormesh kwargs
    kwargs.pop('cmap')
else:
    # No cmap is specified, so guess one from the data properties
    kwargs.setdefault("cmap", cmap(data))
```

This would allow for `librosa.display.specshow(..., cmap=None)` to always use the "default" colormap.  If `cmap` is not provided as a kwarg, it is inferred from `data` just as before.  If `cmap` is otherwise not None (e.g., a string or color map object), it is passed through as is to `pcolormesh`.

@lostanlen and I discussed various options offline, including adding a new parameter to disable cmap selection, or using a token value for `cmap` to indicate the intended behavior.  At the end of the discussion, we agreed that `None` does the job well here with minimal added complexity.

kianelbo:
#### Reference Issue
Fixes #1524 

Document `cmap` kwarg of `specshow`

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -909,6 +909,9 @@ def specshow(
             - ``shading='auto'``
             - ``edgecolors='None'``
 
+        The ``cmap`` option if not provided, is inferred from data automatically.
+        Set ``cmap=None`` to use matplotlib's default colormap.
+
     Returns
     -------
     colormesh : `matplotlib.collections.QuadMesh`


bmcfee (approved):
LGTM, thanks for this!

PR #1565, opened 2022-09-06
use rfft when appropriate in autocorrelate, fixes #1526

### LINKED ISSUE #1526: Improve autocorrelation implementation

bmcfee:
#### Reference Issue
Fixes #1526 

This PR rewrites autocorrelate to use (i)rfft when provided with real-valued inputs.  This should cut the memory usage approximately in half.

#### Any other comments?

This PR is definitely an improvement, but I'm not 100% satisfied yet as noted in the discussion thread of #1526.  We still are computing a full inverse DFT and then cropping the result, rather than only computing what is actually needed.  Perhaps this is the best we can do, given how the fft API is defined.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -847,24 +847,30 @@ def autocorrelate(y, *, max_size=None, axis=-1):
 
     max_size = int(min(max_size, y.shape[axis]))
 
-    # Compute the power spectrum along the chosen axis
-    # Pad out the signal to support full-length auto-correlation.
     fft = get_fftlib()
-    powspec = util.abs2(fft.fft(y, n=2 * y.shape[axis] + 1, axis=axis))
 
-    # Convert back to time domain
-    autocorr = fft.ifft(powspec, axis=axis)
+    # Pad out the signal to support full-length auto-correlation.
+    n_pad = 2 * y.shape[axis] - 1
+
+    if np.iscomplexobj(y):
+        # Compute the power spectrum along the chosen axis
+        powspec = util.abs2(fft.fft(y, n=n_pad, axis=axis))
+
+        # Convert back to time domain
+        autocorr = fft.ifft(powspec, n=n_pad, axis=axis)
+    else:
+        # Compute the power spectrum along the chosen axis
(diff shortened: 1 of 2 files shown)


bmcfee:
Quick memray benchmark results with the following snippet:
```python
import numpy as np
import librosa


x = np.random.randn(60 * 22050)
librosa.autocorrelate(x, max_size=22050)
```

Current main:

```
Start time: 2022-09-06 15:09:05.431000
End time: 2022-09-06 15:09:13.142000
Total number of allocations: 552301
Total number of frames seen: 12694
Peak memory usage: 547.2 MB
Python allocator: pymalloc
```

This PR:
```
Start time: 2022-09-06 15:07:34.459000
End time: 2022-09-06 15:07:41.300000
Total number of allocations: 551834
Total number of frames seen: 12703
Peak memory usage: 261.4 MB
Python allocator: pymalloc
```

bmcfee:
After a bit of homework and reading https://www.fftw.org/pruned.html - I'm content to leave this one as it currently is.  Hacking the inverse to be more memory-efficient seems like more hassle than it's worth at this point.

bmcfee:
Unless there are objections, I'll merge this today.

lostanlen (approved):
Very nice. thanks for this

PR #1566, opened 2022-09-06
Changing default resampler

### LINKED ISSUE #1556: Sample rate conversion going forward

We at this point support a pretty wide variety of sample rate conversion utilities: scipy, libsamplerate, resampy, and soxr.  For historical reasons relating to ease of cross-platform packaging, resampy has been the default for quite a while now.  However, resampy is also not the best choice for most uses: it doesn't support stream processing, and it's not the most efficient implementation for high-quality conversion.  (And, as we saw with the recent 0.3 series shenanigans, using parallel compilation for faster resampling can have bizarre interactions with BLAS libraries and cause serious performance degradations.)

I think it's worth at this point discussing whether it's worth changing our default resampling parameters and dependency structure.  If we do this, I'd like to have it done for the 0.10 series so that it can become a permanent change in the 1.0 release.

I'm open to arguments here, but I think the best option is probably soxr.  It's both the most efficient and most accurate in my benchmarks, it has a stream processing interface (for when that becomes important #1518 ), and it has a pretty robust build distribution at this point on both pypi (static-linked wheels) and conda-forge (dynamic link with libsoxr) (thanks @dofuuz!).

We would of course continue to support alternative resamplers as we do now, all that would change is the default parameters (`res_type='soxr_hq'` for example) and the core requirements would swap out resampy for soxr(-python).

We could also use libsamplerate by default.  It's been around longer than soxr-python and might be more stable (from a development perspective), but otherwise I don't see an argument for it in terms of either efficiency or accuracy.

I see no arguments for using the scipy.signal implementation, other than it would reduce our core dependency stack by one.

**Additional context**

Some other things to consider in this decision:

1. It will have implications for any function that depends on resampling, eg cqt/vqt, iirt, etc.
2. It will also have implications for any wasm-based distributions in the future.  That is, if we want to run librosa in pyodide, we will need to make sure that soxr can also run in that environment.  Sticking with resampy by default would sidestep this issue, as resampy depends only on numba, and librosa will need numba anyway.

### LINKED ISSUE #1558: Simplify CQT/VQT and remove special cases

Our CQT/VQT implementation includes an efficiency hack that processes the first (highest) octave separately from the others (in some situations).  The idea was to use fast downsampling at low octaves, and high-quality downsampling at high octaves:

https://github.com/librosa/librosa/blob/5f58277c919c6526dc45df5cf56c9af8907aa550/librosa/core/constantq.py#L960-L1000

This hack dates from a time when sample rate conversion was a significant computational bottleneck.

Note that forcing a specific resampling algorithm (eg `res_type='polyphase'`) will bypass this logic entirely, and use the same resampler at every octave.

I think at this point, and especially in the context of #1556, it could make sense to streamline this process.  If we move to a faster (default) resampling library, it is probably not necessary anymore, and we can simplify the implementation a bit. 

:shrug: 

**Additional context**

The main downside I see to this is that it will become difficult to exactly (numerically) reproduce old computations performed with the adaptive downsampler selection.

bmcfee:
#### Reference Issue
Fixes #1556
Fixes #1558 

This PR changes the default resampling engine throughout the package to use `soxr_hq` rather than resampy.

Accordingly, resampy has been relegated to an optional dependency, and soxr has been promoted to a core dependency.  Package requirements and testing environments have been updated accordingly.

#### Any other comments?

The adaptive resampling behavior of VQT has also been "simplified", though I'm not yet sure how we should proceed here.

The code has been simplified by removing the special case code block for the top octave filter.  We now only do a single early downsampling operation before processing, and handle all octaves with the same code.  This simplifies the implementation a bit.  

For reference, "early downsampling" here is the idea of an up-front downsampling operation so that Nyquist is within an octave of the highest frequency.  This is essentially an efficiency hack to avoid wasting computation.

The tricky part here is that early downsampling was only ever active when the resampling mode was `kaiser_fast`.  In principle, this logic could be applied to any resampler that supports fractional rates (ie, anything but polyphase).

Alternatively, we could simply remove the early downsampling logic entirely.  There are cases where it's essentially a no-op (eg when hop-length is odd), and eliminating it entirely would get us out of having a magic number for the roll-off of the resampling filter.

---

One last comment: this PR will deprecate the option of supplying `res_type=None` to select adaptive resampling mode.  We'll remove this shim in the 1.0 series.

### DIFF
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -16,6 +16,7 @@ dependencies:
   - pysoundfile>=0.10.2
   - pooch>=1.0
   - packaging>=20.0
+  - soxr-python>=0.3.0
 
   # optional, but required for testing
   - matplotlib>=3.3.0
@@ -26,5 +27,4 @@ dependencies:
   - coverage
   - ffmpeg
   - pip:
-    - soxr
     - samplerate
--- a/.github/environment-docs.yml
+++ b/.github/environment-docs.yml
@@ -12,6 +12,7 @@ dependencies:
   - joblib>=0.14.0
   - decorator>=4.0.10
   - resampy>=0.2.2
+  - soxr-python>=0.3.0
   - numba<0.50
   - numpydoc>=1.1.0
   - ipython>=7.0
--- a/.github/environment-minimal.yml
+++ b/.github/environment-minimal.yml
@@ -11,11 +11,11 @@ dependencies:
   - scikit-learn==0.19.1
   - joblib==0.14.0
   - decorator==4.0.10
-  - resampy==0.2.2
   - numba==0.45.1
   - pysoundfile==0.10.2
   - pooch==1.0.0
   - packaging==20.0
+  - soxr-python==0.3.2b1
 
   # optional, but required for testing
   - matplotlib==3.3.0
@@ -26,9 +26,9 @@ dependencies:
   - coverage
   - ffmpeg
   - tomli<2.0
+  - resampy==0.2.2
 
   # For python 3.6 we limit this dependency
   - typing-extensions<4.2
   - pip:
-    - soxr
     - samplerate
--- a/.github/environment-minimal.yml
+++ b/.github/environment-minimal.yml
@@ -15,7 +15,6 @@ dependencies:
   - pysoundfile==0.10.2
   - pooch==1.0.0
   - packaging==20.0
-  - soxr-python==0.3.2b1
 
   # optional, but required for testing
   - matplotlib==3.3.0
@@ -31,4 +30,5 @@ dependencies:
   # For python 3.6 we limit this dependency
   - typing-extensions<4.2
   - pip:
+    - soxr>=0.3.0
     - samplerate
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -9,7 +9,7 @@ import soundfile as sf
 import audioread
 import numpy as np
 import scipy.signal
-import resampy
+import soxr
 
 from numba import jit
 from .fft import get_fftlib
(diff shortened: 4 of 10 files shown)


bmcfee:
Just as a quick point of reference, here's how CQT looks with `soxr_hq` vs `kaiser_best` in the new, simplified implementation (`amplitude_to_db(np.abs(C_soxr - C_kaiser))`):

image

Discrepancies are south of -85dB, so I expect very little (if any) practical regressions going forward with default parameterizations.

bmcfee:
> Alternatively, we could simply remove the early downsampling logic entirely. There are cases where it's essentially a no-op (eg when hop-length is odd), and eliminating it entirely would get us out of having a magic number for the roll-off of the resampling filter.

I gave this a spin on a 1-minute noise clip with sr=44100 and default CQT parameters.  Early downsampling is saving us about 10% peak memory usage and about a half second of run time.  This seems worth it to me, though I think it should be possible to streamline things a bit and make it work for all resamplers.

bmcfee:
I cleaned up the early downsampling code a bit so that it now works with any resampling mode (including polyphase).  I've also removed the magic number to account for filter roll-off in the sample rate converter.  This means that there could be some fiddly cases where the results are not numerically identical between 0.9.2 and dev, and we could conceivably get a bit of aliasing at the top octave.  I'm not too worried about this though, as soxr has a tighter roll-off and better attenuation anyway.

As early downsampling was only possible when `res_type=None / kaiser_fast` in 0.9.2, my prototyping regression tests are only with that mode.  We get numerical equivalence, eg with this impulse test:
image

The more interesting cases are early downsampling with better resamplers.  For example, here's soxr_hq with and without early downsampling:
image
this example uses a low set of frequencies to highlight the early downsampling, and we can see a much better noise floor when it's enabled.  Part of this is due to the basis sparsification, so if we set sparsity=0, we get:
image

This is all to say - I'm tentatively in favor of keeping early downsampling, and keeping this change to allow it to work for any resampler.  What do y'all think?

lostanlen (approved):

PR #1570, opened 2022-09-12
updated requirements and tests to py3.7 minimum

### LINKED ISSUE #1569: CI failure on windows with resampy 0.4.1
**Describe the bug**

This popped up yesterday after releasing resampy 0.4.1.  Our test suite generally works fine, except for one test on windows that's failing.

The relevant part of the error log is

```python
        shape[axis] = int(shape[axis] * sr_new / sr_orig)
    
        if shape[axis] < 1:
>           raise ValueError(
                "Input signal length={} is too small to "
                "resample from {}->{}".format(x.shape[axis], sr_orig, sr_new)
            )
E           ValueError: Input signal length=720200 is too small to resample from 22050->4000
```

Actual build log is here: https://github.com/librosa/librosa/runs/8276453091?check_suite_focus=true

Note that the same test works fine on all other platforms.  Scrolling up the error log, what I'm seeing is:

```python
tests\test_core.py:1880: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
librosa\core\spectrum.py:1591: in iirt
    y_resampled.append(resample(y, orig_sr=sr, target_sr=cur_sr, res_type=res_type))
librosa\core\audio.py:648: in resample
    y_hat = resampy.resample(y, orig_sr, target_sr, filter=res_type, axis=axis)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = array([0., 0., 0., ..., 0., 0., 0.]), sr_orig = 22050, sr_new = 4000
axis = -1, filter = 'kaiser_best', parallel = False, kwargs = {}
sample_ratio = 0.18140589569160998, shape = [-64134]
```

So it looks like somehow the shape calculation is overflowing and becoming a negative int (?!).  I don't see any what that this makes sense though, as the resulting shape (-64134) is not what you would get be a 16-bit signed integer overflow of the true shape `int(720200 * 4000 / 22050) = 130648`.  At any rate, all numbers in the shape calculation of resampy are Python ints and not numpy ints (up until the division anyway), and this would not overflow in the first place.  So I'm pretty puzzled.

My first instinct was to push this back as a resampy issue, but I don't think there's actually a problem there.  Rather, I think there's something subtly broken in the windows CI environment to trigger this.

As a side note, #1566 would obviate this issue in the tests, but I'd still like to figure out what's going on here.

bmcfee:
#### Reference Issue

This is (hopefully) to fix #1569 

Includes changes from #1551 

This PR updates our requirements to python 3.7 minimum, and adds tests on 3.10.

### DIFF
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -6,7 +6,7 @@ dependencies:
   # required
   - pip
   - audioread>=2.1.9
-  - numpy>=1.17.0
+  - numpy>=1.23.0
   - scipy>=1.2.0
   - scikit-learn>=0.19.1
   - joblib>=0.14.0
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -22,7 +22,6 @@ dependencies:
   - pytest-mpl
   - pytest-cov
   - pytest
-  - contextlib2
   - coverage
   - ffmpeg
   - pip:
--- a/.github/environment-minimal.yml
+++ b/.github/environment-minimal.yml
@@ -22,7 +22,6 @@ dependencies:
   - pytest-mpl
   - pytest-cov
   - pytest
-  - contextlib2
   - coverage
   - ffmpeg
   - tomli<2.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,16 +21,11 @@ jobs:
             fail-fast: false
             matrix:
                 include:
-                    - os: ubuntu-18.04
-                      python-version: "3.6"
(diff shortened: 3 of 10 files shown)


bmcfee:
#1569 persists after upgrading python on the windows environment to 3.10, still not sure what's going on there.

PR #1572, opened 2022-09-14
vectorized note and svara converters properly

### LINKED ISSUE #1553: Vectorize note converters

Our unit converters are generally pretty liberal about preserving shapes of input.  For example, a matrix of midi numbers can be converted into a matrix of frequencies (Hz):

```python
In [4]: midis = np.random.randint(0, 128, size=(5, 10))

In [5]: midis
Out[5]: 
array([[ 12,  51,  30,  56,  54,  12,  44,  72,  95,  82],
       [ 48,  91,  68,  97,  65,  15,  93, 102,  86,  59],
       [  0,  32, 116,  94, 115, 108,  30, 112,  88,  54],
       [ 75,   4, 121,  31,  91,  22,  92, 127,  15,  23],
       [ 96,  37,   2, 105, 106,  84,  99, 109,  46,  87]])

In [6]: librosa.midi_to_hz(midis)
Out[6]: 
array([[1.63515978e+01, 1.55563492e+02, 4.62493028e+01, 2.07652349e+02,
        1.84997211e+02, 1.63515978e+01, 1.03826174e+02, 5.23251131e+02,
        1.97553321e+03, 9.32327523e+02],
       [1.30812783e+02, 1.56798174e+03, 4.15304698e+02, 2.21746105e+03,
        3.49228231e+02, 1.94454365e+01, 1.76000000e+03, 2.95995538e+03,
        1.17465907e+03, 2.46941651e+02],
       [8.17579892e+00, 5.19130872e+01, 6.64487516e+03, 1.86465505e+03,
        6.27192698e+03, 4.18600904e+03, 4.62493028e+01, 5.27404091e+03,
        1.31851023e+03, 1.84997211e+02],
       [6.22253967e+02, 1.03008612e+01, 8.86984419e+03, 4.89994295e+01,
        1.56798174e+03, 2.91352351e+01, 1.66121879e+03, 1.25438540e+04,
        1.94454365e+01, 3.08677063e+01],
       [2.09300452e+03, 6.92956577e+01, 9.17702400e+00, 3.52000000e+03,
        3.72931009e+03, 1.04650226e+03, 2.48901587e+03, 4.43492210e+03,
        1.16540940e+02, 1.24450793e+03]])
```
This works because the conversion is defined entirely by numpy operations.

When it comes to note conversion, we're not quite as flexible.  For example, in midi_to_note, we detect if the input is scalar, and if not, we construct a list and recursively apply the function to each element of the list:
https://github.com/librosa/librosa/blob/04b9d62a9abdacdf025674413545dafb84e09a1a/librosa/core/convert.py#L648-L652

The result is a list-of-lists:

```python
In [7]: librosa.midi_to_note(midis)
Out[7]: 
[
    ['C0', 'D♯3', 'F♯1', 'G♯3', 'F♯3', 'C0', 'G♯2', 'C5', 'B6', 'A♯5'],
    ['C3', 'G6', 'G♯4', 'C♯7', 'F4', 'D♯0', 'A6', 'F♯7', 'D6', 'B3'],
    ['C-1', 'G♯1', 'G♯8', 'A♯6', 'G8', 'C8', 'F♯1', 'E8', 'E6', 'F♯3'],
    ['D♯5', 'E-1', 'C♯9', 'G1', 'G6', 'A♯0', 'G♯6', 'G9', 'D♯0', 'B0'],
    ['C7', 'C♯2', 'D-1', 'A7', 'A♯7', 'C6', 'D♯7', 'C♯8', 'A♯2', 'D♯6']
]
```

This is okay for most applications, but A) it's a little hacky, and B) it makes the output types difficult to reason about.  

Instead of handling this directly, we could use `np.vectorize` to handle the dispatching automatically.  For example, midi_to_note could be rewritten as a scalar helper function that vectorizes as follows:

```python
def _midi_to_note(midi, *, octave=True, cents=False, key="C:maj", unicode=True):
    
    if cents and not octave:
        raise ParameterError("Cannot encode cents without octave information.")

    note_map = notation.key_to_notes(key=key, unicode=unicode)

    note_num = int(np.round(midi))
    note_cents = int(100 * np.around(midi - note_num, 2))

    note = note_map[note_num % 12]

    if octave:
        note = "{:s}{:0d}".format(note, int(note_num / 12) - 1)
    if cents:
        note = "{:s}{:+02d}".format(note, note_cents)

    return note

midi_to_note = np.vectorize(_midi_to_note, otypes='U', excluded=set(['octave', 'cents', 'key', 'unicode']))
```
Then, continuing the example above, we would have
```python
>>> midi_to_note(midis)
array([['C0', 'D♯3', 'F♯1', 'G♯3', 'F♯3', 'C0', 'G♯2', 'C5', 'B6', 'A♯5'],
       ['C3', 'G6', 'G♯4', 'C♯7', 'F4', 'D♯0', 'A6', 'F♯7', 'D6', 'B3'],
       ['C-1', 'G♯1', 'G♯8', 'A♯6', 'G8', 'C8', 'F♯1', 'E8', 'E6', 'F♯3'],
       ['D♯5', 'E-1', 'C♯9', 'G1', 'G6', 'A♯0', 'G♯6', 'G9', 'D♯0', 'B0'],
       ['C7', 'C♯2', 'D-1', 'A7', 'A♯7', 'C6', 'D♯7', 'C♯8', 'A♯2',
        'D♯6']], dtype='<U3')
```
(unicode-dtype numpy array instead of list-of-lists).

There's still a bit of a snag here, in that a scalar input will produce a (0-dimensional) array output rather than collapsing back to scalar.  I don't know if there's a recommended workaround for this, but it's a little annoying.  Worst case, we could make a little decorator-wrapper that detects this case and returns `.item()` if the input was not a np array.

This would be an API-breaking change, but I think it's worth considering, especially as we may start adding more numeric->string converters in the near future.

bmcfee:
#### Reference Issue
Fixes #1553

This PR replaces the optional recursion in `midi_to_note` (and in `_svara_*`) by a proper numpy vectorization.  This simplifies both the implementation of these functions, and allows them to operate more cleanly on arbitrary array shapes.

It also introduces a function `librosa.util.decorators.vectorize`, which operates similarly to `np.vectorize` but preserves scalar inputs.  (Numpy vectorize, as noted in #1553, maps everything to arrays, including scalars.)

Example usage:

With a multi-dimensional array (list of lists), we now produce a proper ndarray output:
```python
In [11]: librosa.midi_to_svara_c([[22, 23, 24], [25, 26, 27]], Sa=60, mela=1)
Out[11]: 
array([['N₂', 'N₃', 'S'],
       ['R₁', 'G₁', 'G₂']], dtype='<U2')
```
with a scalar input, we produce a scalar output as exepcted:
```python
In [12]: librosa.midi_to_svara_c(22, Sa=60, mela=1)
Out[12]: 'N₂'
```

This is a very slightly API-breaking change, in that we now produce ndarrays instead of lists.  I'm okay with doing this without a full deprecation cycle, but we should definitely highlight it in the release notes.

#### Any other comments?

A few things I'd like to improve in this PR:

1. The vectorize helper is not a proper decorator.  It doesn't necessarily need to be, but it would be a bit more elegant if it was.  I just couldn't get my head through the extra level of indirection involved in making a decorator factory.
2. Tests should be expanded to cover the new use cases.

### DIFF
--- a/librosa/core/convert.py
+++ b/librosa/core/convert.py
@@ -598,13 +598,15 @@ def midi_to_note(midi, *, octave=True, cents=False, key="C:maj", unicode=True):
     >>> librosa.midi_to_note(104.7, cents=True)
     'A7-30'
 
-    >>> librosa.midi_to_note(list(range(12, 24)))
-    ['C0', 'C♯0', 'D0', 'D♯0', 'E0', 'F0', 'F♯0', 'G0', 'G♯0', 'A0', 'A♯0', 'B0']
+    >>> librosa.midi_to_note(np.arange(12, 24)))
+    array(['C0', 'C♯0', 'D0', 'D♯0', 'E0', 'F0', 'F♯0', 'G0', 'G♯0', 'A0',
+           'A♯0', 'B0'], dtype='<U3')
 
     Use a key signature to resolve enharmonic equivalences
 
     >>> librosa.midi_to_note(range(12, 24), key='F:min')
-    ['C0', 'D♭0', 'D0', 'E♭0', 'E0', 'F0', 'G♭0', 'G0', 'A♭0', 'A0', 'B♭0', 'B0']
+    array(['C0', 'D♭0', 'D0', 'E♭0', 'E0', 'F0', 'G♭0', 'G0', 'A♭0', 'A0',
+           'B♭0', 'B0'], dtype='<U3')
 
     Parameters
     ----------
@@ -628,7 +630,7 @@ def midi_to_note(midi, *, octave=True, cents=False, key="C:maj", unicode=True):
 
     Returns
     -------
-    notes : str or iterable of str
+    notes : str or np.ndarray of str
         Strings describing each midi note.
 
     Raises
@@ -736,7 +738,7 @@ def hz_to_note(frequencies, **kwargs):
 
     Returns
     -------
-    notes : list of str
+    notes : str or np.ndarray of str
         ``notes[i]`` is the closest note name to ``frequency[i]``
         (or ``frequency`` if the input is scalar)
 
@@ -751,7 +753,7 @@ def hz_to_note(frequencies, **kwargs):
     Get a single note name for a frequency
 
     >>> librosa.hz_to_note(440.0)
-    ['A5']
+    'A5'
 
     Get multiple notes with cent deviation
 
@@ -1707,7 +1709,7 @@ def midi_to_svara_h(midi, *, Sa, abbr=True, octave=True, unicode=True):
 
     Returns
     -------
-    svara : str or list of str
+    svara : str or np.ndarray of str
         The svara corresponding to the given MIDI number(s)
 
     See Also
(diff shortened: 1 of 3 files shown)


bmcfee:
I think this one is ready to :shipit: - anyone want to check it over?

bmcfee:
I'd like to merge this one today,  unless there are objections.

lostanlen (approved):

Issue #1564, opened 2022-08-31
Add type annotations


matangover:
I would like to have code completions and type checking for code that uses librosa. Adding type annotations will significantly improve the coding experience for users who use type checkers / typing-based completion engines.

Currently some type checkers like Pylance are able to infer some types, however this is often wrong and leads to false positives (errors on working code) or false negatives (missing errors on invalid code). Other type checkers like Mypy cannot do anything with code that uses librosa, and must treat it as 'Any'.

bmcfee:
Sure, this would be great to have. (Especially for the 1.0 release.)

It's also an enormous undertaking that I don't have time for in the short term.  Help on this would be much appreciated.

matangover:
Cool. I know some tools that make it easier to add type annotations for existing code using static inference / runtime introspection. If I have time I'll try running one of them and inspecting the results. How exhaustive are the existing tests in `tests/`? Could they be sufficient to verify that the generated typings are correct (by checking that the tests pass a type check)? Otherwise, do you have other sources of "a bunch of code that uses librosa" which can be used for that purpose?

bmcfee:
> How exhaustive are the existing tests in `tests/`? Could they be sufficient to verify that the generated typings are correct (by checking that the tests pass a type check)?

We aim for full code path coverage, though of course that's not the same thing as being exhaustive with respect to types since the input types are generally not bound.  I do think that the test suite is probably your best bet though, as most of the external code that's out there will only use a small subset of the library.

I think the bigger headache here will be going back through and implementing proper [numpy typing](https://numpy.org/devdocs/reference/typing.html#number-precision) where appropriate (which will bump our min numpy up to 1.20).

matangover:
Gotcha. Precise numpy typing is definitely nice to have but we could also do that as a second step

bmcfee:
Note: #1577 bumped our minimum python up to 1.20, so consider the numpy typing issue unblocked.

bmcfee:
Following up on this, #1579 adds stub files (pyi) for the top-level package and submodules (anything with a `__init__.py` gets a `__init__.pyi`).

While hacking this out, I also went down the rabbit hole of generating full stub declarations for all modules using some automated tooling ([stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html)).  This didn't by any means get the job done, but it at least roughed out the function declarations enough that we could go in and insert proper types.  This seems like the easiest route to follow in implementing type hints for this PR.

[...]

PR #1577, opened 2022-09-21
CI updates, dropping python 3.6

### LINKED ISSUE #1517: Update historical documentation builds
Note: this should be among the last issues implemented in 0.10 because of how our doc build CI works.

When we release 0.10, we should drop the historical build of 0.7.2 and only keep the latest releases from the 0.8 and 0.9 series. (0.8.1, 0.9.2).

Dropping the 0.7 series from doc builds should allow us to update the documentation environment dependencies, specifically for numba, matplotlib, and numpy.

Our doc build uses numba 0.48 because 0.7.2 predates #1164 (0.8.0) to use the newer API.  Removing the 0.7 build should relax this constraint.

#1393 notes the issue with numpy versions - dropping 0.7 allows us to move up to numpy > 1.17 if we want.  This could allow us to go into mypy / typing (numpy >= 1.20) for example.

Likewise, #1324 (0.8.1) supports matplotlib>=3.5, so we no longer have to build on 3.3.  Supporting 3.5 in the doc build removes a potential conflict with some new features slated for 0.10, eg #1426 #1422 #929.  That said, this does present a race condition for merging those features vs. maintaining historical doc builds until the 0.10 release.  I'm not sure how best to handle that yet - maybe we just have to drop 0.7 docs before the release?

bmcfee:
#### Reference Issue
Fixes #1517 

This PR updates the documentation build environment following #1566 .

Adding soxr-python as a core dependency turned out to lead to an unresolveable environment specification (envrionment-docs.yml).  Fixing this required bumping our max version cap on numpy.  However, this breaks the 0.7 doc build - which we were planning to remove anyway #1517 to accommodate new features in the 0.10 series.  As a result, I've removed the 0.7 build, and updated our environment spec.

The 0.7.2 documentation is archived at the wayback machine: https://web.archive.org/web/20220615102306/https://librosa.org/doc/0.7.2/

This PR also updates our minimum supported python from 3.6 to 3.7, and adds 3.10 to the support list and CI build matrix.

#### Any other comments?

I've also updated a couple of other minor inconsistencies in the other CI environments and package requirements.

I've tested the build locally using the new environment spec, and everything seems good.  As long as our new CI environments work cleanly, we can merge this ASAP.


bmcfee:
Ok, we're now deep in dependency hell with the minimal environment.  Losing steam on working this out for now, will resume later on.

bmcfee:
I got this to work in a 3.7 environment by bumping the following minimum versions:
```
  - numpy==1.20.3   # up from 1.17.0
  - scikit-learn==0.20.0  # up from 0.19.1
  - decorator==4.3.0  # up from 4.0.10
  - numba==0.51.0  # up from 0.45.1
```
In each case, there isn't a direct technical reason for librosa to require a newer version than the old minimum.  Rather, the bumps have to do with finding a consistent conda environment from the available package builds with py37.  (Think of it this way: we had old minimum versions that pre-date python 3.7, so they won't exist in the conda channels.)

I'm basically fine with these updates.

- Going to numpy 1.20 actually drops a blocker for #1564 , and could potentially allow us to remove some backported code eg for frame construction.
- sklearn 0.20 is also ancient (2018!) - I wouldn't mind jumping up to a 1.0 min version (2021) if there's any reason to do so.
- decorator 4.0 to 4.3 will not affect us, as those changes primarily have to do with either adding async support (unused) or dropping support for old python versions below our minimum spec anyway.
- ditto numba; 0.51.0 is from 2020, and we've already been future-proofed for a while against changes in the late 0.4x series.

If there are no objections, I'll propagate these minimum versions to all environments and setup.cfg, give it one more spin through the doc builder test (locally) and :shipit: when the dust settles.

bmcfee:
Looks like all tests pass and the local doc build works.  I think this is ready to :shipit:

Issue #1184, opened 2020-06-25
Display module not available


Anriod:
**Describe the bug**
In both versions 0.7 and 0.7.2 it tells me that librosa has no attribute display. I've installed via pip.

**To Reproduce**
<!--
Example:
```
import librosa

librosa.display.specshow(D, y_axis='linear')
-->
Expected behavior : 
Displays spectogram
Actual behaviour: 
`AttributeError: module 'librosa' has no attribute 'display
`

```

import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import librosa; print("librosa", librosa.__version__)
librosa.show_versions()
Darwin-18.7.0-x86_64-i386-64bit
Python 3.6.8 (default, Jan 13 2019, 14:36:45) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.10.44.4)]
NumPy 1.16.4
SciPy 1.4.1
librosa 0.7.0
INSTALLED VERSIONS
------------------
python: 3.6.8 (default, Jan 13 2019, 14:36:45) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.10.44.4)]

librosa: 0.7.0

audioread: 2.1.6
numpy: 1.16.4
scipy: 1.4.1
sklearn: 0.21.2
joblib: 0.13.2
decorator: 4.3.0
six: 1.12.0
soundfile: 0.10.3
resampy: 0.2.2
numba: 0.43.1

numpydoc: None
sphinx: None
sphinx_rtd_theme: None
sphinxcontrib.versioning: None
sphinx-gallery: None
pytest: None
pytest-mpl: None
pytest-cov: None
matplotlib: 3.0.1
presets: None

```

lostanlen:
Hello @Anriod,

`import librosa.display` will solve your problem. The display submodule is not imported automatically because it has matplotlib as an external dependency, unlike the rest of the library.

I hope this helps!

Anriod:
Thanks!

PR #1579, opened 2022-09-22
lazy imports

### LINKED ISSUE #1523: lazy-import for submodules

Way back in the past, we made the `display` submodule "optional".  This means that you can use librosa without importing any of the display submodule, or even installing its specific dependencies (ie matplotlib).  This was done primarily to avoid memory bloat, but also to keep the environment dependencies down when display is not needed, eg in embedded device applications.  The memory bloat issue is largely gone at this point due to improved usage of matplotlib, but the environment dependency issue persists.

The drawback of this optional design is that users need to explicitly import `librosa.display`.  This is a minor nuisance, but it does lead to one of our [most googled questions](https://www.google.com/search?channel=fs&client=ubuntu&q=attributeerror+module+%27librosa%27+has+no+attribute+%27display%27).  This particularly hits novice users.

The upcoming scipy 1.9 release uses a lazy-loader to manage submodules: https://github.com/scientific-python/lazy_loader

Rather than loading everything at import time, submodules are only loaded when accessed.  This would make it possible to drop the requirement of explicitly importing `librosa.display` but maintain the benefits of not importing display if it's not needed.

Additionally, if a user *does* try to use `librosa.display` without having a proper environment for it, they'll get a slightly more useful error message that matplotlib is not installed.

We could leave it as is.

**Additional context**
This will add a dependency to librosa.  However, scipy 1.9 will also require the lazy-loader package, so on up-to-date environments we'll get this for free.

This may come with some performance hits - I'm not sure, as I haven't tried it out yet.

Additionally, lazy-loader could have some unpredictable interactions with my `presets` package that we might need to sort out.

This is all to say, I'm not 100% sold on this idea yet, but wanted to raise it for discussion.

bmcfee:
Fixes #1523 , and #1184 

This PR implements lazy loading for faster imports, following the recent examples out of scipy and skimage and using the `lazy_loader` package.

Quick demo of the benefits.  Importing librosa before (current main branch):
```
In [1]: %time import librosa
CPU times: user 1.09 s, sys: 88.1 ms, total: 1.17 s
Wall time: 1.19 s
```

and after, with lazy loading:
```
In [1]: %time import librosa
CPU times: user 1.81 ms, sys: 151 µs, total: 1.96 ms
Wall time: 1.21 ms
```

As an additional benefit, we no longer have to separately import `librosa.display`!  This is a big win for usability.

#### Any other comments?

A couple of things to be aware of here:

1. `lazy_loader` is on pypi, but not yet on conda forge.  We should hold off on merging until the packaging catches up.
2. I've only implemented this at the top-level package import.  I don't think there's much benefit to drilling down further to thread laziness throughout the package, but I could be persuaded otherwise.
3. lazy loader does not like `*` imports, so the top-level `__init__` now explicitly lists out each element being imported from core (and other submodules).  This is probably a good thing, but it does mean that we need to explicitly update the top-level index any time we add or change a core function.


bluenote10:
How does that affect auto-completion in IDEs? I could imagine that it stops working unless they already have explicit support for lazy-loader?

bmcfee:
> How does that affect auto-completion in IDEs?

It should be fine - the lazy-loader package populates `__dir__` and `__all__` for us.  A quick check with ipython works as expected, but if you have something else in mind it would be great if you could clone this PR and test-drive it.

Side note: the latest scipy release is already using this package, so you could also try with that.

bluenote10:
> It should be fine - the lazy-loader package populates __dir__ and __all__ for us.

The problem is that it populates it dynamically. Type checkers and IDEs cannot execute Python module-scope code for obvious reasons. Therefore they have agreed an a set of simple allowed operations that they can support statically as [documented in the typing specs](https://github.com/python/typing/blob/master/docs/source/libraries.rst#library-interface-public-and-private-symbols).

> A quick check with ipython works as expected

ipython has runtime information, so this doesn't apply for IDEs like VSCode or Pycharm. I can give it a try later.

Don't you rely on typing `librosa.` and wait for the IDE to tell you about all its functions including all their parameters? It would be quite a reductions in productivity to me if that no longer works.

Somehow the move towards lazy-loading feels a bit sad. Running non-trivial code including side effects at import time is an anti-pattern in general, not only from a performance perspective, but also in terms of error handling. Libraries should be encouraged to avoid doing so and postpone execution of such logic, shifting it into first function executions. In my opinion introducing lazy-loading reinforces that anti-pattern and may be detrimental to the ecosystem. It felt cleaner to me to let the interpreter validate an entire package tree eagerly (for instance lazy-loading could hide syntax errors in a module that isn't accessed). Perhaps we will soon see many more crashes import time, if more and more logic will move out of currently existing "proper" lazy loading implementations into import time execution.

bmcfee:
> EDIT: The [lazy-loading specs](https://scientific-python.org/specs/spec-0001/) mentions that by default it does break IDE support, but suggests to introduce a `.pyi` file that contains the corresponding imports. Perhaps this is a reasonable work-around?

Hah, you beat me to it - but yes, this is known and the workaround seems reasonable to me.


> Don't you rely on typing `librosa.` and wait for the IDE to tell you about all its functions including all their parameters?

No, we don't implement type hints (yet, see #1564).  As such, I don't see how adding lazy-loader here would constitute a meaningful change from an IDE's perspective.

bluenote10:
> No, we don't implement type hints (yet, see https://github.com/librosa/librosa/issues/1564).

There is no need for that! Because of the conventions mentioned above and the magic included in modern IDEs you already get that today:

Peek 2022-09-23 14-52

I can only recommend to give a modern IDE a try, this functionality is super convenient and a productivity boost.

For comparison, I have checked out the feature branch. As expected, all the nice auto-complete and inline docstring support is gone:

Peek 2022-09-23 14-54

For people depending on it this would be a significant regression.

bmcfee:
Ok, thanks for checking this.  I tried installing vscode, but could not figure out how to make it work with my conda environments.  Y'all can pry (neo)vim from my cold, dead fingers :rofl: 

Seriously though, this PR is marked as WIP.  Does your tab completion IDE magic work with scipy 1.9?  If so, I'm happy to tack in whatever other hints are needed to retain this functionality.

bmcfee:
This vscode nonsennse appears to be at least partially down to a known issue in pylance: https://github.com/microsoft/pylance-release/issues/3262

bmcfee:
Alright, after hacking this thing to manually point to the source path of the git repo, I can confirm that adding a stub file at the top level does partially resolve the autocomplete issue.  Autocomplete works on all simple submodules, but breaks for the complex ones (feature, util, core).  Probably this can be fixed by putting in submodule stubs.  I'll give that a whirl after lunch.

bluenote10:
> Is there maybe some simpler, minimal tooling that we could use to test this, rather than an ten-ton IDE?

Under the hood, VSCode uses [pyright](https://github.com/microsoft/pyright), which has a command line version. Perhaps this could be used to infer what pylance/VSCode "sees", but I've never tried it.

> This vscode nonsennse appears to be at least partially down to a known issue in pylance: https://github.com/microsoft/pylance-release/issues/3262

Sorry for causing all the trouble. Properly working auto-completion is unfortunately indeed a bit brittle, and experiences can indeed vary (the variety of Python project setup approaches isn't easy to handle for IDEs of course) :confused:

bmcfee:
> Sorry for causing all the trouble. Properly working auto-completion is unfortunately indeed a bit brittle, and experiences can indeed vary (the variety of Python project setup approaches isn't easy to handle for IDEs of course) confused

No worries, it's good to catch these things early!  Also, in retrospect, I misunderstood your initial comment:

> Don't you rely on typing `librosa.` 

as meaning "typing as in :duck:", not "typing as in :keyboard:". This is only exacerbated by the fact that in python, we use the former to assist the latter! :facepalm: 



> Under the hood, VSCode uses [pyright](https://github.com/microsoft/pyright), which has a command line version. Perhaps this could be used to infer what pylance/VSCode "sees"

Indeed, it seems like that's how it all works according to the [FAQ](https://github.com/microsoft/pylance-release/blob/main/FAQ.md):

> **What is the relationship between Pylance, Pyright and the Python extension?**

> Pylance is a language server, which provides features like IntelliSense, logical linting (e.g., errors and warnings), code actions, code navigation, semantic colorization, etc. Pyright > is the open-source type checker which Pylance uses under the hood to provide its functionality. Pylance provides a superset of Pyright’s functionality. To use Pylance, you must > have the core Python extension installed in Visual Studio Code as it builds upon that experience.

bmcfee:
Ok, I've tried threading lazy loads throughout the submodules, and it's turned into a massive yak shave.  I'm not sure if it's my install of vscode continuing to be broken (quite possible), if I'm doing something wrong with the stub attacher, or if I'm hitting unimplemented functionality, but it doesn't seem to be working yet.  (I'd love it if someone who actually uses this IDE magic could try out the most recent commit.)

Basically, everything works fine if you use a full path, eg `librosa.core.audio.load`, but exposed bits from submodules (eg, `librosa.core.load`, `librosa.load`) are not working for me.

Now, in the worst case, we roll this idea back.  However, I think we could still implement some lazy external loads for optional dependencies (matplotlib, samplerate, resampy).  This would have a few benefits: namely efficiency at runtime, and we'd be able to expose `display` at the top level again and retain a fix for #1184.

bmcfee:
Following up with the dump from pyright:

[pyright-librosa-fail.txt](https://github.com/librosa/librosa/files/9648087/pyright-librosa-fail.txt)

bschnurr:
one way to help pylace/pyright know about lazy loaded imports is to wrap the original imports under an `if TYPE_CHECKING`

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import samplerate 
else:
    samplerate = lazy.load('samplerate')
```

stefanv:
A solution by the pyright developers!

https://github.com/microsoft/pyright/issues/3989

stefanv:
[Lazy loader docs updated](https://github.com/scientific-python/specs/pull/158)

Stub format should be either:

```
from .something import my_func as my_func
```

or

```
__all__ = ['my_func']
from .something import my_func
```

bmcfee:
Note: this is still blocked by getting an updated lazy-loader (0.1) package merged into conda-forge.  Once that happens, I'll update our test environment specs here and we can unblock.

stefanv:
[conda-forge PR](https://github.com/conda-forge/lazy_loader-feedstock/pull/1) by @bmcfee

bluenote10:
> @bluenote10 I think the most recent commit should fix this for you, could you please verify?

Yes, I gave it a short try: There is some auto-completion now, although behaving differently then before.

What no longer works is that the docstrings are already shown during auto-completion as visible in the GIFs above. Whether that is an effect of the type stubs or rather a VSCode mood I don't know. It's not unlikely thought that it is a real effect, because now the docstrings and the type annotations are separated, which may mean that the docstrings are ignored at auto-completion time. It's not too much of an issue, because after auto-completion it is still possible to see the docstrings.

Also due to the explicit stubs, the annotations themselves are different, basically because now it is no longer Pyright which does the inference, but just the inference result of the mypy stubgen. In general my experience with the stubgen in other contexts was that Pyright's inference system can infer more, but on first glance the stubgen didn't look too bad either.

stefanv:
@bluenote10 I can't quite visualize how the docstrings appear for you now, but if that is satisfactory then :+1: 

Re: the stubs, I am trying to understand why you see different type inference results than before, since @bmcfee removed the function-specific stubs (i.e. there are no mypy stubs present in the last commit).

bluenote10:
> Re: the stubs, I am trying to understand why you see different type inference results than before, since @bmcfee removed the function-specific stubs (i.e. there are no mypy stubs present in the last commit).

Interesting, I need to double check if I was on the right commit then.

The obvious thing was that everything that was previously annotated as `Any` now appeared as `Unknown` which is typically a trademark of the mypy stubgen (and totally fine of course).

bmcfee:
Just checking back on this, since I'm not really clear whether the last round of changes put things in a state that we're happy with.  @bluenote10 are you still having problems with tab completion, or do you think it's useful enough at this point?

bluenote10:
> are you still having problems with tab completion, or do you think it's useful enough at this point?

Yes definitely useful enough in my last experiment.

bmcfee:
Great, thanks @bluenote10 !  In that case, I think I'm ready to declare this one done and merge it.

stefanv:
Excellent, thanks for trying this out and helping figure out how it should work! 🚀

PR #1594, opened 2022-10-21
Replace numpy deprecated types with python types.

mattiadg:
#### Reference Issue
Complete fixing of issue #1350

Replaces deprecated np types complex and float with the equivalent Python types

### DIFF
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -2214,7 +2214,7 @@ def dtype_c2r(d, *, default=np.float32):
     mapping = {
         np.dtype(np.complex64): np.float32,
         np.dtype(np.complex128): np.float64,
-        np.dtype(complex): np.dtype(np.float).type,
+        complex: float,
     }
 
     # If we're given a real type already, return it
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -2214,7 +2214,7 @@ def dtype_c2r(d, *, default=np.float32):
     mapping = {
         np.dtype(np.complex64): np.float32,
         np.dtype(np.complex128): np.float64,
-        complex: float,
+        np.dtype(complex): np.dtype(float).type,
     }
 
     # If we're given a real type already, return it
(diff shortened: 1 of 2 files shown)


librosa/util/utils.py
```diff
@@ -2214,7 +2214,7 @@ def dtype_c2r(d, *, default=np.float32):
     mapping = {
         np.dtype(np.complex64): np.float32,
         np.dtype(np.complex128): np.float64,
-        np.dtype(complex): np.dtype(np.float).type,
+        complex: float,
```
bmcfee, line 2217: I believe this should be `np.dtype(complex): np.dtype(float).type`.  That way we always return a `dtype`, and it maintains precision parity with the input complex type.  This is how it was done in `dtype_r2c` above.

mattiadg: I've done this, but why do we need to use `np.dtype(float).type` and not just `np.dtype(float)`?

bmcfee: For consistency - the first two cases return either `np.float32` or `np.float64`, and the third case is meant to map to one of these two as well.  If we return `np.dtype(float)` we get an output of type:
```python
In [9]: type(np.dtype(float))
Out[9]: <class 'numpy.dtype[float64]'>
```
This is a subclass of `type`, but it's not the same type that we'd get if we return `np.float32` directly:
```python
In [11]: type(np.float32)
Out[11]: <class 'type'>
```
Returning `np.dtype(float).type` gives
```python
In [10]: type(np.dtype(float).type)
Out[10]: <class 'type'>
```

mattiadg: Thank you for your clarification.
I didn't notice I didn't push my change. Is there anything else needed?

bmcfee (approved):

librosa/util/utils.py
bmcfee: Nope, LGTM, thanks!  I'll let the CI run and if it's all clear, we can merge.

bmcfee:
Hrm, not sure what's going on with the CI here - it runs fine on my dev machine (py39). I tend to think the CI failures are unrelated, but I'll try to make some time today to verify that.

bmcfee:
Update: it looks like the test failures are due to an update in numpy.  The py3.7 environment (passing) is using numpy 1.20, the later versions are on 1.23 (failing).  Since this is unrelated to the current PR, I think we're good to merge this.

mattiadg:
Great, thanks!

Issue #758, opened 2018-08-07
BibTeX entry on Zenodo has some non-standard author names


lostanlen:
#### Description
I'm citing librosa 0.6.1 in a paper today and use the BibTeX file automatically generated by Zenodo. Frustratingly, it is not a "plug and play" experience: some authors do not follow a "Firstname (Middlename) Lastname" template. I would prefer if they did.
After talking with @bmcfee, he told me that:
`Seth` is `Seth Kranzler`
`Erik` is `Erik Peterson`
`JackieWu` is `Jackie Wu`

Other manual change that were necessary were for authors with non-ASCII characters in their name:
`Carl Thomé`, `Fabian-Robert Stöter`, `João Felipe Santos`.

I imagine that this is largely an `upstream` issue to Zenodo, and possibly a `wontfix`, but is there any way we can hack into our database of past contributors and make the relevant changes before it gets scraped by Zenodo?

Even if this is not the case, the issue is worth raising, if only because it gives me an opportunity to share the expected BibTeX entry, with manual fixes applied (see below), for everyone to use.

#### Steps/Code to Reproduce
Link: https://zenodo.org/record/1252297

#### Expected Results
```
@misc{brian_mcfee_2018_1252297,
  author       = {Brian McFee and
                  Matt McVicar and
                  Stefan Balke and
                  Carl Thom\'{e} and
                  Colin Raffel and
                  Dana Lee and
                  Oriol Nieto and
                  Eric Battenberg and
                  Dan Ellis and
                  Ryuichi Yamamoto and
                  Josh Moore and
                  Rachel Bittner and
                  Keunwoo Choi and
                  Pius Friesch and
                  Fabian-Robert St\"{o}ter and
                  Vincent Lostanlen and
                  Siddhartha Kumar and
                  Simon Waloschek and
                  Seth Kranzler and
                  Rimvydas Naktinis and
                  Douglas Repetto and
                  Curtis ``Fjord'' Hawthorne and
                  CJ Carr and
                  Waldir Pimenta and
                  Petr Viktorin and
                  Paul Brossier and
                  Jo\~{a}o Felipe Santos and
                  Jackie Wu and
                  Erik Peterson and
                  Adrian Holovaty},
  title        = {librosa/librosa: 0.6.1},
  month        = may,
  year         = 2018,
  doi          = {10.5281/zenodo.1252297},
  url          = {https://doi.org/10.5281/zenodo.1252297}
}
```

#### Actual Results
```
[...]

PR #1615, opened 2022-12-05
Adds `CITATION.cff` file

cr2007:
Used for easily citing the software elsewhere

#### Reference Issue
No Issues to reference

This PR implements the addition of a **`CITATION.cff`** file, which is used for easily citing the software directly from [GitHub](https://github.com/librosa/librosa) without leaving the page.
I have referenced the author list provided from [Zenodo](https://zenodo.org/record/6759664#.Y432DRRBz30).

More Info:
- [What is a `CITATION.cff` file?](https://citation-file-format.github.io/)
- [About CITATION files - GitHub Docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files)
- [citation-file-format/citation-file-format](https://github.com/citation-file-format/citation-file-format) - GitHub Repo

#### Any other comments?

You can have a preview of it in my [forked repo](https://github.com/cr2007/librosa), where you can directly cite the repository from there.

Feel free to add any other edits that you see fit.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,84 @@
+# Referenced from https://zenodo.org/record/6759664#.Y43yzhRBz30
+
+cff-version: 1.2.0
+title: librosa
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+license: "ISC"
+version: 0.9.2
+date-released: 2022-06-27
+authors:
+  - given-names: Brian
+    family-names: McFee
+    affiliation: New York University
+  - given-names: Alexandros
+    family-names: Metsai
+  - given-names: Matt
+    family-names: McVicar
+  - given-names: Stefan
+    family-names: Balke
+  - given-names: Carl
+    family-names: Thomé
+    affiliation: '@xlnaudio'
+  - given-names: Colin
+    family-names: Raffel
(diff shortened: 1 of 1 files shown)


bmcfee:
Thanks for this, but I don't want to hard-code the reference in this way, and I'm generally unsure about how we should be using citation.cff in the first place.

For the first point, the citation file should be synchronized (ideally, automatically) to the most recent release.  I wonder if there's a github action for zenodo to push citation.cff changes upon release?

For the more general point, we do have explicit instruction on how to cite librosa.  Unfortunately, almost nobody adheres to our instructions: most people cite the scipy paper, rather than the specific version that they're using.  The original thinking behind splitting the citation instruction in this way was A) to keep track of specific version usage, especially for newer papers using older versions, and B) to make sure that contributors are properly attributed.  For example, if someone contributed to version 0.8, and people only cite 0.7 even when using 0.8 or later, that new contributor gets no credit.

Now, I generally worry that making citation.cff available and easy to use will just encourage more bad habits.  This would be great if there's a separate citation file for each release (like zenodo already does), but automatically exposing the most recent release (and only the most recent release) would result in similar attribution problems.

lostanlen:
Related: #758

PR #1616, opened 2022-12-05
Syntax highlighting

DavidRomanovizc:
#### Any other comments?
Now in all commands and codes examples have syntax highlighting. It is more beautiful than without highlighting

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,29 +52,29 @@ of things that librosa can do.
 ### Using PyPI
 
 The latest stable release is available on PyPI, and you can install it by saying
-```
+```py
 python -m pip install librosa
 ```
 
 ### Using Anaconda
 
 Anaconda users can install using ```conda-forge```:
-```
+```py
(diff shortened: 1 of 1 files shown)


bmcfee:
I think this is incorrect - almost all of the code blocks are shell script, not python.  Using ` ```python ` will result in incorrect syntax highlighting.

DavidRomanovizc:
> I think this is incorrect - almost all of the code blocks are shell script, not python. Using `` ```python `` will result in incorrect syntax highlighting.

I know, but if we use `sh` or `bash`,  shell scripts will not be highlighted

bmcfee:
Ok, then I think we should leave it as is.

cr2007 (commented):
Not all commands are necessarily Python commands though

Ones like `apt-get install ffmpeg` are done in Shell/Bash

```shell
apt-get install ffmpeg
```

PR #1617, opened 2022-12-09
Parabolic interpolation speedup for yin

### LINKED ISSUE #1608: Accelerate parabolic interpolation in yin/pyin

In the general theme of the 0.10 milestone, I noticed a step in yin/pyin that could be done much more efficiently.  The parabolic interpolation function:
https://github.com/librosa/librosa/blob/0a60e0a87888d1bc8b2c6098e17d2b98536a28a0/librosa/core/pitch.py#L427-L450
works by slicing the input array and doing some vectorized arithmetic ops to compute the intermediate values.  Each intermediate value (`parabola_a` and `parabola_b`) will therefore require allocating almost as much space as the input array (`y_frames`) - so we're tripling memory usage unnecessarily.

The operation here is entirely local, and could be easily implemented by a numba stencil operation. This is very similar to what we did for localmax/localmin in #1533,  eg:
https://github.com/librosa/librosa/blob/0a60e0a87888d1bc8b2c6098e17d2b98536a28a0/librosa/util/utils.py#L983-L1003

This isn't the biggest speed bottleneck in yin/pyin, but it would be easy to improve so we should do it anyway.

bmcfee:
#### Reference Issue

Fixes #1608 

This PR replaces our parabolic interpolation helper with a faster implementation using numba stencils. There are some very minor changes to numerical behavior, but overall users should see no practical difference.  Speedup on this operation is about 80-90% in my (not very rigorous) benchmarks.

#### Any other comments?

If this passes CI, we should be fine to merge.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,6 +8,9 @@ omit =
 exclude_lines =
     @numba.jit
     @jit
+    @numba.stencil
+    @numba.guvectorize
+    @numba.vectorize
     pragma: no cover
     if TYPE_CHECKING:
     @overload
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -327,20 +327,9 @@ def piptrack(
     # Do the parabolic interpolation everywhere,
     # then figure out where the peaks are
     # then restrict to the feasible range (fmin:fmax)
-    avg = 0.5 * (S[..., 2:, :] - S[..., :-2, :])
-
-    shift = 2 * S[..., 1:-1, :] - S[..., 2:, :] - S[..., :-2, :]
-
-    # Suppress divide-by-zeros.
-    # Points where shift == 0 will never be selected by localmax anyway
-    shift = avg / (shift + (np.abs(shift) < util.tiny(shift)))
-
-    # Pad back up to the same shape as S
-    padding = [(0, 0) for _ in S.shape]
-    padding[-2] = (1, 1)
-    avg = np.pad(avg, padding, mode="constant")
-    shift = np.pad(shift, padding, mode="constant")
-
+    avg = np.gradient(S, axis=-2)
+    shift = _parabolic_interpolation(S, axis=-2)
+    # this will get us the interpolated peak value
     dskew = 0.5 * avg * shift
 
     # Pre-allocate output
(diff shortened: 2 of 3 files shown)


bmcfee:
Looks like coverage drops are coming from numba stencils - I'll add a coverage skip for these.

bmcfee:
Any objections to merging this one?  @lostanlen ?

bmcfee:
Did a bit more unpacking of these two functions, and TLDR is that I think piptrack should use the accelerated helper here with a second helper to finish the magnitude interpolation.

The implementation in pyin has explicit logic to keep the estimated peak within a ±1-bin range.  (This is what the magnitude check for abs(b) < abs(a) is doing.)  The implementation in pyin has similar, but more convoluted logic: it first avoids division-by-zero (fine) but does not do a bounds check otherwise; it then later discards estimates outside the spectral peaks.  I think we can probably streamline this by baking the peak filtering into the helper stencil.

The piptrack implementation is also a little tangled in that it overrides the `shift` variable: first equivalent to `-a` in the pyin helper, then later equivalent to `-b/a` (modulo divide-by-zero protections).  The final magnitude interpolation is equivalent to `S - 1/2 * b * b/a`, or if we call `out = -b/a` the result of the pyin helper, `S[k] ← S[k] + 1/4 * (S[k+1] - S[k-1]) * out`.  So if we tack on another helper that computes the skip-one difference (just the `b` part), we can do this all with much less intermediate storage.

bmcfee:
Simplified piptrack is implemented now as well.  Speedup is tidy (20s chirp):

```python
>>> %timeit librosa092.piptrack(S=S)
29.6 ms ± 3.34 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
>>> %timeit librosa.piptrack(S=S)
18.7 ms ± 501 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
>>> pip1, mags1 = librosa092.piptrack(S=S)
>>> pip2, mags2 = librosa.piptrack(S=S)
>>> np.allclose(mags2, mags1)
True
>>> np.allclose(pip2, pip1)
True
```

bmcfee:
Ok, I think this is :shipit: once CI passes - @lostanlen wanna take one more look?

lostanlen (approved):

PR #1625, opened 2022-12-15
Soundfile exception handling, version bump

### LINKED ISSUE #1456: Discussion: deprecating / removing audioread?
Opening a conversation here to plan out how we want to handle deprecating and removing audioread as a dependency.

## Background

When librosa development began, audioread was the best option available for handling audio input in diverse formats.  We eventually added `soundfile` (libsndfile) as a secondary option for decoding, and later promoted that to a first-class dependency.

The current state of the project is that soundfile is the primary and preferred decoder package, and we fall back to audioread only when necessary (i.e., when given a file outside of soundfile's codec support).  We now have some functionality that is *only* supported when using soundfile: notably stream processing, but there are requests for other features (e.g. #1434 ) which could expand this list.

For quite a while, we've been planning to deprecate and remove audioread as a backend, and streamline the library to only use soundfile for I/O.  The primary roadblock to this has been the lack of mp3 support in libsndfile, which should be changing soon.  So now seems like a good time to discuss how we plan to approach the deprecation.

## Audioread vs soundfile codecs

It's worth discussing what we stand to lose by abandoning audioread (and ffmpeg, gstreamer, pymad, etc).  Since soundfile's object interface is a bit richer than audioread's, I believe this entirely comes down to supported audio codecs.

Undoubtedly, the combination of gstreamer, ffmpeg, mad, and coreaudio provides greater format coverage than what is implemented directly in [libsndfile](https://libsndfile.github.io/libsndfile/formats.html).  (See, e.g, [libavcodec](https://en.wikipedia.org/wiki/Libavcodec#Implemented_audio_codecs) for ffmpeg audio codecs.) My hunch is that for the common uses of librosa, it would be exceedingly rare to encounter a format (not including mp3) that would not be supported by libsndfile.  The exception here might be pulling audio directly from audio-video streams (mp4, aac, etc).  

## Why deprecate audioread?

There are three reasons that come to (my) mind in favor of deprecation:

1. minimizing environment complexity
2. functionality (e.g. stream processing)
3. stability, speed, and (upstream) test coverage

### Environment complexity

Audioread is designed to multiplex over different decoder backends, such as gstreamer or ffmpeg.  This maximizes the supported audio codecs, but at the cost of having complex environments that are not easily managed by python.  Support also differs across platforms, which can lead to unexpected differences in behavior.  (GStreamer and ffmpeg both, in turn, multiplex over different modules, so not all environments using these packages will necessarily support the same formats.)  Because these differences can be hidden by audioread (and further obscured by librosa), it can be challenging to diagnose problems.

### Functionality

As far as I can tell, the soundfile API is strictly more complete than audioread in terms of what operations are supported.  If all I/O is conducted through soundfile, this could open up some possibilities for more functional and streamlined API from the librosa side as well.  (If nothing else, it would be more internally consistent by virtue of not needing to support multiple decoder interfaces.)

### Software health

I don't mean to suggest that audioread is unstable, but its test coverage is extremely minimal.  

Moreover, its design inherently interacts with multiple different backend libraries and programs, often through subprocess pipes.  This makes some operations like offset-based load potentially inefficient compared to soundfile's seek-based implementation.

## What timeline makes sense?

It would be great if we could have this deprecation cycle set to complete by the 1.0 release (aiming for 2023.01).  This would mean deprecation in the 0.10 series (summer 2022).  This is an aggressive timeline, but not impossible depending on how quickly things happen on the libsndfile side.

If that is not possible, I would suggest to deprecate in 1.0 and remove in a future 2.0 release.  (We haven't yet begun discussing 2.0 features, but this seems like the kind of change that would be appropriate for a major revision.)  This gives a longer runway for deprecation.

### LINKED ISSUE #1624: Better soundfile exception handling
The python-soundfile package changed its error reporting in the 0.11 release, and we should keep up.

Previously, soundfile would throw a `RuntimeError` for things like unsupported codecs, and we catch this here:
https://github.com/librosa/librosa/blob/e4edf2b98ce540633fd68040e8c62760c2087b7d/librosa/core/audio.py#L168-L177

In the 0.11 release, this has changed to a custom exception class `SoundFileRuntimeError`.  This class does inherit from `RuntimeError`, so our current implementation will continue to work, but it would be better to be more precise here and avoid catching unrelated runtime errors.

Doing this will require a minimum version bump to 0.11.  Our current minimum version is 0.10.2.  I don't see any problem with this version bump.

**Additional context**
[Soundfile documentation on error reporting](https://python-soundfile.readthedocs.io/en/0.11.0/#error-reporting)

Work on this one had previously been started in #1228 , but I think at this point it would be best to start fresh and not try to unwind and rebase against the changes from the last two years.

### LINKED ISSUE #1627: get_duration: filename -> path
This is a very minor thing, but `get_duration` supports `filename=` as an argument, while related functions `load`, `get_samplerate`, `stream` all call it `path`.

We should deprecate the `filename=` parameter and rename it to `path=`.

### LINKED ISSUE #1628: Deprecations should use FutureWarning, not DeprecationWarning
Title says it all - guess I never looked too closely at the core documentation! :laughing: 

https://docs.python.org/3/library/exceptions.html#FutureWarning

TLDR: deprecation warnings are usually hidden, future warnings are not.

Replace uses of `DeprecationWarning` by `FutureWarning`.

bmcfee:
#### Reference Issue
- Fixes #1624 
- Replaces #1228 
- Fixes #1456 
- Fixes #1628
- Fixes #1627 

This PR implements the changes proposed in #1624 to narrow the exception catch from soundfile.

It bumps the minimum supported soundfile version to 0.11.

Documentation has also been updated to reflect the current state of affairs.

- [x] TODO: deprecate audiread backend

#### Any other comments?

I'm not exactly sure how to implement a test to cover audioread loading.


bmcfee:
Codecov failures are deeply annoying, but tests are now up to snuff.

Next step is audioread deprecations, then I'll get on to #1628 in a separate PR.

bmcfee:
CI is all clear, and docs are updated.  I think this is good to go.

README.md
```diff
@@ -108,12 +108,11 @@ librosa (0.x.x, /path/to/librosa)
 
 `librosa` uses `soundfile` and `audioread` to load audio files.
 
-📝 Note that `soundfile` does not currently support MP3, which will cause librosa to
-fall back on the `audioread` library.
+📝 Note that older releases of `soundfile` (prior to 0.11) do not support MP3, which will cause librosa to fall back on the `audioread` library.
```
lostanlen, line 111: For MP3 decoding, librosa depends on `soundfile` v0.11 or later if available. For earlier versions, librosa falls back on `audioread`.

lostanlen (commented):
a small comment regarding the README.
"prior to" is not recommended for ESOL readers ;)

README.md
bastibe: Please be aware that `soundfile` relies on `libsndfile` 1.1 for MP3 decoding. Most Linux distributions do not yet ship `libsndfile` 1.1. You can `soundfile.check_format('MP3')` to see whether MP3 decoding is available.

PR #1637, opened 2022-12-31
Minor typing improvements

bluenote10:

Just a few minor things I had in mind when reviewing the other typings PR.

### DIFF
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -18,7 +18,7 @@ dependencies:
   - soxr-python>=0.3.2
   - msgpack-python>=1.0
   - lazy_loader>=0.1
-  - typing_extensions>=4.0.0
+  - typing_extensions>=4.1.0
 
   # optional, but required for testing
   - matplotlib>=3.3.0
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -18,7 +18,7 @@ dependencies:
   - soxr-python>=0.3.2
   - msgpack-python>=1.0
   - lazy_loader>=0.1
-  - typing_extensions>=4.1.0
+  - typing_extensions>=4.1.1
 
   # optional, but required for testing
   - matplotlib>=3.3.0
--- a/.github/environment-docs.yml
+++ b/.github/environment-docs.yml
@@ -25,7 +25,7 @@ dependencies:
   - sphinx-gallery>=0.10.1
   - sphinx_rtd_theme>=1.0.0
   - lazy_loader>=0.1
-  - typing_extensions>=4.1.0
+  - typing_extensions>=4.1.1
   - ffmpeg
   - pip:
     - sphinx-multiversion==0.2.4
--- a/.github/environment-docs.yml
+++ b/.github/environment-docs.yml
@@ -25,7 +25,7 @@ dependencies:
   - sphinx-gallery>=0.10.1
   - sphinx_rtd_theme>=1.0.0
   - lazy_loader>=0.1
-  - typing_extensions>=4.0.0
+  - typing_extensions>=4.1.0
   - ffmpeg
   - pip:
     - sphinx-multiversion==0.2.4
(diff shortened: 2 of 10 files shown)


bmcfee:
Thanks for this - everything looks okay to me, with a couple of caveats:

1. There doesn't appear to be a typing-extensions==4.1.0 in conda for python 3.7, so our minimal build is failing.  Can we bump this to 4.1.1?  From what I can tell, 4.1.0 didn't work on 3.7 anyway.
2. I understand the change in the hpss window spec parsing, but it is now more restrictive than it was previously.  Eg it will now fail on something like `hpss(margin=[15, 15])` because it uses a list instead of a tuple.  It's true that the type annotation specifies tuple here, and it at present silently works with other sequence types, but changing over to an explicit check against tuple might break things for downstream users.  This is why we used `isscalar()` instead.  I wonder if there's another option here?

bluenote10:
> Can we bump this to 4.1.1?

Yes sure, thanks for looking into this!

> Eg it will now fail on something like hpss(margin=[15, 15]) because it uses a list instead of a tuple. 

If supporting `List` as input is on purpose, we probably should broaden the type explicitly to communicate that and to avoid dropping support for `List` accidentally. I have simply added lists now to both the type and the check.

In the long term, we could also wrap things like `np.isscalar` and other runtime type checks into [user defined type guards](https://mypy.readthedocs.io/en/stable/type_narrowing.html#user-defined-type-guards). This could make sure that the runtime checks are always exactly aligned with the types we are using (i.e., the types in `_typing.py`). I haven't tested it yet, but roughly something like that:

```py
def _is_iterable_like(x: object) -> TypeGuard[_IterableLike[_T]]:
    return isinstance(x, list) or isinstance(x, tuple) or isinstance(x, types.GeneratorType)
    
def _is_bool_like(x: object) -> TypeGuard[_BoolLike_co]:
    ...

def _is_int_like(x: object) -> TypeGuard[_IntLike_co]:
    ...

def _is_float_like(x: object) -> TypeGuard[_FloatLike_co]:
    # I guess the semantics of np.isscalar corresponds to our _FloatLike_co type.
    return np.isscalar(x)
    
```

Getting all the scalar type guards right might require a bit of experimentation. For instance, `_is_int_like` could be implemented either
- **additively**, i.e., based on `_is_bool_like` and adding checks for `int` and `np.integer` (similar to how the types themselves are defined).
- **subtractively**, i.e, based on `np.isscalar`, but since `np.isscalar` is `True` for floats as well, one would have to "subtract" the floating point types to get a pure integer check.

Once we have such functions we could replace many `instance` or `np.isscalar` by them.

For this PR I wanted to keep the diff minimal, but if you're interested in adding such type guards I could look into it.

bmcfee (approved):

bmcfee:
> If supporting `List` as input is on purpose, we probably should broaden the type explicitly to communicate that and to avoid dropping support for `List` accidentally. I have simply added lists now to both the type and the check.

:+1:



> For this PR I wanted to keep the diff minimal, but if you're interested in adding such type guards I could look into it.

Agreed, that would be a much bigger undertaking than makes sense for this PR, but it's worth thinking over.

Meanwhile, everything else looks good here.  Something in the last update is fouling up the lint CI though.

bluenote10:
> Something in the last update is fouling up the lint CI though.

Any ideas what is happening there? Looks like it is failing in the _Install conda environment_ action, substep _Adding tools to 'base' env"_ and errors with:

```
  PackagesNotFoundError: The following packages are missing from the target environment:
    - defaults/linux-64::conda==22.11.1=py310h06a4308_4[md5=07e2e8a698cf1e7f580b9efcae72184e]
```

Could this be an unrelated issue?

bmcfee:
Ok, the issue turned out to be with using mamba instead of conda in the lint environment setup.  I've updated on main - if you rebase, it should go through.

bmcfee:
@bluenote10 thanks again for this one - anything else to add, or shall I merge?

bluenote10:
Can be merged, thanks again for looking into the CI issue!

PR #1643, opened 2023-01-07
Zero crossings rewrite

### LINKED ISSUE #1635: Drop support for threshold=None in zero_crossings
This came up in implementing #1632 - the threshold parameter in zero_crossings is currently optional, but it doesn't need to be.  If threshold is None, it is converted to 0 automatically.

Threshold=0 is equivalent behavior, and dropping the None support would simplify some type annotations and a bit of the implementation.

bmcfee:
#### Reference Issue
Fixes #1635 

This PR removes support for `threshold=None` in zero crossings, since the behavior was equivalent to `threshold=0`

#### Any other comments?

The zero crossings function has been rewritten to use a numba stencil.  The result is numerically equivalent to the old implementation, but avoids extraneous copying and slicing.

I think this should be ready to merge as long as CI passes.  Some CR would be nice, but not strictly necessary IMO.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -14,7 +14,7 @@ import scipy.signal
 import soxr
 import lazy_loader as lazy
 
-from numba import jit
+from numba import jit, stencil, guvectorize
 from .fft import get_fftlib
 from .convert import frames_to_samples, time_to_samples
 from .._cache import cache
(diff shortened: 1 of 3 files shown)


bmcfee:
Coverage drops are due to numba, but all paths are indeed covered.

PR #1646, opened 2023-01-17
Vertical waveshow

### LINKED ISSUE #1496: Vertical waveshow

Branching this issue out from a comment in #1370 and some offline discussions.

TLDR: it could be useful if `waveshow` could draw transposed waveforms, rather than only plotting with time as the horizontal axis.  The main use-case here is probably in things like cross-similarity plots, where it might be nice to have subplots on the top and left sides corresponding to the signals being compared.

The API could look like `waveshow(..., orient='vertical')`, though we'd probably need to consider revising other parameter names like `offset=` and `x_axis=` to be more flexible and align with what's provided by `specshow`.

We could achieve some of this by an axis transformation, but given the headaches we've already seen in melodyplot with this (fill_between does not currently play nicely with axes transforms), I'd rather not.

bmcfee:
#### Reference Issue
Fixes #1496 

This PR extends the `waveshow` function to support transposed wave display.

A new boolean flag is added `transpose=[True|False]` which controls the display direction.  The `x_axis` parameter has been renamed to `axis` with a deprecation cycle expiring at 1.0.  The `AdaptiveWaveplot` class has been similarly extended.

Fancy example:

```python
chroma = librosa.feature.chroma_cqt(y=y, sr=sr)
sim = librosa.segment.recurrence_matrix(chroma)

fig, ax = plt.subplot_mosaic(
    """
    hSS
    hSS
    .vv
    """)

librosa.display.specshow(sim, ax=ax['S'], x_axis='time', y_axis='time', auto_aspect=False)
ax['S'].label_outer()
ax['S'].sharex(ax['v'])
ax['S'].sharey(ax['h'])

librosa.display.waveshow(y, ax=ax['v'])
ax['v'].label_outer()

librosa.display.waveshow(y, ax=ax['h'], transpose=True)
ax['h'].label_outer()
```
Result:
image

#### Any other comments?

I think this should be good to go.


bmcfee:
Note, I wound up going with the boolean `transpose=` parameter here instead of the `orient='h'|'v'` idea mentioned in #1496.  It just seemed simpler, and I can't imagine ever needing to extend this beyond two options.

oriolcolomefont:
The following ERROR raises when running the sample code provided [here](https://github.com/librosa/librosa/pull/1646#issue-1536896353).

Captura de pantalla 2023-01-26 a les 15 34 23

However, when I checked the documentation, the attribute appeared. Could you please shed some light here?

Thanks!!

bmcfee:
Your environment is likely not up to the minimum specifications:

https://github.com/librosa/librosa/blob/71077174b9e73ae81d268f81551bb9667bf3693b/setup.cfg#L42-L92

You need matplotlib >= 3.3.

oriolcolomefont:
Yes! Thank you @bmcfee !

oriolcolomefont:
When running your code [here](https://github.com/librosa/librosa/pull/1646#issue-1536896353)

`librosa.display.waveshow(y, ax=ax['h'], transpose=True)
ax['h'].label_outer()`

The following error raises: `AttributeError: Line2D.set() got an unexpected keyword argument 'transpose'`

Do you know if this argument `transpose` to plot vertically is no longer valid? Thanks!

bmcfee:
It looks like you're using the release version of librosa (0.9.2).  This feature was only recently merged in development and has not yet been released.

oriolcolomefont:
I am indeed using the release version of librosa (0.9.2). Looking forward to the release! Thank for everything!!!

PR #1647, opened 2023-01-20
f0 harmonic interpolation

### LINKED ISSUE #1500: Tempogram ratio features

This was alluded to in #1426 , but it would be handy to finally provide an implementation of the tempogram ratio feature from (Peeters, 2005).  It would look something like the following (bottom subplot):

image

- Peeters, Geoffroy. "Rhythm Classification Using Spectral Rhythm Patterns." ISMIR. 2005.

The basic idea is to take a tempogram, extract a (time-varying) tempo estimate (corresponding to quarter-notes), and then use harmonic interpolation to measure tempogram energy for each frame at all musically important durations.  The benefit of this over a raw tempogram is that it could locally normalize for tempo variation.

The underlying algorithm is somewhat similar to our `interp_harmonics` function, except that we want to pull out a different subset of frequencies for each frame.  I imagine the implementation would use a vectorized interpolator in a similar fashion to what we do for reassigned spectrogram harmonics:

https://github.com/librosa/librosa/blob/7ac022a8496126e95b46a7d56ad880a328359bda/librosa/core/harmonic.py#L253-L265

but of course the details will be slightly different.

We could also support having a single, global tempo (much simpler), as well as aggregation over frames.

bmcfee:
#### Reference Issue
Fixes #1500 

This PR adds the f0 harmonic interpolation method described in #1500.

It's a work in progress, but the core functionality is there.  I've informally tested the following cases:

- single channel, static frequency grid (`librosa.fft_frequencies()`)
- single channel, dynamic frequency grid (`librosa.reassigned_spectrogram`)
- multi-channel, static frequency grid
- multi-channel, dynamic frequency grid

It all seems to work to spec, though I still need to work up some unit tests.  I'm opening the PR now just to get some early feedback on the API and maybe some eyeballs on the implementation.  It's possible that we could simplify the code a bit by factoring out the interpolation helper and just changing how it gets vectorized, but I'm not expert enough with `np.vectorize` to see how to do that immediately.

Once the core is stable, I'll cook up a rhythm descriptor feature following Peeters'05.  We can also add in the implementation of Rafii'21's timbre descriptor, though I think this might work better as an advanced example.  (This is because there are some finnicky details in f0 estimation that make me hesitant to wire in as a library function.)

#### Any other comments?

TODO:

- ~~[ ] support for scalar (static) f0 (useful in tempogram ratios)~~
- [x] unit tests
- [x] multichannel integration tests
- [x] docstring examples
- [x] tempogram ratio feature extractor
- [x] timbre extractor demo
- [x] tempogram ratio tests

### DIFF
--- a/docs/examples/plot_spectral_harmonics.py
+++ b/docs/examples/plot_spectral_harmonics.py
@@ -1,21 +1,21 @@
 # coding: utf-8
 """
-==================
-Spectral harmonics
-==================
+=================
+Harmonic spectrum
+=================
 
-This notebook demonstrates how to extract spectral harmonics from an audio signal.
+This notebook demonstrates how to extract the harmonic spectrum from an audio signal.
 The basic idea is to estimate the fundamental frequency (f0) at each time step,
-and extract the energy at integer multiples of f0 (*harmonics*).
-This representation can be used to compactly encode timbral content, either for
-resynthesis [1]_ or downstream analysis [2]_.
+and extract the energy at integer multiples of f0 (the *harmonics*).
+This representation can be used to represent the short-term evolution of timbre,
+either for resynthesis [1]_ or downstream analysis [2]_.
 
 .. [1] Bonada, Jordi, X. Serra, X. Amatriain, and A. Loscos.
     "Spectral processing."
     DAFX Digital Audio Effects (2011): 393-444.
 
 .. [2] Rafii, Zafar.
-    "The Constant-Q Harmonic Coefficients: A timbre feature designed for music signals [Lecture Notes]."
+    "The Constant-Q Harmonic Coefficients: A timbre feature designed for music signals."
     IEEE Signal Processing Magazine 39, no. 3 (2022): 90-96.
 """
 
@@ -48,7 +48,17 @@ Audio(data=y, rate=sr)
 #
 # We'll constrain `f0` to lie within the range 50 Hz
 # to 300 Hz.
-
+#
+# pyin returns three arrays:
+#   - `f0`, the sequence of fundamental frequency estimates
+#   - `voicing`, the sequence of indicator variables for whether
+#     a fundamental was detected or not at each time step
+#   - `voicing_probability`, the sequence of probabilities that each
+#     time step contains a fundamental frequency
+#
+# For this application, we'll only be using `f0`.  Note that by default,
+# pyin will set `f0[n] = np.nan` (not a number) whenever `voicing[n] == False`.
+# We'll handle this situation later on when resynthesizing the signal.
 f0, voicing, voicing_probability = librosa.pyin(y=y, sr=sr, fmin=50, fmax=300)
 
 ####################################################################
@@ -126,27 +136,26 @@ ax[1].set(ylabel='Harmonics')
 # nans with a frequency of 0.
 f0_synth = np.nan_to_num(f0)
 
-yout = np.zeros_like(y)
+y_out = np.zeros_like(y)
 
 for i, (factor, energy) in enumerate(zip(harmonics, harmonic_energy)):
     # Mix in a synthesized pitch contour
-    yout = yout + mir_eval.sonify.pitch_contour(times, f0_synth * factor,
+    y_out = y_out + mir_eval.sonify.pitch_contour(times, f0_synth * factor,
                                                 amplitudes=energy,
                                                 fs=sr,
                                                 length=len(y))
 
-Audio(data=yout, rate=sr)
+Audio(data=y_out, rate=sr)
 
 
-###################################################
-# The synthesized audio is by no means a perfect
-# reconstruction of the input signal.  Notably,
-# it is derived from a sparse sinusoidal modeling
-# assumption, and will therefore not do well at
-# representing transients.  The result is still
-# largely intelligible, however, and the decoupling
-# of f0 from harmonic energy allows us to modify
-# the synthesized signal in various ways.
+#######################################################
+# The synthesized audio is not a perfect reconstruction
+# of the input signal.  Notably, it is derived from a
+# sparse sinusoidal modeling assumption, and will
+# therefore not do well at representing transients.
+# The result is still largely intelligible, however,
+# and the decoupling of f0 from harmonic energy allows
+# us to modify the synthesized signal in various ways.
 #
 # For example, we can synthesize the same utterance
 # with a constant f0 to produce a monotone effect.
--- a/docs/examples/plot_spectral_harmonics.py
+++ b/docs/examples/plot_spectral_harmonics.py
@@ -7,8 +7,8 @@ Spectral harmonics
 This notebook demonstrates how to extract spectral harmonics from an audio signal.
 The basic idea is to estimate the fundamental frequency (f0) at each time step,
 and extract the energy at integer multiples of f0 (*harmonics*).
-This representation can be used to compactly encode timbral content, either for resynthesis [1]_ or
-downstream analysis [2]_.
+This representation can be used to compactly encode timbral content, either for
+resynthesis [1]_ or downstream analysis [2]_.
 
 .. [1] Bonada, Jordi, X. Serra, X. Amatriain, and A. Loscos.
     "Spectral processing."
--- a/docs/examples/plot_spectral_harmonics.py
+++ b/docs/examples/plot_spectral_harmonics.py
@@ -1,0 +1,166 @@
+# coding: utf-8
+"""
+==================
+Spectral harmonics
+==================
+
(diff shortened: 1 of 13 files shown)


bmcfee:
Thinking through some package organization business relating to tempogram ratio features:

- `librosa.feature.tempogram_ratio` makes sense as a place for this function to live.  TGR is a feature descriptor.
- TGR depends on tempo estimation.  This lives in the `beat` submodule, `librosa.beat.tempo`
- tempo estimation itself depends on tempogram calculation, which is back in `feature`.  This introduces a circular dependency.

I think the reasonable solution here is to deprecation-move `beat.tempo` to `feature.tempo` to eliminate the circular dependency.  This is only a little awkward, since `tempo` could be considered both a "feature" and a specific measurable outcome (like f0).  After a quick skim of the package, I don't think this move would cause any significant trouble beyond some cosmetic rewriting of demo code.

If this sounds good to folks, I'll go ahead with the move.  If not, I'd like to hear some counter-arguments or alternative proposals.

bmcfee:
I'm dropping the requirement of scalar f0 inputs.  For TGR features, we can force it to be time-varying up front, and it will simplify the code substantially to remove that feature.

I've gone ahead and prototyped out the TGR feature with a refactor to move `tempo` into the `feature.rhythm` submodule.  Everything appears to be working here.  A couple of caveats:

1. The ratios employed by Peeters'05 seemed a little strange to me.  I've instead implemented the default ratio collection following Prockup'15, which essentially looks at standard note durations (whole, half, quarter, eighth, sixteenth) along with dots and triplets derived from each.  Whole notes do not get a dot, and sixteenth notes do not get a triplet, leaving 13 ratios by default.
2. Typesetting the docstring here is nightmarish.  Ideally, we could use standard notation in unicode to describe each of these durations.  I'm running into trouble getting this to render properly over web interfaces (sphinx, jupyter) due to combining symbols for dots and triplets.  Actually, even bare notes like 𝅘𝅥  seem to cause trouble.  The workaround that I've put in place so far is to leave trailing white space, and that renders okay (not perfectly):
 
image

The problem here is that velin now complains about trailing whitespace in the docstrings.  I could maybe hack around that by formatting them to live inside a table, but it feels hacky.

bmcfee:
Went a little overboard / silly with the advanced example gallery.  But hey, now we have a clunky autotune with relatively few lines of code...

bmcfee:
Ok, I think this one is ready for review.

bmcfee:
I gave this a read over on my own, and cleaned up a couple of small things.  I think the tests and implementation are fine here, but I'd like some high-level feedback on the documentation (docstrings + example gallery) if folks have any.  Otherwise I'll try to merge this by end of week.

docs/examples/plot_spectral_harmonics.py
```diff
@@ -0,0 +1,166 @@
...
+    yout = yout + mir_eval.sonify.pitch_contour(times, f0_synth * factor,
+                                                amplitudes=energy,
+                                                fs=sr,
+                                                length=len(y))
+
+Audio(data=yout, rate=sr)
+
+
+###################################################
+# The synthesized audio is by no means a perfect
```
lostanlen, line 142: by no means -> not

docs/examples/plot_spectral_harmonics.py
```diff
@@ -0,0 +1,166 @@
+# coding: utf-8
+"""
+==================
+Spectral harmonics
```
lostanlen, line 4: "spectral harmonics": 8k hits on Google
"harmonic spectrum": 192k hits on Google

docs/examples/plot_spectral_harmonics.py
```diff
@@ -0,0 +1,166 @@
...
+and extract the energy at integer multiples of f0 (*harmonics*).
+This representation can be used to compactly encode timbral content, either for
+resynthesis [1]_ or downstream analysis [2]_.
+
+.. [1] Bonada, Jordi, X. Serra, X. Amatriain, and A. Loscos.
+    "Spectral processing."
+    DAFX Digital Audio Effects (2011): 393-444.
+
+.. [2] Rafii, Zafar.
+    "The Constant-Q Harmonic Coefficients: A timbre feature designed for music signals [Lecture Notes]."
```
lostanlen, line 18: [Lecture notes] ?

docs/examples/plot_spectral_harmonics.py
```diff
@@ -0,0 +1,166 @@
+# coding: utf-8
+"""
+==================
+Spectral harmonics
+==================
+
+This notebook demonstrates how to extract spectral harmonics from an audio signal.
+The basic idea is to estimate the fundamental frequency (f0) at each time step,
+and extract the energy at integer multiples of f0 (*harmonics*).
+This representation can be used to compactly encode timbral content, either for
```
lostanlen, line 10: can be used to describe the short-term evolution of timbre

docs/examples/plot_spectral_harmonics.py
```diff
@@ -0,0 +1,166 @@
...
+Audio(data=y, rate=sr)
+
+#####################################################
+# Next, we'll estimate the fundamental frequency (f0)
+# of the voice using `librosa.pyin`.
+#
+# We'll constrain `f0` to lie within the range 50 Hz
+# to 300 Hz.
+
+f0, voicing, voicing_probability = librosa.pyin(y=y, sr=sr, fmin=50, fmax=300)
```
lostanlen, line 52: Consider explaining what `voicing_probabilty` is

docs/examples/plot_spectral_harmonics.py
```diff
@@ -0,0 +1,166 @@
...
+# sine waves at the chosen set of frequencies and scaled
+# appropriately by the amount of energy at that frequency
+# over time.
+#
+# Since the f0 contour is a time-varying frequency measurement,
+# the synthesis will need to support variable frequencies.
+# Luckily, the `mir_eval.sonify` module does exactly this.
+# We'll generate a separate signal for each harmonic separately,
+# and add them into a mixture to better approximate the original
+# signal.
```
lostanlen, line 122: not sure why this is on its own line

docs/examples/plot_spectral_harmonics.py
lostanlen: especially since you're going to do `nan_to_num` later

docs/examples/plot_spectral_harmonics.py
bmcfee: It doesn't need to be, but it doesn't hurt anything either.  It's kind of a quirk of how sphinx-gallery infers notebook / markdown from python comments, using `###...` barriers.
It looks best if the text under the barrier does not overflow, but the barrier length itself is arbitrary.

docs/examples/plot_spectral_harmonics.py
bmcfee: Yeah, fair enough :joy:

bmcfee: Though I would like to avoid too much potential for confusion between this and the harmonic *component* of the spectrogram via hpss.

docs/examples/plot_spectral_harmonics.py
lostanlen: i understand! ok

lostanlen:
Thanks. I approve the docs changes. 

Some tests are failing now? Seems like an issue with ops not dev

```
[2023-02-01T15:37:18.272Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
[2023-02-01T15:37:18.273Z] ['verbose'] The error stack is: Error: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
    at main (/snapshot/repo/dist/src/index.js)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
[2023-02-01T15:37:18.273Z] ['verbose'] End of uploader: 454 milliseconds
Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255
```

bmcfee:
> Some tests are failing now? Seems like an issue with ops not dev

The tests pass, but codecov is failing due to API rate limiting.  It's a known and annoying issue, but it shouldn't block us.

lostanlen (approved):
Let's go

PR #1652, opened 2023-02-03
Allow isolated vertices in bandwidth estimation sometimes

### LINKED ISSUE #1651: Regression: bandwidth estimation now requires all vertices to have edges
**Describe the bug**

In 0.9 and earlier, the bandwidth for `affinity` mode in recurrence matrix calculation was inferred as the median distance to the k'th nearest neighbor of each data point:
https://github.com/librosa/librosa/blob/b9830b4a335b1e7e4be66ee81870573bf3d516d0/librosa/segment.py#L524-L526

Because we were using `nanmedian(rec.max(axis=1))`, any empty rows of `rec` would be implicitly ignored in the median calculation.  This, I believe, is a good and correct thing.

After merging #1495 , the behavior has changed:
https://github.com/librosa/librosa/blob/e18d544cbebfb590da34710ba2de86292179ec39/librosa/segment.py#L1357-L1367

we now explicitly construct an array of the k'th neighbor distance for each point, and then derive bandwidths later on.  If a point does not have any neighbors, an exception is raised.  This is correct and necessary if we're doing individual or pairwise bandwidth estimates (eg `mean_k` or `gmean_k`), but I think the requirement is too strict if we're doing global bandwidth estimation.

**To Reproduce**

Behavior is triggered by our segmentation example in the advanced gallery, resulting in the following:

```
Unexpected failing examples:
/home/bmcfee/git/librosa/docs/examples/plot_segmentation.py failed leaving traceback:
  File "/home/bmcfee/git/librosa/docs/examples/plot_segmentation.py", line 77, in <module>
    R = librosa.segment.recurrence_matrix(Csync, width=3, mode='affinity',
  File "/home/bmcfee/git/librosa/librosa/segment.py", line 685, in recurrence_matrix
    aff_bandwidth = __affinity_bandwidth(rec, bandwidth, bandwidth_k)
  File "/home/bmcfee/git/librosa/librosa/segment.py", line 1363, in __affinity_bandwidth
    raise ParameterError("The sample at time point {} has no neighbors".format(i))
librosa.util.exceptions.ParameterError: The sample at time point 0 has no neighbors
```

**Expected behavior**

I think that if we're estimating scalar bandwidths, we can support empty rows in the recurrence matrix.  In this case, we can fill in `nan` if the link set is empty, and then use nan-friendly estimators if we're producing a scalar bandwidth.

Does this sound reasonable?  If so, I'll make a quick PR for it.

bmcfee:
#### Reference Issue
Fixes #1651 

This PR fixes the regression for supporting scalar bandwidth estimation in graphs with isolated vertices.

#### Any other comments?

I've added a couple more tests to cover the bases.  I'd like @tomxi's eyeballs on this quickly, but I think it's probably good to go.

### DIFF
--- a/librosa/segment.py
+++ b/librosa/segment.py
@@ -289,7 +289,7 @@ def cross_similarity(
 
     # using k for bandwidth estimation also and decouple k for full mode
     bandwidth_k = k
-    if full and (mode != 'connectivity'):
+    if full and (mode != "connectivity"):
         k = n
 
     # Build the neighbor search object
@@ -614,7 +614,7 @@ def recurrence_matrix(
 
     # using k for bandwidth estimation also and decouple k for full mode
     bandwidth_k = k
-    if full and (mode != 'connectivity'):
+    if full and (mode != "connectivity"):
         k = t
 
     # Build the neighbor search object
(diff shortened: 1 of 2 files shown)


tomxi:
Thanks for fixing my styles! all looks good to me

PR #1677, opened 2023-02-23
Vqt octave boundaries

### LINKED ISSUE #1676: VQT wavelet filters with more graceful octave boundaries

In https://github.com/librosa/librosa/issues/1073#issuecomment-968752759 , we sorted out how to handle bandwidth calculation for wavelet filters at the boundaries of the frequency range.  This resulted in the following chunk of code:

https://github.com/librosa/librosa/blob/1f770e8fa778a548111204192757043d0313a010/librosa/filters.py#L797-L804

where we estimate the "local bins per octave" in the neighborhood of the center frequency for each filter.  For "interior frequencies" (i.e. not fmin or fmax), we use one above and one below to do this; for the boundary frequencies, we just use one above (fmin) or one below (fmax).

This is fine if the frequencies span the entire range we care about, but we also often generate filters one octave at a time.  In this case, we'll get slightly different results between the full-range and octave-range calculations when the frequency spacing is uneven.

We could refine this a little by allowing the length calculator to assume octave equivalence above or below (or both) the frequency range.  That is, instead of

```python
bpo[0] = 1 / (logf[1] - logf[0]) 
```
for the BPO estimate at `f[0]`, we could do
```python
bpo[0] = 2 / (logf[1] - (logf[-1] - 1))  # divide fmax/2 to get one freq down from f[0]
```
and likewise for the top frequency:
```python
bpo[0] = 1 / (logf[1] - logf[0]) 
```
becomes
```python
bpo[-1] = 2 / (logf[0] + 1 - logf[-2]) 
```

These behaviors should be controlled by flags to enable/disable octave equivalence on each side.
The result should ensure consistent behavior between full-range and octave-range frequency generation, except at the bottom of the lowest or top of the highest octaves.

Note: when using equal spacing (ie CQT or VQT with intervals='equal'), none of this should matter, and the behavior should be identical in all cases.

:shrug: 

**Additional context**
Since JIT is experimental right now, I think we can tweak this behavior in a minor without jumping revisions.

bmcfee:
#### Reference Issue
Fixes #1676 

This PR implements the changes described in #1676 to produce consistent behavior when constructing non-uniform wavelet bases for VQT.

There should be no change for uniformly spaced frequencies.

The following figure demonstrates the change in behavior (filter length) for a 36-BPO ji5 frequency set:
image

where the solid line shows the correct lengths calculated from the entire frequency set in one shot, and the dashed line shows what happens when each octave is calculated separately and using reflection at the boundaries.

#### Any other comments?

This PR also bumps the version to 0.10.1dev

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -309,8 +309,11 @@ def hybrid_cqt(
     # Get all CQT frequencies
     freqs = cqt_frequencies(n_bins, fmin=fmin, bins_per_octave=bins_per_octave)
 
-    # Compute an alpha parameter, just in case we need it
-    alpha = __bpo_to_alpha(bins_per_octave)
+    # Pre-compute alpha
+    if n_bins == 1:
+        alpha = __et_relative_bw(bins_per_octave)
+    else:
+        alpha = filters._relative_bandwidth(freqs=freqs)
 
     # Compute the length of each constant-Q basis function
     lengths, _ = filters.wavelet_lengths(
@@ -486,7 +489,10 @@ def pseudo_cqt(
 
     freqs = cqt_frequencies(fmin=fmin, n_bins=n_bins, bins_per_octave=bins_per_octave)
 
-    alpha = __bpo_to_alpha(bins_per_octave)
+    if n_bins == 1:
+        alpha = __et_relative_bw(bins_per_octave)
+    else:
+        alpha = filters._relative_bandwidth(freqs=freqs)
 
     lengths, _ = filters.wavelet_lengths(
         freqs=freqs, sr=sr, window=window, filter_scale=filter_scale, alpha=alpha
@@ -656,7 +662,10 @@ def icqt(
 
     # truncate the cqt to max frames if helpful
     freqs = cqt_frequencies(fmin=fmin, n_bins=n_bins, bins_per_octave=bins_per_octave)
-    alpha = __bpo_to_alpha(bins_per_octave)
+    if n_bins == 1:
+        alpha = __et_relative_bw(bins_per_octave)
+    else:
+        alpha = filters._relative_bandwidth(freqs=freqs)
 
     lengths, f_cutoff = filters.wavelet_lengths(
         freqs=freqs, sr=sr, window=window, filter_scale=filter_scale, alpha=alpha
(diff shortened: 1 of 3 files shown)


bmcfee:
There's a small coverage drop here due to not covering `pseudo_cqt` and `icqt` cases where `n_bins=1`.  I'm not too worried about it.

librosa/filters.py
```diff
@@ -825,6 +815,40 @@ def wavelet_lengths(
     return lengths, f_cutoff
 
 
+def _relative_bandwidth(*, freqs: np.ndarray) -> np.ndarray:
```
lostanlen, line 818: plural, maybe? `_relative_bandwidths`?

librosa/core/constantq.py
```diff
@@ -656,7 +662,10 @@ def icqt(
 
     # truncate the cqt to max frames if helpful
```
lostanlen, line 663: `# truncate the cqt to max frames if helpful` ? 🤔

librosa/core/constantq.py
```diff
@@ -1478,17 +1491,24 @@ def griffinlim_cqt(
     )
 
 
-def __bpo_to_alpha(bins_per_octave: int) -> float:
-    """Compute the alpha coefficient for a given number of bins per octave
+def __et_relative_bw(bins_per_octave: int) -> np.ndarray:
```
lostanlen, line 1494: not sure why `_relative_bandwidth` and `__et_relative_bw` are not held to the same privacy status here (one vs. two leading underscores)

bmcfee: It's only because relative_bandwidth may be called from a separate submodule but et_bandwidth can't.

librosa/core/constantq.py
bmcfee: Looks like a comment that got misplaced after inserting some code by copy paste..

librosa/filters.py
bmcfee: We could, though I don't think it matters much as a private function.  The meaning seems clear enough either way, assuming we're okay with the "relative bandwidth" terminology?

lostanlen: it's just that the input (`freqs`) is plural and we can't make the output name plural (`alpha`) since it's sometime singular. so pluralizing the method name would be helpful to understand that this is a many-to-many operation

PR #1703, opened 2023-05-04
fixed bins_per_octave's docs type float to int

kyaryunha:

I just edited one line.
The docs type of bin_per_octave of the pitch_shift function and the actual type do not match.

#### Any other comments?

Thanks for the nice library.

### DIFF
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -282,7 +282,7 @@ def pitch_shift(
     n_steps : float [scalar]
         how many (fractional) steps to shift ``y``
 
-    bins_per_octave : float > 0 [scalar]
+    bins_per_octave : int > 0 [scalar]
         how many steps per octave
 
     res_type : string


kyaryunha:
Or is it not that the docs should change, but that the type hinting in the code should change?

```python
# librosa/effects.py pitch_shift
if not util.is_positive_int(bins_per_octave):
    raise ParameterError(
        f"bins_per_octave={bins_per_octave} must be a positive integer."
    )
```

As it is written in the code, the intention of the code is like int,

This is the only place where bin_per_octave is used, so as long as it is positive, it can be float, too.

```python
# librosa/effects.py pitch_shift
rate = 2.0 ** (-float(n_steps) / bins_per_octave)
```

On the other hand, I'm just curious, I only use the value of 12, but I'm curious about examples of when other numbers are used.

bmcfee:
Thanks for catching this.  PR looks good to me, so I'm happy to merge once CI passes.

> Or is it not that the docs should change, but that the type hinting in the code should change?

I think the type hints are correct, at least for how the code presently works.

We've only ever considered integer bins-per-octave, as this originally came from chroma and CQT implementations, where equal temperament and octave equivalence implies a whole number of divisions for each octave.  This matters in pitch_shift because it's parametrized in terms of semitones, so we need to know how to map semitones onto a frequency ratio.

I suppose it's possible to have a non-integer number of bins per octave, eg I guess the spiral of perfect fifths would properly be non-integral since `3**12 / 2**18 ~ 2.0273 ≠ 2`.  But in this case, the size of a semitone is non-uniform anyway, so the API would break down regardless.

We provide the option to set bins_per_octave to things other than 12 primarily so that people can easily do things like "shift up by one quartertone" (bpo=24), or work naturally in other equal temperament systems.

lostanlen (approved):

kyaryunha:
Thanks for explaining why! 

But there's something I didn't understand in your answer.

`But in this case, the size of a semitone is non-uniform anyway, so the API would break down regardless.`

I don't understand where it's broken...
Could you please explain more?

If it's broken when it's an integer, should I fix the PR by fixing the type hint to int to float, not the docs to float to int?

lostanlen:
@kyaryunha i think your current PR is good. our API assumes an integer number of bins per octave

bmcfee:
> But there's something I didn't understand in your answer.
> 
> `But in this case, the size of a semitone is non-uniform anyway, so the API would break down regardless.`
> 
> I don't understand where it's broken... Could you please explain more?
> 
> If it's broken when it's an integer, should I fix the PR by fixing the type hint to int to float, not the docs to float to int?

First, I agree with @lostanlen - this PR is fine and does not need further changes.  Integer is correct.

With the point above, I was referring specifically to situations beyond the scope of the present API where the number of bins per octave could be fractional.

Issue #1700, opened 2023-04-20
time_stretch does not return the same signal when rate=1


zafarrafii:
**Describe the bug**
time_stretch does not return the same signal when rate=1.

**To Reproduce**
```
import librosa
import matplotlib.pyplot as plt

audio_signal, _ = librosa.load(librosa.ex('fishin'))
audio_signal_stretched = librosa.effects.time_stretch(audio_signal, rate=1)

plt.plot(audio_signal, label='original')
plt.plot(audio_signal_stretched, label='stretched')
plt.plot(audio_signal_stretched-audio_signal, label='difference')
plt.legend(loc='upper left')
plt.show()
```
![image](https://user-images.githubusercontent.com/22797461/233508393-60dd85aa-24d1-4339-b1f7-3fad2a601ff5.png)

**Expected behavior**
time_stretch should return the same signal when rate=1.

**Software versions**
```
Linux-5.4.0-146-generic-x86_64-with-debian-buster-sid
Python 3.7.13 (default, Mar 29 2022, 02:18:16) 
[GCC 7.5.0]
NumPy 1.18.5
SciPy 1.4.1
librosa 0.9.1
INSTALLED VERSIONS
------------------
python: 3.7.13 (default, Mar 29 2022, 02:18:16) 
[GCC 7.5.0]

librosa: 0.9.1

audioread: 2.1.9
numpy: 1.18.5
scipy: 1.4.1
sklearn: 1.0.2
joblib: 1.1.0
decorator: 5.1.1
soundfile: 0.9.0
resampy: 0.2.2
numba: 0.55.1

numpydoc: 1.4.0
sphinx: 5.0.2
sphinx_rtd_theme: None
sphinxcontrib.versioning: None
sphinx-gallery: None
pytest: 7.1.2
pytest-mpl: None
pytest-cov: None
matplotlib: 3.5.1
presets: None
```

Perhaps a simple fix would be to add an if statement which would return the input if rate=1, just like in the time_stretch function of the rubberband Python wrapper: https://pyrubberband.readthedocs.io/en/stable/generated/pyrubberband.pyrb.time_stretch.html Thanks for that wrapper by the way, it works great! I also noticed that some values go out of the [-1, 1] range, but this seems to happen when resampling the audio..

bmcfee:
Thanks for catching that.  It would be an easy enough flag to add, either in the effects wrapper functions or in the phase_vocoder itself.  (Or maybe both?)

[...]

PR #1704, opened 2023-05-04
add volume adjustment(rms-mean) at pitch_shift

kyaryunha:
#### Reference Issue
It's simple, so I left an opinion as a pull request without any issues.

If the volume is changed due to pitch_shift, the average value of the volume can be adjusted.

For compatibility, the default value is None,
Since the volume control method can be various other than rms average, it can be received as a string like res_type.

#### Any other comments?
Thanks for reading and adding if you like!

Or if there is a better way to recalculate volume, please let me know!
I'm still learning about audio, so I don't know much about it.

### DIFF
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -40,6 +40,7 @@ from . import core
 from . import decompose
 from . import feature
 from . import util
+from .feature import rms
 from .util.exceptions import ParameterError
 from typing import Any, Callable, Iterable, Optional, Tuple, Union, overload
 from typing_extensions import Literal
@@ -265,6 +266,7 @@ def pitch_shift(
     n_steps: float,
     bins_per_octave: int = 12,
     res_type: str = "soxr_hq",
+    volume: Union[str, None] = None,
     **kwargs: Any,
 ) -> np.ndarray:
     """Shift the pitch of a waveform by ``n_steps`` steps.
@@ -289,6 +291,10 @@ def pitch_shift(
         Resample type. By default, 'soxr_hq' is used.
 
         See `librosa.resample` for more information.
+
+    volume: string or None
+        Volume adjustment. By default, None is used.
+        Current List - [rms-mean]
 
     **kwargs : additional keyword arguments.
         See `librosa.decompose.stft` for details.
@@ -338,6 +344,11 @@ def pitch_shift(
         target_sr=sr,
         res_type=res_type,
     )
+
+    if volume == 'rms-mean':
+        original_rms_volume = np.mean(rms(y=y))
+        shifted_rms_volume = np.mean(rms(y=y_shift))
+        y_shift = y_shift * (original_rms_volume / shifted_rms_volume)
 
     # Crop to the same dimension as the input
     return util.fix_length(y_shift, size=y.shape[-1])


bmcfee:
Thanks for this.  I'm not sure this is actually the best way to implement this functionality, but I agree that it's better if we can preserve energy through pitch shifts.

There are two fundamental operations in our pitch shifter: 1) time stretching (phase vocoding), and 2) resampling.

Both of these steps can affect signal energy independently, and I'd rather the control over energy preservation be integrated at each stage rather than as a post-processing cleanup.

Resample in fact already has this as an option, but it is disabled by default (`scale=False`), and the `pitch_shift` effect uses the default.  It would be easier to just pass `scale=True` into the resample call:

https://github.com/librosa/librosa/blob/aacb4c134002903ae56bbd4b4a330519a5abacc0/librosa/effects.py#L335-L340

Time-stretching is a bit more subtle, see the recent discussion in #1700 - but most of the time, resampling should be the primary source of energy scaling issues.

I'd be okay with changing the pitch_shift effect function to use `scale=True`, and perhaps exposing this as a pass-through parameter so that users can recover the old behavior (`scale=False`) if necessary.

kyaryunha:
Thanks for your reply!

I modified it by putting scale in resample.

The number below is the value obtained by outputting the average volume of audio using get_volume in the code below.

<table>
<tr>
<td>Original</td>
<td>0.083548285</td>
</tr>
<tr>
<td>resample(scale=False)</td>
<td>0.0576097</td>
</tr>
<tr>
<td>resample(scale=True)</td>
<td>0.06282382</td>
</tr>
<tr>
<td>just multiplying the audio(my pull request)</td>
<td>0.08354831</td>
</tr>
</table>

resample scale=True is still down by 0.0207.


Is get_volume wrong?
(The get_volume function is the code I wrote to get the average volume. So it could be wrong. But get_volume is only np.mean using rms.)

I think the get_volume I wrote to get the average volume of the audio might be wrong. Because to my ears they all seem the same (...)

I am curious about the correct way to find the average volume of audio.


My playground code. 
```python
from typing import Any
import numpy as np
import soundfile
from librosa import core, util
from librosa.effects import time_stretch
from librosa.feature import rms
from pydub import AudioSegment
import librosa


def get_volume(y):
     rms_volume = librosa.feature.rms(y=y)
     average_rms_volume = np.mean(rms_volume)
     return average_rms_volume


def custom_pitch_shift(
     y: np. ndarray,
     *,
     sr: float,
     n_steps: float,
     bins_per_octave: int = 12,
     res_type: str = "soxr_hq",
     # volume: Union[str, None] = None,
     scale: bool = False;
     **kwargs: Any,
) -> np.ndarray:
     rate = 2.0 ** (-float(n_steps) / bins_per_octave)

     # Stretch in time, then resample
     y_shift = core. resample(
         time_stretch(y, rate=rate, **kwargs),
         orig_sr=float(sr) / rate,
         target_sr=sr,
         res_type=res_type,
         scale=scale,
     )

     # if volume == 'rms-mean':
     # original_rms_volume = np.mean(rms(y=y))
     # shifted_rms_volume = np.mean(rms(y=y_shift))
     # y_shift = y_shift * (original_rms_volume / shifted_rms_volume)

     # Crop to the same dimension as the input
     return util.fix_length(y_shift, size=y.shape[-1])


def pitch_shift(file_name, output_filepath, semitones):
     audio = AudioSegment. from_file(file_name)
     audio.export(f"{output_filepath}-original.wav", format="wav")

     y, sr = librosa.load(f"{output_filepath}-original.wav", sr=None)
     original_volume = get_volume(y)
     print(original_volume)

     # pitch shift
     y_shifted = custom_pitch_shift(y, sr=sr, n_steps=semitones, res_type="kaiser_best", n_fft=2048, scale=True)
     print(get_volume(y_shifted))

     # shifted_volumn = get_volume(y_shifted)
     # y_shifted = y_shifted * (original_volume / shifted_volumn)

     soundfile.write(f"{output_filepath}_shifted.wav", y_shifted, sr)
     y2, sr2 = librosa.load(f"{output_filepath}_shifted.wav", sr=None)

     print(get_volume(y2))


if __name__ == "__main__":
     file_name = "./resources/i3wk3BTRgU0.mp3"
     semitones = 3
     pitch_shift(file_name, "transposed", semitones)
```

bmcfee:
> Is get_volume wrong?

No, I don't think it's wrong, but I would urge some caution in interpreting RMSE as "volume".  The two concepts are related, but not identical, and there are other ways that you could think of estimating loudness than average RMS value.  Eg, peak RMS instead of average, or total signal RMS, or better perceptually-motivated models like LUFS... there's lots that could be done here!

Since the scaling that you're proposing here can easily be implemented as a post-processing step, I'd prefer to leave it out here.  A user could always implement the kind of scaling that you're proposing without having to modify the internals of the effect functions.

I do, however, think it makes sense to revise the call to resample to include `scale=True`.  I'm curious for inputs from others - @lostanlen ?

lostanlen:
I agree with @bmcfee and will add that this RMS-based stage will cause subtle unwanted effects: the pitch shift of two concatenated sounds might not correspond to the concatenation of their putch shifts anymore

kyaryunha:
Thanks for the good methods. I will study the methods you told!
I changed to pass a scale argument to resample.

bmcfee (approved):

lostanlen (changes requested):
this PR doesn't do anything, it just adds new keyword arguments ... 🤔

kyaryunha:
Yes, just added a new argument to pass to resample!

What I want to do is, when I pitch_shift the music, I don't want the volume of the result to change.

I wanted to preserve loudness before and after pitch_shift.
PR initially proposed a method of maintaining the mean of rms_mean,
People said that it is better to modify the scale of resample, so I modified that part.

Currently, the code cannot modify the resample's scale when calling pitch_shift.

Or if know a better way about how the volume of the audio does not change significantly when use pitch shift, please let me know!!

lostanlen (approved):
Thanks for clarifying

PR #1718, opened 2023-06-06
CI for documentation builds

### LINKED ISSUE #1717: CI for documentation testing
The recent doc deployment failure #1715 was a result of our documentation build script incorrectly guessing the target of the `latest` symbolic link:

https://github.com/librosa/librosa/blob/edb8a86cf9dbf260a5c9237b83f5905214d42daf/.github/workflows/docs.yml#L66-L72

(corrected version here)

In the PR to fix the previous issue #1714 , I added some hooks to bypass the more computationally intensive parts of doc building for debugging purposes.  This should bring the doc build time down from 30ish minutes to something more reasonable for a CI workflow.  This means we could plausibly add some more robust testing of the documentation build before merge and deploy.

In doing so, we should also add a line to the workflow that verifies that no broken symbolic links exist.

While fixing this, we should also update our analytics tracking.

bmcfee:
#### Reference Issue
Fixes #1717 

This PR adds a debug doc build action to the PR workflow, as described in #1717 

#### Any other comments?

- [x] Add multi-version build
- [x] Add check for broken symbolic links
- [x] pydocstyle

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,19 +77,15 @@ following rules before submitting a pull request:
 You can also check for common programming errors with the following
 tools:
 
--  Code with good test coverage (at least 80%), check with:
+-  Code with good test coverage, check with:
 
           $ pytest
 
 -  No pyflakes warnings, check with:
 
-           $ python -m pip install pyflakes
-           $ pyflakes path/to/module.py
+           $ python -m pip install flake8
+           $ flake8 librosa
 
--  AutoPEP8 can help you fix some of the easy redundant errors:
-
-           $ python -m pip install autopep8
-           $ autopep8 path/to/pep8.py
 
 Some tests in tests/test_display.py use baseline images for output comparison.
 If existing images need to be updated or new ones should be added:
--- a/.github/environment-lint.yml
+++ b/.github/environment-lint.yml
@@ -12,6 +12,7 @@ dependencies:
   - mypy
   - numpy==1.23.4
   - pytest
+  - pydocstyle
 
   # Dependencies for velin
   - numpydoc>=1.1.0
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -18,8 +18,8 @@ env:
   LIBROSA_DOC_DEBUG: True  # Disable docstring example execution
 
 jobs:
-    doc:
-        name: "CI Documentation build"
+    setup:
+        name: "Doc environment setup"
         runs-on: ${{ matrix.os }}
 
         strategy:
@@ -66,11 +66,21 @@ jobs:
           shell: bash -l {0}
           run: python -m pip install --upgrade-strategy only-if-needed -e .[docs]
 
+    linkcheck:
+        name: "Link checking"
+        runs-on: ${{ matrix.os }}
+        needs: "setup"
+        steps:
         - name: Link checking
           shell: bash -l {0}
           working-directory: docs
           run: make linkcheck
 
+    multiversion:
+        name: "Multi-version build"
+        runs-on: ${{ matrix.os }}
+        needs: "setup"
+        steps:
         - name: Build multi-version site
           shell: bash -l {0}
           run: |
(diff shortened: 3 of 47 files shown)


bmcfee:
Along with the sphinx business, I've also added a doc style checker to the linting action.  This supplements our velin and codespell checks.

For now, I'm using the numpydoc convention settings, plus a few extra skips that we could remove over time.  I've corrected the docstrings locally so that it passes with the current config, and this was actually pretty helpful for identifying some badly phrased documentations and a few formatting errors.  The additional skips are (at present):

- D400: First line should end with a period
- D205: 1 blank line required between summary line and description
- D202: No blank lines allowed after function docstring

D205 I've skipped because it does not reliably work with reST formatting in our autodoc sections.

D202 is a strange one, in that it has nothing to do with the documentation - it is only about the spacing (or not) between the docstring and code.  I got about 1/3 of the way through the package at hacking these out before giving up and suppressing the errors so I could focus on more substantive issues.  I'm okay with finishing the job here and not skipping D202 though.

D400 is a style thing, and would require a bit more rewriting than I had the stomach for this afternoon.  I don't feel strongly about keeping it or not.

bmcfee:
I think this is good to go.  Anyone care to do a quick check of the revised docstrings?  The changes are almost entirely cosmetic.

bmcfee:
Note: the linkcheck passes now, and did identify many broken links.  I've fixed the paper links wherever possible with (hopefully) permanent Zenodo DOI links.

The only one that really caused me trouble was the `chroma_ansyn` link on the labrosa page, which apparently has gone 401 on us.  (@dpwe any idea what's going on there?)  For the time being, I've put in a wayback machine link, but it seems reasonable to make that the permanent link anyway.

We do have a number of redirects popping up, mostly github PR links in the changelog, and a handful of resolved DOIs.  I think it makes sense to leave these as is.

bmcfee:
One lingering question for this PR: do we want to add a [build artifact](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) for the doc CI?  This would make it possible to review a version of the doc build on PR prior to merging.

I expect that it's probably overkill for us, but doc previewing is a nice feature of RTD that we're missing by rolling our own sphinx-multiversion instance.  If we *do* want to add a doc artifact, I suspect it will be more useful to base it on the single-version build without debug mode so that we can check example figures.  This would require a separate job in the workflow, and would probably drag out our build time quite a bit.

bmcfee:
I think I'll punt on the build artifact idea for now.  It would be nice to have, but there's already enough in this PR for now.

I believe the contents of this PR are generally uncontroversial, and I think we can safely merge.  The only point that probably warrants some discussion is the slight change to our coding style that's implied here, since pydocstyle passing now becomes a blocker for future PRs.  If we're all okay with that, we can merge.

bmcfee:
Last commit makes some minor simplifications to the contributing guidelines.  I was reading through and noticed that they didn't really match what our lint CI does, which could lead to confusion.  (Ie our code doesn't pass autopep8, so why should a PR?)

bmcfee:
Ok, I think this one is ready to merge.  To summarize:

- The changes to code / docstrings here are almost entirely cosmetic or grammatical, so no need to review.
- There are some minor corrections, primarily in bibliography or broken links. Nothing controversial here.
- I would like some feedback on the changes to CONTRIBUTING.md and the [selected rules for pydocstyle](https://github.com/librosa/librosa/pull/1718#issuecomment-1579260530) before merging, just in case I'm missing anything.

Otherwise I'd like to get this merged so we can have corrected doc builds deployed ASAP.

CONTRIBUTING.md
```diff
@@ -77,19 +77,15 @@ following rules before submitting a pull request:
 You can also check for common programming errors with the following
 tools:
 
--  Code with good test coverage (at least 80%), check with:
+-  Code with good test coverage, check with:
```
lostanlen, line 80: 👍🏻

lostanlen (approved):

PR #1719, opened 2023-06-08
Advanced example for patch generation

bmcfee:
#### Reference Issue

This PR adds an advanced example gallery entry to demonstrate spectrogram patch generation.

It ought to be fairly self-contained, but any feedback on the content would be appreciated.

### DIFF
--- a/docs/examples/plot_patch_generation.py
+++ b/docs/examples/plot_patch_generation.py
@@ -1,0 +1,130 @@
+"""
+==========================
+Efficient patch generation
+==========================
+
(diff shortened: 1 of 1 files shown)


bmcfee:
This should block until #1717 merges

stefanv:
FWIW, I found the example very clear and readable :)

It also made me wonder what it would take to generate n seconds of a "representative" sample, given a spectrum.

bmcfee:
Thanks @stefanv !

PR #1722, opened 2023-06-15
reworked lazy loading of matplotlib

### LINKED ISSUE #1721: lazy_loading matplotlib submodules can cause errors
Librosa uses [lazy_loader](https://github.com/scientific-python/lazy_loader/) to lazily load some libraries. This implementation has the known flaw (see [here](https://github.com/python/cpython/issues/86439), [here](https://peps.python.org/pep-0690/#import-side-effects) and [here](https://github.com/networkx/networkx/issues/5838)) of not triggering certain import side effects, effectively breaking libraries that rely on this, like matplotlib. This can cause weird and hard to replicate bugs, like [this](https://github.com/matplotlib/matplotlib/issues/25506) one, which was filed initially against [matplotlib](https://github.com/matplotlib/matplotlib/issues/25506) and then to [matplotlib-inline](https://github.com/ipython/matplotlib-inline/issues/24).

My best guess is that bugs related to this are probably more widespread than would be apparent, as users would report these bugs to the wrong projects most of the time.

It's not clear to me there is a good solution to this. As far as I know, the only reliable solution would be to patch matplotlib itself to support lazy loading, like [numpy did](https://github.com/numpy/numpy/pull/22045).

**To Reproduce**

In the following case, one should be able to access pyplot (or run one of its methods): [^1]

```python
import librosa.display

import matplotlib
import matplotlib.pyplot

matplotlib.pyplot
```

Instead, one gets:
```
AttributeError: module 'matplotlib' has no attribute 'pyplot'
```

Just removing the `librosa.display` import makes it work again.

**Edit:** Removed reference to importlib's implementation, as I thought Librosa was using it (and not lazy_loader).

[^1]: Yes, both imports are necessary to replicate because the python import system is weird.

bmcfee:
#### Reference Issue
Fixes #1721 (sort of)

This PR partially reverts our use of lazy-loading to import matplotlib in the display module.  The reasoning here is that due to some quirks in how matplotlib depends on state manipulation triggered by import, lazy-loading can actually break things downstream in some unpredictable ways.

The display module overall is still attached lazily at the top-level `__init__.py(i)` though.  This means that we should retain the intended behavior: display (and matplotlib) are optional, mpl is only loaded if the user explicitly does so, but an explicit import of `librosa.display` is not necessary.

#### Any other comments?

This is probably not a good long-term solution, but I think it will help in the short term until the issues around lazy-loading settle down upstream.

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -41,7 +41,10 @@ from itertools import product
 import warnings
 
 import numpy as np
-import lazy_loader as lazy
+import matplotlib.cm as mcm
+import matplotlib.axes as mplaxes
+import matplotlib.ticker as mplticker
+import matplotlib.pyplot as plt
 
 from . import core
 from . import util
@@ -52,21 +55,11 @@ from ._typing import _FloatLike_co
 
 if TYPE_CHECKING:
     import matplotlib
-    import matplotlib.cm as mcm
-    import matplotlib.axes as mplaxes
-    import matplotlib.ticker as mplticker
-    import matplotlib.pyplot as plt
     from matplotlib.collections import QuadMesh, PolyCollection
     from matplotlib.lines import Line2D
     from matplotlib.path import Path as MplPath
     from matplotlib.markers import MarkerStyle
     from matplotlib.colors import Colormap
-else:
-    matplotlib = lazy.load("matplotlib")
-    mcm = lazy.load("matplotlib.cm")
-    mplaxes = lazy.load("matplotlib.axes")
-    mplticker = lazy.load("matplotlib.ticker")
-    plt = lazy.load("matplotlib.pyplot")
 
 
 __all__ = [


bmcfee:
I spent a bit of time trying to get a package-level mock test working, but it turned out to be a little trickier than anticipated.

A much simpler way to verify this is to create a dummy file `matplotlib.py`:
```python
raise ImportError("Tried to import matplotlib!")
```
in the current directory.  This would take precedence over an installed package when imported, allowing us to detect whether the lazy loading is working properly.  All seems to check out.

With the poisoned matplotlib.py in CWD:
```python
In [1]: import librosa

In [2]: librosa.__version__
Out[2]: '0.10.1dev'

In [3]: import librosa.display
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ in <module>:1                                                                                    │
│                                                                                                  │
│ /home/bmcfee/git/librosa/librosa/display.py:44 in <module>                                       │
│                                                                                                  │
│     41 import warnings                                                                           │
│     42                                                                                           │
│     43 import numpy as np                                                                        │
│ ❱   44 import matplotlib.cm as mcm                                                               │
│     45 import matplotlib.axes as mplaxes                                                         │
│     46 import matplotlib.ticker as mplticker                                                     │
│     47 import matplotlib.pyplot as plt                                                           │
│                                                                                                  │
│ /home/bmcfee/matplotlib.py:1 in <module>                                                         │
│                                                                                                  │
│ ❱ 1 raise ImportError("Tried to import matplotlib!")                                             │
│   2                                                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
ImportError: Tried to import matplotlib!
```
without it, but having a working mpl installed:
```python
In [1]: import librosa

In [2]: librosa.__version__
Out[2]: '0.10.1dev'

In [3]: import librosa.display

In [4]:
```

I've also verified that `librosa.show_versions()` works properly with the poisoned matplotlib file:
```python
In [1]: import librosa

In [2]: librosa.show_versions()
INSTALLED VERSIONS
------------------
python: 3.9.15 | packaged by conda-forge | (main, Nov 22 2022, 08:45:29) 
[GCC 10.4.0]

librosa: 0.10.1dev

audioread: 3.0.0
numpy: 1.23.5
scipy: 1.10.1
sklearn: 1.2.2
joblib: 1.2.0
decorator: 5.1.1
numba: 0.56.4
soundfile: 0.12.1
pooch: v1.6.0
soxr: 0.3.5
msgpack: 1.0.5

numpydoc: 1.5.0
sphinx: 6.2.1
sphinx_rtd_theme: 1.2.1
matplotlib: None
sphinx_multiversion: 0.2.4
sphinx_gallery: 0.13.0
mir_eval: 0.6
ipython: None
sphinxcontrib.rsvgconverter: installed, no version number available
pytest: 7.3.1
pytest_mpl: 0.16.1
pytest_cov: 4.1.0
samplerate: 0.1.0
resampy: 0.4.2
presets: 0.1.3
packaging: 23.1
```
With a working install:
```python
...
matplotlib: 3.7.1
...
```
I'm happy enough with this that I don't think we need to augment the test suite, so this is good to merge.

PR #1726, opened 2023-06-19
Consistent conversion of time duration/offset to sample counts

### LINKED ISSUE #1709: Expected number of samples from load()?
When duration*sample_rate is not an integer, it seems that the behavior of __soundfile_load and __audioread_load are different with respect to how many samples will be returned: 

__soundfile_load performs int(duration*sample), ie a 'floor' operation:

https://github.com/librosa/librosa/blob/f21bda6b9fd7d9f1276749862247a6daf59e792d/librosa/core/audio.py#LL215C13-L215C13

while __audioread_load performs int(np.round(duration*sample)): 

https://github.com/librosa/librosa/blob/f21bda6b9fd7d9f1276749862247a6daf59e792d/librosa/core/audio.py#LL247C9-L247C9

Is this behavior intentional? 

Furthermore, neither of these behaviors seem to explain observed results: (sample rate is 22050) 
```
samples, sr = librosa.load(path=audio_path,offset=0,duration=5/22050) #5 samples
samples, sr = librosa.load(path=audio_path,offset=0,duration=5.1/22050) #5 samples
samples, sr = librosa.load(path=audio_path,offset=0,duration=5.9/22050) #6 samples
samples, sr = librosa.load(path=audio_path,offset=0,duration=5.51/22050) #5 samples (?!)
```
Is this a floating point precision issue? `int(np.round(5.51/22050*22050))` gives me 6, not 5, so I don't know why this last line gives 5 samples.

bmcfee:
#### Reference Issue
Fixes #1709 

This PR modifies signal loaders and generators to always use floor when converting fractional sample counts to integers.

#### Any other comments?

The only remaining instance of non-floor conversion (rounding or ceiling) that I could find was

- resample; we take ceiling here to be conservative in case a resampling backend uses something other than floor.  IIRC some use rounding, but most use floor.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -242,12 +242,12 @@ def __audioread_load(path, offset, duration, dtype: DTypeLike):
         sr_native = input_file.samplerate
         n_channels = input_file.channels
 
-        s_start = int(np.round(sr_native * offset)) * n_channels
+        s_start = int(sr_native * offset) * n_channels
 
         if duration is None:
             s_end = np.inf
         else:
-            s_end = s_start + (int(np.round(sr_native * duration)) * n_channels)
+            s_end = s_start + (int(sr_native * duration) * n_channels)
 
         n = 0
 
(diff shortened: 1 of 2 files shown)


bmcfee:
I'm not worried about the coverage drop here, as it's in a deprecated code path and almost certainly will behave correctly.

This PR is good to go / review IMO.

PR #1727, opened 2023-06-20
support pooch >= 1.7

### LINKED ISSUE #1688: Rebuild pooch registry to support 1.7.0.
**Describe the bug**

Pooch 1.7.0 slightly changed the syntax for registry files so that quote characters now need to be escaped.  As a result, we are currently incompatible with pooch >=1.7.

As a stop-gap solution, I've pushed post-release packages to upper bound the version requirements on the current release (0.10.0), but going forward we should rebuild the pooch registry file so that it will work with the more recent versions.

I'm not sure if this is something we can put in the 0.10.1 release, or if it should wait for a major version bump.  It's technically not a breaking change from the librosa API.  However, it would require a later version of pooch since the revised registry file will not work in 1.6 or earlier.  So bumping the minimum requirement here could be construed as forcing an update into the user's environment that may not preserve backward compatibility.

Realistically, I doubt this is much of an issue since very few packages rely on pooch to begin with, and those that do are unlikely to be sensitive to this specific change in the registry syntax (yay we're special!), but it's worth noting here for discussion.

bmcfee:
#### Reference Issue
Fixes #1688 

This PR rewrites the example data registry file to use the new format introduced in pooch 1.7.0.

It removes the upper pin on the pooch dependency, and bumps our minimum supported version to 1.7.

#### Any other comments?

Per my comments in #1688, I'm still ambivalent on this, as it would force an update to user environments that is not backward-compatible.  But, since it's extremely unlikely to affect anyone, I'm leaning toward rolling it into 0.10.1 anyway.

No functionality changes here, so as long as the CI passes and builds cleanly, it should be good to merge.


bmcfee:
Two ideas here:

1. (already implemented just now) is to leave the upper version pin in the docs environment.  This way, old versions will continue to build with pooch 1.6, but we can still test the latest on 1.7.
2. Instead of just reformatting the registry, we can rename the files server-side to avoid apostrophes.  This can be done with a few symbolic links, and old versions will continue to work fine.  Doing this will allow us to have a single registry file that works with pooch 1.6 or 1.7, so the minimum version requirement can stay where it was and we don't force environment updates on users.  However, we will still need the upper version pin on the doc environment.

bmcfee:
I've updated the librosa data repo to move the offending filenames, but add backward-compatible symbolic links so that old versions continue to work.  I think this ought to do the trick.

PR #1747, opened 2023-09-05
Code refactoring 

tosemml:
#### Reference Issue

NA

These minor code adjustments aim to make the code more concise by using optimized NumPy functions.

### DIFF
--- a/scripts/create_intervals.py
+++ b/scripts/create_intervals.py
@@ -66,9 +66,7 @@ def main():
     factorized = dict()
     for interval in all_intervals:
         # Compute the interval
-        log_value = 0
-        for p in interval:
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks @tosemml for this PR.  However, at a glance I'm inclined to not accept it.  The modifications here are in either one-time use scripts (ie the interval constructor) or in unit tests, and not in main library code that will be traversed often.

While the modifications here might be a little faster in practice, they replace explicit loops with list comprehensions.  It's not obvious that this will be faster though, as the list comprehension still uses loops, and now we're iterating twice independently instead of once.  Given that, I'd prefer the more easily human-readable forms we already have over the more terse list comprehensions.

PR #1753, opened 2023-09-19
Fix audio load sampling rate type

### LINKED ISSUE #1748: `librosa.load` returns wrongly-typed int object
**Describe the bug**
`librosa.load` returns `int` object instead of `float`.  

**To Reproduce**
```python
import librosa

import librosa
wave, sr = librosa.load("test.wav", sr=None, mono=True)
print(type(sr))
# <class 'int'>
```

**Expected behavior**
Typing of `.load` says that it returns `float`.  
https://github.com/librosa/librosa/blob/09e4a622456bbad314e2e6d5c75879f56d728b9b/librosa/core/audio.py#L66C1-L66C31  
So, the function should return `float` object.  

**Software versions**
```
Linux-5.15.109+-x86_64-with-glibc2.35
Python 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
NumPy 1.23.5
SciPy 1.10.1
librosa 0.10.1
INSTALLED VERSIONS
------------------
python: 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]

librosa: 0.10.1

audioread: 3.0.0
numpy: 1.23.5
scipy: 1.10.1
sklearn: 1.2.2
joblib: 1.3.2
decorator: 4.4.2
numba: 0.56.4
soundfile: 0.12.1
pooch: v1.7.0
soxr: 0.3.6
msgpack: 1.0.5

numpydoc: None
sphinx: 5.0.2
sphinx_rtd_theme: None
matplotlib: 3.7.1
sphinx_multiversion: None
sphinx_gallery: None
mir_eval: None
ipython: None
pytest: 7.4.0
pytest_mpl: None
pytest_cov: None
samplerate: None
resampy: None
presets: None
packaging: 23.1```

### PR DESCRIPTION
**Pull request by tarepan**
#### Reference Issue
Fixes #1748 

Fix `librosa.load`'s actual return type as described in above issue.  

Two semantic changes are implemented:  

- default value (int -> float)
- casting (int OR float -> float)

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -57,13 +57,13 @@ def load(
         str, int, os.PathLike[Any], sf.SoundFile, audioread.AudioFile, BinaryIO
     ],
     *,
-    sr: Optional[float] = 22050.0,
+    sr: Optional[float] = 22050,
     mono: bool = True,
     offset: float = 0.0,
     duration: Optional[float] = None,
     dtype: DTypeLike = np.float32,
     res_type: str = "soxr_hq",
-) -> Tuple[np.ndarray, float]:
+) -> Tuple[np.ndarray, Union[int, float]]:
     """Load an audio file as a floating point time series.
 
     Audio will be automatically resampled to the given rate
@@ -195,7 +195,8 @@ def load(
         sr = sr_native
 
     # SR Cast
-    sr = float(sr)
+    if isinstance(sr, float) and sr.is_integer():
+        sr = int(sr)
 
     return y, sr
 
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -194,10 +194,6 @@ def load(
     else:
         sr = sr_native
 
-    # SR Cast
-    if isinstance(sr, float) and sr.is_integer():
-        sr = int(sr)
-
     return y, sr
 
 
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -57,7 +57,7 @@ def load(
         str, int, os.PathLike[Any], sf.SoundFile, audioread.AudioFile, BinaryIO
     ],
     *,
-    sr: Optional[float] = 22050,
+    sr: Optional[float] = 22050.0,
     mono: bool = True,
     offset: float = 0.0,
     duration: Optional[float] = None,
@@ -193,6 +193,9 @@ def load(
 
     else:
         sr = sr_native
+
+    # SR Cast
+    sr = float(sr)
 
     return y, sr
 

### CODE REVIEW

**Comment by bluenote10** [comment]
I not sure this change is needed.

According to [PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower):

> when an argument is annotated as having type float, an argument of type int is acceptable

In general in the type system, `int` is considered a subtype of `float`. Explicit conversion is rather an anti-pattern, because everywhere where something is annotated as `float` you may actually get an `int` (think of users passing a plain int literal to an argument that may allow for `float`). Code needs to handle that properly anyway.

**Comment by tarepan** [comment]
At first, I partially agree with @bluenote10 's statement.  

There is a semantical relationship of "integer ⊂ float".  
As PEP 484 defined, Python type hint use the word `float` as "[`int` class](https://docs.python.org/3/library/functions.html#int) | [`float` class](https://docs.python.org/3/library/functions.html#float)".  
This means that programmer should handle `float`-typed variable as `int` | `float` so that should avoid using `int`/`float`-specific methods directly toward the `float`-typed variable (e.g. `.is_integer()`).  
Based on this fact, this PR (internal float cast) has no semantical meaning, so could be redundant.  

But, at the same time, this PR is practically beneficial in current Python ecosystem.  

Current popular type checker (e.g. [pyright](https://github.com/microsoft/pyright)) do NOT follow PEP484 strictly.  
For examle, pyright do NOT raise any errors toward below code.  

```python
def ret_float() -> float:
    return 1

ret_float().is_integer()
```

`is_integer()` is `float` class' method, so this should cause type check error (`int` class do not have corresponding method).  
This means that, in current Python ecosystem, type hint `float` is interpreted as `float` class.  

If the user is *Python-master*, he will check `float` type hint and can write correct "`int` class or `float` class" handler.  
But if the user is *non-Python-master*, he will check variable's class (`float`) and variable's type(`float`), then may misunderstand that "OK, this variable is float class", finally cause error.  
This is totally correct Python error, but is practically so annoying for signal processing (librosa) users because this problem is a Python language detail which even major type checker is not supporting.  
This PR practically prevent this ***correct-but-annoying*** error.  

If librosa think that "librosa users should correctly fall down on `flaot` handling and be a *Python-master*", this PR should be rejected.  
If librosa think that "librosa follow Python de facto (not PEP) and avoid corner-case pitfall for usability", this PR should be accept.  

In this meaning, I partially agree with @bluenote10, but think this PR is meaningful.

tarepan:
Thanks @bluenote10 and @bmcfee for constructive discussion.  

(I think that) now we agree that PEP 484 `float` type simplification have the limitation (false-negative type check) and needs some improvements.  

Remaining issues are:

- type as cencept expression / semantics (*sampling rate* is essentially float value) (@bmcfee)
- `int` class related backward-compatibility (@bluenote10)

(are there missing issues?)  

## Type as concept
I strongly agree with @bmcfee 's approach, expression a concept with type.  
`load(...) -> float` clearly means that "sampling rate is float value".  

## `int` backward-compatibility
I agree with @bluenote10 's suggestion that this PR will break some current user codes.  

But there is a wrong assumption:  
> integer sample rates are actually return as int (which it currently does anyway I guess).

Current librosa implementation do NOT run auto-cast with sr fractionality-check.  
librosa just return `sr_native` which is returned by external library, or use `sr` argument without fractionality-check.  
`load()` could return `float(16000.0)`, which is non-fractional but `float` class.  

Because of this behavior, `assert isinstance(sr, int)` have bugs (false positive toward `float(n.0)`), so is historically invalid.  
We can see the bug as below:  

```python
wave, sr = librosa.load("test.wav", sr=8000.0)
assert isinstance(sr, int), "Fractional sampling rates are not supported"
# AssertionError: Fractional sampling rates are not supported
```

Programmers should implement as below in current librosa:

```python
assert isinstance(sr, int) or (isinstance(sr, float) and sr.is_integer()), "Fractional sampling rates are not supported"
```

If correctly-implemented, this PR never break the fractionality-check.  
In this meaning, this PR do not break the backward-compatibility.

bluenote10:
>  This is a pet peeve of mine, where people generally assume that sampling rates need to be integer valued (they don't!). 

From a theoretical perspective that is certainly true. There can be practical/external constraints however that prohibit fractional sampling rates. For instance:

- To my knowledge there are few file formats that allow for fractional sampling rates, because their specs constrain specifying the sample rate e.g. as a 4 byte integer only.
- Similarly, common audio hardware, drivers and sound server systems may prohibit fractional sample rates by the native interfaces they offer.

Out of curiosity: Are you aware of systems that actually support fractional sample rates? Perhaps my experience with these technologies is a bit outdated, and I should think beyond good old WAVE specs :wink:

In any case, when facing this "integer sample rate constraint", the most simple way to handle it without the need for costly resampling is to propagate them back through to the input.

bluenote10:
> Because of this behavior, assert isinstance(sr, int) have bugs (false positive toward float(n.0)), so is historically invalid.

Mind you that the type annotations for librosa have only been introduced recently. Up to that point users had to use `load` without being aware that it can even return a `float`, because at runtime it simply always returned an `int` when loading from typical integer-based file formats. That's why I would assume there is a lot of code out that assume the return `sr` is an `int` and explicit conversion will break all these historical usages.

> Programmers should implement as below in current librosa:

It is rather unlikely that many people would have written it like that, no? Not only because the expression is rather verbose and requires awareness of the subtleties of PEP 484, but also, because the runtime behavior in the past didn't make it necessary.

Of course, one can argue that code that naively expects to get an `int` an runtime is broken to begin with, but I don't see a huge benefit in forcefully breaking all these "bad" usages.

bmcfee:
> In any case, when facing this "integer sample rate constraint", the most simple way to handle it without the need for costly resampling is to propagate them back through to the input.

Yeah, you raise valid points about hardware and encodings.  I'm thinking more from the perspective of algorithms (e.g. cqt or multirate filterbank) and unit conversion (samples→time or frames), where propagating integer types can silently introduce round-off errors in strange places.  We could probably get away with relying on python / numpy to handle type promotion as necessary here, but I think it's much clearer to be explicit rather than implicit about sampling rates being floats.



> Mind you that the type annotations for librosa have only been introduced recently. Up to that point users had to use `load` without being aware that it can even return a `float`, because at runtime it simply always returned an `int` when loading from typical integer-based file formats. That's why I would assume there is a lot of code out that assume the return `sr` is an `int` and explicit conversion will break all these historical usages.

I'd be pretty surprised if many people (if any) are relying on / assuming `sr` to be integer typed rather than just some-kind-of-number.. but I could be wrong.

tarepan:
> the type annotations for librosa have only been introduced recently.  

Oh, I totally forgot this fact.  
It could be natural (before type introduction) that people assume that `sr` is `int` type because `librosa.load` return `int` class object in popular setup.  

> It is rather unlikely that many people would have written it like that, no?  

There could be 3 type of people:  

- α. Think `sr` as `int` class
- β. Think `sr` as `int` class or `float` class
  - β1. Think `sr==float` as "fractional" (fractionality is checked and auto-casted by librosa)
  - β2. Think `sr==float` as "number, can be non-fractional"

α people never check fractionality with assertion because he never think about `float` (if notice, he is β).  
β1 people will write `assert isinstance(sr, int)`.  
β2 people will write `assert isinstance(sr, int) or (isinstance(sr, float) and sr.is_integer())`.  

@bluenote10 is β1 people.  
I and @bmcfee is β2 people.  
In my intuition, there are many α, some β1 and some β2.  

This PR will break the correct codes of α and β1, and it is backward compatibility break.  
In this meaning, this PR has big downside.  

At the same time, this discussion reveal that current implementation has this big ***ambiguity***.  
As I / @bluenote10 / @bmcfee admitted, `sr` is theoretically `float` class object.  
This PR will align implementation with the theory, and resolve the ***ambiguity***.  
In this meaning, this PR has big upside for future with the pain of backward incompatibility.  

In my opinion, It depends on the policy of librosa.

bluenote10:
> I'd be pretty surprised if many people (if any) are relying on / assuming sr to be integer typed rather than just some-kind-of-number..

If they have adopted type annotations before librosa was annotated they had to make a choice between either assuming `sr: int` or `sr: float`. For instance, as a very crude data point, here is a full text search over GitHub for the two:

- [`sr: int`](https://github.com/search?q=%22sr%3A+int%22+lang%3APython+&type=code) has 3.8 k matches
- [`sr: float`](https://github.com/search?q=%22sr%3A+float%22+lang%3APython+&type=code) has 576 matches

>  This PR will align implementation with the theory, and resolve the ambiguity.

This is indeed the benefit. I just wanted to point out that there is the alternative of making the ambiguity explicit, i.e., simply aligning the type annotation with the current runtime behavior (returning `int | float`), because it also solves the underlying issue without breaking runtime behavior.

bmcfee:
> For instance, as a very crude data point, here is a full text search over GitHub for the two:
> 
>     * [`sr: int`](https://github.com/search?q=%22sr%3A+int%22+lang%3APython+&type=code) has 3.8 k matches
> 
>     * [`sr: float`](https://github.com/search?q=%22sr%3A+float%22+lang%3APython+&type=code) has 576 matches

I was going to say something snarky about how most of those (in the first case) are not actually using librosa, just similar variable names... and then I noticed that one of the type annotation hits for integer sr is from us in the unit conversion code :facepalm: https://github.com/librosa/librosa/blob/43d4427f7a389acc79b5be3c924646701f19bb46/librosa/core/convert.py#L601-L632

tarepan:
I summarize upside/downside of two proposals:

| proposal    | TypeHint | BackCompat | semantics |
| ----------- | -------- | ---------- | --------- |
| `int\|float` | ✔️       | ✔️         | ❌       |
| all `float` | ✔️       | ❌         | ✔️       |

Both proposal will resolve type hint issue (false negative type check).  
`int|float` cannot express the concept 'sr is a number (float)', but can keep backward compatibility.  
all `float` express the cencept clearly, but break the popular sr check notation.  

After this discussion, I agree with @bluenote10 's proposal, `int|float`.  
Some people may feel weird that "sampling rate is just number, why annotated as int union...?", but this proposal resolve all practical problems.  

If @bmcfee agree, I will rewrite this PR. How do you think?

bmcfee:
Yeah, I can live with that solution - thanks for putting in the thought here!

As an aside, this popped up on twitter over the weekend, and it made me laugh a bit:
https://x.com/QyubeySan/status/1705462273966289403

Translation:
image

Which is to say, there are very real and annoying consequences when we force integer sampling rates and induce round-off errors:
image

(To be fair, up-and-down sampling like OP describes is only necessary for some resampling algorithms, and our default choice should not invoke that specific problem... but still!)

tarepan:
I also have experienced same kind of sr problem lol

Anyway, I make a new commit which reflect above discussion, can you review it @bmcfee ?

librosa/core/audio.py
```diff
@@ -194,6 +194,10 @@ def load(
     else:
         sr = sr_native
 
+    # SR Cast
+    if isinstance(sr, float) and sr.is_integer():
+        sr = int(sr)
```
bmcfee, line 199: I don't think this part should be here.  Imagine the case where a user explicitly calls with an integer-valued float target sr:
```
y, sr = librosa.load(somefile, sr=8000.0)
```
With this logic in place, the output sr will be integer typed, which is not what the user specified.

I think it's best to leave it as it was.  If the underlying decoder produces an int type, that's fine to annotate, but we shouldn't coerce it here.

tarepan: It makes sense, I will remove this part.

bmcfee:
Thanks @tarepan !

While we're on this topic, I think it would make sense to check the type annotations for sampling rates elsewhere in the package so that they agree.  I remember finding a couple that were int :grimacing: and the rest were float, but we should try to make this all consistent where possible.

tarepan:
> it would make sense to check the type annotations for sampling rates elsewhere in the package so that they agree.

I agree with the idea, and will list up and fix them.

tarepan:
## Wrongly-typed sr

- `sr: int`
  - `blocks_to_time()` [1](https://github.com/librosa/librosa/blob/96da0eb228f591740ecbaa369cbc03fc92ea8185/librosa/core/convert.py#L603), [2](https://github.com/librosa/librosa/blob/96da0eb228f591740ecbaa369cbc03fc92ea8185/librosa/core/convert.py#L610), [3](https://github.com/librosa/librosa/blob/96da0eb228f591740ecbaa369cbc03fc92ea8185/librosa/core/convert.py#L621), [4](https://github.com/librosa/librosa/blob/96da0eb228f591740ecbaa369cbc03fc92ea8185/librosa/core/convert.py#L631)

bmcfee:
:wave: just checking in on this, since it seems to have stalled out.  Anything we can do to help get this unstuck?

bmcfee:
Closing this PR to finish the work over in #1789

PR #1761, opened 2023-09-26
Expose `pad_mode` in CQT chromagram

carlthome:
Minor thing but chroma_cqt doesn't expose padding mode so couldn't set `reflect` instead of `constant` so easily after https://github.com/librosa/librosa/commit/e964f672f5529ac5069df9cee6176bbb9ec7064d

Maybe nice to expose?

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -1299,6 +1299,7 @@ def chroma_cqt(
     window: Optional[np.ndarray] = None,
     bins_per_octave: Optional[int] = 36,
     cqt_mode: str = "full",
+    pad_mode: _PadMode = "constant",
 ) -> np.ndarray:
     r"""Constant-Q chromagram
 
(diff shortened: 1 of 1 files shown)


bmcfee:
We could expose this, but I am a little worried about signature bloat here.  The behavior you're looking for could be achieved by pre-computing the CQT with the desired parameters and then passing it into `chroma_cqt`, and this is the generally preferred pattern we take with the less-often used parameters to spectral transforms in feature calculations.

I suppose there's an argument to be made here that chroma features do a little more work than normal in setting up the CQT params (eg tuning estimation, bin spacing, etc), and it's therefore more work for the user to lift the CQT out as a precomputation.

A couple of thoughts, if we do want to go down this road:

1. We should expose the same parameters in like-minded functions (cens, tonnetz, chroma_vqt)
2. Are there other parameters to expose here?  If so, might it make more sense to turn this into a kwarg passthrough?

carlthome:
> We could expose this, but I am a little worried about signature bloat here.

Yeah, this makes sense to me! Perhaps we can revisit this if there are more users than just me, interested in setting pad mode for CQT chromagrams.

My need surfaced when digging around for numerical differences in https://github.com/urinieto/msaf/issues/86, though after changing pad mode between zero padding and reflection, all SALAMI numbers were the same (as expected, but still wanted to double check) so my personal need for exposing this setting isn't relevant anymore.

PR #1765, opened 2023-10-09
Switch from pkg_resources to importlib.resources

mattpitkin:

When using librosa, `interval.py` gives a deprecation warning about the use of `pkg_resources`, which is described at the top of the `pkg_resources` documentation [here](https://setuptools.pypa.io/en/latest/pkg_resources.html). This PR fixes this by switching (as suggested in the documentation) to using [`importlib`](https://docs.python.org/3/library/importlib.html), in particular, the ~[`resources.files`](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files)~ [`resources.path`](https://docs.python.org/3.9/library/importlib.html#importlib.resources.path) function.

### DIFF
--- a/librosa/util/example_data/__init__.py
+++ b/librosa/util/example_data/__init__.py
@@ -1,1 +1,1 @@
-'''Resources for loading example data'''
+"""Resources for loading example data"""
--- a/librosa/util/example_data/__init__.py
+++ b/librosa/util/example_data/__init__.py
@@ -1,0 +1,1 @@
+'''Resources for loading example data'''
--- a/librosa/core/intervals.py
+++ b/librosa/core/intervals.py
@@ -12,9 +12,10 @@ from .._cache import cache
 from .._typing import _FloatLike_co
 
 
-with open(resources.files("librosa.core") / "intervals.msgpack", "rb") as _fdesc:
-    # We use floats for dictionary keys, so strict mapping is disabled
-    INTERVALS = msgpack.load(_fdesc, strict_map_key=False)
+with resources.path("librosa.core", "intervals.msgpack") as imsgpack:
+    with imsgpack.open("rb") as _fdesc:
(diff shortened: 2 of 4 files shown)


mattpitkin:
@bmcfee no problem, I can make those changes.

mattpitkin:
I've updated `files.py` and `setup.cfg`. Note that I've also had to add a `__init__.py` file for `example_data` to make it a module, so that the `resources.path` function can find the files (this is a slight annoyance with using `importlib.resources` compared to using `pkg-resources`)

bmcfee:
> ote that I've also had to add a `__init__.py` file for `example_data` to make it a module, so that the `resources.path` function can find the files (this is a slight annoyance with using `importlib.resources` compared to using `pkg-resources`)

Hrm.  Is it possible to avoid that by including the folder in the resource path directly?  eg

```python
with resources.path("librosa.util", "example_data/index.json") as index:
```
(or fancier with `Path("example_data") / "index.json"` or somesuch to support windows)?

bmcfee:
> s it possible to avoid that by including the folder in the resource path directly?

Nevermind, docs answered my question for me: 

> package is either a name or a module object which conforms to the Package requirements. resource is the name of the resource to open within package; it may not contain path separators and it may not have sub-resources (i.e. it cannot be a directory).

librosa/util/example_data/__init__.py
bmcfee, line 1:
```suggestion
'''Resources for loading example data'''
```

mattpitkin: For some reason, the github interface wasn't allowing me to commit the suggestion directly, so I've done it myself.

bmcfee (approved):

bmcfee:
Thanks again for this @mattpitkin ! :shipit:

PR #1766, opened 2023-10-10
Multichannel peak, onset, and beat detection

### LINKED ISSUE #44: variable tempo beat tracking

### LINKED ISSUE #1695: Multi-channel beat and onset detection

When we implemented #1130, we made a choice to not implement multi-channel support for detection functions that produce ragged outputs (eg, beat tracking or onset detection).  This was a practical choice forced primarily by the output format, specifically arrays of frame indices, and secondarily by the nature of the computations which don't naturally vectorize.

There are at this point plenty of examples elsewhere in the code where we use numpy / numba vectorization to distribute otherwise sequential functions over independent channels, such as viterbi.  As such, the dynamic programs used in onset detection and beat tracking could, without  much hassle, also be vectorized, and the only remaining constraint is the output format.

The core problem here is that the outputs for detector methods are *sparse*.  If we were to add a (less spatially efficient) dense output mode, it would be possible to support multichannel detections.

We could add a `sparse: bool` parameter to onset detection and beat tracking.
- If `sparse=True` (default), it would behave as it currently does, and not support multichannel.
- If `sparse=False`, the output arrays would have trailing shape of `N_frames` (matching the underlying spectrogram), and `dtype=bool` to indicate the presence or absence of a detection.  Each channel's detections would be computed independently of the others, and with common parameters.

Additionally, when `sparse=False`, the `units` parameter would be ignored.

I suppose we could extend `units="dense"` to avoid bloating the signatures, but I think a dedicated `sparse` parameter is much clearer.

**Additional context**

We might be able to do something similar with DTW paths, but the motivating use-case there is much less clear.

### LINKED ISSUE #1773: Beat tracker has always trimmed one too many beats
**Describe the bug**
This surfaced in #1766 , but creating a separate issue here for historical purposes.

In the beat tracker, when `trim=True`, we enter the following path:

https://github.com/librosa/librosa/blob/5ca70f5d57fc597660452f011ffff55f9c36ed4b/librosa/beat.py#L505-L516

The idea here is to trim the detected beat positions down to between the first and last positions which exceed a threshold.

The error here is that we use slice notation, which excludes the upper limit.  So we're actually trimming down to between the first and penultimate position that exceed the threshold.

**Expected behavior**

Not excluding that last good beat :shrug: 

**Software versions***

I've traced the git blame back to the beginning of the repo, so this is at least 11 years old.  I'm not sure if it's inherited from a similar thinko in dpwe's original matlab implementation.  However, it does seem that this logic was added to get regression tests to pass, so I expect this is the case.

https://github.com/librosa/librosa/commit/dbb54543023c4020f529b61599a7acd3f504fa98

bmcfee:
#### Reference Issue
Fixes #1695 
Follows up on #1130 
Fixes #1773 
Fixes #44 

This PR implements the extensions to various detection methods proposed in #1695 to support (dense, boolean) output and multi-channel input.

- [x] peak picking
- [x] onset detection
- ~[ ] onset backtracking (maybe out of scope, but could be improved)~
- [x] beat tracking

#### Any other comments?

Everything here should be backward-compatible, and at least as efficient as the existing implementations.


bmcfee:
Thinking through the beat tracking implementation a bit, I'll likely make the following changes:

1. Convert the dynamic program to populate a boolean array instead of position information.  After the beat positions have been filled in, we can use np.flatnonzero to identify their locations.
2. The above should allow us to jit-compile the DP, which should provide a healthy speed boost in the beat tracker.
3. Once the first two steps are done, then we can use numba guvectorize to run the tracker independently across channels.
4. Finally, update the top-level beat tracker function to appropriately handle units / sparse-dense like we do now in onsets.

bmcfee:
While working on the beat tracker, I'm taking some time to clean up some internal API and comments to be more useful and transparent.

bmcfee:
Circling back on a very old issue #44 (closed back in 2019 when we merged PLP) while my head is back in this code.  And yes, I recognize that it's somewhat a moot point, as the beat tracker we have is far behind SotA these days.  However, I still think there's value in at least documenting where the difficulties are, if not improving things to the point where it remains useful.

### Local scores

The first step of the beat tracker, after computing the onset strength envelope (OSE) and estimating tempo is to compute a "local score" for each frame.  This is essentially the OSE convolved with a very sharp window over a ±1 beat window:

https://github.com/librosa/librosa/blob/fb7013bab2d0270381f50bc6989224b0f53f2f91/librosa/beat.py#L447-L450

This is all easily implemented by a 1d conv when the tempo is fixed.  When the tempo is not fixed, we'd need either a separate filter for each frame, or some kind of aggregated filter.  The former is inefficient, but could possibly be computed quickly by sparsification.  The latter seems problematic to me.

### Back-searching

The next place where variable tempo could be problematic is in DP score calculation:

https://github.com/librosa/librosa/blob/fb7013bab2d0270381f50bc6989224b0f53f2f91/librosa/beat.py#L458-L475

where each frame looks back to a window around where we'd expect the previous beat to land.

bmcfee:
Update on beat tracking before i break on this for a bit.

The multichannel extension / refactor is almost done.  I'm running into some numba dispatching problems though, as it's unhappy about the following:

```python
@numba.guvectorize(
        [
            "void(float32[:], float32, float32, int32[:], float32[:])",
            "void(float64[:], float64, float32, int32[:], float64[:])",
        ],
        "(t),(),()->(t),(t)",
        nopython=True, cache=True)
def __beat_track_dp(localscore, frames_per_beat, tightness, backlink, cumscore):
```
where in general, `localscore` has shape `[n,m,p,..., n_frames]` and `frames_per_beat` has shape `[n,m,p...]`, matching `localscore` up to but not including the trailing dimension.  I suppose there's not enough in the layout string to enforce this agreement, and I'm not sure yet how to work around it (maybe extend the dimension of `frames_per_beat`?).

This might need to be reworked a little as a jitted 1d function that gets wrapped up in a numpy vectorize instead of directly using numba vectorize :sweat:.

bmcfee:
This still needs testing, but multichannel beat tracking is now functional.  With jit optimizations, we ended up being just marginally faster than the old implementation, so it's possible that there are more gains to be had here.

Next step is to expand the tests to cover the multichannel cases.

bmcfee:
:laughing: Rewriting the beat tracker has revealed an off-by-one error that has been trimming one too many beats since ... forever!

Here's the problem.  The current (main branch) implementation does the following when trimming beats:
https://github.com/librosa/librosa/blob/5ca70f5d57fc597660452f011ffff55f9c36ed4b/librosa/beat.py#L505-L516
that is, it finds all beat positions where the score exceeds a threshold, and then slices the index array from the first to the last.

The multi-channel branch rewrites this as an explicit loop:
https://github.com/librosa/librosa/blob/7f07f0b6c6fdb442ed98b95b7bd27ee22bd2a5ed/librosa/beat.py#L540-L553

The idea is the same, but note that when we use the array slice notation (main branch), it excludes the upper end of the range, even though the score at that index exceeds the threshold.  This is verified by printing out `smooth_boe, threshold` on an example track from our test suite (multichannel branch):
```python
In [1]: import librosa

In [2]: y, sr = librosa.load('data/test1_22050.wav')

In [3]: bpm, beats = librosa.beat.beat_track(y=y, sr=sr)
[7.36539798 8.16666805 8.83107379 8.38428126 4.82421986 1.50669484
 0.78669973 1.10853638 1.30460965 3.12087935 3.75983776 1.46726352] 2.599477926489801
```
The values printed out are numerically identical to those calculated on the main branch.  However, the main branch excludes the last valid beat (score 3.75983776) while the multichannel branch includes it.  :facepalm: 

This turns out to be the cause of our current test failures, where a display example using beat coords is triggering an error.

bmcfee:
I think aside from linting and final cleanups, this is ready for CR.

One thing that we should discuss though is how we want to handle user-provided tempo in multichannel beat tracking.  Previously, tempo could only ever be a scalar, so there was no ambiguity.  Scalar `bpm=` still works in the multichannel case, with broadcasting working as you might expect.

You can also provide a multichannel array into `bpm=` where a separate tempo is given for each channel.  If we're interpreting this as a scalar, it should have `bpm.ndim == onset_envelope.ndim-1` and `bpm.shape == onset_envelope.shape[:-1]` ; ie excluding the time dimension, but matching everywhere else.

The hitch here is that `feature.tempo` does not provide output in this shape.  Rather, it preserves the trailing dimension with `bpm.shape[-1] == 1`, meaning that each bpm is technically a 1d vector and not a scalar.  This is why we slice out the last dimension when estimating tempo internally:

https://github.com/librosa/librosa/blob/42ca94c6db3faf9384ada16bfaf5d7bbbed23a43/librosa/beat.py#L196-L203

The problem I'm anticipating is that we'd like to be able to do something like:

```python
>>> bpm = librosa.feature.tempo(y=y, sr=sr, ...)
>>> _, beats = librosa.beat.beat_track(y=y, sr=sr, bpm=bpm, ...)
```
but this will fail to vector dispatch properly:

```python
ValueError                                Traceback (most recent call last)
Cell In[5], line 1
----> 1 librosa.beat.beat_track(y=y, sr=sr, sparse=False, bpm=bpm)

File ~/git/librosa/librosa/beat.py:206, in beat_track(y, sr, onset_envelope, hop_length, start_bpm, tightness, trim, bpm, prior, units, sparse)
    197     bpm = _tempo(
    198         onset_envelope=onset_envelope,
    199         sr=sr,
   (...)
    202         prior=prior,
    203     )[..., 0]
    205 # Then, run the tracker
--> 206 beats = __beat_tracker(onset_envelope, bpm, float(sr) / hop_length, tightness, trim)
    208 if sparse:
    209     beats = np.flatnonzero(beats)

File ~/git/librosa/librosa/beat.py:454, in __beat_tracker(onset_envelope, bpm, frame_rate, tightness, trim)
    452 tail = __last_beat(cumscore)
    453 beats = np.zeros_like(onset_envelope, dtype=bool)
--> 454 __dp_backtrack(backlink, tail, beats)
    457 # Discard spurious trailing beats
    458 if trim:

File ~/miniconda3/envs/py311/lib/python3.11/site-packages/numba/np/ufunc/gufunc.py:171, in GUFunc.__call__(self, *args, **kwargs)
    167 def __call__(self, *args, **kwargs):
    168     # If compilation is disabled OR it is NOT a dynamic gufunc
    169     # call the underlying gufunc
    170     if self._frozen or not self.is_dynamic:
--> 171         return self.ufunc(*args, **kwargs)
    172     elif "out" in kwargs:
    173         # If "out" argument is supplied
    174         args += (kwargs.pop("out"),)

ValueError: output operand requires a reduction along dimension -1, but the reduction is not enabled. The dimension size of 1 does not match the expected output shape.
```

We could detect this case easily enough and fix it internally, eg:
```python
if bpm.ndim == onset_envelope.ndim:
    if bpm.shape[-1] != 1:
        raise ParameterError("blah blah blah")
    # Slice out the trailing time axis
    bpm = bpm[..., 0]
```
but it seems a little hacky.  Curious if other people have suggestions here.

bmcfee:
In pinning down the vectorized tempo API, it's probably worth thinking about the (hypothetical) variable-tempo case as well.  Mainly I'm trying to avoid painting ourselves into a corner here in case we do at some point want to revisit this idea.

The hunch lurking in the back of my mind is that the DP-based beat tracker could still be very useful with a sufficiently strong onset estimator, and most modern beat trackers do include some kind of DP, Viterbi, or otherwise sequential decoding for exactly this reason.  As such, I think it might be worth maintaining and even extending the beat tracker logic here even if the core method itself is rather antiquated, because it could find use as a utility function in more sophisticated models.


> Circling back on a very old issue #44 
> [...]
> ### Back-searching
> 
> The next place where variable tempo could be problematic is in DP score calculation:
> 
> https://github.com/librosa/librosa/blob/fb7013bab2d0270381f50bc6989224b0f53f2f91/librosa/beat.py#L458-L475
> 
> where each frame looks back to a window around where we'd expect the previous beat to land.

Now that we've rewritten the backlink search as a jitted function, we actually have a bit more flexibility to support a variable search window for each time step.  The existing implementation uses a preallocated window that gets copied and truncated at each step, and really is a bit of a kludge to maintain as much vectorization as possible for efficiency purposes.  If instead, we unroll this into an explicit loop (which would be fast under numba), we can let each frame have a different tempo and hence a different search window.

> ### Local scores
> 
> This is all easily implemented by a 1d conv when the tempo is fixed. When the tempo is not fixed, we'd need either a separate filter for each frame, or some kind of aggregated filter. The former is inefficient, but could possibly be computed quickly by sparsification. The latter seems problematic to me.

This part is still not great, but I do think implementing this quasi-convolution as a sparse near-toeplitz matrix multiply would be a viable solution here.  We can retain the conv1d version for static tempo of course.

### Multi-channel, multi-tempo :scream: 

The API difficulty here creeps in when we now have to consider the different ways in which a `bpm=` parameter can be scalar or multidimensional.  There are four cases to consider supporting.  In general, let's assume a multichannel onset envelope of shape `(m, n)`.

- scalar -  shared tempo across all time (and channels)
- ndarray, shape=(m,) - fixed tempo across time, different tempo for each channel
- ndarray, shape=(n,) - variable tempo across time, fixed across channels
- ndarray, shape=(m,n) - variable tempo across time and channels

There's an inherent ambiguity with cases (2) and (3) that I'm not wild about.  The obvious way to resolve this is to require expanded singleton dimensions in whichever direction makes the use case explicit.  However, this goes against the way we currently support tempo arguments (see above comment), so we should think carefully about how exactly we want to handle that.

bmcfee:
Remaining TODOs:
- [ ] Add docstring examples for sparse / multichannel mode
- [ ] Consider refactoring the beat tracker dp inner loop to explicitly unroll the vectorized cumulative score calculation
- [ ] Figure out a consistent way to handle tempo arrays

bmcfee:
I've rewritten the DP backsearch as an explicit loop rather than a vectorized argmax.  Speed is basically unaffected, and it would now be trivial to support time-varying tempo in this part of the algorithm.

I'm still struggling with getting vector dispatch to work consistently when `bpm.ndim == onset_envelope.ndim`, as described above.  Numba's guvectorize should be a little better about this, but so far I haven't had much luck - it always results in an all-pair broadcast, so that an input signal of shape `(2, n)` with a tempo estimate of `(2, 1)` turns into a localscore array of size `(2, 2, n)` and then the whole thing blows up on us.  In principle, guvectorize is supposed to handle the special case where the trailing dim has length 1, but maybe i'm doing something wrong here.

bmcfee:
> In principle, guvectorize is supposed to handle the special case where the trailing dim has length 1, but maybe i'm doing something wrong here.

I've now rewritten this using guvectorize and it seems to essentially work, at least in the sense that calling with fully vectorized tempo inputs does properly generalize independent single calls, eg with a stereo input:

```python
In [24]: onsets = librosa.onset.onset_strength(y=y, sr=sr, aggregate=np.median)

In [25]: bpm = librosa.feature.tempo(onset_envelope=onsets, sr=sr)

In [26]: t1, b1 = librosa.beat.beat_track(onset_envelope=onsets, bpm=bpm, sparse=False)

In [27]: t1a, b1a = librosa.beat.beat_track(onset_envelope=onsets[0], bpm=bpm[0], sparse=False)

In [28]: t1b, b1b = librosa.beat.beat_track(onset_envelope=onsets[1], bpm=bpm[1], sparse=False)

In [29]: t1, t1a, t1b
Out[29]: 
(array([[ 89.10290948],
        [117.45383523]]),
 array([89.10290948]),
 array([117.45383523]))

In [30]: np.allclose(b1[0], b1a)
Out[30]: True

In [31]: np.allclose(b1[1], b1b)
Out[31]: True
```

A problem does arise if you pass in a tempo vector that is not fully formed, eg if we were to flatten the `bpm` array to shape `(2,)` instead of shape `(2, 1)` as above:

```python
In [32]: t2, b2 = librosa.beat.beat_track(onset_envelope=onsets, bpm=bpm.flatten(), sparse=False)

In [33]: t2
Out[33]: array([ 89.10290948, 117.45383523])

In [34]: np.allclose(b1, b2)
Out[34]: False
```
In this case, it appears to be taking the first tempo (`bpm[0]`) and applying it to both channels.

The proper fix here, I think, is to use `expand_dims` to ensure that `bpm` has a trailing singleton dimension before calling into vectorized subroutines.

bmcfee:
I put in the expand_dims hack, and it appears to be functional now:

```python
In [1]: import librosa

In [2]: import numpy as np

In [3]: y, sr = librosa.load(librosa.ex('fishin', hq=True), mono=False)

In [4]: onsets = librosa.onset.onset_strength(y=y, sr=sr, aggregate=np.median)

In [5]: bpm = librosa.feature.tempo(onset_envelope=onsets, sr=sr)

In [6]: t1, b1 = librosa.beat.beat_track(onset_envelope=onsets, bpm=bpm, sparse=False)

In [7]: t2, b2 = librosa.beat.beat_track(onset_envelope=onsets, bpm=bpm.flatten(), sparse=False)

In [8]: np.allclose(b1, b2)
Out[8]: True
```

bmcfee:
Seems like this fails with 3-dimensional inputs.. will need to investigate further.

bmcfee:
> Seems like this fails with 3-dimensional inputs.. will need to investigate further.

Following up: I believe the implementation is correct.  The disagreement I was seeing came down to onset envelope calculation differing when given multichannel input.  This eventually traces back to the amplitude_to_db calculation, which is not channel-dependent.

bmcfee:
I'm going to take a crack at extending this to support time-varying tempo, because why not?

bmcfee:
Quick stab at a time-varying local score function:
```python
from librosa.util.exceptions import ParameterError
@numba.guvectorize(
        [
            "void(float32[:], float32[:], float32[:])",
            "void(float64[:], float64[:], float64[:])",
        ],
        "(t),(n)->(t)",
        nopython=True, cache=False)
def bls_numbav2(onset_envelope, frames_per_beat, localscore):

    N = len(onset_envelope)
    
    if len(frames_per_beat) == 1:
        # Static tempo mode
        window = np.exp(-0.5 * (np.arange(-frames_per_beat[0], frames_per_beat[0] + 1) * 32.0 / frames_per_beat[0]) ** 2)
        K = len(window)
        # This is a vanilla same-mode convolution
        for i in range(len(onset_envelope)):
            localscore[i] = 0.
            # we need i + K // 2 - k < N ==> k > i + K //2 - N
            # and i + K // 2 - k >= 0 ==>    k <= i + K // 2
            for k in range(max(0, i + K // 2 - N), min(i + K // 2, K)):
                localscore[i] += window[k] * onset_envelope[i + K//2 -k]
                
    elif len(frames_per_beat) == len(onset_envelope):
        # Time-varying tempo estimates
        # This isn't exactly a convolution anymore, since the filter is time-varying, but it's pretty close
        for i in range(len(onset_envelope)):
            window = np.exp(-0.5 * (np.arange(-frames_per_beat[i], frames_per_beat[i] + 1) * 32.0 / frames_per_beat[i]) ** 2)
            K = 2 * int(frames_per_beat[i]) + 1
            
            localscore[i] = 0.
            for k in range(max(0, i + K // 2 - N), min(i + K // 2, K)):
                localscore[i] += window[k] * onset_envelope[i + K // 2 - k]
    else:
        raise ParameterError(f"Invalid bpm shape={len(frames_per_beat)} does not match onset envelope shape={len(onset_envelope)}")
```

For static tempo, this is about 200μs slower than the straight ahead scipy implementation (70μs on fishin example with standard params) we previously had, but about 200μs faster than a numba-wrapped np.convolve (420μs).  The latter case is because until recently, numba's convolve implementation did not support modes, so we have to explicitly copy out the same-mode slice if using np.convolve.  The above implementation avoids this copy and works in-place on the pre-allocated output array.

For time-varying tempo, it takes about 1.3ms.  We could probably shave that down a bit, but I don't actually think it would be worth the added code complexity to do so.

bmcfee:
Dynamic tempo beat tracker is now implemented in this branch.  I haven't implemented tests yet, but so far it seems to be behaving exactly as I'd expect.  On the brahms example:

image

Red and blue show the current tracker (static tempo), orange and green show how it works with dynamic tempo.  Dashed lines are tempo estimates (Autocorrelation of oenv), solid lines are inter-beat intervals converted to BPM after running the detector.  The static tempo tracker is allowed to fluctuate from the static tempo, and it does so pretty severely when the tempo shifts.  The dynamic tracker follows the estimated tempo, as expected.

Neither of these are necessarily great, though I suspect most of the problems in the dynamic case are due to bad local tempo estimation, and not a failure of the tracking per se.  A better dynamic tempo estimator would probably improve the tracking results.

bmcfee:
Doing some more thorough testing here (static tempo case), I noticed that the DP behavior is now slightly different from what we had in prior stable versions.  For identical inputs (onset envelope saved to disk) in this branch vs 0.9.2, I've verified that the computation agrees numerically up to the local score step. The next step is the DP, which in this branch produces the following backlink table:
```
[-1, -1, -1, -1, -1, -1, -1, -1, -1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3, 6, 7, 7, 7, 7, 7, 7, 14, 14, 14, 14, 14, 14, 14, 14, 14, 23, 23, 24, 24, 24, 25, 25, 25, 30, 31, 31, 31, 32, 33, 33, 34, 34, 34, 41, 42, 42, 42, 42, 42, 42, 43, 48, 48, 49, 49, 50, 51, 52, 53, 53, 53, 58, 58, 60, 60, 61, 61, 62, 62, 62, 62, 67, 68, 69, 70, 70, 71, 72, 72, 76, 76, 77, 77, 80, 80, 81, 81, 81, 81, 81, 81, 87, 88, 88, 89, 90, 90, 93, 94, 96, 97, 97, 98, 99, 99, 100, 100, 100, 100, 101, 106, 106, 108, 108, 109, 109, 113, 114, 116, 116, 116, 116, 117, 117, 117, 118, 118, 120, 123, 126, 126, 127, 127, 127, 132, 132, 133, 133, 134, 134, 135, 135, 136, 136, 136, 137, 143, 144, 144, 144, 144, 145, 149, 150, 151, 151, 151, 153, 153, 154, 154, 154, 154, 160, 160, 161, 161, 161, 162, 163, 167, 168, 168, 170, 170, 170, 171, 171, 171, 173, 173, 178, 178, 179, 179, 179, 180, 181, 185, 186, ...
```
where the special token value of -1 indicates no prior link is possible.
In the 0.9.2 branch, it produces the following:
```
[-1, -1, -1, -1, -1, -13, -12, -11, -10, -9, -8, -7, -6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 6, 7, 7, 7, 7, 13, 13, 13, 13, 13, 13, 14, 14, 14, 14, 14, 14, 23, 24, 24, 24, 28, 29, 29, 30, 30, 31, 31, 31, 33, 33, 34, 34, 34, 36, 36, 42, 43, 43, 43, 45, 47, 47, 48, 49, 49, 49, 51, 52, 53, 53, 53, 54, 54, 60, 61, 62, 62, 63, 63, 64, 66, 67, 67, 68, 70, 70, 71, 72, 72, 72, 73, 73, 80, 80, 80, 81, 81, 81, 81, 81, 86, 87, 88, 88, 89, 90, 90, 91, 91, 96, 97, 98, 98, 99, 99, 100, 100, 100, 100, 105, 106, 106, 108, 108, 109, 109, 109, 114, 116, 116, 116, 116, 117, 117, 117, 118, 118, 123, 123, 126, 126, 127, 127, 127, 128, 132, 133, 133, 134, 134, 135, 135, 136, 136, 136, 140, 143, 144, 144, 144, 144, 145, 145, 150, 151, 151, 151, 153, 153, 154, 154, 154, 160, 160, 160, 161, 161, 161, 162, 163, 163, 168, 168, 170, 170, 170, 171, 171, 171, 173, 176, 178, 178, 179, 179, 179, 180, 181, 181, 186, ...
```
which is not only different, but clearly incorrect because we should never have negative values here apart from -1.  I'm not sure exactly where these came from, but it does seem like a bug.

This never appeared in the final output because the backtracking loop discards any negative values.  However, it does seem to corrupt the cumulative score computation, and the resulting detections are slightly different.

bmcfee:
I've tracked the change in behavior down to this bit:
https://github.com/librosa/librosa/blob/27d532911d144682faa5d87d454d4f1de54ed0c5/librosa/beat.py#L579-L584
which only accumulates a previous score if a valid backlink exists.  Otherwise, it just takes the localscore at the current frame, essentially treating it as the beginning.

In the old implementation, the cumulative DP score was updated whether or not the backlink was going to be populated:
https://github.com/librosa/librosa/blob/5ca70f5d57fc597660452f011ffff55f9c36ed4b/librosa/beat.py#L473-L486

The effect of this is subtle, but it does add up, and does produce slightly different detections at the end of the process.

I do think the new behavior is more correct though: if we know that a backlink will not be taken, it should not factor into the cumulative score.  Whether this constitutes a "bug" or a "thinko" is debatable, and that has implications for whether we schedule this PR for 0.10.2 or the next major revision.

bmcfee:
Circling back on this, i benchmarked the beat tracker before and after the proposed bug fix in the dp algorithm, using them isophonics set.

Tldr, the proposed change results in a slight dip in pretty much all metrics using the default parameters. Will have to investigate this further.

bmcfee:
Circling back on this, i benchmarked the beat tracker before and after the proposed bug fix in the dp algorithm, using them isophonics set.

Tldr, the proposed change results in a slight dip in pretty much all metrics using the default parameters. Will have to investigate this further.

bmcfee:
> Will have to investigate this further.

Digging back in now, and it seems like the score drop is probably coming from changes in the trimming behavior.  With trimming enabled (default), this PR is trimming one beat more than main.  With trimming disabled, it continues to estimate beats to the end of the track where the old implementation did not.  In both cases, I think this was erroneous behavior in the old implementation that nonetheless achieved higher scores on average.

bmcfee:
> With trimming disabled, it continues to estimate beats to the end of the track where the old implementation did not.

Aha, I see what's happening here.  The trimming logic was a little too clever by half:

https://github.com/librosa/librosa/blob/e8ffe19e66a37e4445d3e6ac0ec26b4dd825434a/librosa/beat.py#L509-L514

When `trim=True`, a nontrivial threshold is set, and things behave as expected.  When `trim=False`, the strict inequality means that beats at positions with 0 onset strength are discarded (ie, we still trimed!).  The new implementation, so far, does not preserve this logic, if `trim=False`, we skip the trim function entirely and all beat positions are returned.

I suppose for the sake of consistency within a minor revision (0.10.1 → 0.10.2) we probably should preserve this behavior, but it might make sense to disable that going forward.

bmcfee:
> I suppose for the sake of consistency within a minor revision (0.10.1 → 0.10.2) we probably should preserve this behavior, but it might make sense to disable that going forward.

This is now implemented.  The new implementation agrees with the old one on all metrics to the first two decimal places (isophonics/beatles, again).  The only substantive difference in behavior now should be the fix of #1773 , and I can live with that even if it dips performance in the third decimal place.

At this point, I think this PR is done.

bmcfee:
> At this point, I think this PR is done.

Ok, done except for a little bit of test polish and some comment cleanup.  But code-wise, this is ready for a CR if anyone's up for it?

bmcfee:
I recognize that this PR is probably un-reviewable. :shrug:

That said, if anyone wants to do a skim of the documentation sections, it would be most helpful.  I think the code is likely fine.

For a bit of extra sanity checking, I've cooked up a little demo notebook that shows how the variable tempo tracker works on a synthetic accelerating pulse: https://gist.github.com/bmcfee/07df182c8336b92f0c801fe01b488630

Here's the rendered click track (synthetic pulse + click) for the static tracker output:

[static_tempo.webm](https://github.com/librosa/librosa/assets/1190540/f89d7352-0166-42fb-89d8-57b92fb9cefb)

(definitely wrong sounding)

and for dynamic:

[dynamic_tempo.webm](https://github.com/librosa/librosa/assets/1190540/7b67e205-7d73-4b10-82ee-38bfe4dbfc32)

The failure at the end is due to mis-estimation of tempo at the end of the track as a result of boundary padding.  Otherwise, this seems to be performing to spec.

bmcfee:
For funsies, I redid the audio example above using a snare drum sample and put together an advanced demo notebook to show how it works.  Punchline image:

image

Is this overkill?  Yeah probably - PLP can already kinda do this job.  But it's nice to remove the (approximately) static tempo requirement of DAn's tracker since we can. :shrug:

bmcfee:
Just a bit more detail on the example, here are the accuracy results from `mir_eval`:

|         |   F-measure |   Cemgil |   Cemgil Best Metric Level |   Goto |   P-score |   Correct Metric Level Continuous |   Correct Metric Level Total |   Any Metric Level Continuous |   Any Metric Level Total |   Information gain |
|:--------|------------:|---------:|---------------------------:|-------:|----------:|----------------------------------:|-----------------------------:|------------------------------:|-------------------------:|-------------------:|
| static  |    0.537313 | 0.433658 |                   0.433658 |      0 |  0.375    |                         0.0833333 |                    0.0833333 |                      0.208333 |                 0.208333 |           0.370998 |
| dynamic |    1        | 0.819831 |                   0.819831 |      1 |  1        |                         1         |                    1         |                      1        |                 1        |           0.594384 |
| plp     |    0.968421 | 0.738856 |                   0.738856 |      1 |  0.979167 |                         0.729167  |                    0.916667  |                      0.729167 |                 0.916667 |           0.445927 |


And the sonified outputs:

[static_tempo](https://github.com/librosa/librosa/assets/1190540/105a7598-8dab-4c4e-be38-4282b2ed6294)

[dynamic_tempo](https://github.com/librosa/librosa/assets/1190540/92aacc71-c7bb-4c31-a3fe-04ac0fbc47fa)

[plp](https://github.com/librosa/librosa/assets/1190540/97e7c684-48fa-4fab-826e-ca61875664f9)

docs/examples/plot_dynamic_beat.py
```diff
@@ -0,0 +1,132 @@
...
+
+#########################################
+# --------------------------
+# Static tempo beat tracking
+# --------------------------
+# If we run the beat tracker on this signal directly and sonify the result as a click
+# track, we can see that it does not follow particularly well.
+# This is because the beat tracker is assuming a single (average) tempo across the
+# entire signal.
+# Note: here we are disabling trimming because we know the beats carry through to
```
lostanlen, line 63:
```python
# Note: by default, `beat_track` assumes that there is silence at the end of the signal, and trims last beats to avoid false detections. Here, there is no silence at the end, so we set `trim=False` to avoid this effect.
```

docs/examples/plot_dynamic_beat.py
```diff
@@ -0,0 +1,132 @@
...
+                             click_duration=0.25, length=len(y))
+
+print(f"Tempo estimate: {tempo[0]:.2f} BPM")
+Audio(data=y+click_track, rate=sr)
+
+#########################################
+# -------------
+# Dynamic tempo
+# -------------
+# Instead of using a single tempo, we can supply an array of tempo estimates
```
lostanlen, line 78: Instead of estimating an aggregated tempo for the whole signal, we now estimate a time-varying tempo. For this purpose, we pass `aggregate=None` to `librosa.feature.tempo`. The return value `tempo_dynamic`, is an array:

docs/examples/plot_dynamic_beat.py
```diff
@@ -0,0 +1,132 @@
...
+# to the beat tracker.  This can be computed by disabling aggregation in the
+# tempo estimator as follows.
+
+tempo_dynamic = librosa.feature.tempo(y=y, sr=sr, aggregate=None, std_bpm=4)
+
+# %%
+# Here, we've expanded the standard deviation of the tempo estimator from its
```
lostanlen, line 85: The parameter `std_bpm` is the standard deviation of the tempo estimator and has a default value of 1. Here, we have increased `std_bpm` to 4.

docs/examples/plot_dynamic_beat.py
```diff
@@ -0,0 +1,132 @@
...
+# default of 1 to 4.  This is to account for the broad range of tempo drift in this
+# particular recording (30 BPM to 240 BPM).
+# We can plot the dynamic tempo estimate, along with the static estimate, to see how
```
lostanlen, line 88: no commas

docs/examples/plot_dynamic_beat.py
```diff
@@ -0,0 +1,132 @@
...
+# they differ.
+
+fig, ax = plt.subplots()
+times = librosa.times_like(tempo_dynamic, sr=sr)
+ax.plot(times, tempo_dynamic, label='Dynamic tempo estimate')
+ax.axhline(tempo, label='Static tempo estimate', color='r')
+ax.legend()
+ax.set(xlabel='Time [sec]', ylabel='Tempo [BPM]')
```
lostanlen, line 96: `Time (s)` 
`Tempo (BPM)`

SI abbreviation, parentheses not brackets for consistency with 
https://librosa.org/doc/0.8.1/generated/librosa.beat.tempo.html

docs/examples/plot_dynamic_beat.py
```diff
@@ -0,0 +1,132 @@
...
+
+# %%
+# The dynamic tempo estimate is computed (by default) using a 8-second sliding
```
lostanlen, line 99: `librosa.feature.tempo` estimates tempo over a sliding window whose duration `ac_size` is equal to 8 seconds by default.
In this example, the result is not perfect: `tempo_dynamic` contains jagged and nonuniform steps. However, it does provide a rough picture of how tempo is changing over time.

docs/examples/plot_dynamic_beat.py
```diff
@@ -0,0 +1,132 @@
...
+
+fig, ax = plt.subplots()
+librosa.display.waveshow(y=y, sr=sr, ax=ax, color='k', alpha=0.7)
+ax.vlines(beats_static, 0.75, 1., label='Static tempo beat estimates', color='r')
+ax.vlines(beats_dynamic, -1, -0.75, label='Dynamic tempo beat estimates', color='b', linestyle='--')
+ax.legend()
+
+
+# %%
+# In the figure above (and sonified click track), it can be observed that the
```
lostanlen, line 128: In the figure above, we observe that `beats_dynamic`, the beat estimates derived from dynamic tempo, align well to the audio recording. On the contrary, `beats_static` suffers from alignment problems. Indeed, it has too many detections towards the beginning of the recording and too few detections towards the end of the recording.

docs/multichannel.rst
```diff
@@ -138,13 +138,14 @@ expectations.
...
-**Detectors** with ragged output, for example beat tracking (`librosa.beat`) and
+**Detectors** with sparse output (i.e., frame indices), for example beat tracking (`librosa.beat`) and
 onset detection (`librosa.onset.onset_detect`) do not support multi-channel inputs.
 This is because the output may have differing numbers of events in each channel, and
 therefore cannot be consistently stored in a `numpy.ndarray` output object.
-In these cases, it is best to either process each channel separately (if they are
-truly independent) or aggregate representations across channels (e.g., by averaging
-features) if they are strongly related.
+
+Onset and beat detection **can** handle multi-channel input **only** if the `sparse=False` parameter is provided,
```
lostanlen, line 146: For multi-channel onset detection or multi-channel beat detection, you must set `sparse=False`. The output will be encoded as dense multi-channel arrays with boolean values.

librosa/beat.py
```diff
@@ -60,37 +63,69 @@ def beat_track(
...
+
     trim : bool [scalar]
         trim leading/trailing beats with weak onsets
-    bpm : float [scalar]
+
+    bpm : float [scalar] or np.ndarray [shape=(...)]
         (optional) If provided, use ``bpm`` as the tempo instead of
         estimating it from ``onsets``.
+
+        If multichannel, tempo estimates can be provided for all channels.
```
lostanlen, line 91: `multi-channel` 
(for consistency within the library)

librosa/beat.py
```diff
@@ -60,37 +63,69 @@ def beat_track(
...
+
+        If multi-channel and ``bpm`` is not provided, a separate
+        tempo will be returned for each channel
+    beats : np.ndarray
+        estimated beat event locations.
+
+        If `sparse=True` (default), beat locations are given in the specified units
+        (default is frame indices).
+
+        If `sparse=False` (required for multichannel input), beat events are
```
lostanlen, line 127: `multi-channel` 
(for consistency within the library)

librosa/beat.py
```diff
@@ -60,37 +63,69 @@ def beat_track(
...
         By default, 'frames' are used.
 
+    sparse : bool
+        If ``True`` (default), detections are returned as an array of frames,
+        samples, or time indices (as specified by ``units=``).
+
+        If ``False``, detections are encoded as a dense boolean array where
+        ``beats[..., n]`` is true if there's a beat at frame index ``n``.
+
+        .. note:: multi-channel input is only supported when ``sparse=False``.
```
lostanlen, line 112: `if` not `when`

librosa/onset.py
```diff
@@ -169,23 +188,27 @@ def onset_detect(
         kwargs.setdefault("delta", 0.07)
 
         # Peak pick the onset envelope
-        onsets = util.peak_pick(onset_envelope, **kwargs)
+        onsets = util.peak_pick(onset_envelope, sparse=sparse, axis=-1, **kwargs)
 
         # Optionally backtrack the events
         if backtrack:
+            if not sparse:
+                raise ParameterError("onset backtracking is only supported when sparse=True")
```
lostanlen, line 196: `if` not `when`

librosa/onset.py
```diff
@@ -69,6 +70,8 @@ def onset_detect(
 
         This is primarily useful when using onsets as slice points for segmentation.
 
+        .. note:: backtracking is only supported when ``sparse=True``.
```
lostanlen, line 73: `if` not `when`

librosa/onset.py
```diff
@@ -80,20 +83,32 @@ def onset_detect(
...
         Otherwise, the onset envelope is left unnormalized.
 
+    sparse : bool
+        If ``True`` (default), detections are returned as an array of frames,
+        samples, or time indices (as specified by ``units=``).
+
+        If ``False``, detections are encoded as a dense boolean array where
+        ``onsets[n]`` is true if there's an onset at frame index ``n``.
+
+        .. note:: multi-channel input is only supported when ``sparse=False``.
```
lostanlen, line 93: `if` not `when`

librosa/onset.py
```diff
@@ -80,20 +83,32 @@ def onset_detect(
 
         Otherwise, the onset envelope is left unnormalized.
 
+    sparse : bool
+        If ``True`` (default), detections are returned as an array of frames,
+        samples, or time indices (as specified by ``units=``).
+
+        If ``False``, detections are encoded as a dense boolean array where
+        ``onsets[n]`` is true if there's an onset at frame index ``n``.
```
lostanlen, line 91: true -> True

librosa/beat.py
```diff
@@ -60,37 +63,69 @@ def beat_track(
...
     units : {'frames', 'samples', 'time'}
         The units to encode detected beat events in.
         By default, 'frames' are used.
 
+    sparse : bool
+        If ``True`` (default), detections are returned as an array of frames,
+        samples, or time indices (as specified by ``units=``).
+
+        If ``False``, detections are encoded as a dense boolean array where
+        ``beats[..., n]`` is true if there's a beat at frame index ``n``.
```
lostanlen, line 110: true -> True

librosa/beat.py
```diff
@@ -167,9 +212,16 @@ def beat_track(
             y=y, sr=sr, hop_length=hop_length, aggregate=np.median
         )
 
+    if sparse and onset_envelope.ndim != 1:
+        raise ParameterError("input array must be one-dimensional if sparse=True")
```
lostanlen, line 216: `"sparse=True (default) does not support {}-dimensional inputs. Consider passing sparse=False or converting to mono.`

with `{}` formatting to `onset_envelope.ndim`

librosa/beat.py
```diff
@@ -406,111 +468,224 @@ def __beat_tracker(
...
-        raise ParameterError("bpm must be strictly positive")
+    if np.any(bpm <= 0):
+        raise ParameterError(f"bpm={bpm} must be strictly positive")
+
+    if tightness <= 0:
+        raise ParameterError("tightness must be strictly positive")
+
+    # TODO: this might be better accomplished with a np.broadcast_shapes check
+    if bpm.shape[-1] not in (1, onset_envelope.shape[-1]):
+        raise ParameterError(f"Invalid bpm shape={bpm.shape} does not match onset envelope shape={onset_envelope.shape}")
```
lostanlen, line 479: untested?

librosa/util/utils.py
```diff
@@ -1205,6 +1205,48 @@ def localmin(x: np.ndarray, *, axis: int = 0) -> np.ndarray:
...
+        maxn = np.max( x[max(0, n-pre_max):min(n+post_max, x.shape[0])])
+
+        # Are we the local max and sufficiently above average?
+        peaks[n] = (x[n] == maxn) 
+        
+        if not peaks[n]:
+            n += 1
+            continue
+
+        avgn = np.mean(x[max(0, n-pre_avg):min(n+post_avg, x.shape[0])])
```
lostanlen, line 1239: i'm going to trust you on this one!

librosa/util/utils.py
```diff
@@ -1292,15 +1348,12 @@ def peak_pick(
...
     if post_max <= 0:
         raise ParameterError("post_max must be positive")
-
     if post_avg <= 0:
         raise ParameterError("post_avg must be positive")
-
-    if x.ndim != 1:
-        raise ParameterError("input array must be one-dimensional")
+    if sparse and x.ndim != 1:
+        raise ParameterError("input array must be one-dimensional if sparse=True")
```
lostanlen, line 1356: same proposal as before

librosa/util/utils.py
bmcfee: The computation here is pretty much line-for-line identical to the old (non-numba) implementation.

The only significant change to the algorithm is the variable increment to `n` (line 1247).  This implements the `wait` period between detections without having to do the intermediate computations on frames that we know will not trigger detection.  The output of the function is unchanged, but it's significantly faster.

bmcfee:
Thanks @lostanlen !  I've updated the text mostly taking your suggestions (with some minor tweaks here and there).  Exception messages have also been expanded per your suggestion, though the phrasing is different in the two cases you identified because one has to do with onsets/signals and the other with arbitrary data arrays.

bmcfee:
Doc CI failure is due to a temporary outage of the scipy procedings archive, nothing to be done from our side.

bmcfee:
> It looks like this conditional branch in `beat.py` in still untested?

Yeah - I'm not too worried about this.  It's a redundant check within a private function, and the user-facing function (`beat_track`) checks are tested.  I left this check in place primarily for debugging purposes, and since it's not hurting anything, I'd prefer to leave it here in case we need it later on.

lostanlen (approved):

PR #1768, opened 2023-10-12
Allow amplitude_to_db to accept scalars

matangover:
`amplitude_to_db` can now accept a scalar. This brings it in line with `db_to_amplitude` and similar methods.

```python
# Without this fix:
>>> librosa.amplitude_to_db(2)
  File "<stdin>", line 1, in <module>
  File "/Users/matan/code/librosa/librosa/core/spectrum.py", line 1874, in amplitude_to_db
    power = np.square(magnitude, out=out_array)
TypeError: return arrays must be of ArrayType

# With this fix:
>>> librosa.amplitude_to_db(2)
6.020599913279624
```

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1462,6 +1462,11 @@ def test_amplitude_to_db_complex():
     assert np.allclose(db1, db2)
 
 
+def test_amplitude_to_db_scalar():
+    assert librosa.amplitude_to_db(1) == 0
+    assert np.isclose(librosa.amplitude_to_db(2), 6.0206)
+
+
 @pytest.mark.parametrize("ref_p", range(-3, 4))
 @pytest.mark.parametrize("xp", [(np.abs(np.random.randn(1000)) + 1e-5) ** 2])
 def test_db_to_power_inv(ref_p, xp):
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """Utilities for spectral processing"""
+from __future__ import annotations
 import warnings
 
 import numpy as np
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1869,7 +1869,8 @@ def amplitude_to_db(
     else:
         ref_value = np.abs(ref)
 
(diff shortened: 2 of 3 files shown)


bmcfee:
Thanks for this - can you add a test case to hit this path?

tests/test_core.py
```diff
@@ -1462,6 +1462,11 @@ def test_amplitude_to_db_complex():
     assert np.allclose(db1, db2)
 
 
+def test_amplitude_to_db_scalar():
+    assert librosa.amplitude_to_db(1) == 0
+    assert np.isclose(librosa.amplitude_to_db(2), 6.0206)
```
bmcfee, line 1467: It looks like this test is now triggering failures in the type checking.  Probably the type annotations need to be revised to allow for scalar inputs.

bmcfee: Looking at this now with the type overloads - do we need to do this also for the other db/amplitude/power converters?

matangover: Not in this PR

bmcfee: Can you clarify why not?  It seems like you're proposing to expand the functionality of one function in a family of related functions, but not the others.  We should keep this all internally consistent, especially since some of these are inverses of each other.

matangover: Sorry, I wasn't clear. I don't have the ability right now to contribute more than this small change. It's the contrary of what you say, I expanded the functionality of amplitude_to_db to match the others, so this change improves the API consistency. Then I added a test as you requested, and then I fixed the typings :)

bmcfee: Ok, thanks for clarifying!

I understand the point about expanding the API to match what's supported by other functions.  My question is about expanding the type annotations of the other functions to make this behavior (scalar input support) explicit, since we would now have it for only one of the set.

bmcfee: (one month later) .. sorry to drop the thread here, but I'm not sure where we stand on this.  I do think that we should implement this change consistently across all relevant functions in this PR.  Is that something you'd be interested in doing, or should I try to chip in contributions toward that as well?

matangover: You are more than welcome to chip in! I'm unfortunately still unable to continue contributing on this.

bmcfee:
Closing this PR to continue over in #1790

Issue #1760, opened 2023-09-25
Should `key_to_notes` print naturals? (♮)


lostanlen:
In the wake of #1756, i have looked more closely at `key_to_notes` today. I have a small dissatisfaction with the way we currently treat some pitch classes outside the scale or mode.

Example from the documentation:
```python
>>> librosa.key_to_notes('G#:maj')
['B♯', 'C♯', 'D', 'D♯', 'E', 'E♯', 'F♯', 'F𝄪', 'G♯', 'A', 'A♯', 'B']
```

Here's my logic: we're in G# major, so the key signature has six sharps and one double-sharp. So presumably`D` sounds like`D#` unless explicitly stated otherwise. More precisely, if a place a notehead at the `D` position on the staff, and we're in the key of `G#` major, what I hear has a pitch of `D#`, not a `D` natural.

Hence my proposal to change the output to 
```python
['B♯', 'C♯', 'D♮', 'D♯', 'E♮', 'E♯', 'F♯', 'F𝄪', 'G♯', 'A♮', 'A♯', 'B♮']
```
(note the presence of natural accidentals as suffixes to `D`, `E`, `A`, and `B`)

Here's is another, more common example: D major. We have two sharps. Current behavior:
```python
>>> librosa.key_to_notes('D:maj')
['C', 'C♯', 'D', 'D♯', 'E', 'F', 'F♯', 'G', 'G♯', 'A', 'A♯', 'B']
```

Proposed behavior
```python
>>> librosa.key_to_notes('D:maj')
['C♮', 'C♯', 'D', 'D♯', 'E', 'F♮', 'F♯', 'G', 'G♯', 'A', 'A♯', 'B']
```

Maybe this has drawbacks i haven't considered, or is considered too pedantic / redundant. I'm open to strike a compromise by introducing a new keyword argument to toggle these naturals in or out.

For the record, i do not suggest that we implement double naturals, which are prone to confusion and outdated in music notation anyway: https://en.wikipedia.org/wiki/Natural_(music)#Double_natural

bmcfee:
One big question here is how we'd typeset things when Unicode=false.

bmcfee:
One big question here is how we'd typeset things when Unicode=false.

.. but otherwise I'm generally in favor of this proposal. I like the idea of making it opt-in with a parameter and preserving backwards compatibility.

[...]

PR #1770, opened 2023-10-16
Natural accidentals for out-of-key notes

erabinov:
#### Reference Issue
Should key_to_notes print naturals? (♮) #1760

As mentioned by @lostanlen, it makes sense to put explicit naturals in keys where the corresponding note has an accidental (e.g. C# appears in the D-major scale, so ``key_to_notes('D:maj')`` should contain ``C♮`` rather than ``C``).

Based on our discussion, I've added a new keyword argument in ``key_to_notes``, ``natural``, which, if set to ``True``, implements this feature.

#### Any other comments?

If printing in ASCII, ``♮`` will be replaced by ``n``. A user who has not seen this convention before could guess what it means, but maybe to be more explicit we could use ``nat`` instead?


bmcfee:
Thanks @erabinov !  At a quick glance this looks great.

The main thing that jumped out to me though is that you've added `natural=True` as default, which is technically an API-breaking change.  I'd be open to making this the default in future versions, but for the 0.10.x series, we should maintain strict backward compatibility (`natural=False` by default).

Otherwise, I'll try to give this a more thorough CR tomorrow.

erabinov:
@bmcfee  Got it. Makes a lot of sense. I'll fix and recommit.

erabinov:
By the way, do you know off the top of your head why the linkcheck in the CI Docs workflow would fail?

lostanlen:
thanks @erabinov 

i agree with @bmcfee that we want `natural=False` by default, at least for the time being.

This docstring is fine but not super clear

> If ``True`` (default), print explicit natural symbol for any note which appears with an accidental in the corresponding scale (e.g. ``C♮`` will appear in ``key_to_notes("D:maj")``).



Suggestion:

> If ``True'', mark natural accidentals with a natural symbol (♮). For example, in the key of D major (`key=D:maj`), the penultimate note name is `C` if `key=False` and `C♮` if `key=True`.

librosa/core/notation.py
```diff
@@ -671,22 +676,22 @@ def key_to_notes(key: str, *, unicode: bool = True) -> List[str]:
     `A♯:min` will use sharps, but spell note 0 (`C`) as `B♯`
 
     >>> librosa.key_to_notes('A#:min')
-    ['B♯', 'C♯', 'D', 'D♯', 'E', 'E♯', 'F♯', 'G', 'G♯', 'A', 'A♯', 'B']
```
bmcfee, line 674: These docstring examples should be reverted if `natural=False` is default, and we should add a case or two to cover the `natural=True` setting.

librosa/core/notation.py
```diff
@@ -131,7 +131,7 @@
 
 NOTE_RE = re.compile(
     r"^(?P<note>[A-Ga-g])"
-    r"(?P<accidental>[#♯𝄪b!♭𝄫♮]*)"
+    r"(?P<accidental>[#♯𝄪b!♭𝄫♮n]*)"
```
bmcfee, line 134: Should we add support for this in the key parser as well?  I know it would be weird to have something like `G♮:maj`, but I could imagine it coming up in some weird circumstances (eg parsing a key derived from a modulation).

At any rate, it doesn't seem obviously weirder than `G𝄫:maj`, unless I'm missing something.

librosa/core/notation.py
```diff
@@ -805,9 +810,17 @@ def key_to_notes(key: str, *, unicode: bool = True) -> List[str]:
 
         notes = notes_flat
 
+    if natural:
+        # Apply natural signs to any note which has no other accidentals and does not appear in the scale for key.
+        scale_notes = key_to_degrees(key)
+        for place, note in enumerate(notes):
+            match = NOTE_RE.match(note)
+            if match and match.group('accidental')=='' and __note_to_degree(note) not in scale_notes:
+                notes[place] = match.group('note')+'♮'
```
bmcfee, line 819: We can skip some unnecessary regex parsing here by early skipping over notes in the scale:
```suggestion
        scale_notes = key_to_degrees(key)
        for place, note in enumerate(notes):
            if __note_to_degree(note) in scale_notes:
                continue
            match = NOTE_RE.match(note)
            if match and match.group('accidental')=='':
                notes[place] = match.group('note')+'♮'
```
I think we might still be doing some redundant work here though, since `__note_to_degree` also does a NOTE_RE.match call.  Maybe there's a clever way to simplify this?

bmcfee:
> This is happening across all builds right now. I think it's a combination of zenodo changing some URLs on us and possibly an update to the sphinx tools we're using. We can ignore it for now, I'll clean it up in a separate PR.

Scratch that - it turned out to be a temporary outage with one of the sites the docs link to.  Things are working again now.

librosa/core/notation.py
erabinov: I'm happy to do this. The simplest change will treat `G♮:maj` as equal to `G:maj`, which may or may not be in keeping with the spirit of the issue raised by @lostanlen. Since this is a completely theoretical key signature, I think it's probably fine to treat `G♮:maj` as `G:maj`.

bmcfee: Yeah, I tend to agree - it seems like the only consistent way to handle it.  Would "G♮" ever mean anything *other* than "G", given how we already parse unaccidented notes?

lostanlen:
@erabinov my bad, thanks for correcting

librosa/core/notation.py
erabinov: I agree, especially because we treat `Gb#:maj` as `G:maj`. The natural sign has a theoretical function only once we've established a key or another context (e.g. a bar in which an accidented note has already appeared). The PR is about making use of this function of the natural sign. When the natural sign appears in a key name (i.e. before a context has been established), it doesn't really have a function in music theory, and therefore I think we're safe to treat it simply as an additive 0 to the key name.

librosa/core/notation.py
erabinov: I like that you check whether the note is in the scale first.

I tried a solution without regex parsing which relies on checking whether `note` has length 1, and if so, just adding a natural sign to it. This relies on the fact that `note` is guaranteed to match the regex, but perhaps there are some robustness issues with that assumption.

librosa/core/notation.py
```diff
@@ -23,7 +23,7 @@
     "list_mela",
     "list_thaat",
     "fifths_to_note",
-    "interval_to_fjs",
+    "interval_to_fjs"
```
erabinov, line 26:
```suggestion
    "interval_to_fjs",
```

librosa/core/notation.py
bmcfee: > checking whether `note` has length 1, and if so, just adding a natural sign to it.

Oh that's clever! :grin: 

I can't think of any reason why `len(note) == 1` would fail, as long as the notes have already been simplified (eg to rule out a failing throw on `Gb#`).  In unicode, accidentals are not combining symbols, so they will increment the string length.

librosa/core/notation.py
```diff
@@ -805,9 +824,18 @@ def key_to_notes(key: str, *, unicode: bool = True) -> List[str]:
 
         notes = notes_flat
 
+    # Apply natural signs to any note which has no other accidentals and does not appear in the scale for key.
+    if natural:
+        scale_notes = key_to_degrees(key)
```
bmcfee, line 829: Since we only use this to check membership, it's much more efficient to make a python set out of the degrees:
```suggestion
        scale_notes = set(key_to_degrees(key))
```
eg:
```
In [9]: degs = librosa.key_to_degrees('Db:min')

In [10]: degs_set = set(degs)

In [11]: %timeit 5 in degs
1.27 µs ± 7.08 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [12]: %timeit 5 in degs_set
11.3 ns ± 0.0256 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
```
A silly speedup, but why not.

librosa/core/notation.py
```diff
@@ -714,15 +733,15 @@ def key_to_notes(key: str, *, unicode: bool = True) -> List[str]:
     if multiple:
         sign_map = {+1: "♯", -1: "♭"}
         additional_acc = sign_map[np.sign(offset)]
-        intermediate_notes = key_to_notes(tonic+additional_acc*(abs(offset)-1)+':'+scale)
+        intermediate_notes = key_to_notes(tonic+additional_acc*(abs(offset)-1)+':'+scale, natural = natural)
         notes = [__simplify_note(note, additional_acc) for note in intermediate_notes]
```
bmcfee, line 737: Thinking about this: do we need to propagate `natural=natural` here?  If it's immediately followed by a call to `__simplify_note`, wouldn't a natural marker be automatically purged?

It seems to me like we should be able to call through explicitly with `natural=False` to bypass unnecessary decorations, and only do one fixup at the top-level call (handled below).

librosa/core/notation.py
```diff
@@ -714,15 +733,15 @@ def key_to_notes(key: str, *, unicode: bool = True) -> List[str]:
     if multiple:
         sign_map = {+1: "♯", -1: "♭"}
         additional_acc = sign_map[np.sign(offset)]
-        intermediate_notes = key_to_notes(tonic+additional_acc*(abs(offset)-1)+':'+scale)
+        intermediate_notes = key_to_notes(tonic+additional_acc*(abs(offset)-1)+':'+scale, natural = natural)
```
erabinov, line 736:
```suggestion
        intermediate_notes = key_to_notes(tonic+additional_acc*(abs(offset)-1)+':'+scale, natural = False)
```

librosa/core/notation.py
erabinov: I think that's right. I implemented this change. 

For the mode implementation, though, we should pass through `natural = natural`, I believe.

bmcfee: > For the mode implementation, though, we should pass through `natural = natural`, I believe.

Sorry, brain isn't parsing this for some reason.  What do you mean by "mode implementation"?

erabinov: When we call `key_to_notes` with parameter `key='G:loc'`, for example. The way this currently works is by simply returning `key_to_notes` with parameter `key='Ab:maj'`, so we need to pass the same value for `natural` through into this call.

bmcfee: Ah, got it - thanks for clarifying!

bmcfee:
CI failures are all codecov related - otherwise this all looks good to me.  Thanks again @erabinov !

I'll hold off on merging in case @lostanlen wants to give it a last pass through.

erabinov:
Happy to help!

bmcfee (approved):

lostanlen (approved):

librosa/core/notation.py
```diff
@@ -643,7 +643,14 @@ def key_to_notes(key: str, *, unicode: bool = True) -> List[str]:
...
 
-            ♯ -> #, 𝄪 -> ##, ♭ -> b, 𝄫 -> bb
+            ♯ -> #, 𝄪 -> ##, ♭ -> b, 𝄫 -> bb, ♮ -> n
+
+    natural : bool
+        If ``True'', mark natural accidentals with a natural symbol (♮).
+
+        If ``False`` (default), do not print natural symbols.
+
+        For example, `note_to_degrees('D:maj')[0]` is `C` if `natural=False` (default) and `C♮` if `natural=True`.
```
lostanlen, line 653: perfect

PR #1780, opened 2023-11-14
Fix docstring for the `writeable` argument of `util.frame` function

### LINKED ISSUE #1779: Docstring on the `writeable` argument of `util.frame` function is wrong
**Describe the bug**

In the docstring for `librosa.util.frame` function, it describes the `writeable` argument as:

```python
"""
...
    writeable : bool
        If ``True``, then the framed view of ``x`` is read-only.
        If ``False``, then the framed view is read-write.  Note that writing to the framed view
        will also write to the input array ``x`` in this case.
...
"""
```

This is the opposite of what it does.  The framed view is only read-write when the `writable` argument is `True`.  It's read-only when the `writeable` argument is `False`.

**To Reproduce**

Example:
```python
import numpy as np
import librosa

# array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
x = np.arange(10)

# Assignment is successful
frames = librosa.util.frame(x, frame_length=3, hop_length=2, writeable=True)
frames[0, 0] = -1

# Raises ValueError: assignment destination is read-only
frames = librosa.util.frame(x, frame_length=3, hop_length=2, writeable=False)
frames[0, 0] = -1
```

**Expected behavior**

The doc string matches what the argument actually does.

**Software versions**

macOS-14.0-arm64-arm-64bit
Python 3.11.5 (main, Aug 24 2023, 12:23:19) [Clang 15.0.0 (clang-1500.0.40.1)]
NumPy 1.26.2
SciPy 1.11.3
librosa 0.10.1
INSTALLED VERSIONS
\------------------
python: 3.11.5 (main, Aug 24 2023, 12:23:19) [Clang 15.0.0 (clang-1500.0.40.1)]

librosa: 0.10.1

audioread: 3.0.1
numpy: 1.26.2
scipy: 1.11.3
sklearn: 1.3.2
joblib: 1.3.2
decorator: 5.1.1
numba: 0.58.1
soundfile: 0.12.1
pooch: v1.8.0
soxr: 0.3.7
msgpack: 1.0.7

numpydoc: None
sphinx: None
sphinx_rtd_theme: None
matplotlib: 3.8.1
sphinx_multiversion: None
sphinx_gallery: None
mir_eval: None
ipython: None
pytest: None
pytest_mpl: None
pytest_cov: None
samplerate: None
resampy: None
presets: None
packaging: 23.2

alvingao:
#### Reference Issue

Fixes #1779 

This fixes the docstring for the `writeable` argument of `util.frame` function to match its actual behavior.

The framed view can only be written when the `writeable` argument is set to `True`.  The previous docstring description is the opposite of this behavior.

### DIFF
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -119,8 +119,8 @@ def frame(
     axis : int
         The axis along which to frame.
     writeable : bool
-        If ``True``, then the framed view of ``x`` is read-only.
-        If ``False``, then the framed view is read-write.  Note that writing to the framed view
+        If ``False``, then the framed view of ``x`` is read-only.
+        If ``True``, then the framed view is read-write.  Note that writing to the framed view
         will also write to the input array ``x`` in this case.
     subok : bool
         If True, sub-classes will be passed-through, otherwise the returned array will be


bmcfee (approved):
LGTM, thanks for catching this!

PR #1782, opened 2023-11-20
Updates matplotlib colormap usage to remove deprecation warnings.

### LINKED ISSUE #1781: Matplotlib colormap deprecation warnings in librosa/display.py
**Describe the bug**

```matplotlib=3.7.0``` deprecated ```matplotlib.cm.get_cmap(name=None, lut=None)``` per https://matplotlib.org/stable/api/cm_api.html#matplotlib.cm.get_cmap. This is causing ```librosa/display``` to throw warnings on downstream ```pytest```. This deprecation will be entirely removed in ```matplotlib=3.9.0```.

**To Reproduce**

Execute a test from ```tests/test_display.py``` (with --disable-pytest-warnings temporarily removed from ```setup.cfg```):

```bash
pytest tests/test_display.py::test_double_chroma
```

**Expected behavior**

Test should pass without warnings.

**Screenshots**

```bash
./librosa/librosa/display.py:859: MatplotlibDeprecationWarning: The get_cmap function was deprecated in Matplotlib 3.7 and will be removed two minor releases later. Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap(obj)`` instead.
    return mcm.get_cmap(cmap_seq)
```

**Software versions**

```python
Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35

Python 3.9.18 | packaged by conda-forge | (main, Aug 30 2023, 03:49:32) 
[GCC 12.3.0]

NumPy 1.26.0

SciPy 1.11.3

librosa 0.10.1

>>> librosa.show_versions()
INSTALLED VERSIONS
------------------
python: 3.9.18 | packaged by conda-forge | (main, Aug 30 2023, 03:49:32) 
[GCC 12.3.0]

librosa: 0.10.1

audioread: 3.0.1
numpy: 1.26.0
scipy: 1.11.3
sklearn: 1.3.2
joblib: 1.3.2
decorator: 5.1.1
numba: 0.58.1
soundfile: 0.12.1
pooch: v1.8.0
soxr: 0.3.7
msgpack: 1.0.6

numpydoc: None
sphinx: None
sphinx_rtd_theme: None
matplotlib: 3.8.1
sphinx_multiversion: None
sphinx_gallery: None
mir_eval: None
ipython: None
pytest: 7.4.3
pytest_mpl: 0.16.1
pytest_cov: 4.1.0
samplerate: 0.1.0
resampy: 0.4.2
presets: None
packaging: 23.2
```

**Additional context**
Nil.

philboske:
#### Reference Issue
Fixes #1781.

Changes colormap calls from ```matplotlib.cm.get_cmap(name)``` to ```matplotlib.colormaps[name]``` per deprecation warnings upstream on ```matplotlib>=3.7.0```.

#### Any other comments?

Reduces ```pytest``` warnings from:

```bash
13989 passed, 2 skipped, 522 xfailed, 10002 warnings in 479.77s (0:07:59)
```

to

```bash
13989 passed, 2 skipped, 522 xfailed, 9891 warnings in 383.95s (0:06:23)
```

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -41,7 +41,7 @@ from itertools import product
 import warnings
 
 import numpy as np
-import matplotlib.cm as mcm
+from matplotlib import colormaps as mcm
 import matplotlib.axes as mplaxes
 import matplotlib.ticker as mplticker
 import matplotlib.pyplot as plt
@@ -844,7 +844,7 @@ def cmap(
     data = np.atleast_1d(data)
 
     if data.dtype == "bool":
-        return mcm.get_cmap(cmap_bool, lut=2)
+        return mcm[cmap_bool]
 
     data = data[np.isfinite(data)]
 
@@ -856,9 +856,9 @@ def cmap(
     min_val, max_val = np.percentile(data, [min_p, max_p])
 
     if min_val >= 0 or max_val <= 0:
-        return mcm.get_cmap(cmap_seq)
+        return mcm[cmap_seq]
 
-    return mcm.get_cmap(cmap_div)
+    return mcm[cmap_div]
 
 
 def __envelope(x, hop):


bmcfee (approved):
LGTM, thanks for this!

librosa/display.py
```diff
@@ -844,7 +844,7 @@ def cmap(
     data = np.atleast_1d(data)
 
     if data.dtype == "bool":
-        return mcm.get_cmap(cmap_bool, lut=2)
+        return mcm[cmap_bool]
```
bmcfee, line 847: This one seems potentially wrong, since we're losing the explicit binary map implied by `lut=2`.  Can we still get that with the newer API?

philboske: If ```matplotlib``` detects what is normally considered binary data, it will binarise the ```cmap``` as this is a very common usage pattern in computer vision:

```python
import numpy as np

fig, ax = plt.subplots()
data = [[0, 1, 0, 1], [1, 0, 1, 0]]
ax.imshow(data, cmap="gray")
fig.savefig("test_int8_bool.png")

fig, ax = plt.subplots()
data = [[False, True, False, True], [True, False, True, False]]
ax.imshow(data, cmap="gray")
fig.savefig("test_bool.png")

fig, ax = plt.subplots()
data = [[0, 255, 0, 255], [255, 0, 255, 0]]
ax.imshow(data, cmap="gray")
fig.savefig("test_int8_max.png")

fig, ax = plt.subplots()
data = [[0.0, 1.0, 0.0, 1.0], [1.0, 0.0, 1.0, 0.0]]
ax.imshow(data, cmap="gray")
fig.savefig("test_float.png")
```

These will all produce:

test_bool

philboske: I would have to go digging through ```matplotlib``` issues and PR's to find the exact reason for removing the ```lut``` kwarg, but personally I would transform (e.g., binarise) my data prior to sending it to ```matplotlib``` to display.

bmcfee: Ok, thanks for clarification!  It's entirely possible that our `lut=2` code predates matplotlib doing this automatically, so it might be entirely moot anyway.  I suppose if the image comparison tests are still passing after this change though, we should be fine.

bmcfee: > but personally I would transform (e.g., binarise) my data prior to sending it to `matplotlib` to display.

Sorry, I missed this part earlier.  What we're actually doing here is selecting a colormap based on the data - if we see dtype=bool (ie it's already binary) then we use black-and-white (the code in question).  So I think we don't actually need to change anything here, if matplotlib is behaving as you say above and setting the number of colors appropriately.

PR #1784, opened 2023-11-22
Better handling of warnings in test suite

bmcfee:
#### Reference Issue

Our test suite currently triggers a pretty large number of warnings.  These are currently suppressed, making it difficult to identify futurewarnings (and other things we should keep tabs on) up front.

This PR changes the pytest config to dump warnings out in the log.  After that, I'll go through and clean up whatever deprecated usages I can find.

### DIFF
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -504,7 +504,7 @@ def __last_beat(cumscore):
 
 def __trim_beats(localscore: np.ndarray, beats: np.ndarray, trim: bool) -> np.ndarray:
     """Remove spurious leading and trailing beats"""
-    smooth_boe = scipy.signal.convolve(localscore[beats], scipy.signal.hann(5), "same")
+    smooth_boe = scipy.signal.convolve(localscore[beats], scipy.signal.get_window("hann", 5), "same")
 
     if trim:
         threshold = 0.5 * ((smooth_boe**2).mean() ** 0.5)
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -504,7 +504,7 @@ def __last_beat(cumscore):
 
 def __trim_beats(localscore: np.ndarray, beats: np.ndarray, trim: bool) -> np.ndarray:
     """Remove spurious leading and trailing beats"""
-    smooth_boe = scipy.signal.convolve(localscore[beats], scipy.signal.get_window("hann", 5), "same")
+    smooth_boe = scipy.signal.convolve(localscore[beats], scipy.signal.get_window("hann", 5, fftbins=False), "same")
 
     if trim:
         threshold = 0.5 * ((smooth_boe**2).mean() ** 0.5)
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -704,7 +704,6 @@ def icqt(
             norm,
             sparsity,
             window=window,
-            dtype=dtype,
             alpha=alpha[sl],
         )
 
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -957,7 +957,7 @@ def __reassign_times(
     # equation 5.23 of Flandrin, Auger, & Chassande-Mottin 2002
     # the sign of the correction is reversed in some papers - see Plante,
     # Meyer, & Ainsworth 1998 pp. 283-284
-    correction = np.real(S_th / (S_h + util.tiny(S_h)))
+    correction = np.real(S_th / S_h)
 
     if center:
         pad_length = None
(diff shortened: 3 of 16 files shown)


bmcfee:
One of the persistent warning-types that I've been seeing here is the following:
```

tests/test_core.py: 3072 warnings
  /home/runner/work/librosa/librosa/librosa/core/spectrum.py:378: ComplexWarning: Casting complex values to real discards the imaginary part
    stft_matrix[..., bl_s + off_start : bl_t + off_start] = fft.rfft(

tests/test_core.py: 1536 warnings
  /home/runner/work/librosa/librosa/librosa/core/spectrum.py:362: ComplexWarning: Casting complex values to real discards the imaginary part
    stft_matrix[..., :off_start] = fft.rfft(fft_window * y_frames_pre, axis=-2)

tests/test_core.py: 1536 warnings
  /home/runner/work/librosa/librosa/librosa/core/spectrum.py:366: ComplexWarning: Casting complex values to real discards the imaginary part
    stft_matrix[..., -off_end:] = fft.rfft(fft_window * y_frames_post, axis=-2)
```
This apparently comes from the griffinlim tests, where the dtype inference might not be working exactly correctly.  I don't have time to dig into this more right now, but am flagging it for later inspection.

bmcfee:
> This apparently comes from the griffinlim tests, where the dtype inference might not be working exactly correctly.

This turned out not to be the case.  The GL code is fine, but the test function was passing the (real) dtype argument to `stft` when it should only have passed it to `griffinlim`.  This is now fixed, and should take care of the ~6000 warnings.

bmcfee:
Cleared out some divide-by-zero warnings in reassigned spectrograms with this logic:
https://github.com/librosa/librosa/blob/7e994fe3311961ffc7412a195011cfee24b1cbac/librosa/core/spectrum.py#L785-L790

I had at one point tried modifying this to be stabilized (eg `S_dh / (S_h + util.tiny(S_h))`) but this wound up breaking our tests.

The above fix allows the 0/0 to go through as a nan value (our current behavior) without raising a warning.  Would appreciate a sanity check from @lostanlen on whether this is intended and/or desired behavior for these functions.

bmcfee:
Having spent a bit more time looking through reassigned spectrogram, i think the fix I've put in place is correct and preserves the intended behavior.

bmcfee:
I wasn't able to reproduce the test failure in my local env (py311), but it looks like the max-abs error in icqt reconstruction between single- and multi-channel is on the order of 1e-7, so I think this is coming down to machine precision issues:
```
 >       assert np.allclose(yboth[0], y0), np.max(np.abs(yboth[0] - y0))
E       AssertionError: 7.4505806e-08
```

Previously, we passed in `dtype=None` from these tests.  This overrode the default (complex64) in `vqt_filter_fft`, resulting in a higher precision filter basis than it would otherwise construct - the dtype would ultimately be inferred in `util.sparsify_rows` from whatever the input dtype is rather than casting.

However, the above behavior was a bug when a real `dtype` is provided for `icqt`, in which case the filter basis is constructed as real instead of complex - the passthrough from `icqt` to `vqt_filter_fft` should not have been there in the first place.

I think the best thing to do here is relax the precision on the multichannel test.

bmcfee:
Ok, I think I've got this down to just (known/expected) pathlib deprecation warnings on our 3.11/latest environment.  I'm happy to handle those as a separate PR, since resolving this warning in a minor revision while retaining py3.7 compatibility will require a bit of fancy footwork.

Our 3.7 / minimal dependency build has a boatload of warnings (378 thousand) - I'm not interested in fixing those.

The intermediate builds also demonstrate a small assortment of other warnings, but I think we can safely ignore them.

I think this is ready for a light CR, if there are any takers.  It should be a pretty easy job: most of the time the fixes amount to ignoring specific warnings that are safe to disregard in our test context (eg, nfft warnings due to our test signals being too short).  There are a couple of actual code changes / bug fixes as documented above.

Apart from code changes, the one other workflow thing I'm considering for this PR is adding an action that can post the warning log back into the PR as a comment.  I think this is doable, and it might be helpful for surfacing new futurewarnings earlier.

bmcfee:
Did a bit of looking into it, and I think the warning log → github comment idea is going to be too much of a hassle to work out.  So I think this PR is feature complete and ready for a quick CR.

bmcfee (commented):
Did a yolo/self-CR on this, and I think we're all set.

bmcfee:
Added the bug tag here, since we did catch a couple of very minor bugs in the process.

Issue #517, opened 2017-02-19
Variable rate phase vocoding


bmcfee:
`phase_vocoder` currently accepts a single scalar parameter controlling the rate of time dilation.  In principle, it shouldn't be difficult to either vectorize this, or support a slice+rate parameterization so that different time regions can be accelerated by different rates.

dpwe:
Once you get to variable rates, you have to make sure you are clear about
how the time base is defined, because now the input and output time bases
have a nonlinear relationship.  What you really want is the mapping from
output timebase to input timebase, since it's not infeasible to want
reversals and loops (non-monotonic relationship).  If you accept a vector
of per-input-window time scale, then effectively you're specifying how much
time each fixed-size input window will occupy in the output, and it's hard
to predict how long the output will be.

I think accepting a vector of (input time, output time) (defining a
piecewise-linear relationship) tuples is a nice, neutral description.  You
could then programmatically figure out the slope (time scaling) and time
center required from the input signal for each output frame.

dpwe:
See for example:

https://github.com/dpwe/pitchfilter/blob/master/resample_map.m

(which, weirdly, has been translated into Matlab somewhere between my
memory and GitHub).  That code does "time scaling" by simple resampling,
but otherwise it's a similar concept.

  DAn.

bmcfee:
> I think accepting a vector of (input time, output time) (defining a
> piecewise-linear relationship) tuples is a nice, neutral description.

Yeah, I was just thinking in terms of slices because then the bounds are explicit rather than implicit, and rate because it's closest to the existing interface.

This issue was brought up at the HAMR in rochester, where someone wanted an easy way to warp audio along a dtw path.  I had implemented something along these lines way back as part of the frankenmasher, but never migrated the code into librosa because it was kludgy as all get out.

stefan-balke:
That would be great!

Referencing this: https://github.com/librosa/librosa/issues/298#issuecomment-201908080
and the implementation in the TSM Toolbox: https://www.audiolabs-erlangen.de/content/resources/MIR/TSMtoolbox/pvTSM.m

If we could the following points, we would pretty much cover the capabilities of the TSM Toolbox:
* HPSS ✅ 
* H-Part: Phase Vocoder ✅ 
* P-Part: Overlap-Add ❎ 
* Add up everything ✅ 

That could mean to have some `tsm(x, factor=None, wp=[tuples], ...)` frontend function.

bmcfee:
[...]

PR #1796, opened 2024-01-15
Rewrite of phase vocoder

bmcfee:
#### Reference Issue
Fixes #1770 

This PR implements the rewrite of the phase vocoder function described in the comment thread of #1770.

The function's behavior is not substantially changed, but it is significantly faster and simpler now.  The outputs are different from the old implementation only in the final frames, where phase advance could not be reliably estimated anyway.

The hop length and nfft parameters have been deprecated, as they are no longer used in the new implementation.

A basic unit test has been implemented to verify that `rate=1` acts as a no-op.  Note that this only passes when using high precision (complex128) input STFTs, but I consider this to be good enough.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -177,7 +177,7 @@ intersphinx_mapping = {
     "matplotlib": ("https://matplotlib.org/stable/", None),
     "sklearn": ("https://scikit-learn.org/stable/", None),
     "resampy": ("https://resampy.readthedocs.io/en/stable/", None),
-    "soundfile": ("https://pysoundfile.readthedocs.io/en/latest", None),
+    "soundfile": ("https://python-soundfile.readthedocs.io/en/latest", None),
     "samplerate": ("https://python-samplerate.readthedocs.io/en/latest/", None),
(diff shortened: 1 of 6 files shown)


bmcfee:
Thinking on this a bit more, the new vectorized implementation makes it almost trivial to implement #517 ; all we need to do is allow for a user-specified `frames_out` instead of linearly interpolating frame indices by `rate`.

As noted in that thread though, the difficulty here would be in the API.  Since this is somewhat advanced functionality, and folks probably shouldn't use it anyway (favoring TSM toolbox or rubberband), I'm okay with not going overboard on the API.  The one point that we should figure out is whether to support a time grid in arbitrary units (seconds, samples, frames) or only as frames.

If we only support frame-valued time grids, then we can deprecate the hop length and nfft parameters (as I've already done here).  But if we need to convert other units to frames, we'll need to keep these parameters around, and add `sr`.

I'm inclined to keep it simple and only support frame-valued time grids, but if anyone objects, now would be the time.

bmcfee:
Upon a bit more inspection, I think this implementation is also bad, and needs another complete rewrite.  Closing this out.

Issue #1478, opened 2022-05-04
Documentation restructuring (core submodule)


bmcfee:
The documentation site is generally pretty navigable, but I think the core submodule (sub-heading "Core IO and DSP") is both too large and at this point, inaccurately named.

At present, it looks as follows:
![image](https://user-images.githubusercontent.com/1190540/166687856-7a1c3198-59ef-472d-86d6-2ed223d8d365.png)
where each of those sub-sub-headings has somewhere between 2 and 12 function entries.

I think we can easily break this up into some more meaningful sections, which can be promoted to the top level of the index.  Originally, I had designed the documentation index to reflect the submodule structure of the package, but I don't think that's terribly relevant anymore.

I propose that we eliminate the "Core IO and DSP" top-level heading, and replace it with the following:

- Signals  (includes load, stream, resample, and any signal generators and helper utilities)
- Spectral representations
- Magnitude scaling
- Unit conversion (time, frequency)
- Music notation
- Pitch and tuning
- Harmonics
- Phase recovery
- Miscellaneous (collects current misc + frequency range generators)

I'm also open to other suggestions for top-level organization, but the above seems at least plausible and a bit easier to navigate IMO.

Are there other sections of the docs that could benefit from restructuring?

dana-gill:
Hey Brian -- Dana here (my old @danagilliann is no longer accessible) 😄  I'm happy to start working on this. I can start of with

> I propose that we eliminate the "Core IO and DSP" top-level heading, and replace it with the following:

bmcfee:
:wave: Thanks @dana-gill , that'd be great!

dana-gill:
I do have to admit: Building the documentation is not very easy since the documentation for it was a bit scattered 😅  Would it make sense to propose the following issues?

1. Have a `Contributing` section within the sidebar. I think the easiest would be to just link it to the [Contributing Page on GitHub](https://github.com/librosa/librosa/blob/8dae20c66353190749486abf6b332d93c4ab2b46/CONTRIBUTING.md). Maybe it could be converted to `.rst` some time in the future.
[...]

PR #1801, opened 2024-01-29
Documentation sidebar restructuring

dana-gill:
#### Reference Issue
Fixes https://github.com/librosa/librosa/issues/1478

This restructures the Core IO section of documentation sidebar. Core IO no longer exists and is replaced by the following:

- [x] Signals (includes load, stream, resample, and any signal generators and helper utilities)
- [x] Spectral representations
- [x] Magnitude scaling
- [x] Unit conversion (time, frequency)
- [x] Music notation
- [x] Pitch and tuning
- [x] Harmonics
- [x] Phase recovery
- [x] Miscellaneous (collects current misc + frequency range generators)

#### Any other comments?
None

### DIFF
--- a/docs/harmonics.rst
+++ b/docs/harmonics.rst
@@ -1,0 +1,11 @@
+.. _harmonics:
+
+---------
+Harmonics
+---------
+
+.. include:: generated/librosa.interp_harmonics.rst
+.. include:: generated/librosa.salience.rst
+.. include:: generated/librosa.f0_harmonics.rst
+
+.. include:: generated/librosa.phase_vocoder.rst
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -42,6 +42,7 @@ If you want to cite librosa in a scholarly work, there are two ways to do it.
 
     spectral_representation
     magnitude_scaling
+    unit_conversion
     phase_recovery
     core
     display
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -40,6 +40,7 @@ If you want to cite librosa in a scholarly work, there are two ways to do it.
     :caption: API documentation
     :maxdepth: 1
 
+    spectral_representation
     magnitude_scaling
     phase_recovery
     core
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -44,6 +44,7 @@ If you want to cite librosa in a scholarly work, there are two ways to do it.
     magnitude_scaling
     unit_conversion
     music_notation
+    pitch_and_tuning
     phase_recovery
     core
     display
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -40,6 +40,7 @@ If you want to cite librosa in a scholarly work, there are two ways to do it.
     :caption: API documentation
     :maxdepth: 1
 
+    signals
     spectral_representation
     magnitude_scaling
     unit_conversion
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -58,6 +58,7 @@ If you want to cite librosa in a scholarly work, there are two ways to do it.
     segment
     sequence
     util
+    miscellaneous
 
 
 .. toctree::
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -43,6 +43,7 @@ If you want to cite librosa in a scholarly work, there are two ways to do it.
     spectral_representation
     magnitude_scaling
     unit_conversion
+    music_notation
     phase_recovery
     core
     display
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -40,6 +40,7 @@ If you want to cite librosa in a scholarly work, there are two ways to do it.
     :caption: API documentation
     :maxdepth: 1
 
+    phase_recovery
     core
     display
     feature
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -48,7 +48,6 @@ If you want to cite librosa in a scholarly work, there are two ways to do it.
     pitch_and_tuning
     harmonics
     phase_recovery
-    core
     display
     feature
     onset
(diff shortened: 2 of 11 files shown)


bmcfee:
Rebasing this from main should fix the build issues.

bmcfee:
Thanks @dana-gill !

A couple of broad comments:

- Can you rebase this on the current main?  I made a couple of changes to the environment and sphinx config lately that this is presently reverting.
- The new structure has lost the autosummary / menu list of subsection contents, eg https://librosa.org/doc/latest/core.html .  Can we recover that?
- We've also lost the module-level docstring, and the linter is now complaining about it.
- I don't think it's correct to directly include generated contents (eg in harmonics.rst).  Is there a way to keep this as using autodoc/ autosummary, but just break out the sections differently from what we had before?  I think this should be as simple as keeping the module-level docstring in `__init__.py`, but just reorganizing the section information in the autosummary.  Maybe I'm missing something?

dana-gill:
> Can you rebase this on the current main? I made a couple of changes to the environment and sphinx config lately that this is presently reverting.

Huh, I thought I rebased it already. These are the commands I ran:
```
$ git fetch --all
$ git rebase origin/main
```
And it says `Current branch documentation_sidebar_restructuring is up to date.`. Are there any specific commands I should run for the repo?

> The new structure has lost the autosummary / menu list of subsection contents, eg https://librosa.org/doc/latest/core.html . Can we recover that?

Yes we can 🥳  Just to make sure I understand you: You would like to the sidebar to have both the content of Core IO and DSP *and* the following (ie: How the branch looks now):

Signals (includes load, stream, resample, and any signal generators and helper utilities)
Spectral representations
Magnitude scaling
Unit conversion (time, frequency)
Music notation
Pitch and tuning
Harmonics
Phase recovery
Miscellaneous (collects current misc + frequency range generators)

Let me know if I understood that properly.

> We've also lost the module-level docstring, and the linter is now complaining about it.

To make sure I understand: The linter would like `librosa/__init__.py` to stay the same and all of this should remain the same? https://github.com/librosa/librosa/blob/main/librosa/__init__.py#L3-L210

> I don't think it's correct to directly include generated contents (eg in harmonics.rst). Is there a way to keep this as using autodoc/ autosummary, but just break out the sections differently from what we had before? I think this should be as simple as keeping the module-level docstring in __init__.py, but just reorganizing the section information in the autosummary. Maybe I'm missing something?

I need to do a bit more digging for this 😅  I think there was a reason why I couldn't just keep using `autosummary`, but I would have do some experimentations and get back to you.

dana-gill:
> > You would like to the sidebar to have both the content of Core IO and DSP _and_ the following (ie: How the branch looks now):
> > Signals (includes load, stream, resample, and any signal generators and helper utilities) Spectral representations Magnitude scaling Unit conversion (time, frequency) Music notation Pitch and tuning Harmonics Phase recovery Miscellaneous (collects current misc + frequency range generators)
> > Let me know if I understood that properly.
> 
> Currently all of those things are buried under the "Core IO and DSP" heading, which is not too accurate of a descriptor at this point. I think the idea here was to lift up the sub-headings of core (currently "audio loading", "time-domain processing", etc) to the top-level and eliminate the "core IO and dsp" section, along with some restructuring of the topics to make it a bit more navigable (as you describe).
> 
> > To make sure I understand: The linter would like `librosa/__init__.py` to stay the same and all of this should remain the same? https://github.com/librosa/librosa/blob/main/librosa/__init__.py#L3-L210
> 
> Well the linter wants _some_ docstring in the top-level module, but it doesn't necessarily care about the contents. In light of the above though, I think what we can do is just drop the "Core IO and DSP" heading, then replace the `---` (second-level subheadings) by `===` (first-level headings) along with some reorganization as noted above. We shouldn't really need to change anything outside of `__init__.py` to get the new sidebar structure.
> 
> All that said, we should also while we're here consider #1823 - this AFAIK would just entail fleshing out the docstring for the core module with some descriptive text beyond just the menu headings. Also happy to take that up in a separate issue to keep things simple.

Ahh. So you would like the sidebar to look how it looks like in this PR *and* also have `core.html` at the same time? 

I also did some investigation to see if it was possible to edit only the `__init__.py` and not add an`rst` files. It is, but you will have the plus icons on the left side. If that works with you, then I can make the changes :)

Screenshot 2024-04-05 at 12 55 57 PM


> All that said, we should also while we're here consider https://github.com/librosa/librosa/issues/1823 - this AFAIK would just entail fleshing out the docstring for the core module with some descriptive text beyond just the menu headings. Also happy to take that up in a separate issue to keep things simple.

Sure! I can make that part of the PR.

bmcfee:
> Ahh. So you would like the sidebar to look how it looks like in this PR _and_ also have `core.html` at the same time?

Yeah I suppose so.  `core.html` is autogenerated from the top-level package import, ie the docstring in `librosa/__init__.py`, so all we're really talking about here is promoting the sub-headings within core up a level by changing the formatting.



> I also did some investigation to see if it was possible to edit only the `__init__.py` and not add an`rst` files. It is, but you will have the plus icons on the left side. If that works with you, then I can make the changes :)

Yeah, that's totally fine! Thanks.

bmcfee:
:wave: @dana-gill just checking in to see how you're doing on this?  It's the last PR we need to put in before pushing out the 0.10.2 release, so if there's anything I can do to help get it moving, please let me know.

dana-gill:
> 👋 @dana-gill just checking in to see how you're doing on this? It's the last PR we need to put in before pushing out the 0.10.2 release, so if there's anything I can do to help get it moving, please let me know.

Hey Brian! Thanks for checking in. I've been incredibly swamped with work so I don't think this will make it to the 0.10.2 release. Is that okay?

bmcfee:
Ok, no worries!  I'll make some today to sort out what actually needs to be there from what's a nice-to-have, and take this one out of the critical path.

dana-gill:
Closed in favor of https://github.com/librosa/librosa/pull/1865

PR #1808, opened 2024-02-14
Adding python 3.12 to build matrix, update tests for recent scipy / numpy

### LINKED ISSUE #1772: Can Python 3.12 installation be supported
I try using`pip3.12 install librosa` is error，May I ask when we can support it
image

bmcfee:
#### Reference Issue

Fixes #1772 

Adds 3.12 to supported versions, updates build matrix.

### DIFF
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -7,7 +7,8 @@ dependencies:
   - pip
   - audioread>=2.1.9
   - numpy>=1.20.3
-  - scipy>=1.2.0
+    #- scipy>=1.2.0
+  - scipy==1.11.4
   - scikit-learn>=0.20.0
   - joblib>=0.14.0
   - decorator>=4.3.0
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -19,6 +19,8 @@ dependencies:
   - msgpack-python>=1.0
   - lazy_loader>=0.1
   - typing_extensions>=4.1.1
+  # Forcing some version pins for debugging
+  - libopenblas==0.3.25
 
   # optional, but required for testing
   - matplotlib>=3.3.0
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -1,7 +1,7 @@
 name: librosa-dev
 channels:
-  - defaults
   - conda-forge
+  - defaults
 dependencies:
   # required
   - pip
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -7,8 +7,7 @@ dependencies:
   - pip
   - audioread>=2.1.9
   - numpy>=1.20.3
-    #- scipy>=1.2.0
-  - scipy==1.11.4
+  - scipy>=1.2.0
   - scikit-learn>=0.20.0
   - joblib>=0.14.0
   - decorator>=4.3.0
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -6,7 +6,8 @@ dependencies:
   # required
   - pip
   - audioread>=2.1.9
-  - numpy>=1.20.3
+    #- numpy>=1.20.3
+  - numpy==1.26.3
   - scipy>=1.2.0
   - scikit-learn>=0.20.0
   - joblib>=0.14.0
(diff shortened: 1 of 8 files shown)


bmcfee:
Fixed the failing nnls test and also a broken link in the intersphinx mapping.  This should be good to go when CI passes.

bmcfee:
Now encountering a platform bug on osx (py3.12, numpy 1.26.4, scipy 1.12.0, numba 0.59.0) with multi-channel pyin. I'm not sure of the cause, though I expect it's down to differences in floating point operations.  The end result is that our centered multi-channel pyin test is failing due to a voicing estimation error.  The test passes on all other platforms / configs.

The most recent OSX passing build used py3.11, numpy 1.26.3, numba 0.58.1, and scipy 1.11.4.

Reading the release notes for the above, I don't see any obvious culprits.  I'll try to find some time to bisect this.

bmcfee:
Regression is confirmed to be caused by scipy 1.12.0; 1.11.4 works.

bmcfee:
Pretty baffled by this regression.  The only place we use scipy in this code path are in computing beta and boltzmann stats, and while the code for that has changed in 1.12.0, the outputs agree to numerical tolerance.

bmcfee:
Ok, I'm going to need someone with access to a mac to help debug this one; bisecting through trial and error over GH actions will take forever.

From what I can tell, there are two places where we use scipy functionality in the pyin implementation:

https://github.com/librosa/librosa/blob/8dae20c66353190749486abf6b332d93c4ab2b46/librosa/core/pitch.py#L816

and

https://github.com/librosa/librosa/blob/8dae20c66353190749486abf6b332d93c4ab2b46/librosa/core/pitch.py#L903-L905

@richa-namballa verified that the beta distribution call is numerically equivalent from 1.11.4 to 1.12.0, so that leaves only the boltzmann call as the most likely source of the problem.  I've tried to dig through the compare log between scipy versions, and nothing jumped out at me as affecting boltzmann here, but it's possible that I missed something.

Could someone with a mac (@lostanlen ? @stefan-balke ?  @richa-namballa ?) take a crack at comparing boltzmann pmf outputs between 1.11.4 and 1.12.0 to see if something's gone wrong here?

stefan-balke:
> Could someone with a mac (@lostanlen ? @stefan-balke ? @richa-namballa ?) take a crack at comparing boltzmann pmf outputs between 1.11.4 and 1.12.0 to see if something's gone wrong here?

Of course. Could you provide a meaningful call? The test case? Should I dump it as npz?

bmcfee:
> Of course. Could you provide a meaningful call? The test case? Should I dump it as npz?

Thanks!  I actually have a hypothesis about what's going on now, see https://github.com/scipy/scipy/issues/20110#issuecomment-1953144749 .  I think what would be most helpful is if you can run the offending test on scipy 1.12.0 distributed on conda-forge (should fail) and then again on the pypi build, but with all other dependencies fixed.

bmcfee:
@stefan-balke nevermind, I think we have this sorted out.  The problem was actually due to a divergence in libopenblas 0.3.26, not scipy.  I'm bypassing this test on OSX for now, but we can revisit later when openblas gets fixed.

This PR should be good to merge once CI passes.

mattip:
> The problem was actually due to a divergence in libopenblas 0.3.26, not scipy. I'm bypassing this test on OSX for now, but we can revisit later when openblas gets fixed

Do you know what function(s) in OpenBLAS are in play? I think this can be checked by running the test under `ltrace` and using the `--library` option (the check can be done on linux).

bmcfee:
> I think this can be checked by running the test under `ltrace` and using the `--library` option (the check can be done on linux).

I'm not getting any usable output from running this on my local machine (linux, conda-forge), but it's also possible that my build isn't using openblas at all.

I don't have an immediate guess for what's going on at the level of blas functions.  The triggering example is, as I mentioned previously, a pretty deep call stack and involves both numba/llvmlite and numpy vectorization.

Here's a gist of the test function isolated to a python script (ltracing pytest directly doesn't work): https://gist.github.com/bmcfee/b685d5234c4e111338c87c14323f2d56

This running on the current librosa main branch should do the job.  The conda environment spec in question is here: https://github.com/librosa/librosa/blob/main/.github/environment-ci.yml  - if you want to take a crack at tracing this directly, that environment and the above script should be enough to do it.

PR #1813, opened 2024-02-28
Resolve the performance issue of autocorrelate

alumkal:

This pull request modifies the pad length in `core.audio.autocorrelate` to speed up FFT calculation. On extreme cases (for example, when `len(y) * 2 + 1` is a prime), the performance gain can be up to 10x.

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -10,6 +10,7 @@ import warnings
 import soundfile as sf
 import audioread
 import numpy as np
+import scipy
 import scipy.signal
 import soxr
 import lazy_loader as lazy
@@ -914,22 +915,27 @@ def autocorrelate(
 
     fft = get_fftlib()
 
-    # Pad out the signal to support full-length auto-correlation.
-    n_pad = 2 * y.shape[axis] - 1
+    real = not np.iscomplexobj(y)
 
-    if np.iscomplexobj(y):
+    # Pad out the signal to support full-length auto-correlation
+    if hasattr(scipy.fft, "next_fast_len"):
+        n_pad = scipy.fft.next_fast_len(2 * y.shape[axis] - 1, real=real)
+    else:
+        # TODO: Bump to scipy>=1.4.0 and remove this branch
+        n_pad = scipy.fftpack.next_fast_len(2 * y.shape[axis] - 1)
+
+    if real:
+        # Compute the power spectrum along the chosen axis
+        powspec = util.abs2(fft.rfft(y, n=n_pad, axis=axis))
+
+        # Convert back to time domain
+        autocorr = fft.irfft(powspec, n=n_pad, axis=axis)
+    else:
         # Compute the power spectrum along the chosen axis
         powspec = util.abs2(fft.fft(y, n=n_pad, axis=axis))
 
         # Convert back to time domain
         autocorr = fft.ifft(powspec, n=n_pad, axis=axis)
-    else:
-        # Compute the power spectrum along the chosen axis
-        # Pad out the signal to support full-length auto-correlation.
-        powspec = util.abs2(fft.rfft(y, n=n_pad, axis=axis))
-
-        # Convert back to time domain
-        autocorr = fft.irfft(powspec, n=n_pad, axis=axis)
 
     # Slice down to max_size
     subslice = [slice(None)] * autocorr.ndim


alumkal:
If the use of `scipy.fft.next_fast_len` is unacceptable, the pad length can be changed to the next power of 2. The performance gain will be smaller but still significant. Here's the run time of `numpy.fft.rfft` measured on my computer via IPython's `%timeit`:

|N|32769|65536|131073|262144|524287|1048576|
|-|-|-|-|-|-|-|
|Time (ms)|1.46|0.226|7.31|2.18|55.0|6.09|

The result shows that the padding is definitely needed. Note that `scipy.signal.fftconvolve` [pads the input too](https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/signal/_signaltools.py#L671).

To be honest, I didn't get why autocorrelation has anything to do with STFT. If you worry that the padding will introduce memory overhead when the autocorrelation is computed on fixed-sized overlapped frames, this won't happen at all, since the windows length is almost always a power of 2 in this type of scenario.

bmcfee:
> the pad length can be changed to the next power of 2 (`1 << int(2 * y.shape[axis] - 1).bit_length()`).

Sure, but that's probably overkill as well.  Efficient FFT plans will exist for other highly composite lengths, and this ultimately depends on what algorithms the backend supports.  Since there's no standardized API for querying efficient lengths --- it's not part of fftpack to my knowledge --- I'm a little hesitant to trying to outsmart the FFT backend.


> The result shows that the padding is definitely needed.

Ok, I can see that now - thanks for including some benchmarks.  The differences are a bit more modest running on my machine (at most 4x difference, using either np.fft or scipy.fft).



> To be honest, I didn't get why autocorrelation has anything to do with STFT.

Some applications of autocorrelation apply to the entire time domain signal, but others apply to a framed signal (hence the target axis) with overlap.  This mostly comes up in rhythm analysis where windowed autocorrelation of the onset envelope is used to infer tempo (and other rhythmic attributes).



> If you worry that the padding will introduce memory overhead when the autocorrelation is computed on fixed-sized overlapped frames, this won't happen at all, since the windows length is almost always a power of 2 in this type of scenario.

I don't understand what you mean here.  Computing an rfft will produce an output of size `2*N-1` along the target axis.  If `N` increases, so to will the output size, and with high overlap, those costs can add up.

alumkal:
> Efficient FFT plans will exist for other highly composite lengths

It's true. As you previously stated, librosa supports arbitrary FFT backends, so the best possible length can't be determined. But at least doing something (use a sub-optimal padding) is better than doing nothing (just use a potentially prime length).

> I'm a little hesitant to trying to outsmart the FFT backend

It is the user who decide whether to use an exact-sized FFT (for example, when computing cyclic autocorrelation) or a padded one. There's nothing wrong to pad the input to power of 2 and it is a widely used process.

> And with high overlap, those costs can add up

I think the memory footprint is hardly an issue (the output of `librosa.feature.tempogram` is not extremely large with a reasonable parameter set), but speed is.

bmcfee:
> But at least doing something (use a sub-optimal padding) is better than doing nothing (just use a potentially prime length).

That point is debatable.  The reason I'm pushing back on this is primarily from the perspective of maintenance: adding logic to outsmart a backend with characteristics that are not fully specified is the type of thing that's likely to backfire on us a few years down the road.  Adding (logical) complexity here for the sake of a marginal speed enhancement is not automatically "better".  (It might be better, but we need to think through how it will impact other backends.)



> Besides, some FFT implementations use an O(n²) algorithm for prime sized FFT.

I'm not sure that any of the libraries we currently support.  PocketFFT (numpy and scipy) uses Bluestein for prime lengths; FFTW uses a variety of O(n·log(n)) methods; cupy doesn't say in the docs what exactly they use, but they do claim n·log(n) in the docs (and likely it's similar to FFTW); MKL-FFT is likely similar, etc.  There might be some exotic naive implementations out there, but I rather doubt it - the efficiency differences we're talking about here are constant factors, not asymptotics.



> It is the user who decide whether to use an exact-sized FFT (for example, when computing cyclic autocorrelation) or a padded one. 

Except that the change you're proposing here does not expose the size parameter to the user.  (I don't think it should, btw, but that's not the point here.)


> There's nothing wrong to pad the input to power of 2 and it is a widely used process.

Sure it's widely used, but that's not much of an argument without accounting for the contexts in which it's being used.  In the case of autocorrelation, we follow the forward DFT by an O(N) squared-magnitude computation, and then an inverse DFT.  It's not obvious to me that a constant factor improvement in the DFT is an automatic win if it inflates the cost of the subsequent computation.  It might well be, but it's not obvious.  I think if we want to adopt this kind of change, we should evaluate it with a holistic benchmark of the autocorrelate function and not just the FFT call. 

There's also still the question of intermediate memory consumption.  Again, we need to consider this not just in the 1D case, but also in the ND case.  The primary consumer of the autocorrelate function, aside from end-user code that we can't predict, is in the rhythm module where we operate on frames of the onset envelope.  As you say:

> (the output of `librosa.feature.tempogram` is not extremely large with a reasonable parameter set), but speed is.

For the parameter range in question for something like tempogram (default N=384 ⇒ n_pad=2*384-1=767), the speedup here is negligible (about 5ms in my benchmarks).  It's also true that the memory consumption using scipy next-length padding is negligible (1 additional sample), but using the next-power-of-2 logic, we're inflating the intermediate memory consumption by about 30%.  For a single frame analysis, we can absorb that easily enough, but if you're analyzing a track with thousands of frames, that 30% inflation can hurt.  (It will also take more time to allocate and populate the array.)



All that said, I do think the scipy next-length function is preferable as it handles small-factor composites (2/3/5) and therefore produces smaller pad lengths that are still efficient for Cooley-Tukey.  I benchmarked this locally and there are some marginal speed improvements for very large lengths (N~=20000) relative to `n=2**(n_pad.bit_length())`, about a factor of 2 for numpy/scipy; much less for mkl_fft.  For these three backends, at least, it does seem preferable to handle general small-factor composites than just powers of 2.  I haven't had a chance to test this on cupy or pyfftw (no gpu on my dev machine, and the conda build of pyfftw is giving me trouble today), but I expect a similar story there.

Here are the numbers I'm seeing btw.  In this case, each `N` corresponds to the length of the input signal; `pad` uses analysis length `2*N-1`, `next` uses scipy to compute the next length up, and `two` uses the next largest power of 2.  Results are timeit trials with 1000 runs, rfft on real-valued input.

## N → 2*N-1

|       |        256 |        384 |        512 |      1024 |      2048 |      4096 |     8192 |    16384 |    20000 |   100000 |
|:------|-----------:|-----------:|-----------:|----------:|----------:|----------:|---------:|---------:|---------:|---------:|
| mkl   | 0.0113593  | 0.0106039  | 0.00961816 | 0.0389059 | 0.0152886 | 0.0961433 | 0.227455 | 0.601552 | 1.22446  |  6.25569 |
| numpy | 0.00857028 | 0.00956341 | 0.00895316 | 0.0261107 | 0.0283906 | 0.385988  | 0.274471 | 0.62479  | 0.964962 | 27.0524  |
| scipy | 0.0112202  | 0.0120243  | 0.0115881  | 0.0267099 | 0.0280873 | 0.266151  | 0.268135 | 0.60602  | 0.926604 | 10.1353  |

## next fastest length (N → next fast length (2 * N-1))

|       |        256 |        384 |        512 |       1024 |       2048 |      4096 |      8192 |    16384 |    20000 |   100000 |
|:------|-----------:|-----------:|-----------:|-----------:|-----------:|----------:|----------:|---------:|---------:|---------:|
| mkl   | 0.00275388 | 0.00325493 | 0.00333689 | 0.00520753 | 0.00824195 | 0.0517613 | 0.0661551 | 0.10087  | 0.1102   | 0.415404 |
| numpy | 0.00332278 | 0.00428938 | 0.00544828 | 0.00913997 | 0.0206751  | 0.0402827 | 0.0804388 | 0.181569 | 0.229742 | 1.42236  |
| scipy | 0.00672377 | 0.00750053 | 0.00819183 | 0.011417   | 0.0203592  | 0.041931  | 0.0788315 | 0.168858 | 0.221343 | 1.36439  |

## next power of two (N → 2**((2*N-1).bit_length()))
|       |        256 |        384 |        512 |       1024 |       2048 |      4096 |      8192 |    16384 |    20000 |   100000 |
|:------|-----------:|-----------:|-----------:|-----------:|-----------:|----------:|----------:|---------:|---------:|---------:|
| mkl   | 0.00277746 | 0.00331489 | 0.0033643  | 0.00497826 | 0.00819439 | 0.048646  | 0.0686287 | 0.112183 | 0.138764 | 0.431393 |
| numpy | 0.00330289 | 0.00589798 | 0.00509357 | 0.0090581  | 0.0181111  | 0.0403112 | 0.0803494 | 0.178908 | 0.375336 | 2.00792  |
| scipy | 0.00663946 | 0.00858057 | 0.00828293 | 0.0116491  | 0.0201862  | 0.0419064 | 0.0835794 | 0.166962 | 0.350384 | 2.22584  |

alumkal:
> Adding (logical) complexity here for the sake of a marginal speed enhancement is not automatically "better".
> For the parameter range in question for something like tempogram (default N=384 ⇒ n_pad=2*384-1=767), the speedup here is negligible (about 5ms in my benchmarks).

The speed enhancement is not marginal at all, as shown by the tests performed by both of us. 5ms here means a 2.5x speedup which can't be easily ignored.

> All that said, I do think the scipy next-length function is preferable

I agree. But note that `scipy.fft` is introduced in 1.4.0, which means we have to update the requirements of librosa. However this shouldn't be a problem: scipy 1.4.0 is released in 2019, while numpy 1.20.3 (the minimum version supported by librosa) is released in 2021.

Or we may use [`scipy.fftpack.next_fast_len`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.fftpack.next_fast_len.html#scipy.fftpack.next_fast_len). It is included in older versions of scipy, and its behavior is determined (the result of `scipy.fft.next_fast_len` may change in accordance with scipy's FFt implementation).

Overall, I think `scipy.fft.next_fast_len` and `scipy.fftpack.next_fast_len` are both good choices. I wonder which one do you prefer.

alumkal:
So at least now we can accept the try-except approach as a temporary solution.

I will make a micro benchmark a bit later.

alumkal:
Sorry for the mistake :(

librosa/core/audio.py
```diff
@@ -914,22 +915,27 @@ def autocorrelate(
 
     fft = get_fftlib()
 
+    real = not np.iscomplexobj(y)
+
     # Pad out the signal to support full-length auto-correlation.
-    n_pad = 2 * y.shape[axis] - 1
+    if hasattr(scipy, "fft") and hasattr(scipy.fft, "next_fast_len"):
```
bmcfee, line 921: This isn't quite right.  Older versions of scipy (specifically 1.2.0) had `scipy.fft` as a function, not a module, so the first hasattr check is not going to do anything here.

This is why I had proposed try-except instead of a hasattr check.

alumkal: Actually only `hasattr(scipy.fft, "next_fast_len")` is needed. I wrote the first condition to make my meaning clearer.

I personally avoid the use of try-except block as a normal control flow.

bmcfee: Yes, in general hasattr is preferable to try-except for this kind of thing.  However, scipy does do some tricks with imports and submodule structure.  But if a single hasattr call actually does work in the older versions, that seems fine.

alumkal:
Are there any pending issue?

librosa/core/audio.py
bmcfee: We should remove the redundant hasattr check.  I think the code is clear enough without it.

bmcfee (approved):
LGTM!  Thanks again for this PR, and accommodating my requests for more information about the impacts of this change.

PR #1816, opened 2024-03-06
add cqa method to spectrum.py

shihkuanglee:
I developed a method that can compute constant q transform either faster or with higher resolution.
[https://github.com/shihkuanglee/ADFA](https://github.com/shihkuanglee/ADFA)

#### Any other comments?
The method I developed can take any length of signal to any resolution of spectrum in different scale.
But I am new to this community, I am not sure which way of method (python) fit the need of the community.
So I simply make the 'cqa' method based on the 'stft' method in spectrum.py.

Here is the figure you can get from the 'Examples' in docstring.
specshow-CQ

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -104,6 +104,7 @@ Contributors
 * Shin Hyun <https://github.com/kyaryunha>
 * Iliya S <https://github.com/zenitismus>
 * Eugene Rabinovich <https://github.com/erabinov>
+* Shih-Kuang Lee <https://github.com/shihkuanglee>
 
 Some feature extraction code was based on <https://github.com/ronw/frontend> by Ron Weiss.
 
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -36,6 +36,7 @@ from .._typing import (
 __all__ = [
     "stft",
     "istft",
+    "cqa",
     "magphase",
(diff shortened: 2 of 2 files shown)


lostanlen:
Hello @shihkuanglee, thank you for this. 

Without looking at your method in much detail, my first impression is that librosa is not a good fit for hosting the ADFA. Over the years, we've tried to stick to implementing classical algorithms in music and audio signal processing. The ADFA algorithm is very new (under review, from what i understand?) so its importance for the community is not something we can establish. It is not, or not yet, a classical algorithm like the STFT, the CQT, or the HCQT. Meanwhile, it's adding quite a lot of to the codebase (340 lines, half of it is code) to we have to balance that against the cost of maintaining it.

Small remarks: your `n_bins` argument should be `bins_per_octave` under librosa's `cqt` convention. I don't see a strong need for exposing `base=2`. Bases other than two are practically never used in audio.

From what i've seen, your ADFA comes close to a continuous wavelet transform (CWT) with constant-Q wavelets. So perhaps your ADFA could be included within the `ltfatpy` toolbox: Large Time–Frequency Analysis in Python. I know they have high-quality implementations of the CWT, allowing inversion and perfect reconstruction. Try speaking with them?
https://gitlab.lis-lab.fr/dev/ltfatpy

PR #1822, opened 2024-03-18
exposed tempo min and max parameters in specshow

### LINKED ISSUE #1805: Fix strongly hardcoded BPM range for tempograms
Spend hours trying to produce proper tempograms with BPM ~ 550. The BPM-axis seems locked to around (16, 512).

Finally found [this](https://github.com/librosa/librosa/blob/main/librosa/display.py#L1368) being called by `specshow`:

    elif ax_type in ["tempo", "fourier_tempo"]:
        mode = "log"
        kwargs[base] = 2
        limit(16, 480)

1. Document this hardcoded "feature" in `tempogram`/`tempogram_fourier` and/or `specshow`.
2. Make the `limit(...)` adjustable or add a note in the documentation on below workaround.

https://github.com/librosa/librosa/blob/main/librosa/display.py#L1368

Calling `ax[1].set_ylim(auto=True)` or `ax[1].set_ylim(8, 1024)` after `specshow` fixes the axis.

bmcfee:
#### Reference Issue
Fixes #1805 

This PR expose minimum and maximum tempo limits for tempogram display in specshow.

#### Any other comments?

Seems to  work, should not impact other code.  I think this is good to go.

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -934,6 +934,8 @@ def specshow(
     win_length: Optional[int] = None,
     fmin: Optional[float] = None,
     fmax: Optional[float] = None,
+    tmin: Optional[float] = 16,
+    tmax: Optional[float] = 480,
     tuning: float = 0.0,
     bins_per_octave: int = 12,
     key: str = "C:maj",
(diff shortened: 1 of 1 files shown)


bmcfee:
I named these parameters `tmin` and `tmax`, but i'm now thinking this might be ambiguous if we implement similar functionality for time axes.  Should we switch to `tempo_min` and `tempo_max`?  Or `bpm_min/max`?

bmcfee:
> Should we switch to `tempo_min` and `tempo_max`?

Implemented.  I think `tempo` is preferable to `bpm` here.

bmcfee:
I'm gonna YOLO this one, as it's pretty straightforward.

I'll just note here that the reason to implement the min/max here in the scale function rather than the coords function is that in spectrograms, we typically want to see all frequencies represented, but tempograms we may not (eg, when there's energy at tempo=∞).  As such, it makes sense to set the axis limits post-hoc for tempo.

PR #1825, opened 2024-03-26
Viterbi: numpy indexing is faster than unrolling numba

twoertwein:
This PR speeds up pyin by about 23% (on a Mac M2, python 3.10).

My test call (run ten times):
```py
librosa.pyin(
            wav, # 528318 elements
            fmin=librosa.note_to_hz("C2"),
            fmax=librosa.note_to_hz("C5"),
            sr=16_000,
            frame_length=1024,
            hop_length=256,
            center=False
)
```

The argmax line is numba-compatible, but just having it with numba decreased speed by about 26%.

### DIFF
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -1096,7 +1096,6 @@ def __rqa_backtrack(score, pointers):
     return np.asarray(path, dtype=np.uint)
 
 
-@jit(nopython=True, cache=True)  # type: ignore
 def _viterbi(
     log_prob: np.ndarray, log_trans: np.ndarray, log_p_init: np.ndarray
 ) -> Tuple[np.ndarray, np.ndarray]:  # pragma: no cover
@@ -1125,6 +1124,7 @@ def _viterbi(
     state = np.zeros(n_steps, dtype=np.uint16)
     value = np.zeros((n_steps, n_states), dtype=np.float64)
     ptr = np.zeros((n_steps, n_states), dtype=np.uint16)
+    index = np.arange(n_states, dtype=np.uint16)
 
     # factor in initial state distribution
     value[0] = log_prob[0] + log_p_init
@@ -1141,11 +1141,9 @@ def _viterbi(
 
         trans_out = value[t - 1] + log_trans.T
 
-        # Unroll the max/argmax loop to enable numba support
-        for j in range(n_states):
-            ptr[t, j] = np.argmax(trans_out[j])
-            # value[t, j] = log_prob[t, j] + np.max(trans_out[j])
-            value[t, j] = log_prob[t, j] + trans_out[j, ptr[t][j]]
+        # Vectorize the max/argmax loop
+        ptr[t] = np.argmax(trans_out, axis=1)
+        value[t] = log_prob[t] + trans_out[index, ptr[t]]
 
     # Now roll backward
 


bmcfee:
To summarize: this PR removes numba jitting from the core viterbi algorithm, and replaces an explicit loop with a vectorized call.

I'd be surprised if this is faster in general, though I can believe it will be faster in certain circumstances.  I'm curious how you're benchmarking it though?  Specifically, is your 23% over multiple trials, and not counting the initial jit compilation?

For a bit of context, I believe the vectorized argmax / fancy indexing was not numba-compatible at the time we implemented this (or with the minimum supported numba that we wanted to retain), hence the explicit for loop.

twoertwein:
Good point, I did not exclude the warm up phase (I only averaged it over ten calls)! It might also be limited to Mac M2 (not sure how well numba works on that).

Here is a new benchmark (10 warmup rounds, followed by 100 rounds for timing):

```py
import time

import librosa
import numpy as np

wav = np.random.rand(528318)


# warm up
for i in range(10):
    librosa.pyin(
        wav,
        fmin=librosa.note_to_hz("C2"),
        fmax=librosa.note_to_hz("C5"),
        sr=16_000,
        frame_length=1024,
        hop_length=256,
        center=False,
    )


start = time.time()
for i in range(100):
    librosa.pyin(
        wav,
        fmin=librosa.note_to_hz("C2"),
        fmax=librosa.note_to_hz("C5"),
        sr=16_000,
        frame_length=1024,
        hop_length=256,
        center=False,
    )
print((time.time() - start) / 100)

# py 3.10, numpy: 1.26.4, numba: 0.59.1
# this PR on top of 0.10.1: 1.8619681692123413
# 0.10.1: 2.543446371555328
# ~26% faster
```

bmcfee:
I just ran my own benchmark locally and performance is indistinguishable between the current main and this PR.

```python
import librosa
import numpy as np

# Force a jit compilation if we didn't have it cached already
librosa.pyin(y=y, sr=sr, fmin=32, fmax=2000);

# Current main
%timeit librosa.pyin(y=y, sr=sr, fmin=32, fmax=2000)
# 1.43 s ± 22 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# PR 1825
%timeit librosa.pyin(y=y, sr=sr, fmin=32, fmax=2000)
# 1.46 s ± 32 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

That's on the 5 second trumpet example.  On a longer example (nutcracker, 119 seconds), we get:

```
# Main
26.1 s ± 259 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# PR 1825
27.3 s ± 266 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

twoertwein:
> I'd be surprised if this is faster in general, though I can believe it will be faster in certain circumstances. 

It would be great if you (and your collaborators) could run the following script on other platforms and other python version to see whether the pure numpy-version is worth it or not. Be-aware my results seem to be a bit noisy - sometimes having a larger range of F0 is faster than with a smaller range, which seems unexpected.

```py
import time

import librosa
import numpy as np

for length in (10_000, 100_000):
    for fmax in ("C8", "C7", "C6", "C5", "C4", "C3"):
        wav = np.random.rand(length)

        # warm up
        for i in range(10):
            librosa.pyin(
                wav, fmin=librosa.note_to_hz("C2"), fmax=librosa.note_to_hz(fmax)
            )

        start = time.time()
        for i in range(100):
            librosa.pyin(
                wav, fmin=librosa.note_to_hz("C2"), fmax=librosa.note_to_hz(fmax)
            )
        print(f"{length=}, {fmax=}", (time.time() - start) / 100)

# py 3.10, numpy: 1.26.4, numba: 0.59.1

# 0.10.1:
length=10000, fmax='C8' 0.15894932985305787
length=10000, fmax='C7' 0.1073629117012024
length=10000, fmax='C6' 0.0850246500968933
length=10000, fmax='C5' 0.05933178901672363
length=10000, fmax='C4' 0.02861675024032593
length=10000, fmax='C3' 0.012354247570037842
length=100000, fmax='C8' 1.0184268498420714
length=100000, fmax='C7' 0.6894018507003784
length=100000, fmax='C6' 0.5472757506370545
length=100000, fmax='C5' 0.3582835292816162
length=100000, fmax='C4' 0.20101286888122558
length=100000, fmax='C3' 0.0643980097770691

# this PR on top of 0.10.1:
length=10000, fmax='C8' 0.11170340061187745
length=10000, fmax='C7' 0.0782641363143921
length=10000, fmax='C6' 0.07668728113174439
length=10000, fmax='C5' 0.04809794187545777
length=10000, fmax='C4' 0.0173807692527771
length=10000, fmax='C3' 0.011447958946228028
length=100000, fmax='C8' 0.598180022239685
length=100000, fmax='C7' 0.41540180921554565
length=100000, fmax='C6' 0.4820923900604248
length=100000, fmax='C5' 0.3346623587608337
length=100000, fmax='C4' 0.1822682499885559
length=100000, fmax='C3' 0.05109306097030639
```

> I believe the vectorized argmax / fancy indexing was not numba-compatible at the time we implemented this

The argmax seems to be compatible now but the fancy indexing is still not supported.

Using just the argmax+jit (and the indexing in the loop) - surprisingly it becomes slower:
```py
length=10000, fmax='C8' 0.197688729763031
length=10000, fmax='C7' 0.14093083143234253
length=10000, fmax='C6' 0.10329761981964111
length=10000, fmax='C5' 0.062146799564361574
length=10000, fmax='C4' 0.02347614288330078
length=10000, fmax='C3' 0.012690157890319824
length=100000, fmax='C8' 1.372423028945923
length=100000, fmax='C7' 0.9433927917480469
length=100000, fmax='C6' 0.6792442488670349
length=100000, fmax='C5' 0.4484334897994995
length=100000, fmax='C4' 0.26358938932418824
length=100000, fmax='C3' 0.06613000869750976
```

bmcfee:
Stepping back from this a bit, the benchmark really shouldn't be on pyin - it should be on viterbi directly.  Benchmarking on pyin is including a lot of other cruft in here that could be clouding the timing information.  Additionally, the benchmarking code here is not all that robust, as it's only measuring average time, and does not properly measure the statistics of individual runs - this is exactly what the timeit package is for, and why I used it above.  (But, for the sake of consistency, I'll run it anyway.)

Now, let's reason about what exactly is proposed here.  The viterbi function is jit accelerated primarily because of the sequential dependency across frames which prevents vectorization in the time dimension.  When we have relatively few frames, this isn't such a big deal, but the benefits will stack up on long signals.  The proposed change removes this compilation, so the time direction loop (how viterbi is used in pyin) will be slower, but each step of the loop could be a bit faster by relying directly on compiled numpy code.  This will really only benefit when you have a large state space and a short sequence.

In principle, the jit-compiled inner loop should be about as fast as the vectorized argmax + fancy index, notwithstanding some constant overhead of going through LLVM.  This is what I'm seeing on my dev machine (linux, x64), and what we should expect in general.  If that's not happening on certain platforms (M2?), that seems like a problem upstream of us (numba, llvmlite) and not something we should try to circumvent directly here.

---
Benchmark results:

## Main

```
length=10000, fmax='C8' 0.3628903889656067
length=10000, fmax='C7' 0.3260782194137573
length=10000, fmax='C6' 0.2542057347297668
length=10000, fmax='C5' 0.20305131435394286
length=10000, fmax='C4' 0.07216196298599244
length=10000, fmax='C3' 0.033560080528259276
length=100000, fmax='C8' 1.2282242941856385
length=100000, fmax='C7' 0.9025627398490905
length=100000, fmax='C6' 0.6297197985649109
length=100000, fmax='C5' 0.43032772779464723
length=100000, fmax='C4' 0.28486912488937377
length=100000, fmax='C3' 0.08577804088592529
```


## PR 1825

```
length=10000, fmax='C8' 0.33808276176452634
length=10000, fmax='C7' 0.31756269931793213
length=10000, fmax='C6' 0.25140995502471924
length=10000, fmax='C5' 0.18975526809692383
length=10000, fmax='C4' 0.14299791812896728
length=10000, fmax='C3' 0.03413930177688598
length=100000, fmax='C8' 1.4511582088470458
length=100000, fmax='C7' 1.067232985496521
length=100000, fmax='C6' 0.7453772830963135
length=100000, fmax='C5' 0.4888764786720276
length=100000, fmax='C4' 0.29793501138687134
length=100000, fmax='C3' 0.0984840989112854
```

twoertwein:
Seems to be a Mac/numba issue - running your timit results (after a warmup):

```py
# with the PR
%timeit librosa.pyin(y=y, sr=sr, fmin=32, fmax=2000)
639 ms ± 10.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# without the PR
%timeit librosa.pyin(y=y, sr=sr, fmin=32, fmax=2000)
1.02 s ± 8.38 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Closing the PR as it is an upstream issue

PR #1829, opened 2024-04-03
librosa.cite()

### LINKED ISSUE #1674: Should librosa retrieve its own DOI on the web?
Since #1666 (which closed #1659), we're doing a better job than before at crediting Librosa contributors on Zenodo: see `scripts/create_author_list.py` and its result `.zenodo.json`.

And what a nice result: https://zenodo.org/record/7650421

But i'm a little concerned of human errors in credit attribution. The Zenodo webpage is not super intuitive: one has to scroll down to the "Cite as" panel or access a suffix page (`/export/hx`) to view the corresponding BibTeX entry.

Besides, the fact that we have a DOI for the latest version + a separate DOI for every version (including release candidates) is nice but makes things even more complicated for users: they need to look up their version of librosa, visit Zenodo, scroll to the appropriate version, and eventually download the appropriate resource (BibTeX or something else).

Hence my proposal: a utility function (say, `librosa.cite()`) which returns all the metadata information (DOI, author list ...) of the currently installed version. This is not entirely obvious because the DOI is not included in the release. It's the other way around: the creation of a GitHub tag triggers a new, random DOI. So implementing `librosa.cite()` would necessarily involve an Internet query.

I have begun to look into the feasibility of this, and found the `pyzenodo3` package. This works:

```python
import ast
import pyzenodo3

zen = pyzenodo3.Zenodo()
record = zen.find_record_by_github_repo('librosa/librosa')
record_dict = ast.literal_eval(str(record))
```

Unfortunately, the `record.get_versions()` method is broken. So i anticipate that we're going to have to write some in-house code for this. Ideas welcome But i would also like to hear everyone's thoughts on what the inputs and outputs of `librosa.cite()` should be, and whether it's a good idea at all.

bmcfee:
#### Reference Issue
Fixes #1674

This PR adds `librosa.cite` (top-level import from util.files).

This function allows users to get the DOI (by doi.org url) for either the current version of librosa (excepting development versions) or any other release known to the index.

Example:

```python
>>> librosa.cite()  # Uses whatever librosa.__version__ is
'https://doi.org/10.5281/zenodo.8252662'

>> librosa.cite("0.9.2")  # Cite an older version
'https://doi.org/10.5281/zenodo.6759581'
```
etc.

Exceptions are thrown if the version string is not in the index.  A more helpful exception is thrown if the version string is a dev version.

#### Any other comments?

Maintaining the release index is a manual job for now.  I've added a helper script that scrapes our zenodo page to build the index file (msgpack format), but this needs to be manually synchronized to the librosa-data repository after each release.

I've also put some hacks into the librosa-data repo (out of band here) to support non-audio data files.  It's a bit crufty right now, as everything is living under "audio/"; we can restructure this later on but it works well enough for now.

There is a slight potential for a security issue here. Pooch uses checksums to verify the integrity of downloaded data files, and this is generally a good thing.  However, if we're updating the version index at each release, this would break old installations because the checksum would fall out of sync.  (We could get around this by having a separate version index for each release, but that seems a little silly and difficult to maintain.)

My solution here is to disable checksums on the version index file, as noted in the original #1674 issue.  The potential for a security issue would arise in a MITM attack on the data repo.  I don't think it's very likely to happen, and even if it did, it would only arise when someone explicitly accesses the citation index.  This is all to say, I'm not worried about it, but I am aware of the possibility.


bmcfee:
Had to bump the minimum pooch to 1.1 to get hash bypass support, but otherwise I think this is good to go.

lostanlen:
woah ... impressive !

Issue #1842, opened 2024-05-29
norm argument in MFCC used in 2 different functions


ssslakter:
**Describe the bug**
I was trying to get consistent results from MFCC in torchaudio and librosa but found inconsistencies. These were related to the issue with normalization in Mel spectrograms discussed in this issue https://github.com/pytorch/audio/issues/1058. The solution was to use `'slaney'` norm in the `MelSpectrogram` method as explained in [this comment](https://github.com/pytorch/audio/issues/1058#issuecomment-778476093). However, if you try to pass the `norm` parameter for MFCC as a MelSpectrogram parameter in kwargs, it is also used as a `norm` parameter for `scipy.fftpack.dct`, which has different valid values.

**To Reproduce**
Use `norm='slaney'` in the MFCC function:
```python
librosa.feature.mfcc(
    y=y,
    sr=sr,
    n_mfcc=n_mfcc,
    n_fft=n_fft,
    hop_length=hop_len,
    win_length=win_len,
    pad_mode="reflect",
    n_mels=n_mels,
    htk=True,
    norm='slaney')
```
This will show an error.

**Expected behavior**
It should be possible to use MFCC with the 'slaney' norm for MelSpectrogram.

**Possible solution**
The simplest solution is to rename the norm for the DCT so that there are no name collisions in function arguments. 
You can see the proposed changes in my fork: https://github.com/ssslakter/librosa/commit/65d2e2bbabee2a0c9fe83baf0f3fb6daffaea380

bmcfee:
Thanks for catching this!  I'm a little confused by the bug report though:



> It should be possible to use MFCC with the 'slaney' norm for MelSpectrogram.

This is the default behavior.  In fact, if you're calling MFCC, there's no direct way to use any Mel basis normalization *other than* slaney, since setting the `norm` parameter will pass to the DCT instead of the mel spectrogram.

The error that you're seeing is becuase DCT doesn't know what to do with "slaney" normalization.  Note that there is no ambiguity here in terms of code: `norm` in MFCC will always go to DCT, `norm` in melspectrogram will always go to the filter constructor.  The variable name collision is an unfortunate accident, but the behavior is well specified.


> The simplest solution is to rename the norm for the DCT so that there are no name collisions in function arguments.

We could also rename the mel basis normalization, but I agree that renaming the DCT norm parameter is probably an overall simpler change.

Meanwhile, you can work around this by building your MFCC in two steps:

```python
[...]

PR #1844, opened 2024-05-31
Fix name collision for norm argument in mfcc

ssslakter:
#### Reference Issue
fixes https://github.com/librosa/librosa/issues/1842

This pr fixes name collisions between norm arguments in `mfcc` for `melspectrogram` and for `scipy.fftpack.dct`.

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -1836,7 +1836,7 @@ def mfcc(
     S: Optional[np.ndarray] = None,
     n_mfcc: int = 20,
     dct_type: int = 2,
-    norm_dct: Optional[str] = "ortho",
+    dct_norm: Optional[str] = "ortho",
     lifter: float = 0,
     **kwargs: Any,
 ) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


bmcfee:
Implementing it in this way would constitute an API-breaking change.

Alternatively, we could add a `mel_norm` parameter, keeping the existing `norm` parameter for DCT usage.  This way, the API change is additive and would not break existing code.

It won't be identical to torchaudio's API, but I can live with that.

ssslakter:
Right! Makes sense to me, `norm` inside of kwargs would be taken away by `norm` keyword argument anyway😅

librosa/feature/spectral.py
```diff
@@ -16,7 +16,7 @@
 from ..core.spectrum import power_to_db, _spectrogram
 from ..core.constantq import cqt, hybrid_cqt, vqt
 from ..core.pitch import estimate_tuning
-from typing import Any, Optional, Union, Collection
+from typing import Any, Literal, Optional, Union, Collection
```
bmcfee, line 19: Literal needs to import from typing_extensions, not typing

ssslakter: Oh, I see... think I fixed it now

bmcfee (approved):

bmcfee:
Thanks @ssslakter !

PR #1850, opened 2024-06-26
Update index.rst corrected broken link

rishabh11336:
#### Reference Issue

image

Changed the broken link to correct one
from 
https://conference.scipy.org/proceedings/scipy2015/pdfs/brian_mcfee.pdf  
image

to 

http://conference.scipy.org.s3-website-us-east-1.amazonaws.com/proceedings/scipy2015/brian_mcfee.html
image

#### Any other comments?
I didn't find any issue no. related to similar issue. Kindly merge my request!
Guide me for future!

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,7 +6,7 @@ blocks necessary to create music information retrieval systems.
 
 For a quick introduction to using librosa, please refer to the :doc:`tutorial`.
 For a more advanced introduction which describes the package design principles, please refer to the
-`librosa paper <https://conference.scipy.org/proceedings/scipy2015/pdfs/brian_mcfee.pdf>`_ at
+`librosa paper <http://conference.scipy.org.s3-website-us-east-1.amazonaws.com/proceedings/scipy2015/brian_mcfee.html>`_ at
 `SciPy 2015 <https://scipy2015.scipy.org>`_.
 
 Citing librosa


bmcfee:
Thanks for this.  I've been punting on updating this long broken link until the scipy proceedings URLs have stabilized.  (We've discussed this "temporary" outage a bit on the scipy discord, and librosa is by far not the only project bumping into this link rot.)

Looking at the URL you're using here, I don't think this is stable yet.

rishabh11336:
Understood, 
I might not be getting a complete gravity of situation but,
I can help with the creating GitHub page for paper, which can be hosted on librosa GitHub account independent of any other service

Thank you for your prompt reply...

bmcfee:
Thanks, but that won't be necessary.  The issue has to do with the permanent archive of scipy proceedings, which was disrupted by a server migration some months back.  There is an existing issue https://github.com/scipy-conference/conference.scipy.org/issues/58 to track this, and once the whole thing is resolved, we can update the link here (if necessary).

bmcfee:
Circling back on this, I should have mentioned that the other reason I don't want to just hack the url in the docs here is that it won't fix the broken link in multiversion doc builds.  So the link would work on the "main" / development build of the docs, but not on the version stamped docs that most people would land on.

PR #1858, opened 2024-07-20
fix onset documentation specshow samplerate

BenedictSt:
If not set the power spectogram and onset strength might not align properly.

Screenshot 2024-07-20 at 16 16 21

### DIFF
--- a/librosa/onset.py
+++ b/librosa/onset.py
@@ -142,7 +142,7 @@ def onset_detect(
     >>> D = np.abs(librosa.stft(y))
     >>> fig, ax = plt.subplots(nrows=2, sharex=True)
     >>> librosa.display.specshow(librosa.amplitude_to_db(D, ref=np.max),
-    ...                          x_axis='time', y_axis='log', ax=ax[0])
+    ...                          x_axis='time', y_axis='log', ax=ax[0], sr=sr)
     >>> ax[0].set(title='Power spectrogram')
     >>> ax[0].label_outer()
     >>> ax[1].plot(times, o_env, label='Onset strength')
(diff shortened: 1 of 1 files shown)


lostanlen:
Hello @BenedictSt, thank you for raising this problem.
I don't understand what happened here. Presumably `sr` is both the return value of `load` and the default value of `specshow`, so it's not clear to me (1) what the problem was (2) how your PR fixed it. Can you please explain from the start? Thanks!

BenedictSt:
When you load audio with a different sample rate than default, you have to specify it in `specshow()`, so they align properly.

Example:
```python
import librosa as librosa
import numpy as np

# note: load with other than default sample rate
y, sr = librosa.load(librosa.ex('trumpet'), sr=44100)

o_env = librosa.onset.onset_strength(y=y, sr=sr)
times = librosa.times_like(o_env, sr=sr)
onset_frames = librosa.onset.onset_detect(onset_envelope=o_env, sr=sr)

D = np.abs(librosa.stft(y))
fig, ax = plt.subplots(nrows=2, sharex=True)
librosa.display.specshow(librosa.amplitude_to_db(D, ref=np.max),
                         x_axis='time', y_axis='log', ax=ax[0])
ax[0].set(title='Power spectrogram')
ax[0].label_outer()
ax[1].plot(times, o_env, label='Onset strength')
ax[1].vlines(times[onset_frames], 0, o_env.max(), color='r', alpha=0.9,
           linestyle='--', label='Onsets')
ax[1].legend()

plt.show()
```
previous
Here the power spectrogram does not align with the onset_strength. Also the clip is 5sec long => x-axis ticks are off.

My change sets the `sr` parameter:
```python
librosa.display.specshow(librosa.amplitude_to_db(D, ref=np.max),
                         x_axis='time', y_axis='log', ax=ax[0],
                         sr=sr) # added: set sr parameter
```
fixed

now the align properly.

bmcfee (approved):
LGTM, thanks!

PR #1866, opened 2024-08-17
Revised dtw docstring, fixes #1741

### LINKED ISSUE #1741: Size of warping path (wp) obtained by Librosa Sub-sequence DTW is larger than expected
Hi,

I've encountered an issue while utilizing the librosa sub-sequence Dynamic Time Warping (DTW) function. It seems that the size of the obtained warping path (wp) is unexpectedly larger than the expected value of N.

I'm using the following code to compute DTW:

D, wp = librosa.sequence.dtw(
            X=patch,
            Y=patch_ref,
            metric='euclidean',
            step_sizes_sigma=[[1, 0], [0, 3], [1, 3]],
            weights_mul=np.array([1, 1, 1]),
            band_rad=0.25,
            subseq=True,
            backtrack=True,
        )

Where X is an np.ndarray with a shape of (K, N), where K is 13 and N is 100, and Y is an np.ndarray with a shape of (K, M), where K is also 13 but M is 1823.

According to the documentation of librosa.sequence.dtw, the expected shape of the warping path wp should be (N, 2). However, in my specific case, the size of wp is (101, 2), which is unexpected. The size of D in my case is (N,M).

I have enclosed the relevant data with this report. Please see the attached zip file: 

[dtw_files.zip](https://github.com/librosa/librosa/files/12408355/dtw_files.zip)

librosa.__version__ = '0.9.2'

Your feedback on this would be greatly appreciated.
Thank you,

bmcfee:
#### Reference Issue
Fixes #1741 

Small revisions to the DTW docstring.

#### Any other comments?

@stefan-balke can you peep this quickly and give your take?

### DIFF
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -272,7 +272,7 @@ def dtw(
     -------
     D : np.ndarray [shape=(N, M)]
         accumulated cost matrix.
-        The final value ``D[-1, -1]`` is the total alignment cost.
+        The value at the final index position ``D[-1, -1]`` is the total alignment cost.
     wp : np.ndarray [shape=(L, 2)]
         Warping path with index pairs.
         Each row of the array contains an index pair (n, m).
--- a/librosa/sequence.py
+++ b/librosa/sequence.py
@@ -243,7 +243,7 @@ def dtw(
         in `scipy.spatial.distance.cdist()`
 
     step_sizes_sigma : np.ndarray [shape=[n, 2]]
-        Specifies allowed step sizes as used by the dtw.
+        Specifies allowed step sizes as used by the DTW.
 
     weights_add : np.ndarray [shape=[n, ]]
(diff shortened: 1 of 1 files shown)


librosa/sequence.py
```diff
@@ -272,12 +272,14 @@ def dtw(
     -------
     D : np.ndarray [shape=(N, M)]
         accumulated cost matrix.
-        D[N, M] is the total alignment cost.
-        When doing subsequence DTW, D[N,:] indicates a matching function.
-    wp : np.ndarray [shape=(N, 2)]
+        The final value ``D[-1, -1]`` is the total alignment cost.
```
stefan-balke, line 275: could you add "at index position" here? My brain was (don't know why) tempted to interpret it as $D \in [-1, 1]$ first.

stefan-balke:
Just a small comment. Otherwise LGTM.

PR #1878, opened 2024-10-02
Various doc updates 0.11

### LINKED ISSUE #1809: effects.trim:  top_db can be negative? doc says top_db is a number > 0.
**Describe the bug**

While trying to figure out why efffects.trim is failing to trim silence I came across an old issue on same topic which had reference working code which I confirmed worked in my case of:

`librosa.effects.trim(audio, top_db=-10, ref=0.0173, frame_length=512, hop_length=1)`

Ref an older issue:
https://github.com/librosa/librosa/issues/1445

This code snippet actually works and solved my issue but I am scratching my head on how this works while doc is saying top_db should never be negative. 

So is doc wrong or incomplete?

**Expected behavior**

Crash and burn? =) The documentation explicitly says top_db is float > 0

documentation:

```
def trim(
    y: np.ndarray,
    *,
    top_db: float = 60,
    ref: Union[float, Callable] = np.max,
    frame_length: int = 2048,
    hop_length: int = 512,
    aggregate: Callable = np.max,
) -> Tuple[np.ndarray, np.ndarray]:
    """Trim leading and trailing silence from an audio signal.

    Parameters
    ----------
    y : np.ndarray, shape=(..., n)
        Audio signal. Multi-channel is supported.
    top_db : number > 0
        The threshold (in decibels) below reference to consider as
        silence
```

**Software versions**

Latest librosa

### LINKED ISSUE #1828: `key_to_notes` with an option to return only the notes from the scale
I was a bit surprised at first when I used `key_to_notes` as I expected the default behavior to return only the seven notes of the scale (with possibly the option for pentatonic scales). I reckon it would be more consistent with the `key_to_degrees` function.

```python
>>> librosa.key_to_notes("C:min")
['C', 'D', 'E♭', 'F', 'G', 'A♭', 'B♭']
```

Of course, it's possible to obtain the same result by resorting to an extra step with the `key_to_degrees` function:
```python
>>> key = "C:min"
>>> degs = librosa.key_to_degrees(key)
>>> np.array(librosa.key_to_notes(key))[degs].tolist()
['C', 'D', 'E♭', 'F', 'G', 'A♭', 'B♭']
```

### LINKED ISSUE #1837: centroid is calculated alwais using hanning window
If spectral bandwidth is requested with other window than hamming, the spectral centroid should be calculated using that window. 

The following code in the function should be fixed:
    # centroid or center?
    if centroid is None:
        centroid = spectral_centroid(
            y=y, sr=sr, S=S, n_fft=n_fft, hop_length=hop_length, freq=freq
        )

should be:
    # centroid or center?
    if centroid is None:
        centroid = spectral_centroid(
            y=y, sr=sr, S=S, n_fft=n_fft, hop_length=hop_length, freq=freq, window = window
        )

bmcfee:
#### Reference Issue
Fixes #1837 
Fixes #1828 
Fixes #1809 

This PR includes a handful of small documentation changes to clarify behaviors in a few functions.  There is no functionality change here, so CI is primarily for linting.

A once-over from other contributors would be welcome.

#### Any other comments?

I also took the opportunity to simplify some slicing syntax in `effects.trim`.  This doesn't change behavior, but it's a bit more readable using ellipses than a list of slices.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -409,3 +409,6 @@ smv_prefer_remote_refs = True
 # Set the latest version via command-line
 #  $ sphinx-multiversion -D smv_latest_version=$(./scripts/get_latest_release.sh) docs build/html
 # smv_latest_version = version
+
+
+linkcheck_allow_unauthorized = True
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,7 +6,7 @@ blocks necessary to create music information retrieval systems.
 
 For a quick introduction to using librosa, please refer to the :doc:`tutorial`.
 For a more advanced introduction which describes the package design principles, please refer to the
-`librosa paper <https://conference.scipy.org/proceedings/scipy2015/pdfs/brian_mcfee.pdf>`_ at
+`librosa paper <https://doi.org/10.25080/Majora-7b98e3ed-003>`_ at
 `SciPy 2015 <https://scipy2015.scipy.org>`_.
 
 Citing librosa
--- a/librosa/core/notation.py
+++ b/librosa/core/notation.py
@@ -707,7 +707,7 @@ def key_to_notes(key: str, *, unicode: bool = True, natural: bool= False) -> Lis
     ['B#', 'C#', 'Dn', 'D#', 'En', 'E#', 'F#', 'F##', 'G#', 'An', 'A#', 'B']
 
     We can combine this with ``key_to_degrees`` to get the notes for a given scale:
-    
+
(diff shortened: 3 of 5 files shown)


bmcfee:
A couple of additions here:
- scipy proceedings have working DOI redirects now, so I've finally updated the link for the 2015 paper.
- Updated the linkcheck configs to work even without auth; this should fix the failing link checks to the columbia server for @dpwe's reference code

PR #1881, opened 2024-10-08
Fix time stretch round error

orkunkinay:
This PR fixes a TypeError in the `time_stretch` function where `round()` was applied to an ndarray inappropriately. The fix replaces `round()` with `np.round()` to handle the numpy arrays correctly.

The error message was: `TypeError: type numpy.ndarray doesn't define __round__ method`.

Changes:
- Updated `time_stretch` function in `effects.py` to use `np.round()` instead of `round()`.
- Updated `AUTHORS.md` to add Orkun Kınay as collaborator.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -108,7 +108,8 @@ Contributors
 * Jize Guo <https://github.com/alumkal>
 * Phillip S.M. Skelton <https://github.com/psmskelton>
 * Matt Pitkin <https://github.com/mattpitkin>
-
+* Orkun Kınay <https://github.com/orkunkinay>
+  
 Some feature extraction code was based on <https://github.com/ronw/frontend> by Ron Weiss.
 
 Large portions of librosa were ported from MATLAB code by Dan Ellis <http://www.ee.columbia.edu/~dpwe/resources/matlab/>.
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -392,7 +392,7 @@ def time_stretch(y: np.ndarray, *, rate: float, **kwargs: Any) -> np.ndarray:
     )
 
     # Predict the length of y_stretch
-    len_stretch = int(round(y.shape[-1] / rate))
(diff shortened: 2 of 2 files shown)


orkunkinay:
I am getting the bpm's of the songs with  `librosa.beat.beat_track` which outputs `Tuple[_FloatLike_co | ndarray, ndarray]`. Then I divide the bpm's of the songs to get the rate which again gives `ndarray` as output. Then I pass that ndarray rate as parameter to `librosa.effects.time_stretch`. I am using librosa's own bpm detecting function which gives ndarray, that is the reason I wanted to open this PR and tell my idea.

bmcfee:
Ah I see - I hadn't considered plugging tempo estimates into the time_stretch like you're doing.  That makes sense.

There are other bits of this function that will not be correct if `rate` is an ndarray though; the docstring and type annotation should be revised if we want to explicitly support it.  Additionally, we should verify that `rate.size == 1`, otherwise the non-negativity check will fail, as will the `istft` call that would have a multidimensional `length` value.

bmcfee:
Finally circling back on this after a hectic couple of months.  I'm actually unable to reproduce the behavior that motivated the change here.  Can you fill in the issue template and document your environment, specifically the version of numpy you're using?

On 1.26.4, I can do the following without issue:

```python
>>> import numpy as np
>>> y = np.random.randn(2, 200)
>>> rate = 1.5
>>> int(round(y.shape[-1] / rate))
133
>>> int(np.round(y.shape[-1] / rate))
133
>>> rate = np.array(1.5)  # Force rate to be an array
>>> int(round(y.shape[-1] / rate))
133
>>> int(np.round(y.shape[-1] / rate))
133
```

Does this not work in numpy 2.x or something?

Or, is it the case that your `rate` parameter is actually multidimensional (and not just an ndarray object of `size=1` and `shape=()`).  If that's the case, then the function is not going to work anyway because it would result in a variable-length output array allocation.

Issue #1425, opened 2021-12-20
Autocorrelation off-by-one error


PhilStem:
https://github.com/librosa/librosa/blob/381efbd684c01ae372220526352d34fd732d3b1d/librosa/core/pitch.py#L402

The flipped array `y_frames[..., win_length::-1, :]` has length `win_length+1` at axis -2. I think the line should be `b = np.fft.rfft(y_frames[..., win_length:0:-1, :], frame_length, axis=-2)`. The code I tried seems to produce more accurate results:

```python
sr = 44100
y = np.sin(100*2*np.pi*np.linspace(0, 1, sr, endpoint=False))
res = yin(y, 60, 1000, sr=sr, center=False)
print(res)
```

Before:
```
[100.00016561  99.98214417 100.00800707 100.00344982  99.98128853
 100.00550897 100.00639164  99.9812107  100.00265195 100.00886901
  99.98193343  99.99954079 100.01079017  99.98343062  99.99629365
 100.01209364  99.98562683  99.99304064 100.01274637  99.98840202
  99.98992114 100.01274092  99.99160123  99.98707952 100.01209245
  99.99504764  99.98465883 100.01083594  99.998557    99.98279243
 100.00902398 100.00195128  99.98159414 100.00672526 100.00506986
  99.98114769 100.00402355 100.00777726  99.98149702 100.00101708
 100.00996745  99.98263923  99.99781788 100.01156496  99.98452163
  99.99455075 100.01252378  99.98704407  99.99135116 100.01282489
  99.99006641  99.98836181 100.0124733   99.99342027  99.98572728
 100.01149502  99.99692308  99.98358661 100.00993467 100.00039234
  99.98206405 100.00785359 100.0036581   99.98125865 100.00532858
 100.00657243  99.981234   100.00245121 100.00901523  99.98200972
  99.99932716 100.0108968   99.98355639  99.99607547 100.01215761
  99.98579538  99.99282693 100.01276643  99.98860398  99.98972145
 100.0127174   99.99182533  99.9869035 ]
```

After:
```
[100.00016561 100.00033933 100.00027569 100.00015915 100.00032482
 100.00029493 100.00015733 100.00030679 100.00031272 100.00016005
 100.00028612 100.00032835 100.00016701 100.00026388 100.00034112
 100.00017777 100.00024128 100.00035042 100.00019176 100.00021953
 100.00035573 100.00020834 100.0001998  100.00035666 100.0002268
[...]

PR #1063, opened 2020-01-18
YIN and pYIN

bshall:
#### Reference Issue
Fixes #527 

#### What does this implement/fix? Explain your changes.
This pull request implements the YIN and pYIN algorithms for pitch tracking. The YIN function is based on @lostanlen's PR #974 (with a few modifications) while the pYIN function is based on this [paper](https://ieeexplore.ieee.org/document/6853678).

#### Any other comments?
Both functions work well but some refactoring is definately needed. I compared the outputs of pYIN to the official [vamp plugin](https://code.soundsoftware.ac.uk/projects/pyin) and the results are comparable. I haven't added any special treatment of low amplitude frames yet so silent frames with periodic noice occasionally give wrong results.

Also, note that I haven't used the `librosa.core.autocorrelate` function. The `librosa.core.autocorrelate` computes equation (2) in the [YIN paper](https://asa.scitation.org/doi/10.1121/1.1458024) which seems to perform worse than the one used here (which computes equation (1)). I tried scaling by the window size (as suggested in the paper) but it didn't improve things much.   

I haven't implemented any tests yet so it would be good to get some guidance there.

Issue #1680, opened 2023-03-01
pyin crashes for low frequencies and narrow fmin/fmax bands


bluenote10:
**Describe the bug**

Currnelty, `pyin` cannot handle (I) low frequencies and (II) narrow bands of `fmin` / `fmax`.

**To Reproduce (issue I)**

Testing pyin for low midi keys:

```py
import librosa

def test_pyin(frequency: float):
    sr = 44100
    y = librosa.tone(frequency=frequency, length=sr, sr=sr)
    librosa.pyin(
        y,
        fmin=0.5 * frequency,
        fmax=2.0 * frequency,
        sr=sr,
    )

for midi_key in reversed(range(20)):
    frequency = librosa.midi_to_hz(midi_key)
    print(f"Testing midi key {midi_key}, frequency = {frequency}")
    test_pyin(frequency)
```

This works up to midi key 17 but then errors. Full output:

```
Testing midi key 19, frequency = 24.499714748859326
Testing midi key 18, frequency = 23.12465141947715
Testing midi key 17, frequency = 21.826764464562746
Testing midi key 16, frequency = 20.601722307054366
Traceback (most recent call last):
  File "/tmp/debug.py", line 18, in <module>
    test_pyin(frequency)
  File "/tmp/debug.py", line 7, in test_pyin
    librosa.pyin(
  File "/home/fabian/git/ZenTabs/.venv/venv_zentabs/lib/python3.8/site-packages/librosa/core/pitch.py", line 819, in pyin
    parabolic_shifts = _parabolic_interpolation(yin_frames)
  File "/home/fabian/git/ZenTabs/.venv/venv_zentabs/lib/python3.8/site-packages/librosa/core/pitch.py", line 485, in _parabolic_interpolation
    shiftsi[..., -1] = 0
IndexError: index -1 is out of bounds for axis 1 with size 0
```

**To Reproduce (issue II)**

I also noticed that pyin doesn't like to incorporate too much prior knowledge about a frequency, i.e., making the `fmin` / `fmax` range relatively narrow. In my use case I knew that I'm searching around a certain expected `f0_expected`, so I figured it is wise to help pyin by limiting the search range symmetrically around that frequency. That fails however for all midi keys:

```py
import librosa

def test_pyin(frequency: float):
    sr = 44100
    y = librosa.tone(frequency=frequency, length=sr, sr=sr)
    librosa.pyin(
        y,
        fmin=0.9 * frequency,
[...]

PR #1882, opened 2024-10-25
Improved CMND implementation for YIN / pYIN

dsuedholt:
**TL;DR**: We can noticeably improve the accuracy of the YIN implementation by getting rid of the `win_length` parameter entirely and simply calculating the autocorrelation function (ACF) over the entire frame. Mathematical reasoning and some benchmarks provided below.

Apologies in advance for the wall of text. The devil is very much in the details with this one and I wanted to document it thoroughly.

## ACF Calculation

The main point of this pull request is to replace the calculation of the ACF: https://github.com/librosa/librosa/blob/d5aa7e1ae443c917fa53bef31055ed2613d63133/librosa/core/pitch.py#L393-L395 with a call to librosa's existing implementation: 

```python
acf_frames = audio.autocorrelate(y_frames, max_size=max_period+1, axis=-2)
``` 

Looking at this [discussion](https://github.com/librosa/librosa/pull/1063#discussion_r435547835), I believe the intention behind the current implementation is to calculate a type I ACF with a constant window size. Given window length $N$, this is defined as

$$r(k) = \sum_{n=0}^{N-1}x[n]x[n + k]$$

By contrast, the type II ACF uses an integration window that shrinks as the lag increases:

$$r'(k) = \sum_{n=0}^{N-1-k}x[n]x[n + k]$$

I'll go into some detail as to why I believe that 

1. the current implementation does not actually correctly calculate the type I ACF
2. the librosa implementation should use the type II ACF anyway (as implemented in `librosa.core.autocorrelate`)

It's true that the authors of the original YIN paper recommend to use the type I ACF, because the type II ACF implicitly tapers to 0 as $k \rightarrow N$. However, in the formulation of the type I ACF, the $n + k$ index will exceed the window size. This is not an issue discussed in the paper, since the authors consider the signal in its entirety, rather than on a frame-by-frame basis, so they can essentially "look into the future". 

As I understand it, the current implementation addresses this by making a distinction between `frame_length` $N$ and `win_length` $W < N$, where `win_length` essentially places an upper bound on the calculated lags. Mathematically,

$$r(k) = \sum_{n=0}^{W-1}x[n]x[n+k] \text{ for } 0\leq k < N - W$$

However, due to the `win_length:0:-1` indexing, the current code instead calculates:

$$r(k) = \sum_{n=0}^{W-1}x[n]x[n+k+1] \text{ for } 0\leq k < N - W$$

This off-by-one error was introduced [here](https://github.com/librosa/librosa/issues/1425). I think the reason that this change actually improved accuracy at the time is that it might have cancelled out another off-by-one error in the following calculation of the cumulative mean normalized difference (CMND) function, which this PR also fixes (see next section).

But this brings me to the second point of the argument: rather than fixing this calculation, I propose to drop the `win_length` parameter entirely and simply use `librosa.core.audio.autocorrelate` to calculate the type II ACF over the entire frame. This has several benefits:

- it simplifies the function call interface by removing a parameter
- using the type II ACF is standard in frame-by-frame implementations of YIN, in e.g. [aubio](https://github.com/aubio/aubio/blob/master/src/pitch/pitchyin.c) and also in the [original pYIN implementation](https://code.soundsoftware.ac.uk/projects/pyin/repository/entry/YinUtil.cpp#L57), so it aligns librosa with expected behavior.

As an example, the different ACFs are illustrated below for a 2048-sample frame from track 2 of the vocadito dataset. The highest peak of the type II ACF (excluding 0 lag) is at the correct period, while the type I ACF's peak at the correct period is slightly lower than the peak at double the period.

ACF comparison

## CMND Calculation

The next step in the YIN algorithm is to calculate the difference function. As mentioned before, the current implementation makes another off-by-one error here: https://github.com/librosa/librosa/blob/d5aa7e1ae443c917fa53bef31055ed2613d63133/librosa/core/pitch.py#L398-L406

Translating line 401 into mathematical notation, we see that the `energy_frame` $E$ is calculated as

$$\begin{align}
E(k) &= \sum_{n=0}^{W + k}x[n]^2 - \sum_{n=0}^{k}x[n]^2\\
&= \sum_{n=k+1}^{W + k}x[n]^2\\
&= \sum_{n=0}^{W - 1}x[n + k + 1]^2
\end{align}$$

Meaning that we end up with the same off-by-one error as seen before in the ACF correlation, which seem to cancel each other out to a degree.

The calculation needs to be adapted in any case to account for switching to the type II ACF $r'(k)$. We'll define 

$$\begin{align}E(k) &= \sum_{n=k}^{N-1}x[n]^2\\
&= \sum_{n=0}^{N-1}x[n]^2 - \sum_{n=0}^{k-1}x[n]^2\\
&= r'(0) - \sum_{n=0}^{k-1}x[n]^2
\end{align}$$

Following the logic of equation (7) from the YIN paper, the difference function can be calculated as

$$\begin{align}
d(k) &= r'(0) + E(k) - 2r'(k)\\
&= r'(0) + \left(r'(0) - \sum_{n=0}^{k-1}x[n]^2\right) - 2r'(k)\\
&= 2\left(r'(0) - r'(k)\right) - \sum_{n=0}^{k-1}x[n]^2
\end{align}$$

And going back from maths to code:

```python
yin_frames = np.cumsum(np.square(y_frames), axis=-2)

# Difference function: d(k) = 2 * (ACF(0) - ACF(k)) - sum_{m=0}^{k-1} y(m)^2
k = slice(1, max_period + 1)
yin_frames[..., 0, :] = 0
yin_frames[..., k, :] = (
    2 * (acf_frames[..., 0, :] - acf_frames[..., k, :]) - yin_frames[..., :k.stop-1, :]
)
``` 

The remaining steps (cumulative mean normalization, peak picking etc) are correct in the current implementation as far as I can tell.

## Benchmarks

I've run some comparisons using `mir_eval` that hopefully show that all this mathematical pedantry was worth it. The full numbers are contained [this Google sheet](https://docs.google.com/spreadsheets/d/15_eYfSo3drQSCM-cWmeYEokGsPsMtqqpg4E5PhnP5bs/edit?gid=0#gid=0), but I'll summarize the highlights here.

Using `fmin=80`, `fmax=800`, `sr=44100` and leaving all other values at their default, the new implementation improves the raw pitch accuracy of YIN (using a tolerance of 50 cents) on the [vocadito](https://zenodo.org/records/5578807) dataset significantly:

- from 95% to 98% using `frame_length=1024`
- from 92% to 98% using `frame_length=2048`

pYIN results also show clear improvements both for raw pitch accuracy and overall accuracy (taking into account V/UV decisions). 

The drawback is an increase in computational cost. The FFT is $O(N\log(N))$, and we've replaced 3 FFT operations with length $N/2$ with 2 operations of length $N$. That means at $N=2048$, we'd expect to take roughly 1.5x longer, which indeed matches the measurements for YIN. However, due to the more costly Viterbi decoding, this does not seem to have the same effect on pYIN running times.

## Additional changes

- I removed this line in both YIN and pYIN which marks $d'(0)$ as a trough if $d'(0) < d'(1)$: https://github.com/librosa/librosa/blob/d5aa7e1ae443c917fa53bef31055ed2613d63133/librosa/core/pitch.py#L612 This check is unnecessary since $d'(0) = 1$ by definition (equation 8 in the YIN paper), and a lag of 0 should never be selected as the period anyway.
- `__check_yin_parameters`
  - Since `win_length` is unused now, I've added a deprecation warning if it is called with any value other than `None`
  - It previously included this check: https://github.com/librosa/librosa/blob/d5aa7e1ae443c917fa53bef31055ed2613d63133/librosa/core/pitch.py#L967 I believe this is an error and should use `fmin`? As I understand it, the intention is an upper bound on the largest possible period, which corresponds to the lowest frequency, not the highest. I've changed this check to throw an error if the largest period `sr / fmin` exceeds `frame_length`.
  - Additionally there's a "softer" check (warning instead of error) for `sr / fmin >= frame_length // 2`, since ideally the frame should fit two periods of the lowest frequency for accurate f0 detection.

### DIFF
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -762,7 +762,8 @@ def pyin(
     if fmin is None or fmax is None:
         raise ParameterError('both "fmin" and "fmax" must be provided')
 
-    __check_yin_params(
+   if not isinstance(win_length, Deprecated):
+       warnings.warn(FutureWarning, "win_length parameter is deprecated...")
         sr=sr, fmax=fmax, fmin=fmin, frame_length=frame_length, win_length=win_length
     )
 
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -963,11 +963,13 @@ def __check_yin_params(
         warnings.warn(
             f"With fmin={fmin:.3f}, sr={sr} and frame_length={frame_length}, less than two periods of fmin "
             f"fit into the frame, which can cause inaccurate pitch detection. "
-            f"Consider increasing to fmin={fmin_optimal} or frame_length={frame_length_optimal}."
+            f"Consider increasing to fmin={fmin_optimal} or frame_length={frame_length_optimal}.",
+            stacklevel=3
         )
 
     if win_length is not None:
         warnings.warn(
-            f"The win_length parameter has been depecrated in version 0.11.0 "
-            f"and has no effect. It will be removed in version 1.0.0."
+            f"The win_length parameter has been deprecated in version 0.11.0 "
+            f"and has no effect. It will be removed in version 1.0.0.",
+            stacklevel=3
         )
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -877,8 +877,7 @@ def pyin(
     t_switch = sequence.transition_loop(2, 1 - switch_prob)
     transition = np.kron(t_switch, transition)
 
-    p_init = np.zeros(2 * n_pitch_bins)
-    p_init[n_pitch_bins:] = 1 / n_pitch_bins
+    p_init = np.ones(2 * n_pitch_bins) / (2 * n_pitch_bins)
 
     states = sequence.viterbi(observation_probs, transition, p_init=p_init)
 
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -394,7 +394,7 @@ def _cumulative_mean_normalized_difference(
     k = slice(1, max_period + 1)
     yin_frames[..., 0, :] = 0
     yin_frames[..., k, :] = (
-        2 * (acf_frames[..., 0, :] - acf_frames[..., k, :]) - yin_frames[..., :k.stop-1, :]
+        2 * (acf_frames[..., 0:1, :] - acf_frames[..., k, :]) - yin_frames[..., :k.stop-1, :]
     )
 
     # Cumulative mean normalized difference function.
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -384,27 +384,25 @@ def _cumulative_mean_normalized_difference(
     yin_frames : np.ndarray [shape=(max_period-min_period+1,n_frames)]
         Cumulative mean normalized difference function for each frame.
     """
-    acf_frames = audio.autocorrelate(y_frames, max_size=max_period+1, axis=-2)
+    acf_frames = audio.autocorrelate(y_frames, max_size=max_period + 1, axis=-2)
 
     # Energy terms.
     yin_frames = np.cumsum(np.square(y_frames), axis=-2)
 
     # Difference function: d(k) = 2 * (ACF(0) - ACF(k)) - sum_{m=0}^{k-1} y(m)^2
-    yin_frames[..., 1 : max_period + 1, :] = (
-        2 * (acf_frames[..., :1, :] - acf_frames[..., 1 : max_period + 1, :])
-        - yin_frames[..., 1 : max_period + 1, :]
-    )
+    k = slice(1, max_period + 1)
     yin_frames[..., 0, :] = 0
+    yin_frames[..., k, :] = (
+        2 * (acf_frames[..., 0, :] - acf_frames[..., k, :]) - yin_frames[..., :k.stop-1, :]
+    )
 
     # Cumulative mean normalized difference function.
     yin_numerator = yin_frames[..., min_period : max_period + 1, :]
     # broadcast this shape to have leading ones
-    tau_range = util.expand_to(
-        np.arange(1, max_period + 1), ndim=yin_frames.ndim, axes=-2
-    )
+    k_range = util.expand_to(np.r_[k], ndim=yin_frames.ndim, axes=-2)
 
     cumulative_mean = (
-        np.cumsum(yin_frames[..., 1 : max_period + 1, :], axis=-2) / tau_range
+        np.cumsum(yin_frames[..., k, :], axis=-2) / k_range
     )
     yin_denominator = cumulative_mean[..., min_period - 1 : max_period, :]
     yin_frames: np.ndarray = yin_numerator / (
@@ -950,7 +948,7 @@ def __check_yin_params(
 
     if sr / fmin >= frame_length - 1:
         fmin_feasible = sr / (frame_length - 1)
-        frame_length_feasible = int(np.ceil(sr/fmin) + 1)
+        frame_length_feasible = int(np.ceil(sr / fmin) + 1)
         raise ParameterError(
             f"fmin={fmin:.3f} is too small for frame_length={frame_length} and sr={sr}. "
             f"Either increase to fmin={fmin_feasible:.3f} or frame_length={frame_length_feasible}"
@@ -958,18 +956,18 @@ def __check_yin_params(
 
     if sr / fmin >= frame_length // 2:
         fmin_optimal = sr / (frame_length / 2)
-        frame_length_optimal = int(np.ceil(sr/fmin) * 2 + 1)
+        frame_length_optimal = int(np.ceil(sr / fmin) * 2 + 1)
 
         warnings.warn(
             f"With fmin={fmin:.3f}, sr={sr} and frame_length={frame_length}, less than two periods of fmin "
             f"fit into the frame, which can cause inaccurate pitch detection. "
             f"Consider increasing to fmin={fmin_optimal:.3f} or frame_length={frame_length_optimal}.",
-            stacklevel=3
+            stacklevel=3,
         )
 
     if win_length is not None:
         warnings.warn(
             f"The win_length parameter has been deprecated in version 0.11.0 "
             f"and has no effect. It will be removed in version 1.0.0.",
-            stacklevel=3
+            stacklevel=3,
         )
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -963,7 +963,7 @@ def __check_yin_params(
         warnings.warn(
             f"With fmin={fmin:.3f}, sr={sr} and frame_length={frame_length}, less than two periods of fmin "
             f"fit into the frame, which can cause inaccurate pitch detection. "
-            f"Consider increasing to fmin={fmin_optimal} or frame_length={frame_length_optimal}.",
+            f"Consider increasing to fmin={fmin_optimal:.3f} or frame_length={frame_length_optimal}.",
             stacklevel=3
         )
 
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -361,8 +361,6 @@ def piptrack(
 
 def _cumulative_mean_normalized_difference(
     y_frames: np.ndarray,
-    frame_length: int,
-    win_length: int,
     min_period: int,
     max_period: int,
 ) -> np.ndarray:
@@ -376,10 +374,6 @@ def _cumulative_mean_normalized_difference(
     ----------
     y_frames : np.ndarray [shape=(frame_length, n_frames)]
         framed audio time series.
-    frame_length : int > 0 [scalar]
-        length of the frames in samples.
-    win_length : int > 0 [scalar]
-        length of the window for calculating autocorrelation in samples.
     min_period : int > 0 [scalar]
         minimum period.
     max_period : int > 0 [scalar]
@@ -523,9 +517,6 @@ def yin(
         length of the frames in samples.
         By default, ``frame_length=2048`` corresponds to a time scale of about 93 ms at
         a sampling rate of 22050 Hz.
-    win_length : None or int > 0 [scalar]
-        length of the window for calculating autocorrelation in samples.
-        If ``None``, defaults to ``frame_length // 2``
     hop_length : None or int > 0 [scalar]
         number of audio samples between adjacent YIN predictions.
         If ``None``, defaults to ``frame_length // 4``.
@@ -543,6 +534,11 @@ def yin(
         ``y`` is padded on both sides with zeros.
         If ``center=False``,  this argument is ignored.
         .. see also:: `np.pad`
+    win_length : Deprecated
+        length of the window for calculating autocorrelation in samples.
+
+        .. warning:: This parameter is deprecated as of 0.11.0 and
+            will be removed in 1.0.
 
     Returns
     -------
@@ -568,10 +564,6 @@ def yin(
     if fmin is None or fmax is None:
         raise ParameterError('both "fmin" and "fmax" must be provided')
 
-    # Set the default window length if it is not already specified.
-    if win_length is None:
-        win_length = frame_length // 2
-
     __check_yin_params(
         sr=sr, fmax=fmax, fmin=fmin, frame_length=frame_length, win_length=win_length
     )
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks for this!  Dropping a note now to acknowledge that it's on my radar, but I won't have time to dig into it for a week or two.

librosa/core/pitch.py
```diff
@@ -959,15 +946,28 @@ def __check_yin_params(
     if fmin <= 0:
         raise ParameterError(f"fmin={fmin:.3f} must be strictly positive")
 
-    if win_length >= frame_length:
+    if sr / fmin >= frame_length - 1:
```
lostanlen, line 949: does this PR introduce `ParameterError`'s for cases which were did not error (at least for some values of `win_length`) up until today?

dsuedholt: In the current state of this PR, `win_length` is actually completely ignored. A deprecation warning (not an error) will be logged if `yin` or `pyin` is called with `win_length` set to any value other than `None`.

I do think the `win_length` parameter should be discarded entirely going forward, and I think this way it's not breaking anyone's existing code.

However, I think it might introduce `ParameterError`s for some values of `fmin` and `fmax`. I believe the current librosa code erroneously checks `fmax` to determine whether the largest period fits into the frame, which this PR changes to checking `fmin` (lowest frequency ≘ largest period).

bmcfee:
:wave: thanks again @dsuedholt for the thorough investigation and writeup of your changes here.  Also thanks for your patience while I cleared my stack enough to put some thought into this PR!

Everything you've described makes total sense, and indeed clarifies a few points that popped up in the development of this implementation.

I've enabled the CI to run, and it looks like the changes here do introduce several regressions.  Some of these are down to changes in estimation - I suspect the disabled trough check here  - and others are due to broken handling of multi-channel.  I'll try to sleuth the latter part out to recommend a fix.



> * I believe this is an error and should use `fmin`? As I understand it, the intention is an upper bound on the largest possible period, which corresponds to the lowest frequency, not the highest. I've changed this check to throw an error if the largest period `sr / fmin` exceeds `frame_length`.

:eyes: Huh  We had this right in the issue thread where this was all derived: https://github.com/librosa/librosa/issues/1680#issuecomment-1487241064 ; this was then reverted in the PR at commit 6923fe70a96883a9f1c5b9378ce9dc62d4d1fa25 but I can't at all remember why this was a problem.

librosa/core/pitch.py
bmcfee: If `win_length` is deprecated, please use the `util.deprecation.Deprecated` helper class to mark it as such.  This way we can distinguish between an explicit `None` setting (which would be an error going forward) and using the default (which would not be an error in the future).

librosa/core/pitch.py
```diff
@@ -597,19 +584,18 @@ def yin(
...
-        y_frames, frame_length, win_length, min_period, max_period
+        y_frames, min_period, max_period
     )
 
     # Parabolic interpolation.
     parabolic_shifts = _parabolic_interpolation(yin_frames)
 
     # Find local minima.
     is_trough = util.localmin(yin_frames, axis=-2)
-    is_trough[..., 0, :] = yin_frames[..., 0, :] < yin_frames[..., 1, :]
```
bmcfee, line 612: I think this line is still needed; it's here to ensure that the localmin calculation is properly defined at the edge of the array.

librosa/core/pitch.py
```diff
@@ -773,10 +762,6 @@ def pyin(
     if fmin is None or fmax is None:
         raise ParameterError('both "fmin" and "fmax" must be provided')
 
-    # Set the default window length if it is not already specified.
-    if win_length is None:
-        win_length = frame_length // 2
-
     __check_yin_params(
```
bmcfee, line 765:
```suggestion
   if not isinstance(win_length, Deprecated):
       warnings.warn(FutureWarning, "win_length parameter is deprecated...")
```

librosa/core/pitch.py
```diff
@@ -885,7 +870,6 @@ def __pyin_helper(
         # 2. For each frame find the troughs.
         is_trough = util.localmin(yin_frame)
 
-        is_trough[0] = yin_frame[0] < yin_frame[1]
```
bmcfee, line 888: as above, i think this part may still be needed

librosa/core/pitch.py
```diff
@@ -389,31 +384,25 @@ def _cumulative_mean_normalized_difference(
...
-    acf_frames[np.abs(acf_frames) < 1e-6] = 0
+    acf_frames = audio.autocorrelate(y_frames, max_size=max_period + 1, axis=-2)
 
     # Energy terms.
-    energy_frames = np.cumsum(y_frames**2, axis=-2)
-    energy_frames = (
-        energy_frames[..., win_length:, :] - energy_frames[..., :-win_length, :]
-    )
-    energy_frames[np.abs(energy_frames) < 1e-6] = 0
+    yin_frames = np.cumsum(np.square(y_frames), axis=-2)
```
bmcfee, line 390: not for this PR, but cumsum(square) composition seems like a good candidate for numba ufunc optimization.

librosa/core/pitch.py
```diff
@@ -389,31 +384,25 @@ def _cumulative_mean_normalized_difference(
...
     yin_numerator = yin_frames[..., min_period : max_period + 1, :]
     # broadcast this shape to have leading ones
-    tau_range = util.expand_to(
-        np.arange(1, max_period + 1), ndim=yin_frames.ndim, axes=-2
-    )
+    k_range = util.expand_to(np.r_[k], ndim=yin_frames.ndim, axes=-2)
 
     cumulative_mean = (
-        np.cumsum(yin_frames[..., 1 : max_period + 1, :], axis=-2) / tau_range
+        np.cumsum(yin_frames[..., k, :], axis=-2) / k_range
```
bmcfee, line 405: somewhere in this passage, I think, is where the multichannel failures are creeping in.  @dsuedholt could you step this on a multi-channel example and make sure that everything retains the proper shapes compared to the current main branch?

librosa/core/pitch.py
dsuedholt: I suppose it's a question of definition whether rigorously speaking, an edge point should be considered a local minimum or not. However, I don't think it makes a practical difference in this case. The algorithm will only pick lag candidates within the search bounds, and a lag of 0 can't be within those bounds since it would correspond to "infinite frequency". 

The `localmin` call returns a boolean array, where `is_trough[..., 0, :] == False` by definition according to the [`localmin` documentation](https://librosa.org/doc/main/generated/librosa.util.localmin.html) - I think this suits the algorithm well and doesn't require that additional check.

But I don't think the extra check impacts performance or accuracy, so I'd be happy to add it back in.

bmcfee: > and a lag of 0 can't be within those bounds since it would correspond to "infinite frequency".

That's true, but `yin_frames` is already trimmed to the min / max period before returning from CNMD, so "0" here actually means min_period.

dsuedholt: Ah my bad, you're completely right of course! I'll revert this change.

dsuedholt:
Hi @bmcfee, thanks a lot for the helpful review! Multichannel processing should be fixed, tests are passing now.

I couldn't find the code that generated the ground truth data for the chirps that's being loaded from `tests/data/pitch-yin.npy` 🤔 I changed the tests to use the average instantaneous frequency for every frame as ground truth instead.

bmcfee:
> I couldn't find the code that generated the ground truth data for the chirps that's being loaded from `tests/data/pitch-yin.npy`

I believe these outputs came from a prior yin implementation (from the vamp plugin, IIRC) so that we could regression-test for compatibility; see discussion in https://github.com/librosa/librosa/pull/1063 .  I'd prefer to keep that test as it was so that we don't lose track of compatibility.  If we need to relax the test somehow to account for differences introduced in this change, we can do that, but retain documentation (comments) about what changed and why.

I'm happy to have a free-standing unit test against the instantaneous frequency as a separate test though.

Thanks again for all this!

dsuedholt:
Makes sense! I got the previous tests against the vamp plugin working now with a few adjustments. I reduced the framesize in the YIN call to 1024; before it was being called with the default `frame_length=2048`, but because of the previous `win_length` implementation, it was effectively only looking at 1024 samples.

For the `pyin` chirp test, I switched to the exact values used in the vamp plugin for fmin, fmax, and resolution, and it seems to pass the test for me.

I did notice one more thing though: the tests assumed that the first frame of the pyin output will always be unvoiced, because the decoding is initialized with a prior of 0 for all voiced bins and `1 / n_pitch_bins` for all unvoiced bins, instead of a uniform `1 / (2 * n_pitch_bins)` for all bins: https://github.com/librosa/librosa/blob/d5aa7e1ae443c917fa53bef31055ed2613d63133/librosa/core/pitch.py#L853-L854

I'm not sure why this decision was made, but it matches neither the original paper (defines uniform V/UV prior in equation 6) nor the Vamp plugin (initializes a uniform prior [here](https://code.soundsoftware.ac.uk/projects/pyin/repository/entry/MonoPitchHMM.cpp#L91)). 

I changed the pyin implementation to a uniform prior, which made the `test_pyin_multi_center` test pass for me when it previously didn't (I'm on a mac, I'm not sure if the mentioned OSX issue still persists).

dsuedholt:
I added tests for the warnings (should fix reduced code coverage) and removed a `numpy>2.0` only feature from one of the new chirp tests (should fix the linter failing).

I'm not sure what to do about the doc workflow failing, I can't get the error to happen on my local machine.

tests/test_core.py
```diff
@@ -1171,43 +1171,74 @@ def test_yin_tone(freq):
...
     y = librosa.tone(freq, duration=1.0)
     f0, _, _ = librosa.pyin(y, fmin=110, fmax=1000, center=False)
-    # Skip the first frame, since the voicing prior does not allow it
-    assert np.allclose(np.log2(f0[1:]), np.log2(freq), rtol=0, atol=1e-2)
+    assert np.allclose(np.log2(f0), np.log2(freq), rtol=0, atol=1e-2)
 
 
 def test_pyin_multi():
-    y = np.stack([librosa.tone(440, duration=1.0), librosa.tone(560, duration=1.0)])
+    y = np.stack([librosa.tone(440, duration=0.1), librosa.tone(560, duration=0.1)])
```
bmcfee, line 1241: I'm curious why this duration was shortened so much in this test?

bmcfee:
> I added tests for the warnings (should fix reduced code coverage) and removed a `numpy>2.0` only feature from one of the new chirp tests (should fix the linter failing).
> 

Thanks for this - it's all looking great!  I had one minor q about the revised tests, but I'm otherwise happy with where things are at.

> I'm not sure what to do about the doc workflow failing, I can't get the error to happen on my local machine.

The doc CI was failing due to a broken link.  This is almost surely unrelated to the PR, and it happens sporadically when sites we link to are temporarily unreachable.  As long as the tests that execute before link check are passing, I won't worry about it.

librosa/core/pitch.py
bmcfee: Actually, we can cut memory consumption in half here with an in-place cumsum.  No further optimization needed.
```suggestion
    yin_frames = np.square(y_frames)
    np.cumsum(yin_frames, out=yin_frames, axis=-2)
```

bmcfee:
> I'm not sure why this decision was made, but it matches neither the original paper (defines uniform V/UV prior in equation 6) nor the Vamp plugin (initializes a uniform prior [here](https://code.soundsoftware.ac.uk/projects/pyin/repository/entry/MonoPitchHMM.cpp#L91)).

That's also a good point.  It must have escaped our attention in the initial CR of #1063 because the test signals generally started on unvoiced frames.  I couldn't find any discussion of it in the comment thread anyway.

tests/test_core.py
dsuedholt: Oops, sorry. That's a debugging remnant that shouldn't have been committed, I'll change that back.

librosa/core/pitch.py
dsuedholt: Thanks! I'll add in this change.

bmcfee (approved):
LGTM

bmcfee:
I think we're all set with this.  I gave it a spin locally and it's all behaving nicely.

We do take a bit of a performance hit here: speed was about the same, but memory usage is up by about 10%.  I expect this comes from some inefficiencies in either `autocorrelate` or `abs2` (or both), relative to how it was previously implemented.  I can live with this for now - we can always revisit internal optimizations in a later point release.

Unless there are objections, I think this is good to merge.  Thanks again @dsuedholt for the deep dive and scrutiny on this!

PR #1897, opened 2025-01-09
declare numba requirement before numpy

### LINKED ISSUE #1896: [Bug]: Support python3.11
### Bug summary

When trying to install librosa macos m1 python3.11:

```console
RuntimeError: Cannot install on Python version 3.13.0; only versions >=3.6,<3.10 are supported.
```

I only need 

https://github.com/thewh1teagle/kokoro-onnx/blob/e6a9c7dd9a9ee2e11b7a3c681a3ea344fe458c32/src/kokoro_onnx/__init__.py#L145

This function for this. do I really need so many dependencies for that?

### Code for reproduction

```Python
pip install kokoro-onnx==0.2.3

https://github.com/thewh1teagle/kokoro-onnx
```

### Actual outcome

Error

### Expected outcome

It should install in python3.11/12

### Additional information

_No response_

### Operating system

MacOS

### software versions

Python3.11/12 UV

### Installation

None

dsuedholt:
Fixes #1896

As mentioned in the issue, the following will currently fail:

```
> uv venv -p 3.12
> uv pip install librosa

error: Failed to prepare distributions
  Caused by: Failed to download and build `llvmlite==0.36.0`
  Caused by: Build backend failed to determine requirements with `build_wheel()` (exit status: 1)

[...]
```

I've run into this issue before and used the workaround of installing `numba` explicilty before librosa, which seems to fix the issue for some reason:

```
> uv venv -p 3.12
> uv pip install numba
> uv pip install librosa
# works
```

I have no idea whether the actual issue lies with `numba`, `llvmlite`, or `uv` itself. But `uv` is becoming very popular very quickly and it would be easy for `librosa` could incorporate this workaround by simply declaring `numba` as a dependency before `numpy`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,12 +63,12 @@ packages = find:
 include_package_data = True
 install_requires =
     audioread >= 2.1.9
+    numba >= 0.51.0
     numpy >= 1.20.3, != 1.22.0, != 1.22.1, != 1.22.2
     scipy >= 1.2.0
     scikit-learn >= 0.20.0
     joblib >= 0.14
     decorator >= 4.3.0
-    numba >= 0.51.0
     soundfile >= 0.12.1
     pooch >= 1.1
     soxr >= 0.3.2


bmcfee:
Huh, what a strange fix.  To my understanding, the order of version specifications should not matter at all, and the extent to which it does seems like a bug in uv.  But it also seems like a benign change otherwise, so if we can verify that this is somehow helpful, I don't mind accepting the change.  Thanks for bringing it up!

bmcfee:
Also worth noting that the CI here will not test this, since we (mostly) set up test environments by conda with a matching spec.

dsuedholt:
It's indeed a pretty strange issue. On Mac, the error complains about failing to build `llvmlite==0.36.1` as a dependency of `numba`, but on Linux the error is directly with `numba==0.53.1`. This is consistent across python version (I checked 3.10, 3.11, 3.12 on both platforms).

It can be confirmed that shuffling around the dependencies indeed fixes the error by installing from the commit directly:

```
uv venv -p 3.12
uv pip install "git+https://github.com/dsuedholt/librosa@patch-1"
# works
```

dsuedholt:
Ah, this is definitely caused by a bug within `uv`, but one that has been fixed already. It only seems to occur in `uv 0.5.9` and older. But `uv 0.5.10` was only released three weeks ago, so it might still be worth including this workaround just in case it saves someone a headache somewhere.

PR #1907, opened 2025-02-04
0.11.0rc0 release prep

### LINKED ISSUE #1867: [Bug]: librosa.beat.beat_track always returns `tempo` as a 1D `numpy.ndarray` and not a regular `float`
### Bug summary

According to docs [here](https://librosa.org/doc/latest/generated/librosa.beat.beat_track.html#librosa-beat-beat-track) there was no change in `librosa.beat.beat_track` on 0.10.2. It should still return a tuple like `Tuple[Union[_FloatLike_co, np.ndarray], np.ndarray]` with first element representing `tempo` and second element being `beats`.

Docs says:

> tempo: float [scalar, non-negative] or np.ndarray
>estimated global tempo (in beats per minute)
>
>If multi-channel and bpm is not provided, a separate tempo will be returned for each channel 

However, I have bumped `librosa` version I use in a script and came across the fact that the tempo is now always a 1D `numpy.ndarray` even when using arguments `bpm=None` and `sparse=True` (default values). I don't know if this is intended, but after bumping `librosa` version data serialization to json of bpm estimates started breaking due to the fact that predicted tempo is not a `float` but a `numpy.ndarray` now.

### Code for reproduction

```Python
❯  pip install librosa
❯  python
                                                                                                                            ─
Python 3.10.14 (main, Mar 19 2024, 21:46:16) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import librosa
>>> y, sr = librosa.load(librosa.ex('choice'), duration=10)
>>> tempo, beats = librosa.beat.beat_track(y=y, sr=sr)
>>> type(tempo)
<class 'numpy.ndarray'>
>>> tempo
array([135.99917763])
>>> beats
array([  3,  21,  40,  59,  78,  96, 116, 135, 154, 173, 192, 211, 230,
       249, 268, 287, 306, 325, 344, 363, 381, 397, 413])
__

❯ pip show librosa                                                                                                                    ─
Name: librosa
Version: 0.10.2.post1
Summary: Python module for audio and music processing
Home-page: https://librosa.org
Author: Brian McFee, librosa development team
Author-email: brian.mcfee@nyu.edu
License: ISC
Location: /Users/carlos.lordelo/.local/share/virtualenvs/publish-9aeiLLBq/lib/python3.10/site-packages
Requires: audioread, decorator, joblib, lazy-loader, msgpack, numba, numpy, pooch, scikit-learn, scipy, soundfile, soxr, typing-extensions
Required-by:
```

### Actual outcome

`tempo` or `type(librosa.beat.beat_track(y=y, sr=sr)[0])` is `numpy.ndarray`

### Expected outcome

According to docs [here](https://librosa.org/doc/latest/generated/librosa.beat.beat_track.html#librosa-beat-beat-track), `tempo` or `type(librosa.beat.beat_track(y=y, sr=sr)[0])` should be a regular `float` and not a `numpy.ndarray`

### Additional information

_No response_

### Operating system

MacOS

### software versions

macOS-14.4.1-arm64-arm-64bit
Python 3.10.14 (main, Mar 19 2024, 21:46:16) [Clang 15.0.0 (clang-1500.3.9.4)]
NumPy 1.26.4
SciPy 1.14.0
librosa 0.10.2.post1

### Installation

pip

### LINKED ISSUE #1906: Preparing for 0.11 release
I'm starting a branch to get the 0.11 release ready once the rest of our blockers (#1883 ) clear up.

In doing so, I think we have some space to update our minimum dependencies and clear the stack a little.  Here are the changes I have in mind:

- Drop python 3.7 support (long past EOL), remove from build matrix
- Add python 3.13 support
- Bump minimum scipy to 1.5, allowing us to clear a dead branch https://github.com/librosa/librosa/blob/ebd878fabb87c588f98f372a94fca5d167c59cf2/librosa/core/audio.py#L924-L925

Some notes:

- Technically we'd only need scipy 1.4, but 1.4 only supported up to python 3.7 which is outside of our support range.  If we min at python 3.8, then scipy 1.5 is the oldest supported version.
- Along similar lines, we would be justified in moving sklearn to 1.1 or later for 3.8 support.
- We could also drop python 3.8 support, though it's only recently past EOL.  The main reason to do this would be to clean up our resource manager https://github.com/librosa/librosa/blob/ebd878fabb87c588f98f372a94fca5d167c59cf2/librosa/util/files.py#L39-L47 .  I don't have strong feelings about this.
- We could bump the minimum numpy to 1.22.3.  There's no functional need for this really, as everything we use is in 1.20.3 and later.  But bumping the min here would let us clean up our dependency spec.

---
Other todos:

- Deprecation expirations
- [ ] `util.valid_audio` mono parameter.  Should have expired in 0.10 but we missed it when cutting that release.

bmcfee:
#### Reference Issue
Fixes #1906 
Fixes #1867 

This PR gets things ready for the 0.11 release.  Changes here include:

- [x] modernize dependency spec
- [x] update environments
- [x] expire deprecations
- [x] update type annotations
- [x] update documentation

#### Any other comments?

This is currently still blocked by audioread, so python 3.13 environments will not yet work.  We can use the CI on this PR to greenlight it when ready though.


bmcfee:
Note: python 3.13 builds are currently failing because of upstream issues in the numba build.  0.61 should be shipping on py3.13 but is not yet.

bmcfee:
Numba packaging is unstuck, but python 3.13 CI still fails due to audioread dependency chain regressions.  I'm okay with this for now - we should merge and cut the pre-release with a note about audioread support, and then when audioread drops we can make the final release.

PR #1913, opened 2025-02-11
PR for 0.11.0 release

bmcfee:
This PR will include all changes for the final 0.11.0 release:

- Version bump
- Changelog date
- AUTHORS.md and zenodo.json updates

#### Any other comments?

At this point I think we're only blocked by audioread.


lostanlen (approved):

bmcfee:
Alright, it seems that audioread is lagging on merging a fix for the deprecated legacy codecs.

I think we can force the issue by implementing my fix https://github.com/beetbox/audioread/pull/148 to the librosa dependency spec.  It's not ideal, but since audioread is a dependency anyway, this patch would basically just be a shortcut in the dependency tree (only on python 3.13) and wouldn't actually amount to meaningful changes in anyone's environments.

I'll roll this in soon and cut a new release candidate.

bmcfee:
New snag encountered: python-samplerate does not ship wheels for 3.13, and does not compile on windows (at least in our CI environment).  This is an optional dependency and not really a blocker, but it would be nice to get CI turning green before we drop.  I've raised an issue over on python-samplerate, but it might also be worth seeing if it's just a compiler issue on our side (builds in ubuntu and osx work fine).

PR #1916, opened 2025-02-19
Deprecate set_fftlib, update fftpack uses

### LINKED ISSUE #1915: [Bug]: Some functions don't respect get_fftlib(), and we generally shouldn't use fftpack
### Bug summary

Librosa allows users to set custom FFT implementations (assuming they adhere to the numpy API) via the `set_fftlib` function.  All uses of fft functionality are intended to run through `get_fftlib` rather than calling, say, `np.fft.<whatever>` directly.

### Code for reproduction

Doing a quick audit, I've found one case that fails to do this:

https://github.com/librosa/librosa/blob/3f88620f36d6726c3b656e8264828799ea7fa8fe/librosa/core/convert.py#L1607

### Actual outcome

always uses np.fft

### Expected outcome

should use `get_fftlib()`

### Additional information

Relatedly, we have a few spots that rely on the legacy `scipy.fftpack` interface for DCT when we should be using the `scipy.fft` implementation.  It does not seem that alternative scipy-compatible fft implementations include DCT though, so I don't think we can run DCT through the `get_fftlib` wrapper.

This actually raises a broader question about whether we should deprecate this get/set fftlib functionality.  I think the combination of array-api and scipy's functionality to set the FFT backend directly make it rather redundant and of limited use anyway.  I think I may be in favor of deprecating it in 0.11 and removing it entirely in 1.0 so we have a clear path forward.  A quick github search turns up only 99 uses of `set_fftlib()`, and a large number of them appear to be either documentation examples or copypasta.

If we deprecate it, we can issue a warning with directions to set backend via scipy instead, and then always go through the `scipy.fft` interface.

### Operating system

all

### software versions

0.11.0rc0

### Installation

git checkout

bmcfee:
#### Reference Issue
Fixes #1915 

This PR deprecates the `set_fftlib` functionality, to be removed in 1.0.

Users are instead instructed to use scipy's fft backend selector to customize FFT computations.

#### Any other comments?

The default fftlib is now scipy instead of numpy.  Both implementations now use pocketfft under the hood, so there should be no observable difference to end users.  Switching to scipy will make the transition to a fully scipy-oriented backend selection smoother though.

I've also migrated all uses of the legacy `scipy.fftpack` to now use `scipy.fft`.  This also should have no observable impact on users, but again allows us to fully swap out backends through scipy's selection (which was independent of fftpack previously).

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,15 +7,16 @@ v0.11.0
 2025-02-XX
 
 Maintenance
-
     - `#1831`_ Numpy 2.0 is now supported.
     - `#1898`_ Updated codecov action. *Brian McFee*
     - `#1860`_ Updated GitHub issue templates. *Brian McFee*
 
 Enhancements
+    - `#1916`_ The default FFT backend has been changed from `numpy` to `scipy`.
     - `#1864`_ Accelerated `abs2` calculation on real-typed inputs. *Brian McFee*
 
 Bug fixes
+    - `#1916`_ Utility decorators now preserve type annotations properly. *Brian McFee*
     - `#1900`_ Fixed an error in pooch registry definition. *Brian McFee*
     - `#1897`_ Accommodate dependency resolution issues with `uv`. *David Südholt*
     - `#1878`_ Improved accuracy for `yin` and `pyin`. *David Südholt*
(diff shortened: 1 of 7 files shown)


bmcfee:
Linter is failing in an interesting way.  Adding a `@deprecated` decorator on `set_fftlib` seems to be messing with the type annotations.  When the decorator on, mypy seems to think that
```python
        librosa.set_fftlib("foo")  # type: ignore
```
does not need the tag to ignore type, though it certainly does.  Weirdly, the type annotations are still present in the mypy data jsons, so maybe there's just a bug in mypy?

PR #1919, opened 2025-03-07
Get samplerate installing from source on windows for python 3.13

bmcfee:

CI has been failing on windows after we bumped it to the python 3.13 environment.  The issue here is that the samplerate package does not ship wheels for 3.13, and installing from source does not work with the default environment config.

This PR tries to fix the windows environment, starting from copilot suggestions on how to get a working c++ compiler (and not 3 or 4 different ones at the same time, which seems to be what is going on currently).

#### Any other comments?

No idea if this will work.  But since the python-samplerate folks are not responding to my request for a new wheel for 3.13 win, I guess it's up to us to try making source builds work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,8 +84,8 @@ jobs:
           if: runner.os == 'Windows'
           shell: pwsh
           run: |
-            choco install visualstudio2022buildtools --version=17.2.4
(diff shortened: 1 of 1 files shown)


bmcfee:
I literally cannot believe how dumb the fix for this problem was.  https://github.com/tuxu/python-samplerate/issues/27#issuecomment-2711724134

Issue #1912, opened 2025-02-10
[ENH]: Simplify and reduce testing


bmcfee:
### Problem

Unit tests are great.  We should test all the things!  But sometimes we go a little overboard, and we're long overdue for some house cleaning.  I'd like this to be done in preparation for the 1.0 release.

I see two major avenues for improvement to our test suite:

1. Reducing or eliminating regression tests, and
2. Reducing parameter grids for test functions.

---

## Regression tests

When this project started, we were an upstart, and there was a strong incentive to provide some evidence that our outputs matched things that people were used to (eg auditory toolbox).  In 2025, this isn't really the case anymore.  Rather than appealing to authority ("mfcc.m gave this output, so we do too!") we should strive for a more concrete and self-contained notion of correctness wherever possible.

The benefits here would be two-fold:

1. Clearer and simpler unit tests
2. If we can eliminate regression tests completely, then we can remove the test-data submodule and make downstream packaging a little easier for everyone involved.  (See #1843 for discussion on this point.)

For the sake of posterity, it would be good to archive these regression tests somewhere, but I think at this point we don't need them running in every CI instance.
If we can't eliminate regression tests entirely, I think we may want to regenerate them with newer recordings that have a clearer license attached to them.  But let's cross that bridge when we get to it; my preference first and foremost is to remove regression tests.

---

## Parameter grids

About ten years ago, I went on a test-coding spree, and generated parameter sweeps that cover essentially all interesting combinations of every function, whether or not the parameters in question interacted meaningfully.  For example:

https://github.com/librosa/librosa/blob/ebd878fabb87c588f98f372a94fca5d167c59cf2/tests/test_constantq.py#L118-L129

Over time I've tried to reduce as many of these as I could to have no more than two values per parameter, but it's still a shotgun approach to testing that I think we could improve upon.  If we can minimize the parameter sweeps here, I think we should be able to shave several minutes off of each test run, while getting more precise and specific information per test by focusing on meaningful parameter interactions.

---

## Implementation plan

- I'm going to hold off on this until after the 0.11 release.
- For step 1 above, it's pretty easy to identify all matlab regression tests by `grep -n "\.mat\"" *.py` in the tests directory.  At a glance, I see the following:
    - test_onset_strength
    - deprecated_test_beat (already skipped)
    - test_load
    - test_stft
    - test__reassign_frequencies_regress (arguably redundant, holdover from our expired ifgram implementation)
    - test_lpc_regress
    - test_cens
[...]

PR #1925, opened 2025-03-25
Test overhaul

bmcfee:
#### Reference Issue
implements #1912 

This PR implements a variety of optimizations on the test suite:

- [x] Reduced parameter grids
- [x] More specific tests when possible
- [x] Eliminating unnecessary fixtures and regressions
- [x] Eliminate cross-module imports for helper functions
- [x] Use proper fixtures for RNG state management
- [x] Push build artifacts for failing display tests
- [x] Remove test data submodule

#### Any other comments?

The goal here is both to make the CI action faster, and reduce dependencies on legacy data and regression fixtures in favor of self-contained correctness tests.

In general, coverage and correctness should not be affected.


bmcfee:
As a reference point for this PR, the CI on the 0.11.0 tag for python 3.13, ubuntu-latest ran as follows:
```
========== 14019 passed, 2 skipped, 523 xfailed in 427.64s (0:07:07) ===========
```
*EDIT*: tests are currently running in about 4min.  This should come down a bit more after we purge the regression fixtures.

bmcfee:
Notes to self on eliminating old audio data from tests.  Here are the use-cases:

- [x] `test_util`, `find_files`, looking for audio files in a target folder
- [x] `test_segment`: needs non-trivial self-similarity structure
- [x] `test_multichannel`: needs a stereo file, probably should have some harmonic structure but content doesn't matter much otherwise
- [x] `test_met_features` unused; everything is in the .mat files
- [x] `test_feature`, only uses test1_22050, content is not important.  Could be a sine sweep or a pulse train.  Probably makes sense to make one of each and use them separately for different tests.
- [x] `test_effects`: should have some real audio, but specifics don't matter
- [x] `test_filters`: 6 uses: hz_to_mel / mel_to_hz, hz_to_octs (all trivial). melfb, melfbnorm, chromafb (not trivial)
- [x] `test_display`: will need lots of attention
- [x] `test_decompose` uses both mono and stereo examples, content should be harmonic and percussive
- `test_core`: lots of uses.  
  - [x] Loading and streaming, get_duration, get_samplerate, etc will need a stored audio file.
  - [x] salience will need something harmonic, but this can be generated
  - [x] iirt has a regression fixture.  Maybe we don't need that.
  - [x] everything else should be doable with a synthetic signal

bmcfee:
I've generated a new recording to be used in tests (`test_audio.ogg`), and updated tests accordingly.

It generally seems to work, except there are now some image comparison failures on osx/windows/arm platforms that I'm not currently able to reproduce.  Near as I can tell, it's not a freetype or matplotlib version issue, since the same tests pass on the ubuntu platform.

bmcfee:
I've added a step to the CI workflow that uploads failed image comparison tests as a build artifact, so we actually have some kind of a chance of debugging these things now!

As it turns out, the failing tests are due to rounding error in binarization.  Eg:

Baseline:
baseline
Result:
result
Diff:
result-failed-diff

This should be easy enough to fix; I'll just remove the binarized subplots.  They aren't necessary for this test anyway.

bmcfee:
Ah I thought this was almost done, but I guess not!

Saving the error log here to pick up later:

```python
FAILED tests/test_core.py::test_get_duration_filename - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/test2_8000.wav'
FAILED tests/test_core.py::test_yin_chirp - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/pitch-yin.npy'
FAILED tests/test_core.py::test_pyin_chirp - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/pitch-pyin.npy'
FAILED tests/test_core.py::test_load_force_audioread - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/test2_8000.mkv'
FAILED tests/test_core.py::test_get_duration_audioread - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/test2_8000.mkv'
FAILED tests/test_core.py::test_get_samplerate_audioread - FileNotFoundError: [Errno 2] No such file or directory: 'tests/data/test2_8000.mkv'
FAILED tests/test_filters.py::test_semitone_filterbank - AttributeError: module 'scipy.signal' has no attribute 'freqz_sos'
FAILED tests/test_filters.py::test_semitone_filterbank_defaults[-1] - AttributeError: module 'scipy.signal' has no attribute 'freqz_sos'
FAILED tests/test_filters.py::test_semitone_filterbank_defaults[0] - AttributeError: module 'scipy.signal' has no attribute 'freqz_sos'
FAILED tests/test_filters.py::test_semitone_filterbank_defaults[1] - AttributeError: module 'scipy.signal' has no attribute 'freqz_sos'
```

bmcfee:
I think this is basically good to go.  The only coverage drop is for the audioread codepaths which have not yet been removed.

A couple of final quirks in the revised test suite:

- I did end up retaining a couple of regression tests, namely the yin and pyin fixtures.  The data here is minimal (weighing under 1KB total) and it seems worthwhile given the complexity of the functions involved.  The data has been migrated to the main repository.
- The `iirt` tests are extremely minimal now, and don't really do anything to check the contents of the output.  I couldn't think of a simple way to design a test for this one, but it also didn't seem worth retaining the regression fixtures.  Maybe @stefan-balke has thoughts?
- Similarly the `melfb` tests are a bit weaker now.  They essentially check for peak frequency response per filter and a few qualitative checks, but otherwise do not verify eg triangular shape etc.  I think this is probably ok, and worth the overall reduction in fixtures.  I did leave a TODO comment in there in case anyone wants to come in and strengthen it later though.  Maybe @lostanlen has thoughts on this?

bmcfee:
I'm going to go ahead and merge this on the 1.0 branch.  Nothing is affecting main yet, so there is still opportunity to comment, but merging this will unstick a few other PRs in progress.

stefan-balke:
I remember that we once checked the filter coefficients against the old Matlab implementation with @fzalkow.
Can we make any assumptions about the input. I mean, we could send test signals at the center frequencies through each filter?

bmcfee:
> Can we make any assumptions about the input. I mean, we could send test signals at the center frequencies through each filter?

I guess we could, but really what we'd want here is to check flat response within the passband, not just the center.  I thought about coding this up, but it seemed somewhat overkill to really do it right, and borders on a full regression test anyway.  At the end of the day, I think we can trust scipy's filter constructors and such well enough here, and it's more important to ensure that the various code paths that hook into it are well tested, rather than doing a full integration test.

stefan-balke:
>  and it's more important to ensure that the various code paths that hook into it are well tested, rather than doing a full integration test.

By that you think of rigorous parameter tests that the code is not breaking?

bmcfee:
> By that you think of rigorous parameter tests that the code is not breaking?

Not sure I parsed this correctly.. can you clarify what you mean?

At this point, the IIRT-related tests are
https://github.com/librosa/librosa/blob/4b6773295484619fb6d5d1c55ca66068fc4f897e/tests/test_core.py#L1891-L1985
and
https://github.com/librosa/librosa/blob/4b6773295484619fb6d5d1c55ca66068fc4f897e/tests/test_filters.py#L578-L630

These are reduced in some ways from the previous regression test (ie we don't do any numerical equivalence against the matlab output), and improved in other ways (explicit testing of the filterbank).

PR #1931, opened 2025-04-17
Variable-rate and non-iterative phase vocoding

### LINKED ISSUE #517: Variable rate phase vocoding

### LINKED ISSUE #1700: time_stretch does not return the same signal when rate=1

### LINKED ISSUE #1930: [Optimization]: Possibly redundant phase wrapping operations in phase_vocoder implementation
I think I've discovered that there are mathematically redundant operations in core.spectrum.phase_vocoder that can be simplified. After analyzing the algebra and testing empirically, I found that three operations effectively cancel each other out:

1. Subtracting phi_advance from the phase difference
2. Performing the modulo 2π operation (phase wrapping)
3. Adding phi_advance back to the wrapped result

The simplified version produces numerically identical results to within floating-point precision. With 1 second of uniform random noise at 44.1kHz sample rate, the sum of squared differences between outputs was approximately 5.7e-13.

Current implementation:
```python
# Compute phase advance
dphase = np.angle(columns[..., 1]) - np.angle(columns[..., 0]) - phi_advance

# Wrap to -pi:pi range
dphase = dphase - 2.0 * np.pi * np.round(dphase / (2.0 * np.pi))

# Accumulate phase
phase_acc += phi_advance + dphase
```

Simplified equivalent implementation:

```python
# Compute phase advance
dphase = np.angle(columns[..., 1]) - np.angle(columns[..., 0])

# Accumulate phase
phase_acc += dphase
```

This simplification is mathematically valid because the phasor function only cares about phase modulo 2π when converting to complex numbers.

If there's a specific reason to keep phase_acc (rather than dphase) within the [-π, π) range for numerical stability (though technically the current code keeps it in [-π, π] range as `np.range` implements banker's rounding), that could be done by applying the current "Wrap to -pi:pi range" operation to phase_acc instead:

```python
# Compute phase advance
dphase = np.angle(columns[..., 1]) - np.angle(columns[..., 0])

# Accumulate phase
phase_acc += dphase

# Wrap to -pi:pi range
phase_acc = phase_acc - 2.0 * np.pi * np.round(phase_acc / (2.0 * np.pi))
```

### LINKED ISSUE #1937: [ENH]: Warping path to time grid
### Problem

PR #1931 adds support for variable-rate time stretching (phase vocoding), where the user can supply a vector of (fractional) frame indices (`t_out`) instead of just a rate.  A natural use-case of this would be synchronization via DTW.

The catch here is that DTW outputs a 2D array of frame indices for the warping path.  `wp[i]` is a tuple `(n, m)` meaning that the `i`th step of the path matches `x1[n]` with `x2[m]`.  This isn't directly usable in the phase vocoder, and converting between the two representations is a bit tedious.

### Proposed solution

We should build a small utility function that translates a warping path to a vector of fractional frame indices.  In my mind, the way this should work is basically as follows:

```python
cost, path = librosa.sequence.dtw(X1, X2)  # align X1 and X2
t_out = librosa.sequence.path_to_times(path)  # to be implemented
X1_match = librosa.phase_vocoder(X1, t_out=t_out)  # X1_match is now synchronized with X2
```

There are two details to be worked out here: 1) how to handle subsequence mode, and 2) what to do with repeated indices in the path.

My initial guess at this would be that we can just build a linear interpolator on `wp[:, 1] → wp[:, 0]` and evaluate it at `arange(wp[:, 1].min(), wp[:, 1].max() + 1)`.

---

Once this is implemented, we should expand the synchronization example in the gallery to include audio playback and timestretching.

bmcfee:
#### Reference Issue
Fixes #517 
Fixes #1700
Fixes #1930 

This PR is a second take at #1796 , implementing the following two changes:

1. The code is non-iterative, relying entirely on vector ops and scipy interpolators.  Results are substantially faster for slow-down (rate<1), and comparable for speedup (rate>1).
2. Variable-rate is finally supported through the `t_out` parameter.

#### Any other comments?

The `n_fft` and `hop_length` parameters are now deprecated, as they are not actually necessary to ensure correct operation.

We still need some unit tests to ensure coverage.  I'm blocking this for now until #1925 is complete so as to avoid redundant rewrites.

Meanwhile, it would be good to get an extra set of eyeballs on it, as I think the functionality is complete.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1440,7 +1440,7 @@ def phase_vocoder(
 
     Returns
     -------
-    D_stretched : np.ndarray [shape=(..., d, t / rate), dtype=complex]
+    D_stretched : np.ndarray [shape=(..., n_bins, t / rate), dtype=complex]
         time-stretched STFT.
         For variable-rate stretching, the output will have `len(t_out)` frames.
 
@@ -1468,6 +1468,9 @@ def phase_vocoder(
 
     if (rate is None) == (t_out is None):
         raise ParameterError("Must specify exactly one of `rate` or `t_out`")
+
+    if (rate is not None) and (rate <= 0):
+        raise ParameterError(f"rate={rate} must be a positive number")
 
     if t_out is None:
         t_out = np.arange(0.0, n_frames, rate)
@@ -1509,8 +1512,8 @@ def phase_vocoder(
         copy=False,
     )
 
-    D_out: np.ndarray = util.phasor(phase, mag=mag_interp(t_out))
-    return D_out
+    D_stretched: np.ndarray = util.phasor(phase, mag=mag_interp(t_out))
+    return D_stretched
 
 
 @cache(level=20)
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1418,8 +1418,8 @@ def phase_vocoder(
         If given, `rate` must be None.
 
     kind : str
-        Interpolation kind for magnitude interpolation.
-        Passed to `scipy.interpolate.interp1d`.
+        Interpolation kind for magnitude interpolation, e.g., 'linear' or 'nearest'.
+        See `scipy.interpolate.interp1d` for a full list of supported options.
 
     hop_length : int > 0 [scalar] or None
         The number of samples between successive columns of ``D``.
@@ -1450,8 +1450,24 @@ def phase_vocoder(
     """
     n_frames = D.shape[-1]
 
+    if not isinstance(hop_length, Deprecated):
+        warnings.warn(
+            "The `hop_length` parameter is deprecated as of 1.0 and will be removed in 1.1. "
+            "It is unused in the current implementation.",
+            FutureWarning,
+            stacklevel=2,
+        )
+
+    if not isinstance(n_fft, Deprecated):
+        warnings.warn(
+            "The `n_fft` parameter is deprecated as of 1.0 and will be removed in 1.1. "
+            "It is unused in the current implementation.",
+            FutureWarning,
+            stacklevel=2,
+        )
+
     if (rate is None) == (t_out is None):
-        raise ValueError("Must specify exactly one of `rate` or `t_out`")
+        raise ParameterError("Must specify exactly one of `rate` or `t_out`")
 
     if t_out is None:
         t_out = np.arange(0.0, n_frames, rate)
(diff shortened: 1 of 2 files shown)


bmcfee:
I think the core functionality of this PR is :shipit:-ready.  

From the bot's CR, there was a recommendation to include a docstring example showing how to use the `t_out` input, which is entirely reasonable.  In thinking it through, I noticed that there isn't an easy and obvious way to construct `t_out` for a self-contained example.

This got me thinking that the best thing to do would be to extend the DTW advanced example to include a variable-rate stretch based on the warping path.  This in turn suggests that a utility for converting between warping paths and time grids would be helpful / necessary for this to be useful.  I think that might make the most sense as a separate PR though, in which case I think this can be merged as is.

PR #1932, opened 2025-05-19
fix: correct typo in core/convert.py line:1658

afadel151:
fix: correct typo in core/convert.py line:1658 ( psychoaoustical -> psychoacoustical )

### DIFF
--- a/librosa/core/convert.py
+++ b/librosa/core/convert.py
@@ -1655,7 +1655,7 @@ def mel_frequencies(
     appear equal in width over the full hearing range.
 
     Because the definition of the mel scale is conditioned by a finite number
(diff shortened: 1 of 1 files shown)


bmcfee:
Thanks for catching this!  It's strange that the spell checker didn't flag it earlier though...

bmcfee:
Ok, it turns out that codespell didn't catch this because it's not supposed to.  Nothing else to change here, so I'm happy to merge.  Thanks again!

afadel151:
You’re welcome! I’m glad I could help. Thanks for reviewing and merging the
PR — looking forward to contributing more in the future!

PR #1936, opened 2025-06-02
Trying a new environment cache setup in CI

bmcfee:
This PR updates the CI workflow to cache the conda environments instead of the downloaded packages.

With any luck, this should bring down our CI time substantially for repeated runs in short succession.


bmcfee:
Looks like we're in business!  The environment caches only last for a day, and are generally pretty large (~500MB each), but on days when we have a lot of action this shaves ~2 minutes off each build.

Issue #1902, opened 2025-01-14
[ENH]: Sparse array migration


bmcfee:
### Problem

Scipy 1.15 introduced a [migration plan](https://docs.scipy.org/doc/scipy-1.15.0/reference/sparse.migration_to_sparray.html#migration-to-sparray) for going from sparse matrix to sparse array data structures.  Heading toward the 1.0 release, we should start the transition to sparse arrays.

### Proposed solution

Sparse matrices occur in relatively few spots for us:
- `util.sparsify_rows` (used in CQT and related)
- `util.shear / __shear_sparse` (used in diagonal filtering, recurrence/lag conversion)
- `decompose.nn_filter` (for neighborhoods)
- `segment.cross_similarity` and `recurrence_matrix`

Doing this migration would require a version bump on our minimum supported scipy.  But it does seem like a long-term good idea since we basically never use matrix features here, and it would open the door to better multi-channel support down the road.  It will likely have some bearing on #1743 as well.

bmcfee:
Note: version bump here would at minimum be to 1.8, which is fine for us according to SPEC0 #1935 .  Probably we can get away with a higher minimum version (1.13) when it comes time for the 1.0 release here.

Daniel-Haas-B:
If no one has started this I'd gladly contribute. Looks straightforward :)

bmcfee:
Sure, thanks @Daniel-Haas-B !  Timing is good for this - I was holding off on this one until the type annotation updates #1980 merged, but that's all set now.

Daniel-Haas-B:
Hey again @bmcfee ! I'm addressing this and there is only a quick question/annoyance left on some typing. I’m getting a mypy/SciPy-stubs complaint in `util.shear` / `__shear_sparse` that I am unsure how to proceed. 

mypy complains about `return X_shear.asformat(fmt)` when we try to type it as same sparse subclass in and out (CSR) via the `TypeVar`. The SciPy stubs don’t seem precise enough for `asformat()` to satisfy that promise.

So should we:
- keep the union-of-concrete-sparrays + `TypeVar` approach and add a small `cast`/ignore (or a format switch), or
- simplify `_SparseArray` to just `scipy.sparse.sparray` and loosen the typing contract to “sparray in,  sparray out” (the tests already enforce `.format` preservation).

Do you have a here?

bmcfee:
[...]

PR #1942, opened 2025-06-21
Include `scipy-stubs` in the test dependencies

jorenham:
I noticed that you care about static typing, so I figured that [`scipy-stubs`](https://github.com/scipy/scipy-stubs) could help with that. I added it to the `test` dependencies because there was another stubs-only package there.


bmcfee:
Thanks for this - does it mean we should then remove this bit from setup?

https://github.com/librosa/librosa/blob/e403272fc984bc4aeb316e5f15899042224bb9fe/setup.cfg#L132-L133

jorenham:
> Thanks for this - does it mean we should then remove this bit from setup?
> 
> https://github.com/librosa/librosa/blob/e403272fc984bc4aeb316e5f15899042224bb9fe/setup.cfg#L132-L133

yea I think so; let's try :)

bmcfee:
Hmm looks like this is failing pretty badly (41 errors).  Some of these look easy to fix (eg str->literal specification), and others are probably not worth the effort as they're in the critical path of code that will be rewritten soon anyway (everything involving sparse matrix, on the block for #1902 ).

I think the best path forward is to block this one until #1902 is fixed, then slate it for 1.0 (as it will likely require some API breakage in terms of user-facing type annotations).

jorenham:
> Hmm looks like this is failing pretty badly (41 errors).

At first glance it looks like many of these errors are caused by issues that have already been fixed in the latest `scipy-stubs` [1.16.0.0 release](https://github.com/scipy/scipy-stubs/releases/tag/v1.16.0.0). But we haven't been able to publish it to conda-forge (https://github.com/conda-forge/scipy-stubs-feedstock/pull/22), because we're currently waiting on `scipy` 1.16.0 (https://github.com/conda-forge/scipy-feedstock/pull/300).

bmcfee:
Gotcha - let's wait and see then.  Thanks!

jorenham:
It wasn't as smooth as I'd hoped it to be, but CI is green now :)

jorenham:
Some functions now require `Literal` input, but if backwards compatibility is a concern there, then I could also revert them and `# type: ignore` around it.

librosa/core/harmonic.py
```diff
@@ -10,10 +10,22 @@
...
+    "linear",
+    "nearest",
+    "nearest-up",
+    "zero",
+    "slinear",
+    "quadratic",
+    "cubic",
+    "previous",
+    "next",
+]
```
bmcfee: Two questions:

1. Is this enum consistent with our minimum supported scipy?  I seem to remember some newer modes showing up (eg nearest-up) but it's worth double-checking to make sure this is consistent.
2. Would it make sense to bump this up to _typing?  We use interpolators in a few different submodules, and we probably don't want to repeat ourselves.

librosa/core/spectrum.py
```diff
@@ -1483,7 +1483,7 @@ def iirt(
     center: bool = True,
     tuning: float = 0.0,
     pad_mode: _PadMode = "constant",
-    flayout: str = "sos",
+    flayout: Literal["ba", "sos", "zpk"] = "sos",
```
bmcfee: I don't think we actually support zpk in this method.  Arguably we should, but the reason we added sos was to improve numerical stability.  I can't think of any obvious use cases for zpk mode here.

librosa/core/spectrum.py
```diff
@@ -2103,7 +2103,17 @@ def fmt(
...
+        "linear",
+        "nearest",
+        "nearest-up",
+        "zero",
+        "slinear",
+        "quadratic",
+        "cubic",
+        "previous",
+        "next",
+    ] = "cubic",
```
bmcfee: replicated interpmode enum

librosa/filters.py
```diff
@@ -1227,7 +1227,9 @@ def get_window(
     elif isinstance(window, (str, tuple)) or np.isscalar(window):
         # TODO: if we add custom window functions in librosa, call them here
 
-        win: np.ndarray = scipy.signal.get_window(window, Nx, fftbins=fftbins)
+        # The `type: ignore` is needed on `scipy-stubs<1.16.0.1`
+        # https://github.com/scipy/scipy-stubs/issues/662
+        win: np.ndarray = scipy.signal.get_window(window, Nx, fftbins=fftbins or False)  # type: ignore[arg-type]
```
bmcfee: I get the type ignore, but why the `or False`?

librosa/filters.py
```diff
@@ -1246,8 +1248,8 @@ def _multirate_fb(
     Q: float = 25.0,
     passband_ripple: float = 1,
     stopband_attenuation: float = 50,
-    ftype: str = "ellip",
-    flayout: str = "sos",
+    ftype: Literal["butter", "cheby1", "cheby2", "ellip"] = "ellip",
+    flayout: Literal["ba", "sos", "zpk"] = "sos",
```
bmcfee, line 1247: same comment as above re: zpk

librosa/feature/rhythm.py
```diff
@@ -472,7 +472,17 @@ def tempogram_ratio(
...
+        "linear",
+        "nearest",
+        "nearest-up",
+        "zero",
+        "slinear",
+        "quadratic",
+        "cubic",
+        "previous",
+        "next",
+    ] = "linear",
```
bmcfee: interp mode again

librosa/core/harmonic.py
jorenham: > 1\. Is this enum consistent with our minimum supported scipy?  I seem to remember some newer modes showing up (eg nearest-up) but it's worth double-checking to make sure this is consistent.

it's consistent :)
https://docs.scipy.org/doc/scipy-1.6.3/reference/generated/scipy.interpolate.interp1d.html

> 2\. Would it make sense to bump this up to _typing?  We use interpolators in a few different submodules, and we probably don't want to repeat ourselves.

Yea good idea, I'll get on that

librosa/util/utils.py
```diff
@@ -2010,10 +2010,19 @@ def __shear_dense(X: np.ndarray, *, factor: int = +1, axis: int = -1) -> np.ndar
...
+_SparseMatrixT = TypeVar(
+    "_SparseMatrixT",
+    scipy.sparse.bsr_matrix,
+    scipy.sparse.coo_matrix,
+    scipy.sparse.csc_matrix,
+    scipy.sparse.csr_matrix,
+    scipy.sparse.dia_matrix,
+    scipy.sparse.dok_matrix,
+    scipy.sparse.lil_matrix,
+)
```
bmcfee, line 2022: repeated sparsematrix type

librosa/filters.py
jorenham: Because `fftbins` could also be `None` at this point, and the `get_window` requires `fftbins: builtins.bool | np.bool_`. And I'm a bit hesitant to also include `None` in there, because it's not always the same as `False` in scipy, and sometimes has a special meaning.

tests/test_decompose.py
```diff
@@ -254,7 +254,7 @@ def test_nn_filter_badselfsim(data, x, y, sparse):
 
     srand()
     # Build a recurrence matrix, just for testing purposes
-    rec = np.random.randn(x, y)
+    rec: np.ndarray | scipy.sparse.csr_matrix = np.random.randn(x, y)
```
bmcfee, line 257: Ah this gets tricky.  I recently merged a huge overhaul of the test suite into the 1.0 dev branch (not yet merged into main), which among other things, involved a total rewrite of RNG usage in the test suite.

It's a little orthogonal to the changes in this PR, but we're headed for a merge conflict...

bmcfee (commented):
This all looks great, thanks for the effort!

I did note a few places where there appear to be some redundant definitions.  This isn't a huge deal, but they do seem like definitions that are likely to need changing in the near future (eg sparse matrix -> sparse array) so I want to minimize the chances for getting out of sync.

A bigger problem is going to be divergence from main to the 1.0 branch.  I think this change will make the most sense to apply in the 1.0 branch (hoping to finish this up over the summer), and there are a few places especially in the tests that are already bound for a merge conflict.  Do you think it would be possible to try a rebase to 1.0.0dev and see how that goes?

librosa/filters.py
bmcfee: Oh interesting.  I wonder why we made this optional in the first place? O_o

bmcfee: Just had a look at the mega PR that set this to Optional, and I think it might just have been a mistake that slipped through the CR.  Probably best to just fix it here than to contort the logic around None support that isn't actually used (this is the only place where `fftbins` is used in the function).

librosa/core/spectrum.py
jorenham: Random fact: Technically speaking a `Literal` is different from an enum: In type-theoretical terms a Python `Literal` is a [union type](https://en.wikipedia.org/wiki/Union_type), and an `Enum` a [variant type](https://en.wikipedia.org/wiki/Variant_type_(COM)), which is a special case of a [tagged union type](https://en.wikipedia.org/wiki/Tagged_union), a.k.a. a *disjoint* or *discriminated* union type. 

To illustrate this difference, note that `LitABC = Literal["a", "b"] | Literal["b", "c"]` is exactly equivalent to `LitABC = Literal["a", "b", "c"]`, while a `EnumABC = EnumAB | EnumBC` cannot be reduced to something simpler, because `EnumAB.B` and `EnumBC.B` are different static types, even though they represent the same `"b"` value.

librosa/util/utils.py
jorenham: These `TypeVar` constraints (which differ from `bound=`) cannot be written in terms of the `_SparseMatrix` union unfortunately.

librosa/core/spectrum.py
bmcfee: Huh, TIL :smile:

tests/test_decompose.py
jorenham: I personally don't mind dealing with merge conflicts. So if that will happen, then I'd be happy to help out.

librosa/filters.py
jorenham: The difference is that in this case, the docstring explicitly mentions that `zpk` is allowed. So sould I keep it, or remove it together with the docstring?

librosa/filters.py
```diff
@@ -1419,7 +1421,7 @@ def semitone_filterbank(
     center_freqs: Optional[np.ndarray] = None,
     tuning: float = 0.0,
     sample_rates: Optional[np.ndarray] = None,
-    flayout: str = "ba",
+    flayout: Literal["ba", "sos", "zpk"] = "ba",
```
jorenham: `"zpk"` is not mention in the docstring here, so I'll remove it

jorenham:
> Do you think it would be possible to try a rebase to 1.0.0dev and see how that goes?

Sure, I'll create a new branch for that.

jorenham:
> > Do you think it would be possible to try a rebase to 1.0.0dev and see how that goes?
> 
> Sure, I'll create a new branch for that.

<details>
<summary>These were the conflicts, and most of them were pretty trivial (trailing whitespace):</summary>

image

image

image

image

image

image

image

image

image

</details>

Relevant PR: https://github.com/librosa/librosa/pull/1946

bmcfee:
I'm going to close this one in favor of #1946

PR #1943, opened 2025-06-22
Update README.md

Alokbpandey:

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,11 @@ The latest stable release is available on PyPI, and you can install it by saying
 ```
 python -m pip install librosa
 ```
+or 
 
+```
+pip install librosa
+```
 ### Using Anaconda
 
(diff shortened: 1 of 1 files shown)


Alokbpandey:
adding pip instalation

bmcfee:
The readme already includes the correct version of this (python -m pip install).

Alokbpandey:
Since most people use Colab or Jupyter Notebook, I thought the pip installation command should be included.

bmcfee:
> Since most people use Colab or Jupyter Notebook, I thought the pip installation command should be included.

That's fine, but the correct pip command is already included, directly above where you added to the readme.

That said, notebook users should not be installing packages by punting to shell. This is a pretty bad anti pattern that colab forces users into because it lacks proper environment management, and I'd rather not encourage it further.

PR #1946, opened 2025-06-26
Include scipy-stubs in the test dependencies - v1.0.0 rebase

jorenham:
Ref  https://github.com/librosa/librosa/pull/1942#pullrequestreview-2962888130


jorenham:
CI appears to be stuck

bmcfee:
> CI appears to be stuck

:face_with_diagonal_mouth: it's been doing that sporadically for the past few days on our minimal env.  I haven't had time to dig into it and see what the problem is...

jorenham:
> > CI appears to be stuck
> 
> 🫤 it's been doing that sporadically for the past few days on our minimal env. I haven't had time to dig into it and see what the problem is...

I suppose you could drop Python 3.8 support, which has been EOL since 2024-10-07

bmcfee:
> I suppose you could drop Python 3.8 support, which has been EOL since 2024-10-07

Yeah, fair enough.  We tend to run a bit slower on dropping support than other packages, but I think dropping 3.8 for the 1.0 version makes sense.

bmcfee (approved):
LGTM, and we're all in the green now.  Thanks again!

PR #1952, opened 2025-07-10
Colorbar helpers

### LINKED ISSUE #1940: [ENH]: Colorbar helpers
### Problem

A common pattern in generating figures is to make a decibel-scaled spectrogram and plot a colorbar along side.  The default colorbar is fine, but it's typically better to display units (dB) in the tick labels.  We have examples that do this throughout the documentation, but i'm thinking it would be handy (and easy) to provide a shortcut.

Similarly, as #1938 adds some functionality for phase plots, it would be nice to have a colorbar constructor that places ticks at meaningful values (in radians) and notates them properly as fractions of π instead of decimal values.

### Proposed solution

It would be fairly easy to make some lightweight wrappers to do the above.  I'm thinking something like:

```python
librosa.display.colorbar_db(...)
librosa.display.colorbar_phase(...)
```
which would wrap either the procedural or object API.

---
Poissibly there are other value scales that could benefit from some custom colorbar annotations, but I can't think of any offhand.

bmcfee:
#### Reference Issue

Fixes #1940 

This PR adds two functions: `display.colorbar_db` and `display.colorbar_phase` which are simple helpers to construct sane colorbars for decibel- and radian-valued plots, respectively.

#### Any other comments?

I think this is basically complete and ready to go, but there are a couple of quirks that popped up while implementing it:

1. We might want to explicitly set the vmin/vmax parameters on phase plots to always span ±π.  This would be a slight modification to #1938 - the reason for doing so is that if we have a plot that for whatever reason has a limited angle values, and we apply a cyclic colormap, then the visual wrapping point will not coincide with the underlying angle continuity point.  I don't think we need to change anything about the colorbar helper for this, but we can definitely lump it into this PR.
2. I'm using a linear locator by default for both plots.  Maybe an AutoLocator would be more appropriate / idiomatic?  I picked linear because it tends to pick nice fractional points for phase plots (π/4, 3π/4, etc), but it does some strange things with decibel plots.  Maybe it's best to use Auto for dB and linear for phase.  This slightly connects to the first point in that linear for phase really only makes sense when the vmin/vmax fully span ±π.
3. There might be a bit of fiddling to do with type annotations on these functions.  I'll let CI and linting take a whack at it and then come back to it.


bmcfee:
Turns out this requires a nudge of matplotlib>=3.6.0 to get https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.6.0.html#new-style-format-string-for-colorbar-ticks

bmcfee:
Seems to be bumping into a problem with colorbar format strings.  This is supposed to work in matplotlib 3.6.0, but evidently it does not.  It's possible that mpl is failing to guess the type of format string, and we may need to bump up to 3.8.1 (2023) to get a fix for it (matplotlib/27007).  I'll try to make some time do bisect this one out probably next week.

bmcfee:
Update: a simpler fix is to just use the old-style % formatter for the db colorbar and roll back the minimum matplotlib version change.

PR #1958, opened 2025-07-16
Warping path to step grid

### LINKED ISSUE #1937: [ENH]: Warping path to time grid

bmcfee:
#### Reference Issue
Fixes #1937 

This PR adds `librosa.sequence.path_to_steps`.

The function accepts a warping path as produced by `dtw` and generates a sequence of fractional frame indices which can be used to align the first sequence to the second.

If the `inverse=` flag is set to `True`, then the alignment is reversed, and the frame mapping goes from the second sequence to the first.

#### Any other comments?

This still needs tests, but works fine.  Getting the PR going to start the CI.

I'm also rolling in some enhancements to the advanced example for music synchronization here.


bmcfee:
Doing a little cosmetic surgery on the music sync notebook in this PR.  Here are some new images:

image


Uses a subplot mosaic to show the feature matrices along with the cost matrix.  Also uses a quiver plot instead of scatter to show the direction of the warping path.

image

Notebook now actually includes synchronization via variable-rate phase vocoding.  I've updated all the plots in the example so that the first signal (x1) is shown in blue and the second signal (x2) is always shown in orange.  This makes it much easier to visually track what's going on.

bmcfee:
Getting a test failure with interp1d on scipy 1.6.0, stemming from repeated x values for the same y value.  #1902 would have us nudge to 1.8 anyway, so I'll try that and see if it gets this unstuck.

Otherwise I'll need to rewrite this to produce unique outputs for the interpolator.

Issue #1914, opened 2025-02-12
[Doc]: Migrate intro/tutorial to a mini-gallery


bmcfee:
### Documentation Link

https://librosa.org/doc/latest/tutorial.html

### Problem

The `tutorial` page in the documentation is currently a hard-coded RST file with code examples.  One issue that can arise here is that the code examples are not executed at any point in our CI or doc build, so when things break (eg due to upstream deprecations) it's difficult to catch.

### Suggested improvement

A solution to this would be to migrate the tutorial into a sphinx-gallery, as we have with our `advanced examples` section.  This would have a couple of benefits:

1. Code executes in CI, so we can detect errors.
2. Tutorial would be immediately downloadable as a notebook.
3. We can split the tutorial section into multiple smaller components to focus on different introductory topics.

The sphinx-gallery extension allows for multiple galleries, so it would be easy to keep the intro example(s) separate from the advanced examples.  We might also want to structure this in a more sequential manner than a flat gallery, perhaps using [mini galleries](https://sphinx-gallery.github.io/stable/configuration.html#add-mini-galleries).

Issue #1819, opened 2024-03-13
Physical units in `specshow` axis labels


lostanlen:
**Is your feature request related to a problem? Please describe.**
Today i was looking at our (lesser-known!) gallery of advanced examples
https://librosa.org/doc/main/advanced.html
and noticed that the x-axis says "Time" without physical unit.

Meanwhile, the y-axis gives the physical unit (`Hz`) but not the quantity (`"Frequency"`).
The same is true for tempograms. The axis says `BPM` not `Tempo (BPM)`.

**Describe the solution you'd like**
It would be good to have consistency through the examples and documentation on how to label axes.

**Describe alternatives you've considered**
Calling `xlabel`, `ylabel` manually after `specshow`.

**Additional context**
Here's (an excerpt) of `__decorate_axis`:

```python
def __decorate_axis(
    axis,
    ax_type,
    key="C:maj",
    Sa=None,
    mela=None,
    thaat=None,
    unicode=True,
    fmin=None,
    unison=None,
    intervals=None,
    bins_per_octave=None,
    n_bins=None,
):
    """Configure axis tickers, locators, and labels"""
    time_units = {"h": "hours", "m": "minutes", "s": "seconds", "ms": "milliseconds"}

    elif ax_type in ["tempo", "fourier_tempo"]:
        axis.set_major_formatter(mplticker.ScalarFormatter())
        axis.set_major_locator(mplticker.LogLocator(base=2.0))
        axis.set_label_text("BPM")

    elif ax_type == "time":
        axis.set_major_formatter(TimeFormatter(unit=None, lag=False))
        axis.set_major_locator(
            mplticker.MaxNLocator(prune=None, steps=[1, 1.5, 5, 6, 10])
        )
        axis.set_label_text("Time")

    elif ax_type in time_units:
        axis.set_major_formatter(TimeFormatter(unit=ax_type, lag=False))
        axis.set_major_locator(
            mplticker.MaxNLocator(prune=None, steps=[1, 1.5, 5, 6, 10])
        )
        axis.set_label_text("Time ({:s})".format(time_units[ax_type]))
```

Related: #1364 #1746

bmcfee:
[...]

PR #1967, opened 2025-07-29
oct3 display modes

bmcfee:
#### Reference Issue

Implements #1819 sort of

This PR adds various `oct3`-based display modes for log frequencies (STFT), cqt, vqt, and mel.

#### Any other comments?

Caveats abound in this PR, as enumerated in the discussion of #1819.  But I think this is essentially functional and useful as is.

I'm starting the CI on this now, but plan to add a blurb about it in the (way too big) specshow advanced example page.  This will almost certainly be refactored into a smaller section via #1914 , but I'd like to add the content now.


bmcfee:
> but plan to add a blurb about it in the (way too big) specshow advanced example page. This will almost certainly be refactored into a smaller section via #1914 , but I'd like to add the content now.

Upon closer inspection, there's not a great way to stuff this into the current example page.  I'll punt this section to when we expand out the display example into a full multipage tutorial section (planned for #1914).

I think we can consider this PR as complete and ready for review.

bmcfee:
I rolled in a bit of axis compatibility code for VQT modes (which had been missing before), and updated the docstring to cover new functionality.  I think this is :shipit: ready.

Issue #1370, opened 2021-08-27
Multichannel display


bmcfee:
We've previously decided that #1130 does not directly include modifications to the display module, and that those functions should remain mono-only.  However, this doesn't rule out the possibility of adding new functionality to support multichannel visualizations easily, and I think this would be quite helpful in general.

**Describe the solution you'd like**

This is easiest to consider starting from wave plots.  In the common case of 2-channel signals, I think it would be helpful to have a plot that renders a joined pair of subplots with one channel each and shared axes, like you might see in a DAW.

Here's some prototype code that does this already:
```python
fig = plt.figure(figsize=(10, 5), constrained_layout=False)

grid = fig.add_gridspec(2, 1, wspace=0, hspace=0)
axes = grid.subplots(sharex=True, sharey=True)

librosa.display.waveshow(y[0], sr=sr, ax=axes[0])
librosa.display.waveshow(y[1], sr=sr, ax=axes[1])
axes[0].label_outer()
axes[0].set(xlabel=None)
fig.show()
```
Which produces the following image with the trumpet example:
![image](https://user-images.githubusercontent.com/1190540/131137697-34c19225-43ad-4962-bcdc-5f9fc7766801.png)

Note that this code would break the current pattern of display functions accepting target axes, since it will need to create an axes for each channel.  I think this is fine, and we can still accept `fig` as a target or create a new one as needed.

We also have to do a little bit of cleanup here to hide the xlabel and ticks. (Side note: maybe waveshow and specshow should have options to not set axis labels.)

This approach would easily generalize to `C>2` channels, though if we have higher order arrangements we might get into trouble and have to flatten the array down in advance.

Some parameters we should expose (with defaults):

- `wspace,hspace=0`: spacing between subplots. I think most people would expect vertical packing as i've done in this example, but it should be configurable.
- `sharex,sharey=True`: subplot axis coupling.  This seems generally desirable, but maybe there are use cases where it would not be helpful.
- waveshow parameters (passthrough)

The function would then return, I guess, the `grid` object?  This seems to be how [seaborn](https://seaborn.pydata.org/generated/seaborn.pairplot.html#seaborn.pairplot) does it for a similar use-case, and they seem pretty on-the-ball.

We could do similar things for `specshow`, though we might not want to default to dense packing there because specshow plots don't have the benefit of negative visual space like waves do.

**Describe alternatives you've considered**
We could also just punt this and leave people to their own devices, but that seems like a disservice.

[...]

PR #1968, opened 2025-07-29
Tutorial restructure and general documentation overhaul

### LINKED ISSUE #1454: Improve readme, documentation landing page
I don't have any clear ideas for this, but I suspect that our main website, front page of the documentation, and project readme could be a bit more inviting (and/or flashy).

Some nice examples we could take inspiration from:

- [sklearn](https://scikit-learn.org/stable/)
- [matplotlib](https://matplotlib.org/)
- [seaborn](https://seaborn.pydata.org/index.html)

As part of this, we might want to consider developing some more complete, tutorial-oriented examples to put on a landing page.  Maybe something like using mirdata to download, librosa for features, sklearn for classification?  I'm open to all kinds of suggestions here.

### LINKED ISSUE #1478: Documentation restructuring (core submodule)

### LINKED ISSUE #1823: Docstring for VSCode preview
Hi,

I just stumbled on the Docstring preview in VSCode and that numpy's version is much nicer whereas librosa only shows the TOC.

Seems that VSCode simply renders the Docstring given in the `__init__.py`.
https://github.com/librosa/librosa/blob/main/librosa/__init__.py

It could simply be a copy-pasta from the docs starting page stating what librosa is about.

image
image

### LINKED ISSUE #1914: [Doc]: Migrate intro/tutorial to a mini-gallery

### LINKED ISSUE #1949: [Doc]: Make advanced example notebook downloads self-contained
### Documentation Link

_No response_

### Problem

I noticed today that some of our advanced example notebooks are not fully self-contained, and do not execute cleanly when downloaded.  This is because they rely on bespoke audio clips that are not accessible through the example gallery, and are not included in the "download notebook" link.

### Suggested improvement

It does not appear that sphinx-gallery has a way to bundle data with generated notebook downloads, and even if it did, I don't think we want to do that since we'd need to figure out the file location for loading.

I think the only reasonable path forward is to use the example downloader we already have, and just add more clips to the repository to support our examples.  At present, we really only have 3 to worry about: the two Sir Duke recordings from @stefan-balke and the accelerating snare drum from myself.

@stefan-balke what do you think, should we promote these up to `librosa-data` so that all downloadables are fully self-contained and replicable?

### LINKED ISSUE #1957: [ENH]: html / repr_html formatting for example info
### Problem

Our helper function to print out metadata about packaged examples (`librosa.util.example_info()`) is fine for REPR and console-based uses.
However, I'm finding myself wanting to use it within the example gallery (and new tutorial sections #1914 ) to expose attribution for examples used in each section.  Using plain text ends up looking pretty clumsy, and there's not a very elegant way to capture and reformat.

### Proposed solution

The example_info() function could accept additional parameters to facilitate rich formatting of metadata.  This doesn't need to be fancy - I think it will be best to keep the HTML formatting here simple, and then bundle some CSS in the doc site to make it look nice in context.

I expect that nobody will use this functionality externally, but anything we can do to make the example attribution a bit clearer would be good.

### LINKED ISSUE #2045: Retire the vocal separation example
It's been deprecated for quite some time #1492 , with a big warning across the top saying to not use it #1602 , but I think 1.0 makes for a good opportunity to formally retire that example.

#1968 will involve some restructuring of the advanced examples to eliminate redundancy with the new tutorial section, so if there are no objections, I'd like to roll this removal into that PR.

@tuomaseerola had previously suggested keeping this example around for historical reasons, but that was 4 years ago already. (Time flies!)  Is there still need for this, or can we really retire it now?

bmcfee:
#### Reference Issue

Fixes #1914 
Fixes #1823 
Fixes #1454 
Fixes #1478
Fixes #2045 
Fixes #1949 
Fixes #1957

This PR implements a substantial overhaul of the intro tutorial section.  The plan is to completely replace the existing "tutorial" section of the docs, and very likely replace some of the "advanced examples".

The intro material is now broken down into multiple gallery pages, each focusing on a different thematically related topic.

The proposed structure is as follows:

- Main intro:
  - Loading audio
  - Frequency analysis
  - Pitch
  - Harmony
  - Onsets
  - Rhythm
  - Timbre
  - Musical notation
- Display overview
  - waveshow in more detail
  - basic spectrogram display
  - different frequency axes
  - color handling
  - non-spectral data (tempo, chroma, tonnetz)
  - non-uniform time displays
  - multi-channel displays (depends on #1370 )
  - mir_eval integration and annotations

#### Any other comments?

This is very much a WIP.  I've been trying to knock out a section per week for the last month or so, but will hopefully be able to pick up the pace once some of the display functionality is merged.  I'm opening the PR now mainly because I keep referring to this branch in other issues, so we may as well have it public for scrutiny.


docs/tutorial.rst
```diff
@@ -1,10 +1,12 @@
...
-
 This section covers the fundamentals of developing with *librosa*, including
 a package overview, basic and advanced usage, and integration with the *scikit-learn*
 package.  We will assume basic familiarity with Python and NumPy/SciPy.
 
+.. toctree::
+   :maxdepth: 2
+
+   auto_tutorials/index
+
```
bmcfee, line 10: Note: the plan is to remove the remainder of this file once the intro section is complete.

bmcfee:
Dropping a note for now: the current status on this is that most of part 1 is there, except for the timbre section, which will require some new example data to do the way that I want.  (I'll work on this after the semester wraps up.)

Part 2 is going to focus on display, and largely replace the advanced specshow page with more digestable chunks.  I'm going to block that by the multichannel display PR since having that pinned down first will make a bunch of the examples easier to code up.

bmcfee:
Coming back to this one as the dust is settling on 1.0 functionality.  I noticed that we now have the opportunity to make the documentation site a bit more accessible by adopting the [`petroff10` colormap](https://arxiv.org/abs/2107.02270) in the doc build, rather than the matplotlib default.  This wouldn't change any code or introduce regressions - just the styling of the figures on the doc site.

For example, our onset strength figure would look like:
image
compared to the matplotlib default (current):
image

Multiplot example now looks as:
image
and remains distinguishable under protanopia simulation:
image

I really see no reason not to move ahead with this switch, so I'll work it into this PR.

bmcfee:
Hitting an interesting snag with the multiversion doc build here.

The single-version build works fine.  But when we go to the multiversion, it fails because the `tutorials/` path does not contain a gallery header in older tags (because it didn't exist).  sphinx-gallery does not allow this, and fails noisily.

However, the sphinx config is a one-shot setup that applies to all historical builds, so there's really no way around it.

I wonder if this is the point at which we finally give up on sphinx-multiversion?

bmcfee:
This is unblocked with the new site theme.  I've put up a temporary build of the new tutorials here: https://librosa.org/web-staging/ in case anyone wants to give it a look.

Overall the docs are restructured a bit; the advanced examples are now a third section under the tutorials, for example.  But I think it more or less works, modulo some weirdness in how the left-hand nav panel is behaving currently.

bmcfee:
I did a soft reset on this one so that history doesn't get totally messed up again.  This had the unpleasant side effect of rolling everything into one giant commit. :upside_down_face:

bmcfee:
@ejwong24 @AsymptotaX @cookesan @emilazy - I'm updating the contributing / authors list, and right now the only attribution I have for you is your github handles.  Please comment here with your preferred attribution, or otherwise give a :eyes: to let me know that you've seen this message and are okay with it as is.

bmcfee:
Merging this now to test drive the new doc build actions.  There will still be time to update attributions if folks want https://github.com/librosa/librosa/pull/1968#issuecomment-5120799449

Issue #1751, opened 2023-09-15
Update mypy check for matplotlib 3.8


bmcfee:
When type annotations were added in #1587 and #1632 , we had to punt on matplotlib imports:

https://github.com/librosa/librosa/blob/43d4427f7a389acc79b5be3c924646701f19bb46/setup.cfg#L135-L136

Matplotlib 3.8 release notes mention that type hints are now provided for most public APIs: https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.8.0.html#type-hints

Once 3.8 finishes rolling out on conda(-forge), we should be able to update the lint CI environment and remove the matplotlib skip from our setup.cfg.  This will require us to bump the minimum mpl version in our CI environment spec.

bmcfee:
Taking a first whack at this, failures are plentiful:

```
librosa/display.py:152: error: Item "None" of "Axis | _DummyAxis | _AxisWrapper | None" has no attribute
"get_data_interval"  [union-attr]
            _, dmax = self.axis.get_data_interval()
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~
librosa/display.py:153: error: Item "None" of "Axis | _DummyAxis | _AxisWrapper | None" has no attribute
"get_view_interval"  [union-attr]
            vmin, vmax = self.axis.get_view_interval()
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
librosa/display.py:245: error: Item "None" of "Axis | _DummyAxis | _AxisWrapper | None" has no attribute
"get_view_interval"  [union-attr]
            vmin, vmax = self.axis.get_view_interval()
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
librosa/display.py:331: error: Item "None" of "Axis | _DummyAxis | _AxisWrapper | None" has no attribute
"get_view_interval"  [union-attr]
            vmin, vmax = self.axis.get_view_interval()
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
librosa/display.py:423: error: Item "None" of "Axis | _DummyAxis | _AxisWrapper | None" has no attribute
"get_view_interval"  [union-attr]
            vmin, vmax = self.axis.get_view_interval()
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
librosa/display.py:476: error: Item "None" of "Axis | _DummyAxis | _AxisWrapper | None" has no attribute
"get_view_interval"  [union-attr]
[...]

PR #1971, opened 2025-07-30
Inverted waveshow

bmcfee:
#### Reference Issue
This PR implements the inverted waveform display functionality as described in https://github.com/librosa/librosa/issues/1370#issuecomment-1147514144

Two new parameters are added to `waveshow`:

- `invert` - boolean, whether or not to invert the display
- `invert_color` - if `invert=True`, then the color to use for the inverted waveform.  If not provided (default), the color is sniped from the axes facecolor (usually white).

#### Any other comments?

This PR implements a rather dirty hack to ensure that legend behavior is consistent.  We have a proxy patch artist that contains the provided label (if any), and its position is set to `nan` so that it renders but does not show up in the figure.  This *works* (mpl 3.10.3), but i'm not yet sure if it's *valid* - currently waiting on a response via mpl discourse.

This isn't a mission critical feature, but I think it's kind of fun and looks neat:

```python
plt.close('all')
fig, ax = plt.subplots()
librosa.display.waveshow(y=y, sr=sr, ax=ax, label='foo', invert=True, invert_color=None)
ax.legend(loc='upper right')
```
image

```python
plt.close('all')
fig, ax = plt.subplots()
librosa.display.waveshow(y=y, sr=sr, ax=ax, label='foo', invert=True, invert_color='#2d2d2d')
ax.legend(loc='upper right')
```
image


bmcfee:
The real fun of this is in the multichannel display prototype #1370 where we get a nice DAWish visual (nevermind the bad legend placement): 

image

bmcfee:
I think this is basically complete, with two caveats:

1. I'm not sure the type annotation for `invert_color` is correct.  This intersects with #1751 - *Edit*: I'll punt worrying about this until #1751.
2. Aforementioned comment about nan positioning - will report back on that when I get a response.

If (2) turns out to be completely unsupported nonsense, we can always roll it back to position within the waveform's time grid.  This is what mir_eval's source separation display does, but that somehow feels even more hacky than using nans to explicitly say "don't draw this".

bmcfee:
For reference, the discourse thread on my hacky nan usage is here: https://discourse.matplotlib.org/t/is-it-ok-to-use-nan-for-patch-rect-coordinates/25993

MLsound:
I believe this is a particularly good feature to have when a user is working in dark or high-luminosity environments, as it makes graphs more readable and allows the user to improve the contrast on their screen. Personally, as librosa user, I would prefer the white-over-blue style to the darker one.

A multichannel view is also a nice-to-have feature, especially when working with multitracks.

@bmcfee Please let me know if there's any task where I can help.

bmcfee:
> I believe this is a particularly good feature to have when a user is working in dark or high-luminosity environments, as it makes graphs more readable and allows the user to improve the contrast on their screen.

Thanks - it's worth noting that if you're really looking for dark mode support, it is probably better to use a [dark mode theme](https://matplotlib.org/stable/gallery/style_sheets/dark_background.html) in matplotlib - librosa will respect property cycler of whatever theme matplotlib is configured to use.



> Personally, as librosa user, I would prefer the white-over-blue style to the darker one.

It's all easy to override. :grin: I personally prefer the dark-over-color mode to white, but the white mode (default) has the benefit that it's actually governed by the theme (see above) and pulls directly from the axes facecolor.  So as a default, it's a bit more idiomatic or "unscented".



> A multichannel view is also a nice-to-have feature, especially when working with multitracks.

That's a separate PR (forthcoming), but the motivation for the inverted mode really came from the multichannel use case where you have a bunch of stems that you want to color-code in the same display.  In particular when each stem might be stereo and you want to share colors, eg the following is 3 stereo stems (6 channels in all):

```python
fig, ax = plt.subplots(nrows=y.shape[0], sharex=True, sharey=True, gridspec_kw=dict(hspace=0), figsize=(10, 6))
multiplot('wave', y, sr=sr, share_properties=np.array([0,0,1,1,2,2]), axes=ax, invert=True, invert_color='#2d2d2d')
```

image

It's also okay with the ordinary display mode (invert=False):

image

but a little less obvious to parse at a glance, especially if zoomed in to silent regions.


> @bmcfee Please let me know if there's any task where I can help.

Thanks - I think this one is basically good to go, though I'm waiting on some confirmation that my proxy patch legend hack is not going to blow up on us before merging.  I'll likely then open the PR for the multichannel display next week.

bmcfee:
Well the discourse thread has gone silent, and I don't expect a response now any time soon.  I'm pretty sure that what we're doing here is legitimate since it relies on (not-quite-documented) functionality that's necessary for scatter plots to work.

I'll go ahead and merge this now, and if a response comes in later on, we can revisit the implementation.

Issue #1972, opened 2025-08-06
[ENH]: Alternative waveform displays


bmcfee:
### Problem

`display.waveshow` is designed for interactive plots where fine details matter.  It's not always that useful or aesthetically pleasing for other situations where visual simplicity may be preferable over fidelity to the original signal.  This leads me to think that another waveform display function might be useful to have at some point.

### Proposed solution

Here's a mockup of what I have in mind for a simplified waveform display.  It shares much of the same logic from `waveshow` in amplitude envelope mode, but rather than using a `fill_between`, it instead creates a patch collection of FancyBBoxes with rounded corners for a modern "web2.0" feel.  (Yeah, the mid 2000s are back!)

<img width="703" height="441" alt="Image" src="https://github.com/user-attachments/assets/5d078185-0116-4c4a-9d15-200007e85383" />

The code to produce the above is quite brittle and needs a lot of work, and so is nowhere near ready for a PR.  For example, the trailing silence results in negative curvature of the bbox corners, resulting in a weird sawtooth pattern.  Likewise the artist properties are not fully carried through.

I hope that the above conveys a sense of what I'm going for with this idea though.

bmcfee:
Did a bit of fine-tuning on the code for the above, and I think I have something pretty usable:

```python
def wavebars(y, *, sr=22050, n_bars=100, gap_ratio=0.4, ax=None, 
                     linewidth=0, label=None, **patch_kwargs):

    if y.ndim == 1:
        y = y[np.newaxis, :]
        
    axes = __check_axes(ax)
    
    hop = max(1, y.shape[-1] // n_bars)
    env = __envelope(y, hop)
    y_bottom, y_top = -env[-1], env[0]

    segment_duration = hop / sr
    bar_width = segment_duration * (1 - gap_ratio)
    gap = segment_duration * gap_ratio
    rounding_size = bar_width / 2
    
    patches = []
    
    for i, a in enumerate(zip(y_bottom, y_top)):
        x0 = i * (bar_width + gap)
        base = min(-rounding_size, a[0])
        top = max(rounding_size, a[1])
        p = mpatches.FancyBboxPatch(
            (x0, base), bar_width, top - base,
            boxstyle=f"round,pad=0,rounding_size={rounding_size}",
        )
        patches.append(p)
    
    coll = mcollections.PatchCollection(patches, linewidth=linewidth, transform=ax.transData, **patch_kwargs)
    axes.add_collection(coll)

[...]

Issue #1935, opened 2025-05-27
SPEC endorsements


bmcfee:
I took some time today to read through the Scientific Python Ecosystem Coordination (SPEC) recommendations as listed here: https://scientific-python.org/specs/ .  This issue is to track and discuss our support (or lack thereof) for each SPEC.

---

### SPEC 0 - Minimum supported dependencies

**TLDR**: https://scientific-python.org/specs/spec-0000/ 
-    Support for Python versions be dropped 3 years after their initial release.
-    Support for core package dependencies be dropped 2 years after their initial release.

**Our position**: We don't currently achieve this, and I don't think we necessarily aspire to either.  Rather, our minimum versions are dictated by required functionality  and API in the dependencies.  

It would be easy for us to adopt this SPEC, but I'm not sure that it's something we want to do necessarily.  I don't think the idea, as written, is entirely compatible with long-term reproducibility, but this is definitely a point for discussion.

### SPEC 1 - Lazy loading of submodules

We already do this and I think we can endorse it as is.

### SPEC 4 - Nightly wheels

There was a start at this during the scipy sprint in 2022, but it never quite got off the ground due to one upstream dependency or another.  I do think it's something we should try to do though.

### SPEC 6 - Keys to the castle

This one is difficult to summarize concisely.  Read more here: https://scientific-python.org/specs/spec-0006/

Of the points listed in the "principles" section in the above link:

1. We could do better on documenting restricted resources, probably.
2. I think we are in compliance with the developer privileges principle, but it could stand some review.
3. We could definitely increase the bus factor on project asset access.
4. I think we're most of the way there via github action publishing?

This one definitely needs discussion.

### SPEC 7 - random number generation

**TLDR**: 

-    standardizing the usage and interpretation of an rng keyword for seeding, and
-    avoiding the use of global state and legacy bitstream generators.

We definitely have a mix of keywords for RNG access in the current release.  This could be standardized through a deprecation cycle at the 1.0 checkpoint.

I believe we're otherwise good on global state / bitstreams in the core library code, and #1925 will resolve it similarly in the unit tests.

### SPEC 8

**TLDR**: Complements SPEC 6 and is more targeted at supply chain security (eg pypa distribution).

Broadly, I think we're pretty good on this front through our github action for publishing, but there may be things we can improve upon.

bmcfee:
SPEC 1 and 7 are now endorsed in the 1.0 branch as of #1960

bmcfee:
[...]

PR #1973, opened 2025-08-08
Wavebars

bmcfee:
#### Reference Issue
Implements #1972 

This PR implements `wavebars` for simplified amplitude envelope display.

#### Any other comments?

- [x] tests


bmcfee:
Figures from the docstring examples render as follows:

image

and
image

bmcfee:
For some reason, the image comparison tests just for wavebar are unhappy with matplotlib < 3.7.  I don't see any obvious causes for this in the release notes between 3.7 and 3.5.

I've bumped mpl to 3.7.0 here, though if we wanted to go SPEC0 on it #1935 we could get away with 3.8.  Doing so would be a good move anyway per #1751 - but i'll leave it at 3.7 for now in this PR.

Otherwise, I think this one is ready to :shipit:

librosa/display.py
```diff
@@ -2461,6 +2504,186 @@ def waveshow(
...
+    **patch_kwargs: Any,
+) -> mcollections.PatchCollection:
+    """Visualize a waveform as a series of bars representing the amplitude envelope.
+
+    This visualization is appropriate for displaying a simplified view of the
+    signal, and is best suited for small figures where simplicity is desired.
+
+    Parameters
+    ----------
+    y : np.ndarray [shape=(n,) or (2,n)]
```
lostanlen, line 2533: It would be good to explain here how to interpret the wavebar of a stereo y (i.e., upper vs lower halves)

lostanlen (approved):

Issue #1528, opened 2022-07-14
Nightly builds for bleeding edge dependencies


bmcfee:
**Is your feature request related to a problem? Please describe.**
It would be great if we had a nightly build CI that could pull the current development versions and release candidates of our dependencies and run our test suite.  This would make it easier to detect problems in advance of future API breakages, rather than being suddenly (and asynchronously!) surprised after they've already happened.

**Describe the solution you'd like**
Probably not all of our dependencies have nightly builds or dev releases, but we can do a best effort thing here.  At the very least, we should have

- scipy https://anaconda.org/scipy-wheels-nightly/scipy/files
- numpy https://anaconda.org/scipy-wheels-nightly/numpy
- sklearn https://scikit-learn.org/stable/developers/advanced_installation.html
- matplotlib https://github.com/matplotlib/matplotlib/pull/22733
- numba?
- soundfile?

It would also make sense to have a doc build action for sphinx (and extensions).

Nightly tests should also be stricter about warnings, especially deprecation warnings.

matthewfeickert:
:wave: @bmcfee you might be interested in following/contributing to the discussion on https://discuss.scientific-python.org/t/interest-in-github-action-for-scipy-wheels-nightly-uploads-and-removals/397 given this Issue. :slightly_smiling_face:

bmcfee:
Thanks @matthewfeickert !  Nothing to add from me right now, but I'll keep an eye on it.

PR #1977, opened 2025-08-27
Fix types in beat tracking code

emilazy:

The development version of Numba 0.62.0 is unhappy about `range` being applied to floating point values here.

### DIFF
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -582,7 +582,7 @@ def __beat_track_dp(localscore, frames_per_beat, tightness, backlink, cumscore):
         # Search over all possible predecessors to find the best preceding beat
         # NOTE: to provide time-varying tempo estimates, we replace
         # frames_per_beat[0] by frames_per_beat[i] in this loop body.
-        for loc in range(i - np.round(frames_per_beat[tv * i] / 2), i - 2 * frames_per_beat[tv * i] - 1, - 1):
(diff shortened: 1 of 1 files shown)


bmcfee:
Thanks for this - it appears to be the result of the fix for https://github.com/numba/numba/issues/9909

I'd like to hold off on doing anything with this until 0.62 is final, or we somehow get #1528 off the ground, as we can't otherwise verify that it works properly in CI.

I think the solution implemented here looks fine, though the explicit cast to int for `frames_per_beat` may be premature since it gets cast back to float in the loop line that triggered the problem here.

emilazy:
Very reasonable, both on waiting on CI and on the cast. I’ve pushed a much simpler fix to preserve the previous behaviour without changing the types.

bmcfee:
Thanks @emilazy  - hitting lint errors because the main branch is a bit stale right now.  I'm rebasing to the 1.0dev branch just to get past that.

bmcfee (approved):

emilazy:
FWIW, it looks like Numba plan to roll back the change due to it also causing issues with integers of mixed signedness. I assume the check for floats will return at some point though (this code doesn’t work in vanilla Python either), so this will still be relevant then.

bmcfee:
Ok, thanks for the heads up @emilazy.

I guess this is now a fairly innocuous change, but I see no reason to delay merging.  Is there anything else to do here for now?

emilazy:
It should be good to go as far as I’m concerned, if you want to hit the button :) Thanks for the review!

bmcfee:
Done - thanks again!

Issue #1798, opened 2024-01-17
pyin seems "slow", maybe add PWVD? 


drscotthawley:
**Is your feature request related to a problem? Please describe.**

HI Brian et al!  librosa is amazing. 
It seems like the [pyin f0 estimator](https://librosa.org/doc/0.10.1/generated/librosa.pyin.html#librosa.pyin) takes a "long" time, like around 4 seconds for about 30 seconds of low-SR audio. Like, all manner of spectrogram calculations and beat analyses can be completed in a fraction of a second, then the f0 estimation just...slogs. 
 Is this...normal?   

**Describe the solution you'd like**

I'd like it to run faster, somehow!  Perhaps there are suggestions already that I'm missing?  e.g. maybe I need to pad my waveform until it's got a power 2 number of samples or something? 

I see that you [already did some work on accelerating pyin and yin](https://github.com/librosa/librosa/issues/1608), but...to me it's still taking longer than I'd prefer. (Not that I have competing code that's any faster, mind you! )

* Could it be parallelized somehow? 
* What about implementing a PWVD method, e.g. in "A Fast and Accurate Pitch Estimation Algorithm Based on the Pseudo Wigner-Ville Distribution", https://arxiv.org/abs/2210.15272? 
(I looked for a Python implementation but haven't found one yet.) 

**Describe alternatives you've considered**

Um...not sure but I'm looking around and checking the internet.  Maybe there are pretrained deep learning models that can run fast in inference mode on CPUs?  ....Not sure.  Seems the core of librosa is classic signal processing, run on the CPU. 

**Additional context**

drscotthawley:
I'm mainly interested in vocals, so maybe I'll check out this method from 2009 or any updates thereof... 
https://www.aes.org/e-lib/browse.cfm?elib=15165
Really I'm a newb in this domain!

bmcfee:
> Like, all manner of spectrogram calculations and beat analyses can be completed in a fraction of a second, then the f0 estimation just...slogs.
> Is this...normal?

Kind of,  yes.  pyin is inherently sequential, since it's doing a viterbi decode over the frame-wise pitch likelihoods.  This creates a pretty severe computational bottleneck that we've done our best to optimize around, but there's a limit to how far you can push it.



> I'd like it to run faster, somehow! Perhaps there are suggestions already that I'm missing? e.g. maybe I need to pad my waveform until it's got a power 2 number of samples or something?

[...]

Issue #1743, opened 2023-08-30
The future, array API, and so on


bmcfee:
This question came up at the SpeechBrain 2023 summit, and I figured it warrants some open discussion and thought.  In short: what will it take for librosa to be compatible with the [Array API](https://data-apis.org/array-api/latest/)?  Motivation here is that doing so would make librosa backend-agnostic and ease interoperability with other packages, platforms (GPU), etc.

The way I see it, the answer is both "a lot" and "not much".

**A lot**

We make pretty extensive use of several upstream libraries.  Specifically:

- Numpy
- Scipy
- numba
- soundfile
- soxr

We also make not-so-extensive use of some other upstream libraries:

- sklearn
- audioread
- matplotlib

I don't think we can plausibly make an attempt at array API until the upstream libraries have implemented it.  There could be some exceptions here, eg if the interaction from librosa is one-directional (reading from soundfile, writing to matplotlib for example), we can probably wedge in some appropriate conversions and not have to worry.

However, the core functionality will require at least numpy, scipy, and numba to be on board.  Some relevant discussion threads:

- Numpy: https://github.com/numpy/numpy/issues/21135
- Scipy: https://github.com/scipy/scipy/issues/18286
- Numba: https://github.com/numba/numba/issues/9164

So if I understand the state of things (at present) correctly, we're talking about at least numpy 2.0, and probably a ways after that while midstream libraries (scipy, numba) catch up.

**Not much**

On the other hand, the work we would have to do once our upstream dependencies are in place ought to be rather superficial.  In most cases, I expect it to involve swapping object methods (`x.sum()`) for package functions (`np.sum(x)`) or similar trivial changes.

There will also have to be some modifications to our input validators and type checkers, obviously.

**Some caveats**

Things that could cause difficulty include:

- Sparse arrays.  For the sake of speed, we have a few spots where we bypass the scipy API and operate directly on the underlying data structures for things like index manipulation (eg util.__shear_sparse).  These might cause trouble for us.
- Type hints: could obviously get ugly.  Uglier.
- String data: string dtypes are technically [out of scope for the array api](https://data-apis.org/array-api/latest/purpose_and_scope.html#scope-includes-out-of-scope-non-goals).  We use these extensively in unit conversion and music notation as this was the cleanest way to get dimension- and type-preserving implementations via numpy.vectorize.  It's not clear yet (to me) how this will shake out with an abstracted array API backend.

---

[...]

Issue #1480, opened 2022-05-07
Parallel jit for faster numba functions


bmcfee:
It's possible that some (but not all) of our numba-accelerated functions could be further optimized by enabling `parallel=True` in the numba jit call.

Some good candidates here might be:

- overlap-add: https://github.com/librosa/librosa/blob/a3095313a2ba9e37e24f6180c4decc77cd6a2b0d/librosa/core/spectrum.py#L455-L464
- feature stacking: https://github.com/librosa/librosa/blob/a3095313a2ba9e37e24f6180c4decc77cd6a2b0d/librosa/feature/utils.py#L266-L303
- interval matching (with a bit of a rewrite): https://github.com/librosa/librosa/blob/a3095313a2ba9e37e24f6180c4decc77cd6a2b0d/librosa/util/matching.py#L47-L109
- dense shear: https://github.com/librosa/librosa/blob/a3095313a2ba9e37e24f6180c4decc77cd6a2b0d/librosa/util/utils.py#L1879-L1893
- window sum-square: https://github.com/librosa/librosa/blob/a3095313a2ba9e37e24f6180c4decc77cd6a2b0d/librosa/filters.py#L1456-L1463

Each of these should be benchmarked with and without parallelism before we decide to change anything, but they all seem like they have some room for improvement.

I don't expect any improvements to the remaining numba code, eg in sequence modeling (viterbi or DTW).

bmcfee:
A couple of notes:

1. We should probably have a mechanism to disable parallelism.  (Helpful for debugging, or if folks are implementing parallelism on their own at a higher level, eg batch processing of tracks.)  I think this might be best accomplished by an environment variable so that parallel= can be set at import time.
2. When resampy 0.3 is released, we should pass through the parallel= flag to make use of its own parallel compilation https://github.com/bmcfee/resampy/pull/93
3. There might be some bad interactions between numba caching and parallelism.  We'll have to investigate.

bmcfee:
> There might be some bad interactions between numba caching and parallelism. We'll have to investigate.

This turns out to be a known issue in numba caching - the hash does not depend on the build parameters, so `jit(cache=True, parallel=False)` collides with `jit(cache=True, parallel=True)`.  Until this is resolved, we can't implement run-time parallel/serial selection.  Implementing this feature will certainly require a version bump for numba.

PR #1985, opened 2025-10-16
Allow alternative viterbi implementations for pyin

xavriley:
#### Reference Issue
Partially addresses #1798

As noted in the above issue, viterbi decoding is the slowest operation for f0 tracking with pyin.

This commit is a fairly minimal change which moves the `sequence.viterbi` call to a parameter, allowing for the possibility of alternative viterbi decoding methods. The intended use case for this at the moment is [torbi](https://github.com/maxrmorrison/torbi) which is a new viterbi implementation optimised for GPU and multithreading.

#### Any other comments?

Example benchmark:

[Example audio: deantown.mp3](https://www.dropbox.com/scl/fi/kruyvp9lbhvg0yda4ekxo/deantown.mp3?rlkey=hzvd73o7vwrb0a74y384bcl1p&st=sedda7rc&dl=0)

```
# clone this repo and branch and cd into librosa folder
uv venv --python 3.11
uv pip install -e .
uv pip install git+https://github.com/maxrmorrison/torbi.git matplotlib
```

```python
import librosa
import timeit
import torch
import torbi
import numpy as np

def torbi_single_thread(observation_probs, transition, p_init):
    return torbi.from_probabilities(torch.from_numpy(observation_probs.transpose(0, 2, 1).astype(np.float32)), 
        transition=torch.from_numpy(transition.astype(np.float32)), 
        initial=torch.from_numpy(p_init.astype(np.float32)),
        num_threads=1,
        log_probs=False).numpy()

def torbi_multithreaded(observation_probs, transition, p_init):
    return torbi.from_probabilities(torch.from_numpy(observation_probs.transpose(0, 2, 1).astype(np.float32)), 
        transition=torch.from_numpy(transition.astype(np.float32)), 
        initial=torch.from_numpy(p_init.astype(np.float32)),
        num_threads=16,
        log_probs=False).numpy()

def torbi_gpu(observation_probs, transition, p_init):
    if not torch.cuda.is_available():
        raise ValueError("CUDA is not available")
    
    device = torch.device("cuda:0")
    return torbi.from_probabilities(
        torch.from_numpy(observation_probs.transpose(0, 2, 1).astype(np.float32)).to(device), 
        transition=torch.from_numpy(transition.astype(np.float32)).to(device), 
        initial=torch.from_numpy(p_init.astype(np.float32)).to(device),
        gpu=0,
        log_probs=False).cpu().numpy()

n_fft = 2048
hop_length = 512

durations = [10, 30, 60, 120, 240]
methods = [librosa.sequence.viterbi, torbi_single_thread, torbi_multithreaded, torbi_gpu]

for method in methods:
    timings = []

    for duration in durations:
        y, sr = librosa.load('deantown.mp3', duration=duration)
        pyin_time = timeit.timeit(
            lambda: librosa.pyin(
                y,
                fmin=librosa.note_to_hz('E0'),
                fmax=librosa.note_to_hz('C6'),
                frame_length=n_fft,
                hop_length=hop_length,
                viterbi_function=method
            ),
            number=1  # Run once for this benchmark
        )
        timings.append(pyin_time)
        print(f"{method.__name__} timeit result for {duration} seconds: {pyin_time:.4f} seconds")
    
    plt.plot(durations, timings, label=method.__name__)

plt.legend()
plt.xlabel('Input Duration (s)')
plt.ylabel('Execution Time (s)')
plt.title('pyin timing comparison')
plt.savefig('pyin_timing_comparison.png')
plt.show()
```

image

| Method | 10s | 30s | 60s | 120s | 240s |
|--------|-----|-----|-----|------|------|
| **librosa.sequence.viterbi** | 2.9153 | 8.2866 | 15.9120 | 32.8650 | 71.5760 |
| **torbi_single_thread** | 3.9254 | 11.7587 | 23.3271 | 46.8374 | 85.3265 |
| **torbi_multithreaded (num_threads=16)** | 0.9531 | 2.7089 | 5.2896 | 10.6621 | 19.3450 |
| **torbi_gpu** | 0.7770 | 0.7975 | 1.5992 | 3.2255 | 6.0070 |

The speed up with GPU is quite dramatic (~10x), but even with CPU and multithreading its a lot faster (70s -> 20s,  3.5x).

### DIFF
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -668,6 +668,7 @@ def pyin(
     fill_na: Optional[float] = np.nan,
     center: bool = True,
     pad_mode: _PadMode = "constant",
+    viterbi_function: Callable = sequence.viterbi,
 ) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
     """Fundamental frequency (F0) estimation using probabilistic YIN (pYIN).
 
@@ -750,6 +751,10 @@ def pyin(
         ``y`` is padded on both sides with zeros.
         If ``center=False``,  this argument is ignored.
         .. see also:: `np.pad`
+
+    viterbi_function : Callable
+        Function to use for Viterbi decoding.
+        Defaults to ``librosa.sequence.viterbi``.
 
     win_length : Deprecated
         length of the window for calculating autocorrelation in samples.
@@ -880,7 +885,7 @@ def pyin(
 
     p_init = np.ones(2 * n_pitch_bins) / (2 * n_pitch_bins)
 
-    states = sequence.viterbi(observation_probs, transition, p_init=p_init)
+    states = viterbi_function(observation_probs, transition, p_init=p_init)
 
     # Find f0 corresponding to each decoded pitch bin.
     freqs = fmin * 2 ** (np.arange(n_pitch_bins) / (12 * n_bins_per_semitone))


xavriley:
p.s. If anyone tries to run this on colab, torbi now has wheels available which resolve the CUDA mismatch errors when installing directly from git, so `pip install torbi` will work.

bmcfee:
Thanks for this - I definitely see the motivation for this kind of improvement.

That said, it's not obvious to me that we should merge PRs of this nature.  (Maybe we should, but it's a point for discussion.)

Here's how I'm thinking of it:

- This particular change is relatively small, but I think we should try to be systematic in how we handle modular functionality like this.  Examples in the past where we tried to do this sort of thing include audio decoding (yikes), sample rate conversion (less bad, but still yikes), and fft implementations (now thoroughly expunged).
- Especially in niche cases like viterbi decoding where there isn't necessarily a consistent API to rely on, we'll end up needing to support a bunch of shims (as in sample rate conversion).  This pretty quickly leads to a lot of unmaintainable code that I think we can avoid.
- Bigger picture, part of the roadmap for us is to extend out to be fully array-api compliant.  This is a long-term plan though (see #1743), as we're blocked by upstream dependencies (eg numba) getting on board first.  However, I do think this is the "correct" approach, in that it should in principle mean that we have fewer competing modules for the same functionality, and more consistent behavior.

That said, one other thing I've been thinking about lately is the possibility of registering hooks for extensions.  I haven't fully fleshed this out yet, but if we can set things up so that torbi (for example) could provide the librosa shim logic and register itself as an extension, then this would simplify the maintainability from our side, and still make it easy for a user to plug in their own backend code at designated points.

I haven't drafted up an issue for this idea yet, but I have been slowly iterating on the idea offline, and there are a number of reasonable use cases for it.  (I had it in mind for, say, v1.1 or 1.2.)  I could prioritize drafting this up though if it would be a more productive discussion now.

bmcfee:
Outside of the above, I'm a little curious why a multi-threaded implementation has such a boost here.  I expect that this gap reduces (vanishes?) when the state space (fmax-fmin) is sufficiently small, and almost all of the gain here is coming from parallelizing over the previous state search in each iteration:
https://github.com/librosa/librosa/blob/e403272fc984bc4aeb316e5f15899042224bb9fe/librosa/sequence.py#L1146-L1150

In principle, numba should be smart enough to dispatch this well for us, but perhaps it isn't?  We at some point had parallel dispatch enabled on jit compiled functions like `_viterbi`, but eventually removed it due to inconsistent efficiency gains and conflicts with byte object caching.  See #1480, but the caching issues is still open for numba and blocking run-time selection of single vs multithreaded jit execution (which would be the right thing to do IMO).

Would it be possible to rerun the above example with the _viterbi jit compilation line modified to include `parallel=True` ?  I don't expect to see it catch up to the gpu implementation necessarily, but we should be able to narrow the gap to the multithreaded cpu version.  (It's kind of mysterious that a gpu version of this would benefit as well, but I again chalk that up to parallelizing the argmax within each iteration.)

Additionally, the benchmark code you have might not be totally fair, because it could be including the first pass jit compilation on librosa's viterbi within the time measurement.  A more fair implementation would run each algorithm multiple times (3 is probably enough here) and report the best case.

xavriley:
Thanks for the detailed consideration! My apologies for the delay - I went on vacation just after posting this.

> the benchmark code you have might not be totally fair, because it could be including the first pass jit compilation on librosa's viterbi within the time measurement

Good point. Running this again and taking the best of 3 it looks like `librosa.sequence.viterbi` improves a bit on the longest run but its not conclusive:

image

```
# librosa viterbi timings
viterbi timeit result for 10 seconds: 3.0610 seconds
viterbi timeit result for 30 seconds: 8.7436 seconds
viterbi timeit result for 60 seconds: 17.1353 seconds
viterbi timeit result for 120 seconds: 34.6463 seconds
viterbi timeit result for 240 seconds: 59.4177 seconds # vs 71 seconds before
```

```
# updated benchmark code - uses timeit.repeat
    for duration in durations:
        y, sr = librosa.load('deantown.mp3', duration=duration)
        pyin_time = min(timeit.repeat(
            lambda: librosa.pyin(
                y,
                fmin=librosa.note_to_hz('E0'),
                fmax=librosa.note_to_hz('C6'),
                frame_length=n_fft,
                hop_length=hop_length,
                viterbi_function=method
            ),
            repeat=3,
            number=1
        )) # best of 3 runs
        timings.append(pyin_time)
        print(f"{method.__name__} timeit result for {duration} seconds: {pyin_time:.4f} seconds")
```

> Would it be possible to rerun the above example with the _viterbi jit compilation line modified to include `parallel=True`?

On my first attempt this seemed to run around 2x slower for the same code above:

```
viterbi timeit result for 10 seconds: 5.5906 seconds
viterbi timeit result for 30 seconds: 17.3912 seconds
viterbi timeit result for 60 seconds: 36.1492 seconds
viterbi timeit result for 120 seconds: 70.1659 seconds
viterbi timeit result for 240 seconds: 123.1444 seconds
```

I added some debugging code for numba (code for that is [here](https://gist.github.com/xavriley/e28cee328cb6bda7697bd83fbc67b92c)) and it looks like numba is only attempting to parallelize the array operations and isn't touching the `for ... in range` statements. I'm not familiar with this but Claude seems to think this is expected behaviour?

After changing the inner loop to use `numba.prange` (https://numba.pydata.org/numba-doc/0.11/prange.html) instead the expected improvements are realised:

image

```
viterbi timeit result for 10 seconds: 0.6644 seconds
viterbi timeit result for 30 seconds: 2.5514 seconds
viterbi timeit result for 60 seconds: 4.1658 seconds
viterbi timeit result for 120 seconds: 8.4040 seconds
viterbi timeit result for 240 seconds: 16.2225 seconds
```

Not bad for a single character change! 😅 I'm not sure what the implications are if the implementation switches to use `prange` by default though.

xavriley:
Specifically the change was here: https://github.com/librosa/librosa/blob/e403272fc984bc4aeb316e5f15899042224bb9fe/librosa/sequence.py#L1147

`for j in range(n_states):`

becomes

```
from numba import jit, prange

...

for j in prange(n_states):
```

bmcfee:
Thanks for putting in the extra benchmarking work!

> I'm not sure what the implications are if the implementation switches to use `prange` by default though.

Right - this is necessary for `parallel=True` to do anything useful to parallelize the loop.

As I noted above, we don't enable parallel jit compilation by default because the speedups were inconsistent across platforms the last time we dug into this.  There's an additional hiccup here where the byte object cache doesn't track compile-time parameters in the hash, so dynamically switching parallel on or off via an environment variable (my personal preference for these things) isn't safe with `cache=True`.  The end result here is that it would only work with caching disabled, so you'd pay jit cost on every startup.  Far from ideal.

I don't think it's harmful to switch it to prange, but without `parallel=True` at compile time, it's unlikely to make any difference.

xavriley:
No worries, thanks again for the explanation.

I spent a bit more time with this to see if I could address the cache vs parallel issue and didn't have any joy. The only thing that worked was to duplicate the code for `_viterbi` into functions for `_viterbi_serial` and `_viterbi_parallel` and then select between them based on an env var at runtime. Keeping the method names distinct prevents them occupying the same space in the cache and avoids potential conflicts.

On the one hand, this approach is obviously super messy but on the other hand there are only a [handful of functions](https://github.com/librosa/librosa/issues/1480) that would benefit from parallel optimizations anyway. Would it be reasonable to implement these dual path functions as a stop gap solution while we wait for progress on the numba side? I wouldn't normally push for this, but the perf improvements for these are so substantial (approx 4x faster for pyin) that it seems a shame not to take them. Would it be worth me making a PR?

bmcfee:
> The only thing that worked was to duplicate the code for `_viterbi` into functions for `_viterbi_serial` and `_viterbi_parallel` and then select between them based on an env var at runtime. Keeping the method names distinct prevents them occupying the same space in the cache and avoids potential conflicts.

Yeah... this is where I wound up with this as well.  I'd really rather not have duplicated code like this, as it easily becomes a maintenance nightmare.  At one point I also looked into trying to keep separate jit objects derived from the same code, but there wasn't an obvious way to do this while maintaining caching.



> Would it be reasonable to implement these dual path functions as a stop gap solution while we wait for progress on the numba side?

My hunch here is that it would not be much more effort to just fix it from the numba side.  I haven't had the bandwidth to do so myself, and it seems like nobody else on the numba team has this as a priority, but it doesn't seem like a hard problem to solve.  Relevant issue is here: https://github.com/numba/numba/issues/6845 - maybe @stuartarchibald or @gmarkall has some insight about the status of that issue 3 years on?

bmcfee:
Coming back to this - I'm not sure if we ever properly resolved this discussion, but I think I'm inclined to close the PR as currently implemented.

I'm still open to revisiting the prange approach, but that should probably be done more systematically across the package.

Does that sound reasonable @xavriley ?

gmarkall:
I failed to spot the previous comment, so apologies for the delay in my reply here. Although I was able to do a quick fix on the Numba side, it wasn't something I was able to pick up and work into a suitable form for a PR, so the issue hasn't really moved on from there.

bmcfee:
Thanks @gmarkall !  That's what I figured, but good to know.

bmcfee:
@gmarkall i was thinking on this a little yesterday, and I might have a not-too-horrible hack that would do the job here without modifications to numba.  Maybe you could give a 5-second think and let me know if it's completely bonkers?

- Problem: we want both parallel and serial versions of the same function.  We want to cache jit-compiled objects.  We do not want to duplicate code.
- The snag: caching is keyed by bytecode without jit parameters, so we'd get collisions.
- Proposed fix: do some functionally trivial manipulation to a copy of the `__code__` field of the function to be jit-compiled.  This should prevent the cache keys from colliding, and allow us to have two cached versions (one parallel, one not) that can be dynamically selected.

I think in principle this works, but it seems like it could be delicate to get right depending on compiler optimizations potentially sifting out the trivial modifications we might insert.

So is this idea worth pursuing at all?

gmarkall:
It's a good idea... I wonder if the same result can be achieved more simply though, with something like:

```python
@njit
def implementation(<args>):
    for i in prange(...):
        ...

@njit(cache=True, parallel=True)
def parallel_version(<args>):
    implementation(<args>)

@njit(cache=True, parallel=False)
def serial_version(<args>):
    implementation(<args>)
```

In this implementation I'd expect `implementation` to always get inlined into `parallel_version` and `serial_version` (so no function call overhead) but should cache differently based on the line number / function name.

Could this work and be a little less "magic"?

bmcfee:
> Could this work and be a little less "magic"?

Hrm, I tried this (numba=0.63.1) and the jit compiler failed to identify any parallelizable transformations, probably because the underlying implementation function was already jitted in advance.  If `implementation` is not jitted, it fails to compile outright.

gmarkall:
Ah, I think what you observed there is correct.

I think it might be possible to get something like your idea above to work, but I suspect it would be tricky and fragile in ways I can't anticipate. 

FWIW my last attempt to make something do the right thing was:

```python
from numba import njit, prange
import numpy as np
import time

N = 10**8

def vector_math(r, x, y):
    for i in prange(len(r)):
        r[i] = np.cos(x[i]) + np.sin(y[i])

parallel_core = njit(parallel=True)(vector_math)
serial_core = njit(vector_math)


@njit(cache=True, parallel=True)
def parallel_version(r, x, y):
    parallel_core(r, x, y)


@njit(cache=True, parallel=False)
def serial_version(r, x, y):
    serial_core(r, x, y)


def test(version):
    np.random.seed(1)
    x = np.random.random(N)
    y = np.random.random(N)

    r = np.zeros_like(x)

    # Warm up: includes compile / cache load
    warmup_start = time.perf_counter()
    version(r, x, y)
    warmup_end = time.perf_counter()

    r = np.zeros_like(x)

    # Execution
    execution_start = time.perf_counter()
    version(r, x, y)
    execution_end = time.perf_counter()

    return (warmup_end - warmup_start, execution_end - execution_start)

parallel_warmup, parallel_run = test(parallel_version)
serial_warmup, serial_run = test(serial_version)

print("Warmup times:")
print(f"Parallel {parallel_warmup}")
print(f"Serial {serial_warmup}")
print("Execution times:")
print(f"Parallel {parallel_run}")
print(f"Serial {serial_run}")
```

For the first run here I get

```
Warmup times:
Parallel 0.36560620900127105
Serial 0.8993439679907169
Execution times:
Parallel 0.10669100799714215
Serial 0.832208043007995
```

then subsequent runs give

```
Warmup times:
Parallel 0.22273038000275847
Serial 0.8614024600101402
Execution times:
Parallel 0.10083083200152032
Serial 0.8643852300010622
```

The times are fairly consistent - the time-to-first run of the parallel implementation is reduced by a lot more than the serial implementation. So it could be that the cache is working for both versions, but the parallel compilation is a lot more time consuming than the serial compilation.

bmcfee:
Interesting.  I hadn't thought about just duplicating the inner compile with cache disabled :smile: 

It does appear to be working, with a modified version of @xavriley 's benchmark above I see about a 3x speedup on my machine, which I think is qualitatively commensurate with his reported results for multithread+cpu:

```
parallel=True timeit result for 10 seconds: 1.4998 seconds
parallel=True timeit result for 30 seconds: 3.7949 seconds
parallel=True timeit result for 60 seconds: 7.4318 seconds
parallel=True timeit result for 120 seconds: 14.4339 seconds
parallel=True timeit result for 240 seconds: 26.5637 seconds

parallel=False timeit result for 10 seconds: 3.9604 seconds
parallel=False timeit result for 30 seconds: 10.4949 seconds
parallel=False timeit result for 60 seconds: 21.3461 seconds
parallel=False timeit result for 120 seconds: 47.4239 seconds
parallel=False timeit result for 240 seconds: 88.7744 seconds
```
image

*EDIT*: disregard below.  I Think this was just jupyter + autoreload messing with things.  On a fresh kernel I'm not seeing this warning.
---
I do get a weird warning though:
```
The keyword argument 'parallel=True' was specified but no transformation for parallel execution was possible.

To find out why, try turning on parallel diagnostics, see https://numba.readthedocs.io/en/stable/user/parallel.html#diagnostics for help.

File "../../git/librosa/librosa/sequence.py", line 1260:
@jit(cache=True, parallel=False)
def _viterbi_serial(log_prob, log_trans, log_p_init):
^
```
weird because `parallel=False` here, not True as claimed in the warning?

bmcfee:
Ok, I've played around with this a bit, and I think this actually does provide a workable solution to #1480 .  I think at this point it makes sense to close this PR out, and open a new one specifically to address exposing parallel controls on certain functions (chiefly pyin / viterbi, but maybe a couple of others).

Thanks @xavriley for the nudging on this point, and @gmarkall for getting it unstuck!

Issue #1909, opened 2025-02-08
Metrogram Transform for Time Signature Detection


danferns:
Hey, first of all thank you for maintaining this very useful library.

I'm working with my team (@ethan89009, @ValerianCoelho) on our college project for time signature (meter) detection. We are planning to use the Metrogram Transform introduced in [this paper](https://ieeexplore.ieee.org/document/10363392). 

In short, it looks for rhythmic ratios in the tempogram, I found that [`librosa.feature.tempogram_ratio`](https://librosa.org/doc/main/generated/librosa.feature.tempogram_ratio.html) does something similar. The major difference is the Metrogram Transform doesn't require the tempo of the track. It works even if the tempo changes with time. 

It would be nice to receive some pointers on code from librosa that we might be able to reuse or draw from when implementing this transform. A lot of functions in librosa are using more basic functions (eg. autocorrelation) from the library instead of coding the math from scratch. Which is neat. We will also try figuring this out.

Would this transform be a good fit for a PR to librosa? There was some related discussion in #194.

bmcfee:
Thanks for the proposal.  I'm reading through the paper now, and here are my thoughts about what a librosa implementation would look like:

1. Everything up to equation 7 is essentially onset strength envelope calculation (`D(t)` in the paper's notation).  I would separate functionality at this point, and implement a method that operates on any precomputed onset strength envelope.
2. Equation 8 looks a bit like a fourier (power) tempogram, except using a wavelet basis instead of a DFT.
3. Equation 11 looks like an adapted version of the Peeters 2011 idea of DFT/AC product.  This is essentially a cleanup step on a tempogram representation to suppress subharmonics.  It might be nice to have a function to implement just this step, but I would keep it separate from the metrogram idea.
4. The key equation here is 12, which is very similar to our `salience` function, which first uses `interp_harmonics` and then aggregates out the harmonic dimension.  Metrogram basically does the same thing, with two differences: 1) the frequency (tempo) dimension is aggregated out, not the harmonic dimension, and 2) the aggregation is a sum-product against the fundamental.  (1) is easy enough to achieve by changing the target axis parameter.  (2) is a little more delicate, but you could achieve it easily by doing something like the following:

```python
# Assume T is the tempogram we begin with, and k are the harmonics/ratios we're interested in

T_interp = interp_harmonics(T, freqs=tempogram_freqs, harmonics=k, ...)  # and other parameters as needed
[...]

PR #1986, opened 2025-10-20
Add Metrogram Transform

danferns:
#### Reference Issue
Reference #1909

This PR implements the Metrogram Transform by adding the following functions:

1. `metrogram`: This function summarizes the presence of rhythmic ratios in a tempogram. For example, a tempogram with two simultaneous energy peaks at 90BPM and 30BPM would have a strong presence of the 1/3 ratio. This makes it possible to perform meter estimation by finding the ratio between the beat’s and downbeat’s frequency.
2. `interp_broadcast`: Interpolates two arrays along a given axis to a common grid, and performs a broadcast operation (eg. np.multiply) to combine them. It is used to create a Fundamental Tempogram that is supplied to `metrogram`. 

Documentation, examples, and tests have been added for both.


bmcfee:
Dropping a note to indicate that I see this PR and will get to review it as soon as I have time.

Meanwhile: would it be possible for you to re-target this PR against the 1.0 dev branch?  I don't expect there to be much in the way of conflicts here, but the plan right now is to prioritize 1.0 above a point release on 0.11.

danferns:
Feel free to take your time! :smile: 

I've switched the target branch and got no effective merge conflicts.

librosa/util/utils.py
```diff
@@ -2683,3 +2696,118 @@ def phasor(
         z *= mag
 
     return z  # type: ignore
+
+
+def interp_broadcast(
+    x1: np.ndarray,
+    x2: np.ndarray,
+    x1_pos: np.ndarray,
+    x2_pos: np.ndarray,
```
bmcfee, line 2724: Some minor quibbles on this one:

1. It probably makes sense for x1 and x1_pos to go together, ditto for x2 / x2_pos.
2. Can we * everything after to force keywording?
3. Is there a better name than `func` that we can use here?
4. Probably there should be a check on `axis` to ensure that it's valid for the input array ndim.

librosa/util/utils.py
```diff
@@ -2683,3 +2696,118 @@ def phasor(
...
+
+    Returns
+    -------
+    result : np.ndarray or (np.ndarray, np.ndarray)
+        The result from combining both arrays after interpolation.
+        If ``func`` is set to ``None``, returns the interpolated arrays separately ``(y1, y2)``.
+
+    Examples
+    --------
+
```
bmcfee, line 2803: Please add usage examples

librosa/feature/rhythm.py
```diff
@@ -655,3 +655,91 @@ def tempogram_ratio(
...
+        factors = np.array(
+            [1 / 3, 1 / 4, 1 / 5, 1 / 7]
+        )
+
+    tg_interp = interp_harmonics(
+        tg, freqs=freqs, harmonics=factors, kind=kind, fill_value=fill_value
+    )
+
+    
+    product = tg_interp * tg[np.newaxis, ...]
```
bmcfee: is this the right place for newaxis in general?  It seems like it should be at position to match the target axis, which might be more readily accomplished with the `expand_dims` helper.

Also, isn't this product what `func` is for in `interp_broadcast`?  It seems like that code isn't being used.

bmcfee (changes requested):
This is a good start!  We're missing test coverage and examples in the docstrings.  I've left a few comments in places that look like they could use some refinement.  Otherwise, please also fix the issues identified by the linter action.

librosa/util/utils.py
danferns: I've addressed 1, 2 and 4.

- [x] 3. How about `op` for broadcast operator / operation?

librosa/feature/rhythm.py
danferns: Thank you! I've switched to `expand_dims` ~~and enabled the `axis` parameter~~. 

> EDIT: While writing examples, I realized this axis isn't intended to be user configurable. Using expand_dims at -3 is working for mono and stereo audio in some rough testing.

As I understand, `func` in `interp_broadcast` extracts the overlap between Fourier/AC tempograms. The tg product also extracts overlap, but between the tempogram and slowed-down versions of itself.

I was intending to let the user create the tempogram, and show the use of `interp_broadcast` through an example. But we can also provide a default pathway which accepts `y` and returns the metrogram. Should we add this?

danferns:
Thank you for the review! I've addressed most of the minor comments. I'll work on the examples next. 

Also, James Cozens (who authored this method) found us working on this PR! He has kindly offered his help if needed and will be sharing a preprint with more details on his implementation.

librosa/util/utils.py
danferns: Have addressed all these points now, thank you!

librosa/feature/rhythm.py
```diff
@@ -655,3 +655,119 @@ def tempogram_ratio(
...
+    >>> # extend the window, to capture the slower downbeat pulses
+    >>> win_length = 384 * 4
+    >>>
+    >>> fourier_tempogram = librosa.feature.fourier_tempogram(y=y, win_length=win_length)
+    >>> fourier_freqs = librosa.fourier_tempo_frequencies(win_length=win_length)
+    >>> ac_tempogram = librosa.feature.tempogram(y=y, win_length=win_length)
+    >>> ac_freqs = librosa.tempo_frequencies(ac_tempogram.shape[-2])
+    >>>
+    >>> # combine Fourier and AC tempo grid, and remove inf to avoid nan results
+    >>> funt_freqs = np.union1d(fourier_freqs, ac_freqs)[:-1]
```
danferns: I tried combining the Fourier and AC frequency axes, so we don't need to choose between them and can preserve more information. I found it gives more reliable results on metronome tracks. 

But now, `funt_freqs` becomes a hybrid frequency axis. I'm not sure how to use arbitrary frequency axes with specshow. For this reason, I've only plotted the final metrogram in the example.

danferns:
I think we're good for a final review now!

bmcfee:
Thanks @danferns !  Sorry for the lag between approving test runs.

Overall I think this is looking good.

* I'm not going to fuss over type annotations right now for this, especially for the literals on interpolation modes.  There's an open PR to deal with that stuff at the package level, and I expect this will merge before that one does so we can just roll it in there.
* We are still missing some code coverage on the tests, looks like just the bounds check on `interp_broadcast`.  If you can add a test to hit that case, I think we'll be good to go.

On the previous comments, I guess I had been assuming that the metrogram call would be wrapping the interp_broadcast call for us, but upon reflection, I think it's cleaner to keep them separate as you've done here.  I would like to add some more documentation to highlight how `interp_broadcast` can be used though, as I think it's generally helpful.  We can make it a separate issue, and maybe add a new feature function that does hybrid tempogram a la (Peeters 2005).  At that point, we may want to revisit the documentation on metrogram to use the hybrid tempogram directly instead of two-stepping it manually, but this shouldn't hold us up on this PR.

Regarding the union frequency axis, it's an interesting idea, as it essentially replaces the interpolation calculations with exact (but ordered) lookups.  I guess the downside here is that it makes visualization of the hybrid tempogram really wonky, since we no longer have a nicely parametrized frequency grid to work with.

danferns:
Thanks for the review! The tests should be covered now, and I've slightly modified the metrogram docs to clarify that a single (non-hybrid) axis can also be used, as a hint for anyone wanting to visualize the process.

I like the idea of a dedicated product spectrum function.

bmcfee (approved):

bmcfee:
@danferns I just threw in one last copilot check to identify any issues that I missed previously, and it looks like there are a couple of very small things here.  Can you take a look and let me know what you think?

librosa/feature/rhythm.py
```diff
@@ -655,3 +655,119 @@ def tempogram_ratio(
...
+    >>> fourier_freqs = librosa.fourier_tempo_frequencies(win_length=win_length)
+    >>> ac_tempogram = librosa.feature.tempogram(y=y, win_length=win_length)
+    >>> ac_freqs = librosa.tempo_frequencies(ac_tempogram.shape[-2])
+    >>>
+    >>> # combine Fourier and AC tempo grid (alternatively, you may use either one)
+    >>> # we remove np.inf from ac_freqs to avoid nan results
+    >>> funt_freqs = np.union1d(fourier_freqs, ac_freqs[1:])
+    >>> fundamental_tempogram = librosa.util.interp_broadcast(
+    ...     x1=ac_tempogram,
+    ...     x1_pos=ac_freqs,
```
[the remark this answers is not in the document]
danferns: Not an issue: `interp_pos` is independent of the inputs. `ac_freqs[0]` is `np.inf`, which results in nans when present in `interp_pos`, but its presence in `x1_pos` does not cause this problem.

librosa/util/utils.py
```diff
@@ -2683,3 +2684,171 @@ def phasor(
...
+    --------
+    librosa.feature.metrogram
+    """
+    if interp_pos is None:
+        interp_pos = x1_pos
+
+    if x1.ndim != x2.ndim:
+        raise ParameterError(
+            f"x1 (ndim={x1.ndim}) and x2 (ndim={x2.ndim}) have a different number of dimensions."
+        )
```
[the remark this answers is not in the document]
danferns: I assumed broadcast compatible meant both arrays must have the same dimensions. But from [numpy docs](https://numpy.org/doc/2.1/user/basics.broadcasting.html#general-broadcasting-rules):

> Two dimensions are compatible when
    1. they are equal, or
    2. one of them is 1.

and

> Input arrays do not need to have the same number of dimensions. 

It is possible that interp_broadcast is able to work with a wider range of inputs than is currently allowed. I'll need some time to try this out...

danferns:
Thank you! I've applied the smaller suggestions. The comment about broadcast compatibility seems to suggest that interp_broadcast is rejecting some inputs that it can actually handle. I will look into this and get back to you!

librosa/util/utils.py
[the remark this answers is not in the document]
bmcfee: This is actually pretty straightforward and will likely simplify the code below.

You can use `np.broadcast_shapes` to check if two array shapes are compatible.  (If they aren't, it will throw a value error.)

Really though, the broadcast compatibility is only needed if `op is not None`.

For example, you could do something like:

```python
x1_interp = ...
x2_interp = ...

y1 = x1_interp(interp_pos)
y2 = x2_interp(interp_pos)


if op is None:
    return y1, y2
try:
   np.broadcast_shapes(y1.shape, y2.shape)
except ValueError as exc:
   raise ParameterError("something informative about shape mismatch") from exc

return op(y1, y2)
```

You could in principle instead check shape compatibility of everything except the target interpolation axis on the input arrays (which will be fixed anyway by interpolating to a common grid before broadcast).  I guess that would be marginally more efficient, but slightly fussier to implement.  Probably the compatibility check is unnecessary anyway because if they're incompatible, it will fail at the `op(y1, y2)` line anyway; but it's nice to have an interpretable error message.

danferns:
I find the solution to just check y1 and y2 very elegant. I've updated the tests to also follow numpy's broadcasting rules. All Copilot comments should be resolved now!

bmcfee:
LGTM!  Thanks for all your work on this - happy to merge now.

danferns:
Great! Happy to see this merged!

Thank you for the reviews and all the help along the way. This was a great learning experience!

PR #1992, opened 2026-01-29
Fix #1970: Prevent AdaptiveWaveplot adaptor from being garbage collected

### LINKED ISSUE #1970: [Bug]: Handling multiple adaptive waveshows
### Bug summary

As noted in #1370, there is currently a problem with matplotlib's event callback framework that prevents multiple attachments to the same event.  This makes it impossible for multiple waveshows on the same axes object (or different axes in the same figure) to properly update when the viewport is changed, eg in interactive mode.

### Code for reproduction

```Python

fig, ax = plt.subplots()

# Plot 1
librosa.display.waveshow(y=y, sr=sr, ax=ax)

# Plot 2
librosa.display.waveshow(y=y, sr=sr, ax=ax, zorder=-1)
ax.set(xlim=[0, 0.025])
plt.show()
```

### Actual outcome

Display stays in envelope view.

### Expected outcome

It should switch to sample view.

### Additional information

I think for now the only way around this is to build our own dispatch container.  This can be done using weakrefs so that handles don't live forever.  I have some prior experience with this in mir_eval lately, so it ought to work.

### Operating system

all os's

### software versions

all versions

### Installation

git checkout

Daniel-Haas-B:
#### Reference Issue
Fixes #1970 

TLDR:
This PR fixes the underlying cause of "stuck envelope view" by retaining adaptors per Axes even if the caller ignores the return value. It also avoids turning that retention into a memory leak by breaking strong adaptor→artist→axes references using weakrefs inside `AdaptiveWaveplot`. Finally, it adds tests that verify both the user-visible behavior (adaptivity survives GC) and the safety property (axes/figures remain collectible and the registry does not pin them).

- Add _WAVESHOW_ADAPTORS WeakKeyDictionary to retain adaptors per-Axes
- Store artists (steps, envelope) and axes as weakrefs in AdaptiveWaveplot
- Add properties to preserve existing attribute API
- Add 3 regression tests

I wished this was more of a surgical PR like we initially planned in #1970, but I had to touch some internal implementation details to avoid the memory leak. Let me know if there are other approaches or if what I did here isn't quite ok.

### 1) What #1970 originally suggested vs what the repro showed

Issue #1970 frames the problem as: "Matplotlib's callback framework prevents multiple attachments to the same event, so multiple adaptive waveshows cannot update correctly."

While exploring the repro, I found the "stuck in envelope mode after `set_xlim`" symptom can happen even with a _single_ `waveshow()` when the return value is not retained. That changed the likely root cause from "multiple callbacks don't attach" to "the adaptor object that owns the callback is getting garbage-collected."

`waveshow()` builds an `AdaptiveWaveplot` adaptor and registers its `update()` as a Matplotlib callback (`xlim_changed` / `ylim_changed`). The callback is a bound method (`self.update`), and Matplotlib stores bound-method callbacks using weak references. In other words, callback registration does not keep the adaptor alive.

So if the user calls:

```python
librosa.display.waveshow(...)
ax.set_xlim(0, 0.025)
```

without assigning the return value, the adaptor can be garbage-collected immediately after waveshow() returns. When that happens, the callback effectively disappears and no longer runs on subsequent xlim changes. The plot stays in whatever mode it initially chose.

### 2)The "minimal retention" plan from the thread (wasn't sufficient as-is)

In the thread, we discussed a "minimal fix" that sounded like it would be enough: keep a module-level `WeakKeyDictionary[Axes -> adaptors]` so adaptors remain alive, and rely on weak keys so entries vanish when the Axes is cleaned up.

I implemented that simple retention and it did fix the functional bug. The problem was that my first attempt at a cleanup test didn't actually test cleanup. I wrote a test that asserted we were using a `WeakKeyDictionary` and that the registry contained the Axes. That only proves "we stored something in a `WeakKeyDictionary`." It does not prove the Axes can become unreachable (which is the entire condition under which `WeakKeyDictionary` entries disappear).

So: `WeakKeyDictionary` semantics only help if the keys are collectible. If the retained values keep the key reachable, the entry will never be removed. Once I rewrote the test to check real reachability and forced GC, the naive retention approach failed.

### 3) Why the naive approach can pin Axes/Figures in memory

To make the GC test meaningful, I removed `pyplot` from the equation because `pyplot` can keep references alive via global state. I constructed figures/axes without `pyplot` (`Figure()` + `FigureCanvasAgg` + `add_subplot`) and explicitly severed the figure→axes relationship (`fig.delaxes(ax)` + `fig.clear()`) before deleting references and forcing `gc.collect()`.

Even in that setup, the Axes remained reachable with the naive retention approach. That strongly suggested the reachability was coming from our own object graph rather than `pyplot` internals.

This is consistent with how Matplotlib wires objects together. `waveshow()` creates two artists: a `Line2D` for the sample view (from `plt.step`) and a `PolyCollection` for the envelope view (from `plt.fill_between`). Matplotlib artists hold a strong reference back to their owning Axes via `.axes`. The original `AdaptiveWaveplot` held strong references to those artists (`self.steps`, `self.envelope`) and maybe also to the Axes itself (`self.ax`). Once we added global retention of the adaptor, that created a strong path from the registry's value back to the registry's key:

registry (value) → adaptor → artist(s) → Axes (key)

In that situation the Axes never becomes unreachable, so the `WeakKeyDictionary` entry never disappears. The registry does not behave like "weak retention" in practice, and it risks pinning figures/axes in memory.

### 4) What this PR does in more detail

I still retain adaptors per Axes, but I also changed `AdaptiveWaveplot` internals to avoid the value→key reachability problem.

`AdaptiveWaveplot` now stores weak references to the connected Axes and to the artists it manages:

`_ax_ref = weakref.ref(ax)`

`_steps_ref = weakref.ref(steps_artist)`

`_envelope_ref = weakref.ref(envelope_artist)`

To preserve compatibility, I added `@property` accessors so existing code can continue to treat these like normal attributes:

`ax` returns `self._ax_ref()` (or `None`)

`steps` returns `self._steps_ref()` (or `None`)

`envelope` returns `self._envelope_ref()` (or `None`)

Because weakrefs can return `None` if the object has been collected or removed, the relevant code paths (notably `update()` and `disconnect()`) now guard against `None`. This makes it safe for a global registry to retain adaptors without pinning the Axes through artist back-references.

### 4)Why the tests look how they look

First, I needed to prove the user-facing fix: adaptivity should survive even if the caller does not retain the return value. The tests `test_waveshow_adaptor_survives_gc_single` and `test_waveshow_adaptor_survives_gc_multi` call `waveshow()` without assignment, force `gc.collect()`, zoom in via `set_xlim`, draw, and then assert that sample lines are visible while envelope `PolyCollection` objects are hidden. They use `ax.lines` and `ax.collections` instead of `ax.get_children()` to avoid false positives from unrelated artists.

Second, I needed to prove the safety property: the registry should not pin Axes objects in memory. The test `test_waveshow_registry_cleanup_on_axes_gc` constructs figures/axes without `pyplot`, explicitly removes the axes from the figure, deletes references, forces GC, and verifies that the Axes is collected and that the registry no longer contains that Axes key. It compares key-id sets before and after rather than asserting the registry is empty, since other tests may populate it.

### DIFF
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -791,14 +791,17 @@ class AdaptiveWaveplot:
     # Preserve the old attribute API by exposing properties with same names
     @property
     def steps(self) -> Optional[Line2D]:
+        """The step plot artist (Line2D), or None if garbage collected."""
         return self._steps_ref()
 
     @property
     def envelope(self) -> Optional[PolyCollection]:
+        """The envelope artist (PolyCollection), or None if garbage collected."""
         return self._envelope_ref()
 
     @property
     def ax(self) -> Optional[mplaxes.Axes]:
+        """The connected Axes, or None if not connected or garbage collected."""
         return None if self._ax_ref is None else self._ax_ref()
 
     def __del__(self) -> None:
--- a/librosa/display.py
+++ b/librosa/display.py
@@ -45,6 +45,7 @@ from itertools import product
 import re
 import warnings
 from fractions import Fraction
+import weakref
 
 import numpy as np
 from matplotlib import colormaps as mcm
@@ -97,6 +98,11 @@ __all__ = [
 ]
 
 # mypy: disable-error-code="attr-defined"
+
+# Keeps adaptors alive as long as their Axes exists, preventing GC
+_WAVESHOW_ADAPTORS: weakref.WeakKeyDictionary[
+    mplaxes.Axes, set["AdaptiveWaveplot"]
+] = weakref.WeakKeyDictionary()
 
 # Nominal center frequencies for oct3 bands
 __OCT3_FREQUENCIES = np.array(
@@ -765,13 +771,13 @@ class AdaptiveWaveplot:
     ):
         self.times = times
         self.samples = y
-        self.steps = steps
-        self.envelope = envelope
+        self._steps_ref = weakref.ref(steps)
+        self._envelope_ref = weakref.ref(envelope)
         self.sr = sr
         self.max_samples = max_samples
         self.transpose = transpose
         self.cid: Optional[int] = None
-        self.ax: Optional[mplaxes.Axes] = None
+        self._ax_ref: Optional[weakref.ref[mplaxes.Axes]] = None
 
         # Only set the label on the patch if we have one to set
         kwargs = dict()
@@ -779,8 +785,21 @@ class AdaptiveWaveplot:
             kwargs["label"] = label
         # This creates an invisible patch to contain the label
         self.label_patch_ = mpatches.Rectangle(
-            (np.nan, np.nan), 0, 0, facecolor=self.steps.get_color(), **kwargs
+            (np.nan, np.nan), 0, 0, facecolor=steps.get_color(), **kwargs
         )
+
+    # Preserve the old attribute API by exposing properties with same names
+    @property
+    def steps(self) -> Optional[Line2D]:
+        return self._steps_ref()
+
+    @property
+    def envelope(self) -> Optional[PolyCollection]:
+        return self._envelope_ref()
+
+    @property
+    def ax(self) -> Optional[mplaxes.Axes]:
+        return None if self._ax_ref is None else self._ax_ref()
 
     def __del__(self) -> None:
         """Disconnect callback methods on delete"""
@@ -812,8 +831,8 @@ class AdaptiveWaveplot:
         self.disconnect()
 
         # Attach to axes and store the connection id
-        self.ax = ax
-        self.ax.add_patch(self.label_patch_)
+        self._ax_ref = weakref.ref(ax)
+        ax.add_patch(self.label_patch_)
         self.cid = ax.callbacks.connect(signal, self.update)
 
     def disconnect(self, *, strict: bool = False) -> None:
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks so much for the thorough work on this!  I like the solution you've landed on here, so I'm happy to merge once the CI goes through.  It does look like there might be some linting issues to clear up though.

bmcfee:
There is one tiny coverage drop here, but I'm not too fussed with it since it seems like a last-ditch safety guard that isn't tested.  Possibly this is difficult-to-impossible to trigger with a precise test due to the vagaries of python gc timing.  @Daniel-Haas-B if you have thoughts on how to hit this line, please go for it, but I'm happy to leave it as is if not.

Daniel-Haas-B:
Awesome! Will adress the failings in CI.
Should I add changelogs to 0.11?

bmcfee:
> Should I add changelogs to 0.11?

Nah, I'll be rolling it in with a huge pile of changes for the 1.0 release.  As long as the PR is tagged appropriately and linked to the issue (which it is), it's easier for me to do this all in one go when prepping the release.

bmcfee:
Thanks again! All looks good to me. I'm not worried about the type checking failures, as there's an open pr to update these across the package anyway.  Anything left to do here, or are we good to merge?

Daniel-Haas-B:
All done from my side! Thank you for creating & maintaining Librosa!

bmcfee:
Thanks again @Daniel-Haas-B !  This one has bugged me for years :grin: glad to have it fixed for the 1.0 release.  Hopefully I'll find time to get that off the ground in the next month or two.

PR #1995, opened 2026-02-12
WIP Parallel jit for viterbi and friends

### LINKED ISSUE #1480: Parallel jit for faster numba functions

bmcfee:
#### Reference Issue
Fixes #1480 

This PR adds run-time selectable parallelism for jit-compiled functions in the viterbi family.

A new parameter "parallel" is introduced to select this option, with the default to true.

The pyin function also exposes this parameter.

Parallelism isn't always beneficial: for example, if there is a higher-level embarrassing parallelism to exploit (eg batch processing independent tracks), then it might be better to disable tight-loop threading to cut down on contention.

Probably this should be documented and illustrated as part of #1968 

#### Any other comments?

Todo:

- [ ] tests
- [ ] documentation
- [ ] other functions?  Maybe overlap_add could be a contender here, along with calling functions like istft, griffinlim, icqt, etc.


bmcfee:
Thinking on this design a bit: I'm not 100% sold on the idea of making `parallel: bool` a parameter.  It makes some limited amount of sense here for viterbi, but in the bigger picture, we might also benefit (even more?) from parallelizing vectorized ufuncs and gufuncs.  The issue here is that these operations are typically buried under a few layers, and touch many points of the API.

It may make more sense to have some kind of global state / context / environment variable that can be queried under the hood, rather than threading `parallel=True` through the entire API.  This would have some additional benefits down the road, eg when we take on #1743 (which may involve rewriting parts without numba for different backend targets); in that case, we'd avoid having extra parameter clutter.

I'll give this more of a think, but in the meantime, I welcome suggestions for how to do this.

bmcfee:
Noodled this out a bit.  I think the plan that makes the most sense is to introduce a singleton object that all modules can query directly at run-time to select between single- and multi-threaded execution globally.  This is along the lines of what we do with the `Cache` object, but much simpler and tighter in scope.

I'm thinking something like:

```python
librosa.set_parallel(True)
# all subsequent calls will use parallelized numba builds
librosa.set_parallel(False)
# back to parallel=False numba builds
```
At import time, we can query the environment for whether parallelism should be enabled, just like we do for cache parameters:
https://github.com/librosa/librosa/blob/e403272fc984bc4aeb316e5f15899042224bb9fe/librosa/_cache.py#L84-L90

I generally don't like hidden stateful things like this, but I think here the tradeoff is worth it since it prevents API bloat.

I do think the undertaking here is enough that we should punt it to a 1.1 feature though.

bmcfee:
Updated this branch to (hopefully) work with sparsified viterbi from #2007

PR #1997, opened 2026-03-01
Optimize pyin with sparse Viterbi

AsymptotaX:
I hope this implementation addresses one of the most significant performance bottlenecks in the library by optimizing the Viterbi decoding step in pyin.

**Before**: pyin built a dense transition matrix and decoded with the generic sequence.viterbi.
**After**: pyin now uses a specialized sparse Viterbi path (__pyin_viterbi + __viterbi_sparse_numba) with cached transition structure (@lru_cache).

**Coverage**: added tests for sparse vs dense decoding equivalence, batched shapes, transition-cache behavior, and new guard branches; tests/test_core.py passes.

Why these functions are in pitch.py (not sequence.py):
This is really a pyin optimization, not a generic Viterbi feature.
sequence.py is used as the shared decoding module, so I didn’t want to put pyin-specific assumptions there.
Keeping it in pitch.py makes the scope clear: we speed up pyin only, and we don’t risk changing behavior for other code paths that rely on sequence.viterbi.

#### Environment 
macOS-26.3-arm64
Python 3.13.5
NumPy 2.4.2
SciPy 1.17.1
librosa 1.0.0dev

#### Profiling Comparison (60s audio)
Test file: test_1min.wav
Sample rate: 44.1 kHz
Duration: 60 seconds
Profiler: cProfile

| Component | Dense (s) | Optimized (s) | Speedup |
| --- | --- | --- | --- |
| Total pyin | 11.797 | 1.230 | **9.6×** :fire: |
| Viterbi core | 11.164 | 0.570 | **19.6×** :fire: |
| numpy.vectorize | 11.567 | 0.488 | **23.7×** :fire: |
| __pyin_helper | 0.420 | 0.487 | ~1.0× |
| YIN CMND | 0.191 | 0.167 | ~1.1× |
| Autocorrelate | 0.129 | 0.115 | ~1.1× |

#### Benchmark Comparison
[benchmark.py](https://github.com/user-attachments/files/25659606/benchmark.py)

Repeats: 3 runs per file with cache warm-up.
| File  | sr | Samples | OLD mean (s) | NEW mean (s) | Speedup |
| --- | --- | --- | --- | --- | --- |
| `test_1min.wav`  | 44100 | 2,646,000 | 11.70 | 1.02 | **11.4×** |
| `test_5min.wav` | 44100 | 13,230,000 | 58.65 | 5.09 | **11.5×** |
| `test_15min.wav` | 44100 | 39,690,000 | 175.72 | 16.81 | **10.4×** |

#### tests
I performed a validation test comparing the new implementation against the previous dense version, and no discrepancies were observed.

INPUT
file: /bench/test_1min.wav
sr: 44100
samples: 2646000
duration_sec: 60.00

DIFF (new vs old)
state_sequence_equal: True
state_sequence_mismatch_frames: 0
f0_exact_equal: True
voiced_flag_equal: True
voiced_prob_exact_equal: True
voiced_prob_allclose: True
shared_voiced_frames: 5164
shared_voiced_nonempty: True
f0_abs_diff_mean_hz: 0.000000000000
f0_abs_diff_max_hz: 0.000000000000
voiced_prob_abs_diff_mean: 0.000000000000
voiced_prob_abs_diff_max: 0.000000000000

### DIFF
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -3,6 +3,7 @@
 """Pitch-tracking and tuning estimation"""
 
 import warnings
+from functools import lru_cache
 import numpy as np
 import scipy
 import numba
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks @AsymptotaX - acknowledging that I've seen this for now.  I won't have time to look at it in depth for a week or two, but it sounds promising from your description!

AsymptotaX:
Implemented tuning updates and sped up pyin, including a faster pYIN Viterbi path.

Further optimized the sparse Viterbi path:
- rolling buffers instead of a full value[t, j] table
- fewer allocations
- fastmath=True
- src_idx narrowed to uint16
- removed the unnecessary uniform p_init

Moved pYIN trough-probability computation into a Numba helper __pyin_helper_probabilities instead of doing that work in Python with util.localmin and scipy.stats.boltzmann.pmf

Why the latest implementation is better:
- It produces the same output with less memory traffic and fewer allocations
- More of the hot path now runs inside Numba, with less Python/SciPy overhead


#### benchmarks
Repeats: 3 runs per file with cache warm-up.
| File  | sr | Samples | OLD mean (s) | NEW mean (s) | Speedup |
| --- | --- | --- | --- | --- | --- |
| `test_1min.wav`  | 44100 | 2,646,000 | 11.70 | 0.58 | **20.1×** :fire: |
| `test_5min.wav` | 44100 | 13,230,000 | 58.65 | 2.79 | **21.0×**  :fire:|
| `test_15min.wav` | 44100 | 39,690,000 | 175.72 | 9.24 | **19.0×**  :fire:|


#### tests
audio: test_1min.wav

OUTPUT (current working tree)
```
sr: 44100
samples: 2646000
duration_sec: 60.0
frames: 5168
voiced_frames: 5168
voiced_ratio: 1.0
elapsed_sec: 0.5552897500019753
```

OUTPUT (reference git ref)
```
sr: 44100
samples: 2646000
duration_sec: 60.0
frames: 5168
voiced_frames: 5168
voiced_ratio: 1.0
elapsed_sec: 13.09254616600083
```

DIFF (current vs reference)
```
f0_exact_equal: True
f0_allclose: True
f0_mismatch_count: 0
voiced_flag_equal: True
voiced_prob_exact_equal: False
voiced_prob_allclose: True
voiced_prob_mismatch_count: 0
shared_voiced_frames: 5168
f0_abs_diff_mean_hz: 0.000000000000
f0_abs_diff_max_hz: 0.000000000000
voiced_prob_abs_diff_mean: 0.000000000000
voiced_prob_abs_diff_max: 0.000000000000
speedup_vs_ref: 23.58x
```

bmcfee:
Starting to look into this now.  I first want to get to the bottom of the (obviously unrelated) test environment failures though.  Looks like something's gone strange in our CI and it's running python 3.14 everywhere.

bmcfee:
Test env is mostly fixed now - we still have an upstream bug to contend with on python 3.14, but i've at least got the builds back into a place where they work on older versions. If you could rebase to the current 1.0.0dev branch we should be able to get on with review.

bmcfee:
Updated python-samplerate package should fix the broken builds on 3.14 now.

bmcfee:
Thanks for your patience on this one.  I've had some time to think on it a bit, and I generally like the ideas here.  I'm quite wary of having a largely redundant viterbi implementation kicking around though, as it invites maintenance overhead down the road that I'd prefer to avoid.

I think broadly the viterbi optimizations can be rolled into the mainline viterbi implementation.  I was able to replace your original ~11x speedup (pyin round trip, 1 min chirp test from 50 to 5000 at sr=44100) with a pretty non-invasive set of changes.  It turns out to be comparably fast to pre-compute the indices of nontrivial transitions up front within viterbi as it is to assume a compact index range as you're doing here.

A quick hack version of it looks as follows:

```python
    n_steps, n_states = log_prob.shape

    state = np.zeros(n_steps, dtype=np.uint16)
    value = np.zeros((n_steps, n_states), dtype=np.float64)
    ptr = np.zeros((n_steps, n_states), dtype=np.uint16)

    # factor in initial state distribution
    value[0] = log_prob[0] + log_p_init

    # Pre-compute transitions with ~nonzero likelihood
    pred_states = []
    for j in range(n_states):
        pred_states.append(np.flatnonzero(log_trans[:, j] >= -86))

    for t in range(1, n_steps):
        # Want V[t, j] <- p[t, j] * max_k V[t-1, k] * A[k, j]
        #    assume at time t-1 we were in state k
        #    transition k -> j

        # Unroll the max/argmax loop to enable numba support
        for j in range(n_states):
            best_cost = -np.inf
            for k in pred_states[j]:
                cost = value[t-1, k] + log_trans[k, j]
                if cost > best_cost:
                    ptr[t, j] = k
                    best_cost = cost
            value[t, j] = log_prob[t, j] + best_cost

    # Now roll backward

    # Get the last state
    state[-1] = np.argmax(value[-1])

    for t in range(n_steps - 2, -1, -1):
        state[t] = ptr[t + 1, state[t + 1]]

    logp = value[-1:, state[-1]]

    return state, logp
```

Obviously `-86` is a magic number that I yoinked from 2×epsilon, and we could make this a controllable threshold (probably not parametrized to the user in log space, but you get the idea).  It would also be pretty easy to bypass this thresholded sparsification with a token value (-inf or whatever) and revert to the current "exact" behavior.

This implicit sparsification + rolling value buffer results in the following benchmark on my laptop:

1.0.0dev branch:
```
In [23]: sr = 44100

In [24]: y = librosa.chirp(sr=sr, fmin=50, fmax=5000, duration=60)

In [25]: %timeit librosa.pyin(y=y, sr=sr, fmin=50, fmax=5000)
56.1 s ± 2.37 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Proposed:
```
In [18]: sr = 44100

In [19]: y = librosa.chirp(sr=sr, fmin=50, fmax=5000, duration=60)

In [20]: %timeit librosa.pyin(y=y, sr=sr, fmin=50, fmax=5000)
5.27 s ± 421 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So, not quite to the 20x speedup you report later, but that's not implementing any of the other changes that you made specifically to pyin.

The question I have at this point: I do think it's better to push whatever speedups we can into the primary viterbi implementation, as that benefits everyone and limits our maintenance costs down the road.  If that sounds reasonable, how hard would it be to isolate the rest of the pyin changes here and remove the specialized pyin_viterbi code?

bmcfee:
@AsymptotaX nudging on the previous comment from a few weeks back - I'm curious for your thoughts on this, especially given the proliferation of PRs to improve pyin lately.

PR #1999, opened 2026-03-13
Python 3.14 support

bmcfee:

This PR adds 3.14 to the build matrix and supported version list.

#### Any other comments?

Something seems to have gone very wrong this week in our test CI with respecting our python version constraints.  This PR will resolve the problem once I figure out what it is.

### DIFF
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -29,5 +29,4 @@ dependencies:
   - resampy>=0.2.2
 
   - pip:
-    - --no-binary samplerate
     - samplerate
--- a/.github/environment-ci.yml
+++ b/.github/environment-ci.yml
@@ -29,4 +29,5 @@ dependencies:
   - resampy>=0.2.2
 
   - pip:
+    - --no-binary samplerate
     - samplerate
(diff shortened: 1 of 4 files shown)


bmcfee:
Current failures appear to be coming from two independent (but not orthogonal) problems:

- Our conda environment setup at some point started ignoring our python-version setting and defaulting to the latest (3.14)
- This was fine, except that now we have incompatibilities popping up with numpy 2.4.  These may be upstream from us or not, I'm not yet sure.  

Right now, the main problem is to fix the python version in our builds.

bmcfee:
Failures are now locked down to numpy 2.4.2 on python 3.14.  It may be due to a bad wheel for python-samplerate?

bmcfee:
Ok, 3.14 failures are attributable to https://github.com/numpy/numpy/issues/30991 , which in turn surfaces a change in the refcount semantics from python 3.14 that makes resize ops tricky.  This is then being triggered in the python-samplerate package.  Recommended from numpy devs to wait on an upstream fix rather than patch it ourselves.

PR #2004, opened 2026-03-29
feat: implement hybrid_tempogram (#1991)

### LINKED ISSUE #1991: [ENH]: Hybrid tempograms
### Problem

Spinning off of #1986 - once the metrogram PR merges, we'll have some basic building blocks to implement hybrid tempograms, ie those which combine autocorrelation and fourier analyses, as was done in Peeters 2005.

### Proposed solution

We can add a simple function to the rhythm submodule `hybrid_tempogram`, which calls out to both basic tempogram feature extractors and does the broadcast interpolation accordingly to merge the representations.

There is already a sketch of how this would work in the metrogram docstring.  When this function is implemented, we should revise that docstring accordingly.

Gopesh111:
#### Reference Issue
Fixes #1991

This PR adds the `hybrid_tempogram` function to `librosa/feature/rhythm.py`. 

Here is what was changed:
- Combined `fourier_tempogram` and `tempogram` (autocorrelation) using a geometric mean on a common frequency grid, as described in Peeters (2005).
- Used `scipy.interpolate.interp1d` to map the autocorrelation lag frequencies onto the Fourier grid.
- Added shape matching (taking the minimum of frames) and used `np.maximum(0, ...)` before the square root to handle any potential `NaN` values or broadcast errors.
- Used `axis=-2` to ensure it works correctly for multi-channel audio.
- Updated `librosa/feature/__init__.py` and `librosa/feature/__init__.pyi` so the lazy loader registers the new function.

#### Any other comments?
I have tested this locally with sample audio. The output shapes match the expected grids, and the values are numerically stable without any warnings.

I am a final-year computer science student working on a deepfake audio detection project, and I use Librosa a lot. This is my first major open-source PR, so please let me know if I need to adjust any formatting or conventions!


bmcfee:
Thanks for this!  It appears to be off to a good start, but I do see a few things that need to be addressed.  I'll leave comments in CR.  Meanwhile, please re-target this PR against the 1.0.0dev branch instead of main.

Regarding styling and formatting: this will mostly be covered by our linting continuous integration checks.  You can run those locally if you want (see the [github action](https://github.com/librosa/librosa/blob/e403272fc984bc4aeb316e5f15899042224bb9fe/.github/workflows/lint_python.yml#L45-L71) for the sequence of operations that we check), or just wait for CI to run once it's approved.

librosa/feature/rhythm.py
```diff
@@ -653,3 +654,76 @@ def tempogram_ratio(
         return aggregate(tgr, axis=-1)  # type: ignore
 
     return tgr
+def hybrid_tempogram(
+    y=None,
+    sr=22050,
+    onset_envelope=None,
+    hop_length=512,
+    win_length=384,
+    **kwargs,
```
bmcfee: I don't think kwargs are appropriate here.  Fourier tempogram and autocorrelation tempogram have slightly different parameter sets, and passing the same kwargs to both may produce errors.

OTOH, the number of remaining parameters is small here, and could easily be included explicitly in the hybrid signature.

librosa/feature/rhythm.py
```diff
@@ -653,3 +654,76 @@ def tempogram_ratio(
...
+
+    onset_envelope : np.ndarray [shape=(..., n)] or None
+        onset strength envelope.
+
+    hop_length : int > 0 [scalar]
+        hop length, number of samples between successive frames.
+
+    win_length : int > 0 [scalar]
+        window length for the tempogram analysis.
+
```
bmcfee, line 803: It might be worth allowing pre-computed tempograms to be passed in here.  I'm on the fence about this, since doing it properly could require quite a bit of additional parameter bloat (eg in case they have different time grids or whatever).

librosa/feature/rhythm.py
```diff
@@ -653,3 +654,76 @@ def tempogram_ratio(
...
+    tempogram : np.ndarray [shape=(..., n_frequencies, n_frames)]
+        the hybrid tempogram
+    """
+
+   # 1. Compute Fourier tempogram
+    tg_f = fourier_tempogram(
+        y=y, sr=sr, onset_envelope=onset_envelope, 
+        hop_length=hop_length, win_length=win_length, **kwargs
+    )
+    freqs = tempo_frequencies(tg_f.shape[-2], sr=sr, hop_length=hop_length)
```
bmcfee: This part is incorrect.  The `tempo_frequencies` function is for autocorrelation tempograms; here we should be using `fourier_tempo_frequencies`.

librosa/feature/rhythm.py
```diff
@@ -653,3 +654,76 @@ def tempogram_ratio(
...
+
+    # 2. Compute Autocorrelation tempogram
+    tg_a = tempogram(
+        y=y, sr=sr, onset_envelope=onset_envelope, 
+        hop_length=hop_length, win_length=win_length, **kwargs
+    )
+    lags = tempo_frequencies(tg_a.shape[-2], sr=sr, hop_length=hop_length)
+
+    # 3. Hybrid Interpolation
+    from scipy.interpolate import interp1d
```
bmcfee: We can just call to `scipy.interpolate.interp1d` here; I don't think there's any benefit to having a local import.

librosa/feature/rhythm.py
```diff
@@ -653,3 +654,76 @@ def tempogram_ratio(
...
+    f_interp = interp1d(lags, tg_a, axis=-2, bounds_error=False, fill_value=0.0)
```
bmcfee: There are a couple of ways this could be improved:

1. Allow user-specified parameters to interp1d (eg to control the interpolation mode).  See `interp_harmonics` for a related example of how this could look.
2. Since autocorrelation tempograms have a 0-lag frequency of +∞, we may as well restrict the interpolation to finite frequencies.  This should remove the need for the nan_to_num below.

librosa/feature/rhythm.py
```diff
@@ -653,3 +654,76 @@ def tempogram_ratio(
...
+
+    # 4. Shape Matching (Crucial Fix)
+    # Sometimes frames differ by 1 due to padding. We take the minimum.
+    n_frames = min(tg_f.shape[-1], tg_a_resampled.shape[-1])
+    
+    # 5. Merging (Geometric Mean)
+    # Ensure values are non-negative before sqrt to avoid NaNs
+    # Also handle potential NaNs from interpolation
+    product = np.abs(tg_f[..., :n_frames]) * tg_a_resampled[..., :n_frames]
+    hybrid = np.sqrt(np.maximum(0, np.nan_to_num(product)))
```
bmcfee: I think geometric mean is a nice default, but I could imagine wanting to do other combination strategies.  If there's a clean way to expose this at the API level, I think we should go for it.

librosa/feature/rhythm.py
Gopesh111: Thanks for pointing this out! I've removed **kwargs and explicitly added the center and window parameters to keep the signatures safe and clean.

librosa/feature/rhythm.py
Gopesh111: I thought about this too, but as you mentioned, it might cause parameter bloat if the time grids don't align perfectly. I've kept it simple for now, but let me know if you'd prefer me to add it in this PR!

librosa/feature/rhythm.py
Gopesh111: Ah, my mistake! I've corrected this to use fourier_tempo_frequencies for the Fourier grid.

librosa/feature/rhythm.py
Gopesh111: Fixed. I've removed the local import and am using the global call now.

librosa/feature/rhythm.py
Gopesh111: Great suggestions. I've added an interp_kwargs parameter (defaulting to bounds_error=False, fill_value=0.0). I also sliced the autocorrelation arrays to drop the 0-lag (+∞ frequency), which completely removed the need for np.nan_to_num.

librosa/feature/rhythm.py
Gopesh111: I agree. For now, I've kept the geometric mean as the default. If you'd like, I can add an aggregate parameter (accepting a callable or string like 'geom' or 'arithmetic') in this PR itself. Let me know what you prefer!

Gopesh111:
Hi @bmcfee, I've addressed all the feedback and also re-targeted the PR to the 1.0.0dev branch. Added a local test to ensure multi-channel audio (axis=-2) and the new finite interpolation logic run smoothly without NaNs. Thanks for the detailed review!

Gopesh111:
Hi @bmcfee, extremely sorry for the multiple pushes. I've fixed the typo and restored the metrogram code. Could you please approve the workflow one last time to verify the CI and coverage? Thanks for your patience!

Gopesh111:
Hi @bmcfee,
I’ve fixed the codespell error by removing the non-English text from the docstring. I also verified the file with flake8 locally, and it’s now returning 0 errors.
The previous CI failure was strictly due to that typo. Could you please trigger the checks one last time? Ready for the final review!

Gopesh111:
Hi @bmcfee, I've refactored the interpolation logic to natively handle multi-channel arrays and ensured that interp_kwargs are safely merged with defaults to prevent out-of-bounds errors. All tests and linting are now passing locally. Ready for your review!

Gopesh111:
Hi @bmcfee,

I’ve cleaned up the trailing whitespaces in the docstring to address the velin linter failure. No changes were made to the core logic.
With the coverage at 100% and all 13 functional CI checks already passing in the previous run, this should now fully clear the linting stage. Ready for the final review!

Gopesh111:
Hi @bmcfee,

I've refactored the interpolation logic to explicitly pass bounds_error and fill_value, which should resolve the mypy type-checking issues. I also removed the duplicate @cache decorator that slipped in during the last merge.

Thank you so much for your patience throughout this process. As this is my first major PR, I really appreciate you sticking with me through these linting and formatting iterations.

All checks are passing locally now, hoping for the final green light!

librosa/feature/rhythm.py
```diff
@@ -657,6 +665,110 @@ def tempogram_ratio(
...
+        The hybrid tempogram combining both representations
+    """
+    import scipy.interpolate
+
+    # Safe defaults that allow extrapolation without out-of-bounds error
+    bounds_error = False
+    fill_value = 0.0
+    if interp_kwargs is not None:
+        bounds_error = interp_kwargs.get("bounds_error", False)
+        fill_value = interp_kwargs.get("fill_value", 0.0)
```
bmcfee: I think we can also put in `assume_sorted=True` and `copy=False` to make this a little faster.

That said, this part is a little clumsy.  We can still use kwargs here for interp1d pass-through, and in that case, we can just use `setdefault` instead of having branches here.  My previous concern with kwargs was how it was used in two different functions, but that's no longer an issue.

librosa/feature/rhythm.py
```diff
@@ -657,6 +665,110 @@ def tempogram_ratio(
...
+    *,
+    y=None,
+    sr=22050,
+    onset_envelope=None,
+    hop_length=512,
+    win_length=384,
+    center=True,
+    window="hann",
+    interp_kwargs=None,
+):
```
bmcfee: This function header should have type annotations.

librosa/feature/rhythm.py
```diff
@@ -657,6 +665,110 @@ def tempogram_ratio(
...
+    """Compute a hybrid tempogram.
+
+    This function computes a hybrid representation by combining the
+    Fourier tempogram and autocorrelation tempogram. The tempograms are
+    aligned onto a common frequency grid and merged using the geometric mean.
```
bmcfee: A previous version of this docstring had the citation to Peeters, which seems to have vanished.  We should add that back in.

librosa/feature/rhythm.py
```diff
@@ -657,6 +665,110 @@ def tempogram_ratio(
...
+        Whether to center the frames
+    window : str
+        Window type
+    interp_kwargs : dict or None
+        Additional keyword arguments for interpolation
+
+    Returns
+    -------
+    hybrid : np.ndarray
+        The hybrid tempogram combining both representations
```
bmcfee, line 710: There should be an examples section here illustrating how to use the function, with figures, as I previously noted.

librosa/feature/rhythm.py
```diff
@@ -657,6 +665,110 @@ def tempogram_ratio(
...
+    """
+    import scipy.interpolate
```
bmcfee: I'd like to avoid local imports.  scipy is imported at the top level and that should be enough.

bmcfee (changes requested):

Gopesh111:
Hi @bmcfee, thanks for the thorough review! I've addressed all the points:

1)Added assume_sorted=True and copy=False with setdefault for the kwargs optimization.

2)Added type annotations to the function header.

3)Restored the Peeters (2005) citation in the docstring.

4)Added a complete Examples section with plotting.

5)Removed the local scipy import.

Let me know if everything looks good now!

Gopesh111:
Hi @bmcfee, it's ready for your review. Please let me know if any changes are required

librosa/feature/rhythm.py
```diff
@@ -657,6 +665,128 @@ def tempogram_ratio(
     return tgr
 
 
+def hybrid_tempogram(
+    *,
+    y: "np.ndarray" = None,
+    sr: float = 22050,
+    onset_envelope: "np.ndarray" = None,
```
bmcfee: These type annotations are incorrect for optional parameters; please see elsewhere in the package (eg tempogram) for the pattern we follow on these.  They also should not be strings.

librosa/feature/rhythm.py
```diff
@@ -657,6 +665,128 @@ def tempogram_ratio(
...
+
+    Returns
+    -------
+    hybrid : np.ndarray
+        The hybrid tempogram combining both representations
+
+    References
+    ----------
+    .. [1] Peeters, Geoffroy. "Rhythm Classification Using Periodicities and the Beat-Histogram."
+        Proceedings of the 6th International Conference on Music Information Retrieval (ISMIR). 2005.
```
bmcfee: I don't think this will build / render correctly.  Please refer to other docstrings to see how to format these.  You don't need to make a separate References section.

librosa/feature/rhythm.py
```diff
@@ -657,6 +665,128 @@ def tempogram_ratio(
...
+
+    Display the result
+
+    >>> import matplotlib.pyplot as plt
+    >>> fig, ax = plt.subplots()
+    >>> img = librosa.display.specshow(htemp, x_axis='time',
+    ...                                y_axis='tempo', hop_length=hop_length,
+    ...                                ax=ax)
+    >>> ax.set(title='Hybrid Tempogram')
+    >>> fig.colorbar(img, ax=ax)
```
bmcfee, line 750: A couple of things here:

1. I think the example would be a lot more informative if there are three subplots, showing the autocorrelation and fourier tempograms first and then the hybrid last.
2. y_axis here should be `fourier_tempo` instead of `tempo` since the target interpolation grid is the Fourier grid, not the autocorrelation grid.

librosa/feature/rhythm.py
```diff
@@ -657,6 +665,128 @@ def tempogram_ratio(
...
+    """
+    kwargs = dict(interp_kwargs) if interp_kwargs is not None else dict()
```
bmcfee: Why go through all this instead of just having  `**kwargs` in the signature?

Gopesh111:
Hi @bmcfee, thanks for the review! I've addressed all the feedback. Ready for the final check!

Gopesh111:
Hi @bmcfee, everything is verified locally.

Functional tests: Passed (including multi-channel and custom kwargs).

Mypy: Core typing issues resolved (any-return fixed via np.asarray).

Linting: Flake8 and Velin passing with 0 errors.

Ready for final review!!

Gopesh111:
Hi @bmcfee, I noticed the CI Docs workflow failed specifically at the make linkcheck step.

Since I haven't added any new external URLs or links in the hybrid_tempogram docstring, I suspect this might be a transient network timeout with an existing link in the documentation. Could you please re-trigger the docs job or let me know if there's anything on my end that needs fixing?

librosa/feature/rhythm.py
bmcfee: I built the docs locally, and the figure generated looks as follows:

Image

I then tried running some examples locally, and every one of them produces an all-zeros hybrid tempogram.  If I override the default `fill_value=0` to anything nonzero (even 1e-10), it seems to work fine.

Digging a bit more, it seems that `assume_sorted=True` is only valid when the input domain is sorted in *ascending order*.  Disabling this and keeping fill_value=0.0 by default recovers correct behavior.  Very strange indeed, but that's my mistake for suggesting it (I had forgotten that we were interpolating the lag domain instead of the frequency domain).

librosa/feature/rhythm.py
```diff
@@ -656,6 +664,145 @@ def tempogram_ratio(
...
+    >>> img = librosa.display.specshow(hybrid_tempogram, x_axis='time',
+    ...                                y_axis='fourier_tempo', hop_length=hop_length,
+    ...                                ax=ax[2])
+    >>> ax[2].set(title='Hybrid Tempogram')
+    >>> fig.colorbar(img, ax=ax)
+    """
+    interp_kwargs_dict: dict[str, Any] = kwargs if kwargs else {}
+    interp_kwargs_dict.setdefault("bounds_error", False)
+    interp_kwargs_dict.setdefault("fill_value", 0.0)
+    interp_kwargs_dict.setdefault("assume_sorted", True)
```
bmcfee:
```suggestion

```

bmcfee:
Don't worry about the linkcheck failure - it does appear to be transient and unrelated to this PR.  There is one lingering bug in the code though (noted in comment) and I've made a patch suggestion to fix it.

librosa/feature/rhythm.py
Gopesh111: Thanks for catching that, @bmcfee! You are totally right—since the lag domain isn't strictly ascending, assume_sorted=True was forcing it to return the fill_value everywhere.

I have removed that line and pushed the fix so the interpolation works correctly now. And thanks for the heads-up on the linkcheck failure!

Let me know if the docs build correctly on your end now.

Gopesh111:
Hi @bmcfee, I wanted to send a separate note to thank you for the thorough and patient review on this PR.

I especially appreciate you taking the time to test the implementation locally and catching that bug. It’s been great working through these iterations with you to get this feature right. Thanks again for the guidance!

bmcfee (approved):

bmcfee:
This all looks good to me now, thanks for addressing all the nitpicks!

I'm gonna let the bot run over it once just to catch anything I may have missed, but otherwise I think this should be merge-ready.

tests/test_features.py
```diff
@@ -1138,3 +1136,52 @@ def test_metrogram_aggregate(y_clicks, tg_ex):
     met1 = librosa.feature.metrogram(tg=tg_ex[1:], freqs=freqs, aggregate=None)
     met2 = librosa.feature.metrogram(tg=tg_ex[1:], freqs=freqs, aggregate=np.max)
     assert np.allclose(np.max(met1, axis=-2), met2)
+def test_hybrid_tempogram():
+    # Load a short snippet of sample audio to keep tests fast
+    y, sr = librosa.load(librosa.ex("trumpet"), duration=2.0)
```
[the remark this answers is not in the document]
bmcfee: This is a good catch - test code should not rely on examples from the main package.  Instead, we have bundled examples in the tests folder that bypass the external data access.

librosa/feature/rhythm.py
```diff
@@ -656,6 +664,144 @@ def tempogram_ratio(
...
+
+    # 3. Restrict to finite frequencies (drop 0-lag / infinite BPM)
+    tg_a_finite = tg_a[..., 1:, :]
+    lags_finite = lags[1:]
+
+    # 4. Hybrid Interpolation
+    f_interp = scipy.interpolate.interp1d(
+        lags_finite, tg_a_finite, axis=-2, **interp_kwargs_dict
+    )
+    tg_a_resampled = f_interp(freqs)
```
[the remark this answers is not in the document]
bmcfee: Quick fix here is to just push `axis` into the kwargs.

librosa/feature/rhythm.py
```diff
@@ -656,6 +664,144 @@ def tempogram_ratio(
...
+    # 2. Compute Autocorrelation tempogram
+    tg_a = tempogram(
+        y=y,
+        sr=sr,
+        onset_envelope=onset_envelope,
+        hop_length=hop_length,
+        win_length=win_length,
+        center=center,
+        window=window,
+    )
```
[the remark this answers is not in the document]
bmcfee: This is reasonable - onset envelope calculation isn't the most expensive thing, but remember that it will also trigger an stft (twice) and so on.  Probably the best course of action here is to replicate the onset envelope calculation branch from either of the two primary tempogram functions so that they don't have to do it redundantly.

Gopesh111:
Hi @bmcfee,

Firstly, I sincerely apologize for the delay in getting these updates pushed. I had some prior commitments that took longer than expected, but I’m back on it now.

I have addressed all the feedback from the Copilot review and your suggestions:

A) Redundant Computation: I’ve replicated the onset envelope calculation branch so that fourier_tempogram and tempogram don't trigger redundant STFTs.

B) Kwargs & Axis: Moved the axis parameter into the interp_kwargs_dict as suggested.

C) Typing & Formatting: Fixed the _WindowSpec type hint and the missing space in the imports.

D) Test Fixtures: Replaced the external example audio in tests with a synthesized deterministic click track to avoid I/O flakiness.

Everything should be in line with the library's standards now. Thanks again for your patience

bmcfee:
Thanks @Gopesh111 !  There's one small coverage failure here, but I'm not worried about it right now.  I did commit the copilot suggestion to expand the docstring here afd0f462fbd1aa034712539ea83b0c7c3187ad40 and otherwise I think we're good to go.

Gopesh111:
Thanks for handling that docstring update, @bmcfee! Really appreciate your help and patience throughout this PR.

bmcfee:
I also fixed the doc build error upstream in the 1.0 dev branch, so this this :shipit: 

Thanks again @Gopesh111 !

Issue #2008, opened 2026-05-08
[ENH]: Extended envelopes in waveshow


bmcfee:
### Problem

In waveshow, we currently alternate between a sample plot and a sample-peak envelope plot, as determined by the size of the view port.  This is all well and good, but the sample-peak approach has always bugged me a little bit due to its sensitivity to single data points.  This is a feature, not a bug, but it got me thinking that we could possibly do other things in addition to peak envelope.

An obvious alternative would be an RMS envelope.  This would especially make more sense for very long signals where the influence of a single peak within a region is probably not representative.

But we don't have to stop there - several other aggregator functions could make sense here (e.g. quantiles).

Pushing this line of thought further: what if we supported multiple simultaneous envelope artists?  If done carefully, this has the potential to show much more information in a pretty clean display.  Some example use cases could be `peak+rms` (which would imply the same content as a crest factor plot), or multiple quantiles [50/75/95] ; these would make sense with stacked alpha values, allowing the overlapped region to show through more intensely, but it should also be possible control styling independently for each envelope.

The reason to do multiple envelope plots here instead of just asking the user to call waveshow multiple times is essentially to eliminate redundancy of the time series plot objects.

### Proposed solution

I think the API here could be a pretty clean extension of the waveshow API.  Something like:
```python
librosa.display.waveshow(y=y, sr=sr, envelope='peak')
librosa.display.waveshow(y=y, sr=sr, envelope='rms')
librosa.display.waveshow(y=y, sr=sr, envelope=['peak', 'rms'])
librosa.display.waveshow(y=y, sr=sr, envelope=('quantile', 0.95))
```
though the quantile tuple packing is a little precarious, and we may do better to do some lightweight parsing as in the `vscale=dB[<controls>]` option to specshow.

Similarly, styling could be inherited from the global options like we currently do (now for each envelope object), or be additionally controlled either for all envelopes or individually:
```python
librosa.display.waveshow(y=y, sr=sr, envelope='peak', env_kw=dict(color='C1'))
librosa.display.waveshow(y=y, sr=sr, envelope=['peak', 'rms'], env_kw=[dict(alpha=0.5), dict(alpha=0.9)])
```
---
## Example

This figure was made manually to illustrate the utility of peak+rms overlays, but it could be automated via the above design:
```python
y, sr = librosa.load(librosa.ex('libri1'), duration=3)
max_points = 11025
hop_length = max(1, y.shape[-1] // max_points)
[...]

Issue #1975, opened 2025-08-13
[ENH]: Even more wave displays - polar version!


bmcfee:
### Problem

I've been thinking it would be nice to include some more extended versions of wave displays, and in particular, polar transforms of waveshow (and wavebars) might be interesting to include.

These are useful for visualizing short loops of audio, eg https://www.ethanhein.com/wp/2023/building-the-amen-break/

### Proposed solution

This turns out to be very easy to do using existing matplotlib functionality, but takes a bit of work to make it look nice.

```python
# Example signal
y, sr = librosa.load(librosa.ex('choice'), duration=5)

T = librosa.get_duration(y=y, sr=sr)                # period in seconds you want to represent around the circle
r0 = 1.0               # base radius
k  = 1.0               # amplitude→radius scale

fig, ax = plt.subplots(subplot_kw=dict(projection='polar'))

# Build a data→(theta,r) pre-transform (linear)
pre = (transforms.Affine2D()
       .scale(2*np.pi / T, k)        # (t, a) -> (θ, k·a)
       .translate(0.0, r0))          # add base radius

librosa.display.waveshow(y, sr=sr, ax=ax, transform=pre + ax.transData)
ax.set_rticks([])
ax.set_thetagrids([])
ax.set_frame_on(False)

## To go clockwise from midnight:
ax.set_theta_direction(-1)
ax.set_theta_offset(np.pi/2)
```
produces

<img width="389" height="389" alt="Image" src="https://github.com/user-attachments/assets/5413974d-b321-4a69-b24c-5b5769de6ffe" />

and here it is with wavebars:

<img width="389" height="389" alt="Image" src="https://github.com/user-attachments/assets/a27fa8f2-34f8-40c9-a949-db819a6dfd14" />

lostanlen:
Extremely cool!

I think that having some angular ticks would be good to visualize microtiming.

How hard would it be to visualize multiple (at least two) tracks as concentric circles? as in the Ethan Hein example

bmcfee:
That should be trivial by accepting the target radius as an argument. 

+100 on the idea of angular ticking for microtiming.  I think it's easy enough to translate the default time locator through the angular mapping, but it might also be worth thinking through supporting this in annotation displays (via mir_eval or any of the proposed fmp toolbox functionality slated for 1.1).

*Update*: here's a quick hack showing two 8-beat slices on the same plot, with angular 24 angular ticks:

[...]

Issue #1538, opened 2022-07-17
Annotation plotting utilities


bmcfee:
**Is your feature request related to a problem? Please describe.**

The display module has traditionally focused on signals and spectrogram-like data, but I'm becoming increasingly convinced that we should also provide some features for visualizing annotations.

I've resisted this in the past because much of that functionality is provided well by [mir_eval](https://craffel.github.io/mir_eval/#module-mir_eval.display), and I'd prefer to avoid duplicating efforts.  In many ways, mir_eval is the logical place for this sort of thing.  However, there are also plenty of use cases outside of MIR and it's not reasonable to expect everyone to know about mir_eval for this purpose.

Some duplication of functionality might be okay and appropriate here, if it makes plotting generally more accessible.

**Describe the solution you'd like**

I don't think we need to fully replicate mir_eval's display module, but the following at least seem useful:

- labeled intervals
- events
- maybe pitch + voicing?
- maybe source separation?

The others I think would be a little more out of scope (multipitch, hierarchical segments, etc).  But the above would cover a pretty wide variety of use cases.

I'm not opposed to matching API and implementation here, but I think we might want to think about this from scratch first.

bmcfee:
Thinking about labeled interval plots again, I remembered that we provide two functions for this in mir_eval: one allows for overlapping intervals and one expects disjoint intervals.  Both of these are important and distinct use cases, and we should aim to support both.

I'm not entirely sure that the approach I took in mir_eval's is the best at this point.  Briefly:

- segment plots use collections of rectangles that span the vertical extent.  I think this is kind of okay, but it would be better if the vertical dimension (or, more generally, non-time-dimension) operated in axes space instead of data space.  This way we could be robust to changes in the data limits after the annotation is drawn.
- overlapping plots use broken barh collections.  This, again, I think is okay, but the non-time dimension should be in axes space and not data space.

The way to implement this, I think, is using [blended transformations](https://matplotlib.org/stable/tutorials/advanced/transforms_tutorial.html#blended-transformations).

This might raise some questions about how annotation plots might be changed or extended (eg when overlaying multiple annotations).

bmcfee:
This might also do weird things with non-uniform coordinates (eg beats).  Or maybe not.. not sure what we should do here.

bmcfee:
> * overlapping plots use broken barh collections.  This, again, I think is okay, but the non-time dimension should be in axes space and not data space.
> ...
[...]

PR #2012, opened 2026-05-18
Multiplot

bmcfee:
#### Reference Issue
#1370 

This PR implements multi-channel display wrappers.

There are two primary functions:

- `librosa.display.multiplot`
- `librosa.display.legend_for_axes`

and a smattering of small helper functions.

The `multiplot` functions serves as a convenience wrapper for plotting multiple related objects of the same kind along a subplot axes array.  It can currently dispatch to either `waveshow`, `wavebars`, or `specshow`, but the dispatch logic is generic enough that this should be easy to expand in the future if needed.

Self-contained example from the documentation:
```python
    >>> y, sr = librosa.load(librosa.ex('choice'), duration=10)
    >>> yh, yp = librosa.effects.hpss(y)

    >>> librosa.display.multiplot('wave', y, yh, yp, sr=sr, share_properties='row',
    ...                           labels=['original', 'harmonic', 'percussive'],
    ...                           sharex=True, sharey=True,
    ...                           label_outer=True,
    ...                           invert=True)

    >>> librosa.display.legend_for_axes()  # Helper to create a single legend across subplots
```

produces the following figure:
image

Note that here the input is variadic (`*args`), but it's also possible to support a pre-stacked multichannel array, e.g.
```python
>>> y_stack = librosa.to_multi(y, yh, yp)
>>> librosa.display.multiplot('wave', y_stack, sr=sr, ...)
```

The example above creates a new figure and axes to draw on, but it can also accept a preconstructed subplot array.  This is helpful for doing linked subplots of mixed type, eg the following with a 3-channel signal and 3-channel stfts uses slices of the subplot array `axes` for different kinds of plots:

```python
>>> fig, axes = plt.subplots(nrows=2, ncols=3, sharex=True, sharey='row', 
... gridspec_kw=dict(hspace=0, wspace=0.05, height_ratios=(0.85, 0.15)), constrained_layout=True)

>>> imgs = librosa.display.multiplot('img', Call[0], Call[1], Call[2], axes=axes[0, :], 
... x_axis='time', y_axis='log', sr=sr, labels=None, vscale='dBFS', label_outer=True)

>>> librosa.display.multiplot('wave', yall, axes=axes[1, :], share_properties=None, 
... labels=labels, sr=sr, invert=True, invert_color='white')

>>> librosa.display.legend_for_axes(axes[1], pad=0)
>>> librosa.display.colorbar_db(imgs[0], ax=axes[0,-1])
```
image

---
The `legend_for_axes` helper is a separate, self-contained utility function that aggregates legend handles across a given set of subplots and creates a single legend attached to the bounding box constructed from the union of the input axes.  (See examples above.)  This is inspired by similar functionality in seaborn's facetgrid plots where it makes more sense to have a unified legend shared across different subplots than to attach legends to individual axes.

The main weirdness in `legend_for_axes` is the nomenclature and logic around the loc parameter.  Right now it's a transparent pass-through to the relevant matplotlib code, but it does lead to some counterintuitive-looking code like `center left` loc actually pinning to the right of the figure and so on.  I don't like it very much, but I think I might like even less the idea of overriding matplotlib's parametrization.  Can't win them all I guess.

#### Any other comments?

This design and implementation have been heavily iterated on by copilot.  I think at this point the design is pretty solid, but it still needs testing.  

This functionality has been lurking in my dev notebooks for almost a year at this point, as pinning down the API and scope has proven to be pretty tricky.  I'm starting the PR now to solicit feedback from humans - I believe the design is in a state where it feels comfortable to use, is properly scoped, and is not overly complex.


bmcfee:
Another fun trick with this, probably mentioned somewhere in #1370 , is property share mapping.  This lets us easily do DAW-style visualizations where styles are linked across some (but not all) subplots.  For example:

```python
y1, sr = librosa.load('/home/bmcfee/data/data/medleydb_tmp/MusicDelta_ModalJazz_STEM_02.wav', mono=False)

y2, sr = librosa.load('/home/bmcfee/data/data/medleydb_tmp/MusicDelta_ModalJazz_STEM_04.wav', mono=False)

y3, sr = librosa.load('/home/bmcfee/data/data/medleydb_tmp/MusicDelta_ModalJazz_STEM_05.wav', mono=False)

y = np.vstack((y1, y2, y3))

fig, ax = plt.subplots(nrows=y.shape[0], sharex=True, sharey=True, squeeze=False, gridspec_kw=dict(hspace=0), figsize=(10, 6))
waves = librosa.display.multiplot('wave', y, sr=sr, share_properties=[0,0,1,1,2,2], axes=ax, invert=True)
```

Does the following:
image

eg the blue plots are left/right for one stem, orange are left/right for another, etc.  This doesn't (yet) play 100% nicely with variadic inputs that are individually multichannel.  I'm not sure if I want to bother trying to force that yet.

docs/conf.py
```diff
@@ -146,6 +146,7 @@
     "figure.subplot.top": 0.85,
     "figure.subplot.wspace": 0.4,
     "text.usetex": False,
+    "savefig.bbox": "tight",
```
bmcfee, line 149: doc config change here is to accommodate some of the quirky side effects of adding legends to subplot arrays.  This change ensures that the full figure is captured in the save.

librosa/display.py
```diff
@@ -2951,3 +2972,635 @@ def colorbar_db(
...
+        else:
+            # User already stacked the inputs into one array.
+            datum = data[0][idx]
+        output.flat[flat_idx] = function(
+            datum,
+            ax=axes.flat[flat_idx],
+            label=labels.flat[flat_idx],
+            **properties.flat[flat_idx],
+            **kwargs,
+        )
```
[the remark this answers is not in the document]
bmcfee: This comment is wrong.  specshow will pass `label` through to pcolormesh, which will accept it but not do anything with it.  I think this is fine as is.

bmcfee:
I think this is basically there, with two caveats:

- legend_for_axes might benefit from some stress testing to identify good default values for padding and the like
- there's currently a bug introduced by the savefig changes in sphinx where the wavebars example no longer builds.  i'll try to fix that later today.

bmcfee:
The aforementioned wavebars bug should be fixed now.

bmcfee:
I think the rough edges here are now thoroughly buffed out.  I'd still like at least one other human to try this out, but it feels pretty done to me.  (And copilot has bashed the PR quite a bit internally outside of CR.)

librosa/display.py
```diff
@@ -2951,3 +2977,680 @@ def colorbar_db(
...
+
+    Returns
+    -------
+    np.ndarray
+        An array of property dictionaries for each subplot in the multiplot grid, with shape compatible with the axes.
+    """
+    properties = np.empty(prop_group.shape, dtype=object)
+    properties.fill(None)
+
+    style_cycler = cycle(plt.rcParams["axes.prop_cycle"])
```
stefanv: Flyby question: what if the user provided a set of axes with a cycle already provided?

librosa/display.py
```diff
@@ -2951,3 +2977,680 @@ def colorbar_db(
...
+    share_properties: Optional[Union[bool, Literal["row", "col"], np.ndarray]] = None,
+    fig_kw: Optional[dict] = None,
+    sharex: bool = True,
+    sharey: bool = True,
+    label_outer: bool = True,
+    labels: Optional[Sequence[Optional[str]]] = None,
+    titles: Optional[Sequence[Optional[str]]] = None,
+    **kwargs,
+) -> np.ndarray:
+    """Create multiple visualizations from a multi-channel input array.
```
stefanv: Hrm, multiple visualization from a multi-channel input array. Multiple visualizations of each, one visualization per "channel"?

librosa/display.py
```diff
@@ -2951,3 +2977,680 @@ def colorbar_db(
...
+
+    Example use cases include:
+        - Displaying multiple waveforms from a multi-channel audio file.
+        - Displaying multiple spectrograms from a multi-channel audio file.
+
+    Parameters
+    ----------
+    func : str
```
stefanv, line 3316: Is there ever a case where you want to broadcast `func` onto the channels, i.e. `[waveshow, wavebars]` onto a 2D dataset?

librosa/display.py
```diff
@@ -2951,3 +2977,680 @@ def colorbar_db(
...
+        The name of the display function to use for the multiplot. Accepted values are 'waveshow',
+        'wavebars', and 'specshow'.
+
+    *data : np.ndarray or multiple ndarrays
```
stefanv: It'd be helpful to be a bit more explicit on which axis contains the channels. It looks, from this description, to be the first. I.e., data[0], data[1], data[2] will be what's plotted.

librosa/display.py
```diff
@@ -2951,3 +2977,680 @@ def colorbar_db(
...
+        `data` can either be a single multi-dimensional array, or multiple separate arrays to plot.  If a single array is provided, the shape of the
+        array will determine the layout of the multiplot grid.  If multiple arrays are provided, they will be plotted in the order they are given, and the
+        number of arrays will determine the layout of the multiplot grid.
+
+    axes : matplotlib.axes.Axes, np.ndarray, or None
+        The axes to use for the multiplot. If None, a new axes array will be created on `fig`.
+        If an array of Axes objects is provided, it must be compatible with the shape of the data.
+        If a single axes object is provided, it will be interpreted as a 1x1 array (i.e. a single subplot).
+
+    fig : matplotlib.figure.FigureBase or None
```
stefanv, line 3332: Do axes not already belong to a figure? What happens if you specify a figure from a different set of axes?

librosa/display.py
```diff
@@ -2951,3 +2977,680 @@ def colorbar_db(
...
+        The figure to use for the multiplot. If None, a new figure will be created if needed.
+
+    orient : str
```
stefanv: Trying to recall NumPy syntax, but in skimage we list the literal options: `orient : {'h', 'v'}`

stefanv: In code, would probably be more readable to use `orientation`, `horizontal`, and `vertical`.

librosa/display.py
```diff
@@ -2951,3 +2977,680 @@ def colorbar_db(
...
+        The orientation of the multiplot grid. Accepted values are 'h' for horizontal and 'v' for vertical. This determines how the
+        subplots are arranged when the input data has a single  non-singleton dimension (e.g., shape (n, k) with k > 1).
+
+    share_properties : bool, str, np.ndarray, or None
+        The property sharing scheme for the multiplot grid. Accepted values are:
+
+        - `None` or `False`: no properties are shared, and each subplot is treated as a unique group.
+        - `True`: all subplots share the same properties and belong to a single group.
```
stefanv, line 3345: What properties are we talking about?

librosa/display.py
```diff
@@ -2951,3 +2977,680 @@ def colorbar_db(
...
+    the figure and axes objects for us.
+
+    >>> import matplotlib.pyplot as plt
+    >>> y, sr = librosa.load(librosa.ex('choice'), duration=10)
+    >>> yh, yp = librosa.effects.hpss(y)
+    >>> librosa.display.multiplot('waveshow', y, yh, yp, sr=sr, share_properties='row',
+    ...                           labels=['Original', 'Harmonic', 'Percussive'],
+    ...                           sharex=True, sharey=True,
+    ...                           label_outer=True,
+    ...                           invert=True)
```
stefanv, line 3404: `invert` isn't documented above?

librosa/display.py
bmcfee: Ah great point.  I don't think such a thing is common, but it's easy enough to support, so why not. :grin:

librosa/display.py
bmcfee: Agreed, this is awkwardly written.  I'll rewrite.

librosa/display.py
bmcfee: I think one could invent cases like this, maybe not for waveshow vs wavebars, but with different parametrizations of one or the other.  At present I don't think we have functionality for which this would make sense, but #2008 discusses some plans for future functionality where it might.

The wrinkle here would be that we would also probably want to broadcast display function parameters (in kwargs) accordingly, and that's getting into pretty fiddly argument handling.  I thought about this a bit in developing this PR, but decided it would be out of scope.

In practice, we never have more than 2 dimensions in the axes array though, so I think supporting broadcast here would veer into "over-engineering".  It's easy enough to just call multiplot separately on slices of the axes array to get the same effect.

librosa/display.py
bmcfee: Right.  This got a little confused when I added variadic `data` support.  But yes, taken as a `*args`, data is properly a tuple of 1 or more ndarrays, so it really is `data[0], data[1], ...` etc.  The catch here is that when there is only one ndarray, then `data[0]` is interpreted as a stack of inputs to be mapped out over the axes array.

Anyway, I'll rewrite this one for clarity.

librosa/display.py
bmcfee: Hah they had better :grin: but this is meant to allow constructing axes on an existing figure that didn't already have them.  We do have some logic to ensure that all axes belong to the same figure, but maybe not to check that a provided `fig` matches `axes.figure`. 

I suspect it doesn't matter since `fig` wouldn't be used if `axes` is provided, so it shouldn't rise to the level of throwing an exception IMO.  I'm not sure if a warning is warranted here either, but maybe.

librosa/display.py
bmcfee: Yup I agree with the readability concern - here i'm aiming for stylistic consistency with other packages (eg seaborn).  

Re: listing literal options, you're totally right.  Do you have thoughts about autodoc-typehints?  I've never been a huge fan of replicating type annotations in the doc string, and this seems like a good compromise at least for the rendered doc site.  I guess someone reading docs in a repl would be able to see type annotations in the header directly.

librosa/display.py
bmcfee: Lol fair :grin:

librosa/display.py
bmcfee: Right, it's a pass-through parameter to waveshow.  Probably good to inline-comment those here.

bmcfee:
Thanks @stefanv for the thorough review!

librosa/display.py
stefanv: That makes total sense! I think I was confused by the example plot I saw — I thought you generated it with a single one of these function calls, but then later saw that you append onto existing axes.

librosa/display.py
stefanv: What I was wondering is whether `fig` is needed at all; you'll know best from experience (it's here, after all :smile:).

stefanv: (I typically use `subplots` to make `fig`, `axes` together.)

bmcfee: Right, I always use subplots the same way.  But technically it's possible to do `plt.figure()` first and then add axes later, so I didn't see any reason to forbid someone from doing that.

librosa/display.py
bmcfee: I added some explanation higher up in the docstring.  (It made more sense that way.)  NBD IMO since this is an internal function anyway.

librosa/display.py
stefanv: Yes, that's a tricky one. We're working on https://github.com/scientific-python/docstub, but it's probably not everyone's cup of tea.

librosa/display.py
stefanv: You could name the argument `func_args` instead of `**kwargs`, then users would call: `func_args={'invert': True}` or `func_args=dict(invert=True)`, which I think is a bit more intuitive.

librosa/display.py
stefanv: Indeed, commenting on a private function was by accident :)

librosa/display.py
bmcfee: Oh interesting.  If I understand right, it seems like docstub is solving the reverse problem of autodoc-typehints.  I.e., autodoc-th would use stub info to populate bits of the docstring (as rendered on the doc build), while docstub would generate the stubs from the docstring?

librosa/display.py
bmcfee: Yeah, I see that.  We generally don't do that kind of pass-through encapsulation though, so it would be a bit of an outlier relative to the rest of the package.  The places where we do isolate pass-through arguments in dicts (rather than kwargs) are those where the parameter names collide with the top-level function and the two need to be kept separate (rather than forced to be the same, like you would want for something like a sampling rate).

I mentioned #2008 elsewhere - that's an example where isolating pass-through arguments into dictionaries would be necessary because there's no other convenient syntactic structure for having arrays of kwargs.  Maybe we'll end up going that way at some point, but for the common use cases I have in mind, I think it will be a bit cleaner to just use flat kwargs.  For example:

```python
>>> multiplot('specshow', D1, D2, D3, D4, x_axis='time', y_axis='log', vscale='dBFS')
```
seems more intuitive to me than
```python
>>> multiplot('specshow', D1, D2, D3, D4, func_args=dict(x_axis='time', y_axis='log', vscale='dBFS'))
```

bmcfee:
I guess one final point to chew on here is whether the API here will be properly future-proofed.  I'm specifically thinking here about other types of visualization functions, eg annotation plots in the long-stalled out #1538 or the more fanciful #1975 and #1976.

I suspect the answer for #1538 is "yes" but the type annotations may need to widen.  For example, if /when we do decide to add annotation plots, most of the natural ways to encode that stuff are as tuples of `(array of times, array of labels)`  (where "times" may mean intervals or events).  So stacked input is directly off the table, but variadic input of a well-formed tuple sequence should be fine.

#1976 is IMO "maybe" but it gets dicey because the transform objects need to be constructed separately for each axes.  This means that it's probably going to end up being the job of the waveshow function to do it, and that means that we'd need to be able to send through some additional numerical data that would be subplot-specific.  (Ditto for #1975.)

librosa/display.py
stefanv: Yes, exactly. Either way, I agree that you ideally only want to specify this once.

bmcfee: :+1: i'll fix it here for now, but start a separate branch testing out autodoc-typehints.

bmcfee: Following up: i tried out autodoc-typehints locally and it interacts very badly with sphinx-rtd-theme.  So redundancy wins for the time being.

bmcfee:
I think this one is good to go now.  Thanks again for the reviews @stefanv - anyone have objections to merging?

PR #2018, opened 2026-05-27
Spec0 2026.06

bmcfee:
#### Reference Issue
Partially addresses #1935 

This PR updates dependencies for SPEC0 compliance in 2026.Q2.

#### Any other comments?

This may not actually work.  Our historical documentation builds require environments that worked ~2 versions back, and that may not be compatible with these revised dependencies.  Part of the purpose of this PR is to test it out and see how it fares in our CI workflows.


bmcfee:
Looks like the environments are broadly consistent.  I'm going to do a local doc build with full example generation to see if 0.10.2 actually runs on numpy 2.1.

bmcfee:
Confirming that local doc build works with historical code runs.  I had to add dead batteries (aifc and sunau) to the historical environment, and these can be removed down the road when we drop support for doc builds on version <1.0.

Requiring numpy >2.1 also bumps our minimum numba to 0.61.0 (jan 2025).  This is pretty recent, but can't really be helped.  I've revised the environments accordingly in this branch.

I'm still not 100% sure it will make sense for us to fully adopt spec0, but at least now we know that it's technically feasible.  Meanwhile I'll do a triage to see what functionality becomes unblocked or could be extended with this updated set of minimum versions.

bmcfee:
Looks like our ci-minimal environment is not actually satisfiable.

bmcfee:
Bumped soxr dependency to 1.0 to fix environment consistency issues.  Technically we could get away with 0.4, but it seemed like we may as well just move up to 1.0 since it's stable now.

PR #2034, opened 2026-06-08
Display Highlighter

### LINKED ISSUE #2033: [ENH]: Highlighter
### Problem

I'm increasingly in the situation where I want to overlay some kind of plot element (line, scatter points, whatever) on a spectrogram image.  We do this kind of thing often in the documentation, e.g. for plotting f₀ contours and the like.  And every time we do it, it takes a bit of manual fiddling to find a color value that works well against the spectrogram.

At the same time, I'm also thinking more about grayscale and print format, where we really don't have a lot of freedom in selecting colors to optimize contrast.

### Proposed solution

Maybe the solution here is not (only) to select colors carefully, but to add decorations to plot elements via path effects to help them stand out.  It's easy to go overboard with this sort of thing, but a little can go a long way.

For example, here's a spectrogram + f0 plot from our current documentation:

Image

and here's the same data rendered with a stroke outline (and using default color, narrower linewidth):

Image

vs here it is with just a default `ax.plot`:

Image

---

The above highlight is implemented easily enough via `plot(..., path_effects=[matplotlib.patheffects.withStroke(foreground='w', alpha=0.75, linewidth=3)])`.

I think we can do just a bit better though.  I used a white stroke above because the image is in magma cmap and mostly dark, but in other situations a dark stroke would be more appropriate.  We could in principle infer this from a provided axes object (with an image drawn on it) and computing, say, the median value, and mapping it to a luminance to determine whether a light or dark shadow should be used.  (This could of course be bypassed.)

The API I have in mind would look something like:

```python
fig, ax = plt.subplots()
librosa.display.specshow(...)
ax.plot(..., path_effects=librosa.display.highlight())
```
or even (not mutually exclusive):
```
artist = ax.plot(...)
librosa.display.highlight(artist)
```

I don't want to over-engineer this, but I think a fairly simple method and helper function could make a lot of plotting code much cleaner going forward, especially for print media.

bmcfee:
#### Reference Issue
Fixes #2033 

`librosa.display.highlight` - a helper function for generating path effects for artists to improve contrast against dense backgrounds (eg images or scatterplots).

This should make it easier to generate reasonable-looking plots of things like f0 or tempo estimation curves overlaid on spectrogram-like images without the line artists blending into the image display.

The logic here is as described in #2033, with a few extensions and embellishments to enhance usability.  Some key features:

- User-specified luminance threshold to determine light vs dark
- Controllable bright and dark highlights (white and black as defaults)
- Explicit override option for bypassing the luminance inference and just making a highlight effect directly
- Can apply directly to an artist (if provided) or return a path effects list to be added to an existing artist later

#### Any other comments?

I've revised many of the docstring examples to make use of the new function.  Otherwise this PR has no impact on functionality in the rest of the package.

Initial drafts of the function were drafted by gemini, with substantial rewriting by myself.  I'd be curious for feedback from anyone else who might use this.


bmcfee:
Two specific points I'd like feedback on:

1. Right now the stroke parameters have a couple of defaults that are set via kwargs.setdefault (linewidth and alpha).  This is mostly to keep the api as simple as possible, since we'll be passing kwargs through anyway.  But I could imagine bumping this up to the function header to make it more transparent.
2. The inference around transparent axes is a little dodgy.  Probably this could be smarter. Probably also this affects essentially nobody in perpetuity throughout the universe.

bmcfee:
Test failures are coming from codecov authentication errors - but coverage on the delta here is 100% and all tests are passing.

PR #2035, opened 2026-06-08
Modernize types in `librosa._typing`

jorenham:
#### Reference Issue

None that as I'm aware of

Since you now require `python>=3.12`, the fancy new PEP 695 `type` syntax can be used. Additionally, I synced the copied numpy type aliases with numpy, which have been improved since  ([numpy src](https://github.com/numpy/numpy/blob/c80693c5c1052e9355fe1a1fc6b3f57dec24c89c/numpy/_typing/_scalars.py)).
I also removed some unnecessary quotes around the numpy scalar types in `_typing.py`, because those are no longer needed since NumPy 1.23 if I remember correctly.

#### Any other comments?

The SciPy type aliases reminded me of https://github.com/scipy/scipy-stubs/issues/441. Is that something you'd be interested in here?

### DIFF
--- a/librosa/_typing.py
+++ b/librosa/_typing.py
@@ -1,53 +1,44 @@
-from __future__ import annotations
+# ruff: noqa: PYI047
 
 from collections.abc import Sequence
-from typing import Any, Callable, Generator, List, Literal, Never, Tuple, TypeVar, Union
+from typing import Any, Callable, Generator, Literal, Never
 
 import numpy as np
 import scipy.sparse as sp
 from numpy.typing import ArrayLike
 
 # RNG types
-SeedLike = Union[int, np.integer, Sequence[int], np.random.SeedSequence]
-RNGLike = Union[np.random.Generator, np.random.BitGenerator]
+type SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
+type RNGLike = np.random.Generator | np.random.BitGenerator
 
 
-_WindowSpec = Union[str, Tuple[Any, ...], float, Callable[[int], np.ndarray], ArrayLike]
-_T = TypeVar("_T")
-_IterableLike = Union[List[_T], Tuple[_T, ...], Generator[_T, None, None]]
-_SequenceLike = Union[Sequence[_T], np.ndarray]
-_ScalarOrSequence = Union[_T, _SequenceLike[_T]]
+type _WindowSpec = str | tuple[Any, ...] | float | Callable[[int], np.ndarray] | ArrayLike
+type _IterableLike[T] = list[T] | tuple[T, ...] | Generator[T, None, None]
+type _SequenceLike[T] = Sequence[T] | np.ndarray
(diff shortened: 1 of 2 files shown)


bmcfee:
Thanks!  LGTM.

CI failures are due to codecov auth failing on the old action, so please disregard.  (I have it fixed in another branch yet to merge.)  All that matters for this PR is the linter action anyway.

I think we have bigger work to do in modernizing the annotations outside of the _typing helper module, but that can be for another day.

*EDIT*: and yeah, I think the issue you linked to could be helpful.  I don't have a great sense of how pervasive it is in our code / how much hackery we've done do dance around it.  I expect not very much, but if the stubs can be strengthened to make our lives easier, I'm for it. :grin:

jorenham:
> I think we have bigger work to do in modernizing the annotations outside of the _typing helper module, but that can be for another day.

That's what my initial plan was. But then I figured that a first PR that touches every single file might not be appreciated, so I started with this module :stuck_out_tongue_closed_eyes:

bmcfee:
> But then I figured that a first PR that touches every single file might not be appreciated, so I started with this module 😝

Hah.  Well I'm happy to merge this one now and deal with the rest in a second PR.  Or, if you have a quick way to change over all the Unions and Optionals (can ruff do this?), we could take it here.  Either way is fine by me - I've been staring at the type annotations enough over the past week while modernizing our linting / type checking that it's not a huge deal to me.

jorenham:
> Or, if you have a quick way to change over all the Unions and Optionals (can ruff do this?), we could take it here.

Something like `ruff check --select=UP007 --select=UP045 --fix` should do the trick, and then another `ruff check --fix` to clean up the unused imports. There are more relevant pyupgrade rules that you might want to use here though: https://docs.astral.sh/ruff/rules/#pyupgrade-up

bmcfee:
Great!  How about I merge this one then, and take over the grunt work myself?

jorenham:
> _EDIT_: and yeah, I think the issue you linked to could be helpful. I don't have a great sense of how pervasive it is in our code / how much hackery we've done do dance around it. I expect not very much, but if the stubs can be strengthened to make our lives easier, I'm for it. 😁

I think it will be a separate package, something like `scipy-typing`, so that you'll also be able to use it at runtime. It'll probably have to become a required dependency though (unless you always import it in `if TYPE_CHECKING`).

jorenham:
> Great! How about I merge this one then, and take over the grunt work myself?

I'd hate to have to review such a PR myself haha, so that's probably the easiest route.

bmcfee:
Alright.  Thanks again!

bmcfee:
Lol.

```
...
Found 768 errors.
[*] 768 fixable with the `--fix` option.
```

jorenham:
> Lol.
> 
> ```
> ...
> Found 768 errors.
> [*] 768 fixable with the `--fix` option.
> ```

Haha yea; enabling all of pyupgrade (`--select=UP`) is even more fun :P

PR #1963, opened 2025-07-18
Expire deprecations for 1.0 release

bmcfee:
#### Reference Issue

This PR expires deprecations scheduled for the 1.0 release.

#### What does this implement/fix? Explain your changes.

- [x] audioread
- [x] `filename` argument in `stream`
- [x] `None` rest-type in VQT
- [x] `set_fftlib` / `get_fftlib` 
- [x] `win_length` in `yin` and `pyin`
- [x] `x_axis` parameter in `waveshow`
- [x] `constant_q` / `constant_q_lengths` in `filters`

#### Any other comments?

I'll also take this opportunity to tighten up our dependency constraints for the historical doc builds, aligning to 0.10, 0.11, and 1.0 releases.

PR #2037, opened 2026-06-09
Fill in missing type annotations

jorenham:
I noticed on the typestats dashboard (https://jorenham.github.io/typestats/dashboard/report/#librosa) that librosa is *almost* at 100% type coverage. Most of the missing type annotations turned out to live in `librosa.display`, and the remaining ones in `core/intervals.py` and `beat.py`. By filling those in, type coverage increased from 97.35% to 100%.

Besides typing, explicit (attribute) annotations can also be considered as a form of documentation. And I suppose that the general "explicit is better than implicit" argument also applies here.

### DIFF
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -20,44 +20,14 @@ import scipy
 from . import core, onset, util
 from .feature import fourier_tempogram
 from .feature import tempo as _tempo
-from .util.decorators import moved
 from .util.exceptions import ParameterError
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
-    from typing import Any
-
     import scipy.stats
-    from typing_extensions import deprecated
 
     from ._typing import _FloatLike_co
 
-__all__ = ["beat_track", "tempo", "plp"]
-
-
-if TYPE_CHECKING:
-    @deprecated(
-        "This function was moved to 'librosa.feature' in librosa version 0.10.0."
-        "This alias will be removed in librosa version 1.0."
-    )
-    def tempo(  # noqa: D103
-        *,
-        y: np.ndarray | None = None,
-        sr: float = 22050,
-        onset_envelope: np.ndarray | None = None,
-        tg: np.ndarray | None = None,
-        hop_length: int = 512,
-        start_bpm: float = 120,
-        std_bpm: float = 1.0,
-        ac_size: float = 8.0,
-        max_tempo: float | None = 320.0,
-        aggregate: Callable[..., Any] | None = np.mean,
-        prior: scipy.stats.rv_continuous | None = None,
-    ) -> np.ndarray: ...
-else:
-    tempo = moved(moved_from="librosa.beat.tempo", version="0.10.0", version_removed="1.0")(
-        _tempo
-    )
+__all__ = ["beat_track", "plp"]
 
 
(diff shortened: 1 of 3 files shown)


bmcfee:
Ah nice, thanks again!

I'm happy to lump this into one PR since it's all conceptually related.  Regarding those three remaining ones:

```
   --> librosa/beat.py:36  tempo
36 | tempo = moved(moved_from="librosa.beat.tempo", version="0.10.0", version_removed="1.0")(
```
This one is an expired deprecation that I haven't removed yet, so not a big deal directly.  However, it does show that our deprecation decorator is not preserving type annotations the way it probably should, so maybe there's something to check there?



> ```shell
>    --> librosa/core/intervals.py:22  imsgpack
> 22 | with _resource_file("librosa.core", "intervals.msgpack") as imsgpack, imsgpack.open("rb") as _fdesc:
> ```

Similar story here - the resources loader wrapper doesn't have an output type, easy fix.



> ```shell
>   --> librosa/core/intervals.py:24  INTERVALS
> 24 |     INTERVALS = msgpack.load(_fdesc, strict_map_key=False)
> ```

This is a static data file, should be `dict[float, dict[int, int]]`.

bmcfee:
As an aside, I also noticed that soundfile 0.14 dropped this week, so we'll eventually be able to remove the ignore-missing-imports line from setup there.  I haven't done so yet because the conda packaging over there hasn't caught up yet.  Just mentioning it here so that I don't forget about it later on.

jorenham:
> I'm happy to lump this into one PR since it's all conceptually related.

Ok great; I'll get on that then :ok_hand: 

> However, it does show that our deprecation decorator is not preserving type annotations the way it probably should, so maybe there's something to check there?

Typestats (or rather `pyrefly report`, which powers it) doesn't do type inference. The heuristic here is that it requires module-level constants to be annotated, unless the RHS is a trivial literal (e.g. `= True`), following the typing spec.

librosa/beat.py
```diff
@@ -24,17 +24,40 @@
...
+        std_bpm: float = 1.0,
+        ac_size: float = 8.0,
+        max_tempo: float | None = 320.0,
+        aggregate: Callable[..., Any] | None = np.mean,
+        prior: scipy.stats.rv_continuous | None = None,
+    ) -> np.ndarray: ...
+else:
+    tempo = moved(moved_from="librosa.beat.tempo", version="0.10.0", version_removed="1.0")(
+        _tempo
+    )
```
jorenham: Clearly overkill; but I mostly wanted to show how functions can be deprecated so that static type-checkers also warn users about it.

jorenham: Let me know if I should revert or simplify this to e.g. `: Callable[..., Any]`.

bmcfee: Oh we should be able to just nuke that line altogether.  I somehow missed it previously in #1963

jorenham:
```
❯ uvx --from git+https://github.com/jorenham/typestats typestats check librosa
coverage:   100.00%
typable:    1656
typed:      1613
any:          43
```

:tada:

bmcfee (approved):

bmcfee:
Awesome!  Thanks again :shipit:

PR #2039, opened 2026-06-09
avoid turning scalars into 0-d arrays in `core.{convert,spectrum}` and `util.utils`

jorenham:
I was taking a look at the typing situation of `librosa.core.convert`, and noticed that many of the overloaded function signatures did not match the runtime behavior. Those overloads describe that for scalar-like input, a numpy scalar (e.g. `np.floating`) is returned. But because of how `np.asanyarray` is used, that wasn't correct; they'd return 0-d arrays in those cases.

I used the `[()]` trick to convert 0-d arrays (back) into "bare" scalars:

```pycon
>>> import numpy as np

>>> np.asanyarray(42)
array(42)
>>> _[()]
np.int64(42)

>>> np.asanyarray([42])
array([42])
>>> _[()]
array([42])

>>> np.int_(42)[()]
np.int64(42)
```

If returning 0-d arrays was intentional, then feel free to close this, and I'll then instead work on a PR that fixes the overloads (and docstrings). 
FWIW; from a static typing perspective, it's a lot easier if everything always returned an array, as that would allow us to remove a bunch of overloads. But perhaps that might become an issue for some users. Because `np.generic` and `np.ndarray` are something like 95% ducktype-compatible, which isn't 100%.

~I didn't look at other modules yet, but I figured I'd start with this one to see what you think, before moving on to other modules.~

### DIFF
--- a/librosa/core/convert.py
+++ b/librosa/core/convert.py
@@ -126,7 +126,7 @@ def frames_to_samples(
     if n_fft is not None:
         offset = int(n_fft // 2)
 
-    return (np.asanyarray(frames) * hop_length + offset).astype(int)
+    return (np.asanyarray(frames) * hop_length + offset).astype(int)[()]
 
 
 @overload
@@ -205,7 +205,7 @@ def samples_to_frames(
         offset = int(n_fft // 2)
 
     samples = np.asanyarray(samples)
-    return np.asarray(np.floor((samples - offset) // hop_length), dtype=int)
+    return np.asarray(np.floor((samples - offset) // hop_length), dtype=int)[()]
 
 
 @overload
@@ -409,7 +409,7 @@ def time_to_samples(
     array([    0,  2205,  4410,  6615,  8820, 11025, 13230, 15435,
            17640, 19845])
     """
-    return (np.asanyarray(times) * sr).astype(int)
+    return (np.asanyarray(times) * sr).astype(int)[()]
 
 
 @overload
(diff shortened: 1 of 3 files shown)


bmcfee:
Ah nice, I never thought of doing that. :grin: 



> FWIW; from a static typing perspective, it's a lot easier if everything always returned an array, as that would allow us to remove a bunch of overloads. But perhaps that might become an issue for some users.

For sure - but I think users actually expect scalar in/scalar out behavior, especially since things like midi numbers are often used as indices into lists (not ndarrays).

bmcfee (approved):
This LGTM; tried it out locally, quacks like I'd expect.  I'm not sure if you were planning to identify other parts of the code where similar changes would make sense, but the only others that came to mind on a quick skim were util.utils.abs2 and util.utils.phasor

jorenham:
> I'm not sure if you were planning to identify other parts of the code where similar changes would make sense, but the only others that came to mind on a quick skim were util.utils.abs2 and util.utils.phasor

I went over all `@overload`ed functions, and the only other cases I found were `core.spectrum.{power_to_db, amplitude_to_db}` and (indeed) `util.utils.phasor`. 

If you don't mind, I'll address those here as well then.

bmcfee:
> the only other cases I found were `core.spectrum.{power_to_db, amplitude_to_db}`

Did the inverse transforms (db_to_power/amplitude) not also have this issue?  Or is it easier there because they're all one-liner expressions?

jorenham:
> > the only other cases I found were `core.spectrum.{power_to_db, amplitude_to_db}`
> 
> Did the inverse transforms (db_to_power/amplitude) not also have this issue? Or is it easier there because they're all one-liner expressions?

They indeed didn't have the same issue, because `np.power` with scalar input already results in scalar output.

bmcfee (approved):

bmcfee:
:shipit: thanks again!

PR #2043, opened 2026-06-10
improved `dtype_r2c` and `dtype_c2r` typing in `util.utils`

jorenham:
While working on #2039 I noticed that the two `dtype_*`  functions `librosa.util.utils` had a unnecessarily broad `DTypeLike` return type. This narrows it, while carefully avoiding a bunch of mypy false positive errors (caused by issues in mypy's overload selection algorithm in the prescence of gradual types).

### DIFF
--- a/librosa/util/utils.py
+++ b/librosa/util/utils.py
@@ -2315,7 +2315,9 @@ def stack(arrays: list[np.ndarray], *, axis: int = 0) -> np.ndarray:
         return result
 
 
-def dtype_r2c(d: DTypeLike, *, default: type | None = np.complex64) -> DTypeLike:
+def dtype_r2c(
+    d: DTypeLike, *, default: type[np.complexfloating] = np.complex64
+) -> np.dtype[np.complexfloating]:
     """Find the complex numpy dtype corresponding to a real dtype.
 
     This is used to maintain numerical precision and memory footprint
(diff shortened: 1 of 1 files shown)


librosa/util/utils.py
```diff
@@ -2414,14 +2417,13 @@ def dtype_c2r(d: DTypeLike, *, default: type | None = np.float32) -> DTypeLike:
     >>> librosa.util.dtype_r2c(np.complex128)
     dtype('float64')
     """
-    mapping: dict[DTypeLike, type] = {
+    mapping: dict[np.dtype[np.complexfloating], type[np.floating]] = {
         np.dtype(np.complex64): np.float32,
         np.dtype(np.complex128): np.float64,
-        np.dtype(complex): np.dtype(float).type,
```
bmcfee, line 2420: @jorenham My recollection of this specific line was to provide a mapping for whatever the system's native precision is (in case it's compiled for something other than float64).  Is this not needed?

jorenham: That's indeed not needed, `np.dtype(complex)` will be either `complex64` or `cmoplex128`.

bmcfee: Hrm.  Just digging back through the blame on this one for my own sake.  I guess what happened is that this was originally a catch-all for `np.complex → np.float`, which then became `complex → float` when the type aliases were deprecated.

Original post is here https://github.com/librosa/librosa/issues/1143#issuecomment-645300595 wherein there was some concern about long type support and variation across platforms.  I guess that's thoroughly by the wayside now.

jorenham: since NumPy 2 a lot of those system-dependent dtype oddities have been improved. `float64` is now always a `double`, and `float32` always a (single) `float`. For floating-point data types, the only remaining system-dependent stuff is related to `longdouble`, which librosa doesn't seem to use (which is a good decision IMO).

bmcfee: Got it, thanks!  I'll admit to not having entirely followed all of the changes in numpy 2 - we bumped the requirement here mainly by adopting spec0, and previously had been setting minimal versions to the oldest version that had all the functionality we actually used.  Makes sense that the transition would leave room for a lot of cleanups like this.

bmcfee:
:shipit: thanks again (again)!

PR #2044, opened 2026-06-11
Waveplot styling

### LINKED ISSUE #2042: AdaptiveWaveplot proxy artist should match steps, not envelope
AdaptiveWaveplot creates a proxy artist for legend labeling:

https://github.com/librosa/librosa/blob/9edf171ffc3755999b930402059468cb77778f61/librosa/display.py#L862-L871

Presently, this is just a simple rectangle patch, and therefore only shows the color and transparency of the waveform.  This is mostly fine, but can get confusing when zoomed in (and moving to steps mode), especially when there are other legend entries derived from line plots, as the waveshow legend entry will appear as a patch instead of a line.

This gets dicier in grayscale displays, where line style indicates the label, not color.  Of course, rectangular patches don't have line styles, so there's a visual mismatch to the legend.

I think it would be unambiguously better to style the proxy artist after the steps plot instead of the fill_between.  Most of the time (ie, in envelope mode), there won't be any real difference.  But if the user has gone to the trouble to set line style, markers, etc, they should be respected in the legend.

bmcfee:
#### Reference Issue
Fixes #2042 

This PR implements a custom legend handler for waveshow.  This fixes #2042, and makes it so that legend entries match the styling of the high-resolution line plot, in both inverted and regular modes.

#### Any other comments?

If CI passes, this should be good to go.


bmcfee:
Note: I regenerated all waveshow image regression baselines, though many of them did not actually change.

PR #2049, opened 2026-06-13
reformat `**/__init__.pyi` re-exports

jorenham:
Ruff doesn't complain about these trailing commas, but I think that without them, things are more readable. The juicy - diff is also just nice I suppose.

Great use of `.pyi` stubs btw.


bmcfee (approved):
LGTM, thanks!

PR #2051, opened 2026-06-13
Multichannel db

### LINKED ISSUE #1778: Better handling of multichannel in decibel scaling

When we implemented multichannel support #1130 , one subtle issue that pops up is the calculation of a reference amplitude in decibel scaling.  This is because when reference is given as a function (eg, `ref=np.max`), it is applied globally to the entire input to compute a scalar value:

https://github.com/librosa/librosa/blob/5ca70f5d57fc597660452f011ffff55f9c36ed4b/librosa/core/spectrum.py#L1866-L1868

This makes sense in the single channel case.  However, in the multichannel case, it can break assumptions of channel independence.  This was first noticed in the mfcc calculation, where deriving mfccs (which depends on db scaling) can produce different results on a multichannel input than if it was applied independently to each channel.  It's also popping up in #1766 by way of onset detection (again, depending on db scaling).

Instead of aggregating globally, we could aggregate only over the trailing dimensions.  Using `keepdims=True` should result in channel-wise independent processing.  This would be a breaking change, but I expect it would be more in line with expected behavior so it could be worth doing.

There is a subtlety here though.  It's not always clear how many trailing dimensions should be included.  Most of the time it would be two (time and frequency); sometimes it could be more (eg if patch processing spectra) or less (db scaling single-channel measurements).  Probably we'd need some API to expose the aggregation dimensions here, and that would require some thought.

bmcfee:
#### Reference Issue
Fixes #1778 

This PR adds channel-independent processing for decibel conversion.  This is implemented by adding target axes to `*_to_db` functions, which are inferred (for ndim>1) as `axes=(-2,-1)`.

This ensures, for example, that calculating mfcc's on multichannel inputs produces identical results to channel-independent calculation.

#### Any other comments?

I had initially planned on exposing the target axes through the api (e.g. in `mfcc`).  I think I'll walk this back though.  I think it's generally more correct to handle this channel-independently (as in this PR now by default).  It's still possible for a user to pass in a precalculated dB spectrum, and if they really want things the way they were before, they can compute the dB scale with `axes=(-3,-2,-1)` (or whatever) up front.  We shouldn't make it too easy for users to do the wrong thing here.

I opted against implementing this ref targeting for `zero_crossings`, at least for now, as it's doing some pretty different things than db scaling.

The one lingering question for me here is how best to handle trimming and splitting.  Probably I just need to think on that a bit more, but i wanted to get this PR off the ground first to make sure the rest of the tests are working properly.

### DIFF
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -2046,7 +2046,7 @@ def amplitude_to_db(
 
     magnitude = np.abs(S)
 
-    if axes is None:
+    if axes == "auto":
         if magnitude.ndim >= 2:
             axes = (-2, -1)
         elif magnitude.ndim == 1:
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -1719,7 +1719,7 @@ def power_to_db(
     ref: float | Callable = ...,
     amin: float = ...,
     top_db: float | None = ...,
-    axes : None | Literal["auto"] | int | Sequence[int] = None,
+    axes: None | Literal["auto"] | int | tuple[int] = ...,
 ) -> np.floating[Any]: ...
 
 
@@ -1730,7 +1730,7 @@ def power_to_db(
     ref: float | Callable = ...,
     amin: float = ...,
     top_db: float | None = ...,
-    axes : None | Literal["auto"] | int | Sequence[int] = None,
+    axes: None | Literal["auto"] | int | tuple[int] = ...,
 ) -> np.ndarray: ...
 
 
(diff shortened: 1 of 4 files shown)


librosa/core/spectrum.py
```diff
@@ -1853,9 +1864,23 @@ def power_to_db(
     else:
         magnitude = S
 
+    if axes is None:
+        if magnitude.ndim >= 2:
+            axes = (-2, -1)
+        elif magnitude.ndim == 1:
+            axes = (-1,)
+        else:
+            axes = None
```
bmcfee, line 1875: This actually makes for a rather strange problem.  We use `axes=None` to trigger default target axes inference, but this makes it impossible to target all axes via `None` up front, which is a reasonable behavior too.  You can hack around this by doing something like `axes=np.arange(x.ndim)`, but that feels very clumsy.

I guess an alternative here would be to use some kind of sentinel default value other than `None`.  Gonna ponder this for a bit.

PR #2053, opened 2026-06-16
Fix decorator callable types in `librosa.util.decorators`

jorenham:
The decorator functions in `util.decorators` all had the same generic type parameter scoping issue, where the generic type variables were bound to the decorator function itself, rather than the returned decorator. The solution is to return a "callback protocol" with a generic `__call__` method, so that the returned callable type *itself* is generic, rather than the "outer" decorator function.

FWIW, this is one of those mistakes I've made a lot of times myself.

### DIFF
--- a/librosa/util/decorators.py
+++ b/librosa/util/decorators.py
@@ -12,19 +12,19 @@ import numpy as np
 from decorator import decorator
 
 if TYPE_CHECKING:
-    from typing import Any, Callable, Iterable, ParamSpec, TypeVar
+    from collections.abc import Callable, Iterable
+    from typing import Protocol, type_check_only
 
     from numpy.typing import DTypeLike
-    P = ParamSpec("P")
-    R = TypeVar("R")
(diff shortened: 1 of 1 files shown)


bmcfee:
This looks fine to me.  I understand the problem, but I'll admit that the solution here completely baffles me :laughing:

jorenham:
I like to think of

```py
class _Decorator(Protocol):
    def __call__[**P, R](self, fn: Callable[P, R], /) -> Callable[P, R]: ...
```

as something like

```
[**P, R](Callable[P, R]) -> Callable[P, R]
```

maybe that helps :shrug:

bmcfee:
Yeah I get it in that it's saying that the decorator preserves the type signature of the function it's decorating.  Just that `[**P, R](Callable...)` looks pretty weird to me.  But if that's the way it's done, I'm not gonna argue with it. :grin:

bmcfee:
Thanks again!

PR #2055, opened 2026-06-17
Stream resampling

### LINKED ISSUE #1518: Stream resampling

We currently have two interfaces to ingesting audio: `load` and `stream`.  Aside from the obvious functional differences between the two, a key difference is that the `stream` interface does not support resample-on-load, while `load` has this enabled by default.  This is primarily because the resampling libraries we support (scipy, resampy, libsamplerate, soxr) do not all support stream-based processing, though the latter two do.

It would be great if we could smooth this gap over, though it does introduce some technical hurdles.

For a restricted class of `res_type` values (namely, samplerate and soxr resamplers), we could in principle support resample-on-load within `stream`.  We could think of this as having a second layer of processing in between the soundfile chunk generator and our block generator which handles the conversion.

From an interface perspective, this ought to be relatively straightforward, with the caveat that block and frame parameters should be understood to operate at the target sample rate, not the native rate.

This does raise a question of whether we could guarantee numerical equivalence (assuming the same resamplers are used) between load and stream.  Getting this right might require some modification of the chunking parameters in soundfile to ensure that the resampler has enough future context to perform the interpolation.  That in turn will depend on the specifics of each resampler.

:shrug: 

**Additional context**

If at all possible, I'd like to strive for API consistency here.  This will require some hard decisions:

- If load resamples by default, should stream also?
  - I'd argue yes, even though we don't yet.  This will break API.
- If stream resamples, should it use the same default implementation as load?
  - Again, I'd argue yes.  However, the default resampler (resampy) does not support streaming (and won't) - so we'll have to change the default.  This will break API.
- If we change the default resampler, what are the implications for package dependencies?
  - At present, resampy is the default because it's simple, reliably packaged, permissively licensed, and cross-platform.  Back in 2016, it was the only library that checked all these boxes, but it's a different world now.
  - Off the cuff, my preference would be to move to soxr as the default.  The core library is very mature, though the python wrapper is young and has a small bus factor.
  - Samplerate would also be fine, though in my experience, soxr is both more accurate and faster.
  - Would we relegate resampy to an optional dependency, and promote (soxr|samplerate) to a strict requirement?  (I think yes - I don't want to increase the number of strict requirements, and would prefer to reduce them when possible.)

bmcfee:
#### Reference Issue
Fixes #1518 

This PR rewrites the simple block generator to support in-place sample rate conversion using soxr.

#### Any other comments?
There is a bit of an API consistency mismatch here, as the default for `sr=` is `None` here to preserve backward compatibility, while the default for `load` is 22K.  We should resolve this at some point, perhaps in 1.1 or 1.2.

The code is feature complete, but I haven't written new tests yet.  Right now I just want to start CI to ensure that the previous test suite passes.

Code co-authored by gemini pro 3.1.

### DIFF
--- a/docs/examples/plot_pcen_stream.py
+++ b/docs/examples/plot_pcen_stream.py
@@ -51,6 +51,7 @@ stream = librosa.stream(filename, block_length=16,
                         frame_length=n_fft,
                         hop_length=hop_length,
                         mono=True,
+                        sr=sr,
                         fill_value=0)
 #######################################################################
 # For this example, we'll compute PCEN on each block, find the maximum
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -219,7 +219,7 @@ def _align_step_size(target_step, target_sr, orig_sr):
     return int(aligned_native_step)
 
 
-@future_default(param_name="sr", old_default=None, new_default=22050, version="1.2.0")
+@future_default(param_name="sr", old_default=None, new_default=22050, version="1.1.0")
 def stream(
     path: str | int | sf.SoundFile | BinaryIO,
     *,
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -215,7 +215,7 @@ def _align_step_size(target_step, target_sr, orig_sr):
             f"fractional sample advance at original sampling rate {orig_sr}."
         )
 
-    return int((target_step * orig_sr) / target_sr)
+    return int(aligned_native_step)
 
 def stream(
     path: str | int | sf.SoundFile | BinaryIO,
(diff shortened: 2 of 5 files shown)


bmcfee:
I think this is ready for a proper review now.  A couple of things to chew on:

1. Should we allow approximate frame alignment?  Right now we require frame size to be integral at the native sampling rate to ensure numerical equivalence to batch processing, but I could imagine relaxing this (or optionally relaxing this) by a `strict=` parameter.  Maybe a user doesn't care about comparison to the batch and just wants something close.
2. ~~Should we already start making noise about using the default `sr=None`?~~ Done.
3. ~~Probably there are some optimizations we could make here, in terms of copying and buffer allocation.  I figured I'd wait until we have a solid test suite in place to start messing with that.~~ done.
4. ~~I think the tests could stand to be refactored into simpler, more focused test functions instead of one gigantor parameter grid.~~ done.

bmcfee:
I think I've addressed everything above that needed addressing.  The implementation seems solid to me now.  I slept on the idea of allowing fractional / rounding sample indices since I think the constraints as they are now are correct, and we really should be enforcing sample-consistent behavior between stream and load.

I put in the future warning about the default `sr`, and a new decorator function to help with this kind of thing. (I expect we'll be seeing it get some use on the run into 2.0, whenever that happens.)

Issue #2056, opened 2026-06-18
[ENH]: Mixing functions should accept out=


bmcfee:
### Problem

This came up in implementing #2055 , but there are circumstances where it would be helpful to be able to give a pre-allocated output buffer to functions like `to_mono`, rather than having it allocate a new buffer every time.

This specifically comes up in `stream` where we do a mono downmix on every block produced by soundfile, and we could in principle reuse a working buffer instead of constantly creating new ones.

### Proposed solution

Add `out=` parameter to `to_mono`, `to_stereo`, and `to_multi`.

PR #2058, opened 2026-06-19
Add out= parameter to to_mono for in-place downmixing

ejwong24:
Implements #2056 against the redesigned 1.0 mixing API.

Adds an optional keyword-only `out=` buffer to `to_mono`, `to_stereo`, and `to_multi`, letting callers reuse a pre-allocated array instead of allocating on every call (e.g. per-block downmix in `stream()`).

### Changes
- New `_init_output(out, shape, dtype)` helper: validates a caller-provided buffer (exact shape + dtype, else `ParameterError`), zero-fills it, and returns it; allocates fresh when `out=None`.
- `out=` added to all three mixing functions, with docstring updates and a "same object as `out`" return note.
- Removed `@cache(level=20)` from the three functions (incompatible with in-place `out=`, per review).
- Tests for correctness, `is out` identity, the zero-fill path (via `np.empty`), and shape/dtype validation errors.

### Verified locally (1.0.0dev gate)
ruff, numpydoc lint, mypy (librosa + tests), codespell, pytest — all green.


bmcfee:
Thanks @ejwong24 for this PR - I've retargeted it to the 1.0 dev branch, and this raised some nontrivial conflicts.  (Note: to_mono and related functions have changed quite a bit since 0.11 due to #1964 ) Please resolve these before we can begin testing and review.

librosa/core/audio.py
```diff
@@ -590,8 +590,32 @@ def loadx(key: str, *, hq: bool | None = None, **kwargs: Any) -> tuple[np.ndarra
...
+    out[...] = 0
+    return out
+
+
 @cache(level=20)
-def to_mono(*signals: np.ndarray, pad: bool = True, norm: bool = True) -> np.ndarray:
+def to_mono(
+    *signals: np.ndarray,
+    pad: bool = True,
+    norm: bool = True,
```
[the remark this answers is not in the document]
bmcfee: This one is a valid critique.  I'd actually just simplify this and remove the caching property on this function altogether; having out= support is a greater benefit and dancing around trying to do both will bring unwanted complexity.

bmcfee:
Thanks for updating this.  Looks good so far, though the caching note from the bot is valid.

I'd also add that we may want to think about how to handle size in the out= array.  Right now, we implicitly assume the size matches exactly, which is consistent with numpy semantics elsewhere.  On the other hand, the primary consumer of this will be `librosa.stream`, where the final block will be under-sized when `fill` is disabled.  This would eventually lead to a problem where we provide an over-sized output array to the function, resulting in an error.

We could fix that at the call site, eg detect the shape mismatch and truncate the output view first.  Or we could handle it gracefully here.  Probably it's better to handle it explicitly at the call site, but I wanted to surface the issue here in case anyone has thoughts on the matter.

librosa/core/audio.py
```diff
@@ -590,8 +590,31 @@ def loadx(key: str, *, hq: bool | None = None, **kwargs: Any) -> tuple[np.ndarra
...
+    If ``out`` is ``None``, return a new zero-filled array of the given shape
+    and dtype.  Otherwise ``out`` must match ``shape`` and ``dtype`` exactly;
+    it is zeroed in place and returned.
+    """
+    if out is None:
+        return np.zeros(shape, dtype=dtype)
+    if out.shape != shape:
+        raise ParameterError(f"out has shape={out.shape}, but expected {shape}")
+    if out.dtype != dtype:
+        raise ParameterError(f"out has dtype={out.dtype}, but expected {dtype}")
```
bmcfee: I think we can actually relax this one.  All we really need is that out.dtype is at least as good as the input dtype.  Presumably if the user has explicitly allocated an array with a given precision, that takes precedence over what we can infer from the input data.

The logic here could be something like

```
if out.dtype is not np.promote_dtypes(out.dtype, dtype): ...
```

ejwong24:
Pushed `ef2c72af` (rebased onto your 1.0.0dev merge).

**Caching:** removed `@cache(level=20)` from all three of `to_mono`/`to_stereo`/`to_multi` (and the stale "caches at level 20" notes). Agreed the `out=` contract is much simpler without it.

**Over-sized `out` from `stream`:** I kept the in-function check strict (exact shape + dtype → `ParameterError`), matching numpy `out=` semantics. Per your lean, the `stream` call site can pass a truncated view (`out[..., :n]`) for the final short block rather than the function silently accepting an over-sized buffer. Happy to wire that into `stream` in this PR, or leave it as a follow-up — your call.

**Scope:** updated the PR description; this intentionally covers all three now (the Copilot note was against the pre-rebase description).

librosa/core/audio.py
```diff
@@ -800,7 +834,7 @@ def to_stereo(
     dtype = np.promote_types(left.dtype, right.dtype)
 
     # Create an empty stereo output buffer
-    output = np.zeros((2, size), dtype=dtype)
+    output = _init_output(out, (2, size), dtype)
     if downmix:
         output[0] = to_mono(left, norm=norm)
         output[1] = to_mono(right, norm=norm)
```
bmcfee: This should pass through the slices of `output` into the to_mono calls, since we've already gone to the trouble of allocating it here.  Otherwise it's a double allocate.

librosa/core/audio.py
bmcfee: This one is tricky.  We have a double allocate as in to_stereo, since out is not passed through.  But the length fixing makes it so that a plain pass-through of out= won't work either.

bmcfee:
> Per your lean, the `stream` call site can pass a truncated view (`out[..., :n]`) for the final short block rather than the function silently accepting an over-sized buffer. Happy to wire that into `stream` in this PR, or leave it as a follow-up — your call.

That would be great, thanks!

Apologies for the somewhat disjointed CR on this one - doing this from an airport after a redeye and brains aren't operating at full capacity :sweat_smile:

ejwong24:
Done — pushed `cc6c1565`, all three addressed:

**`to_stereo` double-allocate:** the internal `to_mono` calls now write straight into the output rows (`out=output[0]`/`[1]`). One subtlety: `output` is the promoted dtype but `left`/`right` may be narrower, so I cast them to the shared dtype first (`copy=False`, free when they already match) — otherwise the pass-through trips the strict dtype check.

**`to_multi`:** pass-through when the signal already matches the output length (`y.shape[-1] == size`); fall back to the old allocate-and-`fix_length` only when a length adjustment is actually needed. Covers the common all-equal-length case without allocating while keeping the padding/trimming path correct. Open to a cleverer always-in-place approach if you'd prefer, but this seemed like the cleanest correctness/complexity trade.

**`stream`:** allocates one reusable mono buffer up front and writes each block's downmix into a truncated view (`mono_buffer[:n]`), so only the largest block's worth of memory is allocated, once. The buffer is consumed into the ring buffer / resampler before the next block, so reuse is safe.

Added a mixed-dtype regression test for the `to_stereo`/`to_multi` pass-through. Full 1.0.0dev gate green locally (ruff, numpydoc, mypy, codespell, pytest).

librosa/core/audio.py
bmcfee: Coming back to this one, since everything else seems to be sufficiently resolved.  @ejwong24  - I tested this idea out locally and it seems to work as expected.

ejwong24:
Done — pushed `151288d5`. Implemented the relaxed dtype check as you suggested: `_init_output` now accepts any `out=` buffer that can hold the result without loss of precision — `np.promote_types(out.dtype, dtype) != out.dtype` raises, otherwise it's accepted. So an explicitly higher-precision buffer is honored, and only too-narrow buffers are rejected.

(Small tweak to your sketch: used `np.promote_types` and `!=` rather than `is`, since dtype objects aren't singletons.)

Updated the docstrings on all three functions to match, and added a test for the newly-allowed case (float32 input into a float64 `out=`). Full 1.0.0dev gate green locally.

bmcfee:
Thanks @ejwong24 !

PR #2063, opened 2026-06-30
Fix RQA single-axis inputs

### LINKED ISSUE #2005: [Bug]: Intermittent segfault with librosa.sequence.rqa
### Bug summary

Running `librosa.sequence.rqa` on matrix input with 1 row produces intermittent segfaults (amongst other errors, seemingly all related to memory allocation/freeing).

### Code for reproduction

Running the following script (call it `rqa_test.py`) will reproduce the error, intermittently:

```Python

import argparse
import librosa
import numpy as np

parser = argparse.ArgumentParser()
parser.add_argument("--rows", type=int, default=1)
parser.add_argument("--cols", type=int, default=8)
args = parser.parse_args()

m = np.random.uniform(0,1,[args.rows,args.cols])

rqa, path = librosa.sequence.rqa(m)

print(rqa, path)
```

Running the above from inside the Python interpreter (obviously without the command line arguments) will also (intermittently) crash Python.

### Actual outcome

Intermittently receiving the following at the terminal after running the script:

`Segmentation fault`

```
munmap_chunk(): invalid pointer
Aborted
```

```
corrupted size vs. prev_size
Aborted
```

```
double free or corruption (out)
Aborted
```

```
python: /root/miniconda3/envs/buildenv/conda-bld/llvmdev_1755759130742/work/llvm/include/llvm/ADT/ilist_iterator.h:322: llvm::ilist_iterator_w_bits<OptionsT, IsReverse, IsConst>::reference llvm::ilist_iterator_w_bits<OptionsT, IsReverse, IsConst>::operator*() const [with OptionsT = llvm::ilist_detail::node_options<llvm::Instruction, true, false, void, true, llvm::BasicBlock>; bool IsReverse = false; bool IsConst = false; llvm::ilist_iterator_w_bits<OptionsT, IsReverse, IsConst>::reference = llvm::Instruction&]: Assertion `!NodePtr->isKnownSentinel()' failed.
Aborted
```

Interestingly it sometimes errors despite the printed result occurring (although sometimes not). Also it only seems to error if `--rows=1` and the value passed to `--cols` is greater than 6!

### Expected outcome

It should just produce the printed output without error.

### Additional information

Running `gdb -ex r --args rqa_test.py --rows=1` consistently  produces the following output:

```
Debug memory block at address p=0x22f8560: API 'r'
    368 bytes originally requested
    The 7 pad bytes at p-7 are FORBIDDENBYTE, as expected.
    The 8 pad bytes at tail=0x22f86d0 are not all FORBIDDENBYTE (0xfd):
        at tail+0: 0xe6 *** OUCH
        at tail+1: 0xd7 *** OUCH
        at tail+2: 0xe5 *** OUCH
        at tail+3: 0x1b *** OUCH
        at tail+4: 0xe0 *** OUCH
        at tail+5: 0xb6 *** OUCH
        at tail+6: 0xe3 *** OUCH
        at tail+7: 0x3f *** OUCH
    Data at p: 00 00 00 00 00 00 00 00 ... 00 00 00 00 00 00 00 00

Enable tracemalloc to get the memory block allocation traceback

Fatal Python error: _PyMem_DebugRawFree: bad trailing pad byte
Python runtime state: finalizing (tstate=0x0000000000a3f4a0)
...

Thread 1 "python" received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=140737350299648) at ./nptl/pthread_kill.c:44
44      ./nptl/pthread_kill.c: No such file or directory.
```

### Operating system

Tested on Ubuntu and macOS

### software versions

```
Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python 3.10.12 (main, Mar  3 2026, 11:56:32) [GCC 11.4.0]
NumPy 2.2.6
SciPy 1.15.3
librosa 0.11.0
INSTALLED VERSIONS

python: 3.10.12 (main, Mar  3 2026, 11:56:32) [GCC 11.4.0]

librosa: 0.11.0

audioread: installed, no version number available
numpy: 2.2.6
scipy: 1.15.3
sklearn: 1.7.2
joblib: 1.5.3
decorator: 5.2.1
numba: 0.65.0
soundfile: 0.13.1
pooch: v1.9.0
soxr: 1.0.0
lazy_loader: 0.5
msgpack: 1.1.2

numpydoc: None
sphinx: None
sphinx_rtd_theme: None
matplotlib: 3.10.8
sphinx_multiversion: None
sphinx_gallery: None
mir_eval: None
ipython: None
pytest: None
pytest_mpl: None
pytest_cov: None
samplerate: None
resampy: None
presets: None
packaging: 26.0
```

### Installation

pip

suhashollakc:
#### Reference Issue
Fixes #2005

This fixes `librosa.sequence.rqa` for similarity matrices with a single row or a single column.

The RQA dynamic-programming implementation initializes the first row and column, then unconditionally reads and writes the `(1, 1)` base case. For inputs shaped `(1, N)` or `(N, 1)`, that access is out of bounds inside the Numba-compiled function. Depending on the runtime, this can surface as intermittent memory corruption or a crash after the function appears to return.

This change returns after first-row/first-column initialization when either axis has length 1, avoiding the out-of-bounds `(1, 1)` access while preserving the expected score and backtracking pointer initialization for those edge cases.

#### Any other comments?
Added regression coverage for both `(1, 8)` and `(8, 1)` inputs with `backtrack=True` and `backtrack=False`.

Validation:

```bash
.\.venv\Scripts\python.exe -m pytest tests\test_sequence.py -k rqa -q

### CODE REVIEW

**Comment by bmcfee** [comment]
Thanks for the PR.  This probably does the trick to avoid whatever problem is occurring upstream to trigger segfaults, but it doesn't resolve the underlying issue (which may not be ours to resolve in the first place).  I'd rather not merge this kind of fix unless / until we're certain that the problem is entirely upstream (in numba) and that the numba devs aren't going to fix it in a timely fashion.

**Comment by bmcfee** [comment]
Ok, I dug into it more and I agree that this is a reasonable solution.  Thanks for the PR!

I do want to rebase this to the 1.0.0dev branch though.  Otherwise I see no problem with merging assuming the CI passes.

File: tests/test_sequence.py [hunk 1]
```diff
@@ -793,6 +793,23 @@ def test_rqa_empty(gap_onset, gap_extend, knight):
...
+    out = librosa.sequence.rqa(rec, backtrack=backtrack)
+
+    if backtrack:
+        score, path = out
+
+        assert score.shape == rec.shape
+        assert path.shape == (1, 2)
+        assert tuple(path[0]) == np.unravel_index(np.argmax(rec), rec.shape)
+    else:
+        assert out.shape == rec.shape
```
bmcfee: mypy isn't quite clever enough to figure out this from the overload, but we can just do it explicitly:

```suggestion
    if backtrack:
        score, path = librosa.sequence.rqa(rec, backtrack=backtrack)

        assert score.shape == rec.shape
        assert path.shape == (1, 2)
        assert tuple(path[0]) == np.unravel_index(np.argmax(rec), rec.shape)
    else:
        score = librosa.sequence.rqa(rec, backtrack=backtrack)
        assert score.shape == rec.shape
```

bmcfee:
There was one minor issue with the type checking on tests here - I took the liberty of fixing it and all looks good.  :shipit: 

Thanks again @suhashollakc !

PR #2066, opened 2026-07-07
Infer n_bins automatically in cqt/vqt

### LINKED ISSUE #1966: [ENH]: cqt/vqt infer n_bins
### Problem

A minor annoyance when working with cqt and related functions is that the `n_bins` and `bins_per_octave` parameters are not directly tied.  This means that if we want to change the frequency resolution but keep the same range, we need to change two parameters at once.

### Proposed solution

We could allow `n_bins=None` as an indicator to calculate the maximum number of bins that will still fit (comfortably, including passband limits) under Nyquist.  "Comfortably" here is doing a bit of work, as this calculation will depend on various other parameters of the filter bank (window function, filter scale), but it's entirely doable.

I see no reason not to add this functionality.

I could also be persuaded to make it the default behavior in a future release (1.1?).

### LINKED ISSUE #2071: Bug: CQT banding artifacts with maximal n_bins
While testing out #2066  I noticed that it surfaces an interesting behavior that I'd never seen before.

Note: none of this has to do with the automatic `n_bins` maximization, but rather is an artifact that would always have happened when setting `n_bins` to be large and we have a hop length of `2**n` (as we do for default).

With each octave we cover, we induce a downsampling operation as long as `hop_length` at the new current rate is even:
https://github.com/librosa/librosa/blob/e403272fc984bc4aeb316e5f15899042224bb9fe/librosa/core/constantq.py#L1020-L1025

The problem here is that each downsampling op includes a low-pass filter with some degree of roll-off.  Even in soxr's vhq mode, this rolloff is non-trivial, and it does add up.  In the example below, you can see banding artifacts close to each octave boundary:

cqt-rolloff

(`cqt` with `n_bins=None, sr=22050, hop_length=512, bins_per_octave=60`)

If we force everything to run at the native sampling rate by setting `hop_length` as an odd number up front (so the downsampling branch never activates), these banding artifacts disappear because we never downsample:

cqt_rolloff_oddhop

(`cqt` with `n_bins=None, sr=22050, hop_length=513, bins_per_octave=60`)

The effect also diminishes if we reduce the number of downsampling ops:
cqt_fewer_bins

(same as above, but with `hop_length=512, n_bins=440`).

---

I think this can be easily resolved by adding a 1-octave offset to the downsampling branch.  That is, we only do a downsample if the hop length is even *and* the highest frequency in the next octave down is at least one octave below the current nyquist.

I expect this will have minimal impact on how CQT functions with lower n_bins cutoffs, but the present behavior does seem like a bug to me that we can correct here.

bmcfee:
#### Reference Issue
Fixes #1966 
Fixes #2071 

This PR implements automatic n_bins detection for vqt (and cqt).  This automatically picks the largest `n_bins` such that filters will remain within the pass-band.

#### Any other comments?

I think this is functionally fine.  The code is a bit messy, in that there's a fair amount of redundant logic here for calculating filter cutoffs.  OTOH it's simple enough that the redundancy doesn't bother me too much.

Some thoughts:

- I don't think this should ever become the default parametrization because it will lead to unpredictable (or, more correctly, surprising) output sizes for users.
- There's a question as to whether this should become the default for *internal* uses of cqt/vqt, such as when computing chroma features.  I'm ambivalent on this one.

### DIFF
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -948,8 +948,7 @@ def vqt(
     fmin = fmin * 2.0 ** (tuning / bins_per_octave)
 
     if fmin >= sr / 2:
-        raise ParameterError(f"fmin={fmin} is too high for sr={sr}")
-
+        raise ParameterError(f"fmin={fmin} must be less than sr/2={sr/2}")
     if n_bins is None:
         # If n_bins is None, we need to compute the number of bins to reach sr/2
         # Over-allocating by one octave buys us enough buffer to be safe.
--- a/librosa/core/constantq.py
+++ b/librosa/core/constantq.py
@@ -244,7 +244,7 @@ def hybrid_cqt(
     fmin : float > 0 [scalar]
         Minimum frequency. Defaults to `C1 ~= 32.70 Hz`
 
-    n_bins : int > 0 [scalar]
+    n_bins : int > 0 or None [scalar]
         Number of frequency bins, starting at ``fmin``
         If `None` the number of bins will be inferred as the maximum that will
         fit below `sr/2`.
@@ -451,7 +451,7 @@ def pseudo_cqt(
     fmin : float > 0 [scalar]
         Minimum frequency. Defaults to `C1 ~= 32.70 Hz`
 
-    n_bins : int > 0 [scalar]
+    n_bins : int > 0 or None [scalar]
         Number of frequency bins, starting at ``fmin``
         If `None` the number of bins will be inferred as the maximum that will
         fit below `sr/2`.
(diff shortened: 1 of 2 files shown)


bmcfee:
Noting for later: the monotonic search from the top down to find the maximum filter frequency will not work if the spacing is non uniform. It needs to find the index after which all frequencies fit, not the highest one such that none above fit.  Our frequency range is monotonic, but alpha and f_cutoff might not be. 

This is easy enough to fix.

bmcfee:
Ok, I think I've made all necessary changes for this one now and it's in good shape.  @lostanlen I'd be happy for your feedback on it if you have a few cycles to spare.

bmcfee:
Gonna round this off by adding nbin inferene for psuedo and hybrid cqt as well.

In doing so, I noticed that we never actually do a Nyquist check for the filter bank in pseudo or hybrid modes.  Whoops.

bmcfee:
This PR now also fixes #2071, which surfaced in development of auto-nbins.

My initial cut of the fix documented in #2071 was to require a P5 below nyquist for the bottom of the current octave before resampling.  I've refined this a little after checking the pass-band behavior of our typical sample rate converters, which are mostly above 80% Nyquist.  (We could do something really clever here and tune it to different converters, but that seems well overkill here.)

The new logic still avoids the banding artifacts and is about as efficient as I can make it without getting much more complex.  I think this PR is ready to :shipit: when CI passes.

bmcfee:
Documenting the provenance of the 80% rule by resampling an impulse to sr/2 and plotting the resulting spectrum:

image

It's definitely more conservative than we need to be for the defaults (soxr_hq).  In fact all soxr modes except lq don't roll off until about 93%.

libsamplerate (medium) hits at about 87%.

PR #2068, opened 2026-07-07
Standardize n_fft parameter docstrings

### LINKED ISSUE #1522: `n_fft` docstrings
**Describe the bug**
`n_fft` is one of the most ubiquitous parameter in librosa, and yet its definition is relatively lapidary.

Here's a quick list i just made by grepping the source:
1. FFT window size for ``S``
2.  FFT window size
3. number of FFT bins to use, if ``y`` is provided.
4. length of the windowed signal after padding with zeros.
5. The number of samples per frame in D.
6. STFT window size
7.  If ``S`` is provided, then ``n_fft`` is inferred from ``S``. Else, copied from input 
8. number of FFT components in the resulting STFT
9. number of FFT components
10. number of FFT bins
11. The length of each analysis frame.
12. FFT window size for use in ``feature()`` if ``S`` is not provided.
(might contain more)

I would love it if someone could audit these definitions and see what can be improved. I'm not saying that we need the same definition everywhere (calling it a window size in some places and a number of bins somewhere else makes perfect sense) but i don't think we need as many as the ones i listed above.

Relatedly, it seems like some of these definitions are off by a factor two, since they refer to (1+n_fft//2) (assuming n_fft even) and not n_fft itself. The ISTFT even/odd disjunction (in the wake of #1244) is not very well explained.

brookcs3:
#### Reference Issue
Fixes #1522

An audit of every `n_fft` parameter docstring found the definitions reduce to three semantic groups; full breakdown in [this comment on #1522](https://github.com/librosa/librosa/issues/1522#issuecomment-4907859317). This PR:

- Standardizes the analysis-frame group on **"length of the FFT frame"**. (Earlier revisions used "window"; per review, `n_fft` is the frame length, and a window can be shorter than the frame, so the whole set including the `frames_to_time` family and `fft_frequencies` now uses "frame".)
- Fixes the bins/components wordings that conflated the frame length with the `1 + n_fft // 2` bin count (`filters.mel`, `filters.chroma`, `piptrack`, `estimate_tuning`, `feature.inverse.*`). Shape claims are stated only where the output or filter bank is actually shaped by them: `filters.mel`, `filters.chroma`, and `mel_to_stft`; the audio-output functions (`mel_to_audio`, `mfcc_to_audio`) reference the intermediate STFT with no shape claim.
- Rewrites `istft`'s frame-length inference note with the explicit formula (`2 * (stft_matrix.shape[-2] - 1)`) and why odd frame lengths need `n_fft` passed.
- Corrects `mel_to_stft`'s documented return shape from `(..., n_fft, t)` to `(..., 1 + n_fft // 2, t)`, matching what `nnls(mel_basis, M)` produces.
- Fixes two frequency-axis references that assumed 2D input (`istft`'s `win_length` line, `phase_vocoder`'s `hop_length` line): axis is `-2` for multichannel.

Docstrings only; no behavior changes. Rebuilt against `1.0.0dev` after retargeting, including the definitions that only exist on that branch.

#### Any other comments?
If you would rather keep #1522 open for the glossary discussion, I am happy to change "Fixes" to "References".

### DIFF
--- a/librosa/core/audio.py
+++ b/librosa/core/audio.py
@@ -1163,7 +1163,7 @@ def get_duration(
         it is better to use the audio time series directly.
 
     n_fft : int > 0 [scalar]
-        FFT window size for ``S``
+        length of the FFT frame used when computing ``S``
 
     hop_length : int > 0 [ scalar]
         number of audio samples between columns of ``S``
--- a/librosa/core/convert.py
+++ b/librosa/core/convert.py
@@ -100,7 +100,7 @@ def frames_to_samples(
     hop_length : int > 0 [scalar]
         number of samples between successive frames
     n_fft : None or int > 0 [scalar]
-        Optional: length of the FFT window.
+        Optional: length of the FFT frame.
         If given, time conversion will include an offset of ``n_fft // 2``
         to counteract windowing effects when using a non-centered STFT.
 
@@ -169,7 +169,7 @@ def samples_to_frames(
         number of samples between successive frames
 
     n_fft : None or int > 0 [scalar]
-        Optional: length of the FFT window.
+        Optional: length of the FFT frame.
         If given, time conversion will include an offset of ``- n_fft // 2``
         to counteract windowing effects in STFT.
 
@@ -229,7 +229,7 @@ def frames_to_time(
     hop_length : int > 0 [scalar]
         number of samples between successive frames
     n_fft : None or int > 0 [scalar]
-        Optional: length of the FFT window.
+        Optional: length of the FFT frame.
         If given, time conversion will include an offset of ``n_fft // 2``
         to counteract windowing effects when using a non-centered STFT.
 
@@ -293,7 +293,7 @@ def time_to_frames(
         number of samples between successive frames
 
     n_fft : None or int > 0 [scalar]
-        Optional: length of the FFT window.
+        Optional: length of the FFT frame.
         If given, time conversion will include an offset of ``- n_fft // 2``
         to counteract windowing effects in STFT.
 
@@ -1374,7 +1374,7 @@ def fft_frequencies(*, sr: float = 22050, n_fft: int = 2048) -> _Array1D[np.floa
     sr : number > 0 [scalar]
         Audio sampling rate
     n_fft : int > 0 [scalar]
-        length of the FFT window
+        length of the FFT frame
 
     Returns
     -------
@@ -2090,7 +2090,7 @@ def times_like(
     hop_length : int > 0 [scalar]
         number of samples between successive frames
     n_fft : None or int > 0 [scalar]
-        Optional: length of the FFT window.
+        Optional: length of the FFT frame.
         If given, time conversion will include an offset of ``n_fft // 2``
         to counteract windowing effects when using a non-centered STFT.
     axis : int [scalar]
@@ -2147,7 +2147,7 @@ def samples_like(
     hop_length : int > 0 [scalar]
         number of samples between successive frames
     n_fft : None or int > 0 [scalar]
-        Optional: length of the FFT window.
+        Optional: length of the FFT frame.
         If given, time conversion will include an offset of ``n_fft // 2``
         to counteract windowing effects when using a non-centered STFT.
     axis : int [scalar]
--- a/librosa/core/convert.py
+++ b/librosa/core/convert.py
@@ -1374,7 +1374,7 @@ def fft_frequencies(*, sr: float = 22050, n_fft: int = 2048) -> _Array1D[np.floa
     sr : number > 0 [scalar]
         Audio sampling rate
     n_fft : int > 0 [scalar]
-        FFT window size
+        length of the FFT window
 
     Returns
     -------
--- a/librosa/core/spectrum.py
+++ b/librosa/core/spectrum.py
@@ -428,7 +428,7 @@ def istft(
         Number of frames between STFT columns.
         If unspecified, defaults to ``win_length // 4``.
 
-    win_length : int <= n_fft = 2 * (stft_matrix.shape[0] - 1)
+    win_length : int <= n_fft = 2 * (stft_matrix.shape[-2] - 1)
         When reconstructing the time series, each frame is windowed
         and each sample is normalized by the sum of squared window
         according to the ``window`` function (see below).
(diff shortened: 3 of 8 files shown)


librosa/core/spectrum.py
```diff
@@ -436,9 +436,10 @@ def istft(
 
     n_fft : int > 0 or None
         The number of samples per frame in the input spectrogram.
-        By default, this will be inferred from the shape of ``stft_matrix``.
-        However, if an odd frame length was used, you can specify the correct
-        length by setting ``n_fft``.
+        By default, this is inferred as ``2 * (stft_matrix.shape[-2] - 1)``,
+        which assumes an even-length ``n_fft``. If the STFT was computed with
+        an odd ``n_fft``, that inference is off by one; pass the original
+        ``n_fft`` to recover the correct frame length.
```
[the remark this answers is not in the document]
brookcs3: Good catch. Fixed in 48d3416, along with the same axis issue in `phase_vocoder`'s `hop_length` description (`(D.shape[0]-1)//2` had the same 2D-only assumption).

librosa/core/convert.py
```diff
@@ -1591,7 +1591,7 @@ def fft_frequencies(*, sr: float = 22050, n_fft: int = 2048) -> np.ndarray:
     sr : number > 0 [scalar]
         Audio sampling rate
     n_fft : int > 0 [scalar]
-        FFT window size
+        length of the FFT window
```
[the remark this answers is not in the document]
brookcs3: Deliberate tradeoff: the repo's parameter descriptions are mixed on capitalization, and `mfcc`/`melspectrogram` already used the lowercase "length of the FFT window" before this PR. I kept that exact string everywhere so the definition is uniform across the codebase; capitalizing it only where a neighboring description happens to be capitalized would trade global consistency for local. Happy to flip the whole set to "Length ..." if the maintainers prefer.

bmcfee:
I'm retargeting this to the 1.0dev branch, as that's where all development recently has occurred.  It's very nearly ready to merge back into main, and may in fact do so before this PR is complete.  But retargeting now should help to avoid conflicts down the road.

librosa/core/audio.py
```diff
@@ -1163,12 +739,12 @@ def get_duration(
         it is better to use the audio time series directly.
 
     n_fft : int > 0 [scalar]
-        FFT window size for ``S``
+        length of the FFT window used when computing ``S``
```
bmcfee: This actually should be "frame" not "window", since a window can be smaller than the frame.

librosa/core/spectrum.py
```diff
@@ -428,7 +427,7 @@ def istft(
         Number of frames between STFT columns.
         If unspecified, defaults to ``win_length // 4``.
 
-    win_length : int <= n_fft = 2 * (stft_matrix.shape[0] - 1)
+    win_length : int <= n_fft = 2 * (stft_matrix.shape[-2] - 1)
```
bmcfee, line 431: very good catch 😁

librosa/feature/inverse.py
```diff
@@ -144,17 +137,17 @@ def mel_to_audio(
     sr : number > 0 [scalar]
         sampling rate of the underlying signal
     n_fft : int > 0 [scalar]
-        number of FFT components in the resulting STFT
+        length of the FFT window of the resulting STFT; the output has ``1 + n_fft // 2`` frequency bins
```
bmcfee: The output in this function is audio, not an STFT.  It has no frequency bins.

librosa/core/audio.py
brookcs3: Fixed in a0fd1a3. Question on scope: should "frame" replace "window" in the standardized phrase everywhere (i.e. "length of the FFT frame")? The canonical string came from the existing wording in `mfcc`/`melspectrogram`, but given the frame/window distinction you're drawing, I'm happy to sweep the whole set.

librosa/feature/inverse.py
brookcs3: Right, thanks — reworded in a0fd1a3. `mel_to_audio` and `mfcc_to_audio` now say "length of the FFT window for the intermediate STFT" with no bins clause; `mel_to_stft` keeps the bins clause since its output is the spectrogram.

brookcs3:
Rebuilt the branch on `1.0.0dev` after the retarget (a0fd1a3 replaces the main-based commits). Re-ran the docstring audit against the new base, which has three `n_fft` definitions that don't exist on `main`; those are standardized too. The review fixes above are folded in.

bmcfee:
Thanks @brookcs3 !

I was going to reply in the issue thread, but since the PR is already open, I left comments here instead.  Generally, where we use `n_fft` we're referring to "frame length" and not "window length".  The two are often used interchangeably, but they aren't exactly the same thing - a window can be shorter than a frame.  I noted one such case of this in a comment above, but it applies more generally.

The other point I noted was that the new docstrings about output size aren't always correct.  I didn't check all of them yet, but the example i noted only uses stft as a subroutine and does not return a spectrogram, making the output shape reference in the docstring incorrect.

brookcs3:
Swept in ad90123: every `n_fft` description now reads "length of the FFT frame" (28 sites, including the `frames_to_time` family and `fft_frequencies`, which predate this PR but describe the same quantity).

On the output-shape claims, the full list after the inverse.py fix is three, and I checked each against its own Returns block or code: `filters.mel` and `filters.chroma` return `(n, 1 + n_fft // 2)` filter banks, and `mel_to_stft` returns the linear spectrogram (its Returns shape is also corrected in this PR). `mel_to_audio` and `mfcc_to_audio` carry no shape claim anymore, just "for the intermediate STFT".

librosa/core/audio.py
bmcfee: I hesitate to say "everywhere" but probably almost everywhere.  We certainly have cases where framing is done without windowing, but in those cases, we use `frame_length` instead of `n_fft` - at least i'm pretty sure.  But the changes here LGTM!

bmcfee:
This all LGTM, thanks again!  Anything else to put in this one?

brookcs3:
Nothing else for this one from me — I've updated the PR description to match the final state (Copilot flagged it as stale after the frame sweep). The one related idea left in #1522 is the glossary revision you mentioned there; that feels like its own issue rather than a rider on this PR.

bmcfee:
> The one related idea left in #1522 is the glossary revision you mentioned there; that feels like its own issue rather than a rider on this PR.

Indeed - I think I'll roll that into #1968.

Thanks again! :shipit:

Issue #2067, opened 2026-07-07
[ENH]: Extend precise return type annotations to the feature and onset modules


brookcs3:
### Problem

#2050 and #2052 replaced bare `-> np.ndarray` returns with precise types like `NDArray[np.float64]` across `beat`, `core`, `filters`, `segment`, `sequence`, and `util`. The `feature` and `onset` modules were not part of that pass, so all 15 public functions in `librosa/feature/spectral.py`, the 4 in `librosa/feature/rhythm.py`, the 4 in `librosa/onset.py`, and `plp` in `beat.py` still return bare `np.ndarray`. A type checker therefore sees a precise return type for `beat_track` but an opaque one for `mfcc` or `onset_strength`, and the benefit of the new annotations stops at the feature/onset boundary.

### Proposed solution

Carry the same convention from #2050/#2052 into the remaining modules:

- `librosa/feature/spectral.py`: `spectral_centroid`, `spectral_bandwidth`, `spectral_contrast`, `spectral_rolloff`, `spectral_flatness`, `rms`, `poly_features`, `zero_crossing_rate`, `chroma_stft`, `chroma_cqt`, `chroma_cens`, `chroma_vqt`, `tonnetz`, `mfcc`, `melspectrogram`
- `librosa/feature/rhythm.py`: `tempogram`, `fourier_tempogram`, `tempo`, `tempogram_ratio`
- `librosa/onset.py`: `onset_detect`, `onset_strength`, `onset_backtrack`, `onset_strength_multi`
- `librosa/beat.py`: `plp` (and the private `__beat_tracker`)

I'd be happy to make the PR for this, as one PR or split per module, whichever you prefer. If @jorenham already has this remainder planned, I'll defer to that.

jorenham:
I was indeed planning on circling back to this :)

bmcfee:
Thanks for noting this - it would be great to fill these parts in.  I think in all the cases you've identified, it should not be too challenging, with the possible snag that many of them will preserve minimal precision depending on input (e.g., a float32 in will give a float32 out, not a float64).  I haven't done a thorough audit for this in quite some time though (see #1143 ), so it might be a bit more work than anticipated.

brookcs3:
[...]

PR #2069, opened 2026-07-08
Precise dtype type annotations in `librosa.beat`

jorenham:
This adds precisie type annotations to the two public functions in `librosa.beat`. And as indeed predicted in https://github.com/librosa/librosa/issues/2067#issuecomment-4907530792, this was already pretty involved. Especially for `beat_track`, where the return types depend on the input *values* in several ways simultaneously.

I managed to avoid the overload complexity for `bpm` from getting out of hand too much here by using the `typing_extensions` backport of `TypeVar` with a [PEP 696](https://peps.python.org/pep-0696/) `default=` that usually would require Python 3.13+ to use. (`typing_extensions` can always be imported in `if TYPE_CHECKING` guards because its stubs are bundled with typeshed as part of the stdlib stubs). Without those defaulting type variables, this would've required 6 overloads instead of 3 (we'd have to handle the default case separately).

This did not change any runtime behavior, and only affects (i.e. improves) static typing.

#### Reference Issue
Towards #2067

### DIFF
--- a/librosa/beat.py
+++ b/librosa/beat.py
@@ -11,7 +11,7 @@ Beat and tempo
 """
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Literal, overload
 
 import numba
 import numpy as np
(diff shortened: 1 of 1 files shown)


jorenham:
cc @brookcs3

bmcfee:
LGTM.  Happy to merge unless you want to add more in this PR.

jorenham:
Merging this sounds good :ok_hand:

bmcfee:
Thanks again!

Issue #2076, opened 2026-07-13
[Bug]: chroma_stft (via estimate_tuning/piptrack) leaves unreturned memory in the glibc arena, growing RSS across repeated calls


imrulhuda:
### Bug summary

Repeated calls to `librosa.feature.chroma_stft()` cause process RSS (resident memory) to grow over time in a long-lived process, especially when input length varies call-to-call (as real audio files typically do). This is not a Python object leak or a growing cache — `chroma_stft` internally calls `estimate_tuning()` → `piptrack()`, which allocates ~6 full-spectrogram-sized temporary arrays (including a redundant `np.abs(S)` copy, since the input power spectrogram is already non-negative) to compute a single scalar tuning offset. These are freed correctly at the Python level every call, but glibc's malloc does not return the freed memory to the OS under this allocation pattern. A single `ctypes` call to `malloc_trim(0)` reliably reclaims the excess, confirming this is arena fragmentation rather than a reference leak.

### Code for reproduction

```python
import ctypes
import numpy as np
import librosa

libc = ctypes.CDLL("libc.so.6")

def vmrss_mb():
    with open("/proc/self/status") as f:
        for line in f:
            if line.startswith("VmRSS:"):
                return int(line.split()[1]) / 1024

sr = 16000
rng = np.random.default_rng(0)
base_len = 3_964_032  # ~248s at 16kHz

# Fixed-length input: RSS should plateau quickly if this is just one-time setup cost.
y0 = rng.standard_normal(base_len).astype(np.float32)
librosa.feature.chroma_stft(y=y0, sr=sr, hop_length=256)  # warmup
print(f"[fixed length] warmup: {vmrss_mb():.1f} MB")
for i in range(1, 11):
    librosa.feature.chroma_stft(y=y0, sr=sr, hop_length=256)
    print(f"[fixed length] call {i}: {vmrss_mb():.1f} MB")

libc.malloc_trim(0)
print(f"[fixed length] after malloc_trim(0): {vmrss_mb():.1f} MB")

# Varying-length input, closer to real-world usage (real files are rarely identical length):
print()
for i in range(1, 21):
    length = base_len + rng.integers(-500_000, 500_000)
    y = rng.standard_normal(length).astype(np.float32)
    librosa.feature.chroma_stft(y=y, sr=sr, hop_length=256)
    print(f"[varying length] call {i} (len={length}): {vmrss_mb():.1f} MB")

libc.malloc_trim(0)
[...]

PR #2080, opened 2026-07-20
Skip redundant np.abs copy in piptrack for non-negative real input

imrulhuda:
Refs #2076

#### Reference Issue
Follow-up to #2076 (partial — addresses one of the allocations, not the full piptrack rework).

piptrack unconditionally ran `S = np.abs(S)`, which allocates a full-size copy of the spectrogram on every call — even when `S` is already a real, non-negative magnitude/power spectrogram (the common case). This is one of the intermediate allocations behind the RSS growth in #2076 (glibc arena fragmentation, confirmed with malloc_trim, not a reference leak). I guarded the rectification so it only runs when it's actually needed:
      
    if np.iscomplexobj(S) or S.min() < 0:
        S = np.abs(S)

Complex input is still converted to magnitude and signed real input is still rectified, so output is unchanged for every input; the copy is only skipped when `S` is already non-negative.

#### Any other comments?

  - On the condition: the issue suggested `S.min() <= 0`, but I used `< 0`. Real spectrograms often contain exact zeros, so `<= 0` would trigger the copy on almost every real input and defeat the optimization (`abs(0) == 0`, so zeros don't need rectifying). Happy to change it if you prefer `<=`.
  - Scope: this removes only one of piptrack's several intermediate allocations. The larger streamlining you mentioned (numba stencil reusing  `_parabolic_interpolation`) is left for a possible follow-up.
  - Tests: added `test_piptrack_magnitude_equivalence` (non-negative real vs complex vs sign-flipped → identical output). piptrack / estimate_tuning / chroma suites pass; ruff, numpydoc lint, mypy, and codespell clean locally.

### DIFF
--- a/librosa/core/pitch.py
+++ b/librosa/core/pitch.py
@@ -314,8 +314,13 @@ def piptrack(
         pad_mode=pad_mode,
     )
 
-    # Make sure we're dealing with magnitudes
-    S = np.abs(S)
+    # Make sure we're dealing with magnitudes.
+    # ``np.abs(S)`` forces a full-size copy of ``S`` on every call.  For a
+    # real, non-negative spectrogram (the common case, e.g. a magnitude or
+    # power spectrogram) that copy is redundant, so only rectify when the
+    # input is complex or actually contains negative values.
+    if np.iscomplexobj(S) or S.min() < 0:
+        S = np.abs(S)
 
     # Truncate to feasible region
     fmin = np.maximum(fmin, 0)
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1655,6 +1655,29 @@ def test_piptrack(freq, n_fft):
     assert np.all(np.abs(np.log2(recovered_pitches) - np.log2(freq)) <= 1e-2)
 
 
+@pytest.mark.parametrize("n_fft", [1024, 2048])
+def test_piptrack_magnitude_equivalence(n_fft):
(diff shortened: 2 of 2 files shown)


bmcfee:
This LGTM, thanks!

CI issues seem to be some combination of the rebase and history rewriting i had to do to clean up some erroneous attributions in main.  (Or maybe github is just having problems today?)  Either way it's unrelated and should not be a blocker on merging this one.

bmcfee (approved):

bmcfee:
I hate to ask this, but would it be possible to redo this PR from a fresh checkout of the repo?

Basically what happened here is that I did an incorrect squash merge last week that led to incorrect attributions for a large number of contributions.  I fixed the history this morning, but this happened after you had cloned and submitted this PR.  As a result, the commit history in this PR includes a number of commits from main which should not be there, and which would undo the attribution correction that I implemented this morning.

It may be possible to fix this by fetching main with a fast-forward, rebasing your branch locally to main's head, and then fast-forward pushing back to github.  But if that doesn't work, I think the only way is to start the branch again on a fresh clone.

Sorry for this!  Otherwise I think the PR is fine and ready to merge.

imrulhuda:
Sure! Will do towards the end of workday.

imrulhuda:
Done!

bmcfee:
Merged. Thanks!

PR #2084, opened 2026-08-06
1.0 final

bmcfee:
#### What does this implement/fix? Explain your changes.
1.0 final release prep.

PR #2087, opened 2026-08-10
Validate audio input in feature.rms (closes #2086)

### LINKED ISSUE #2086: feature.rms skips valid_audio, so one NaN sample makes effects.trim() silently return an empty signal
### Description

`librosa.feature.rms` is the only feature function that does not validate its input with
`util.valid_audio`. As a result one non-finite sample makes `librosa.effects.trim` return an
empty array: the entire signal is silently deleted, with no warning and no exception.

Every neighbouring function raises a clear `ParameterError` on the same input.

### Steps to reproduce

```python
import numpy as np, librosa

rng = np.random.default_rng(0)
y = (rng.standard_normal(16000) * 0.05).astype(np.float32)
bad = y.copy()
bad[8000] = np.nan          # one non-finite sample in 16,000

print("librosa", librosa.__version__)
print("clean  trim ->", librosa.effects.trim(y)[0].shape,
      " split ->", librosa.effects.split(y).shape)
print("1 NaN  trim ->", librosa.effects.trim(bad)[0].shape,
      " split ->", librosa.effects.split(bad).shape)

print("rms NaN frames:", int(np.isnan(librosa.feature.rms(y=bad)).sum()), "/ 32")
for fn in ("stft", "zero_crossing_rate"):
    try:
        (librosa.stft if fn == "stft" else librosa.feature.zero_crossing_rate)(bad)
        print(f"{fn}: no error")
    except Exception as e:
        print(f"{fn}: {type(e).__name__}: {e}")
```

### Actual behaviour

```
librosa 0.11.0
clean  trim -> (16000,)  split -> (1, 2)
1 NaN  trim -> (0,)      split -> (0, 2)
rms NaN frames: 4 / 32
stft: ParameterError: Audio buffer is not finite everywhere
zero_crossing_rate: ParameterError: Audio buffer is not finite everywhere
```

### Expected behaviour

Either a `ParameterError`, consistent with `stft` and `zero_crossing_rate`, or a trim that
still returns the signal. Not a silent `(0,)`.

### Why it happens

`feature.rms` frames the signal directly and never calls the validator:

```python
x = util.frame(y, frame_length=frame_length, hop_length=hop_length)
```

`'valid_audio' in inspect.getsource(...)` is `True` for `stft` and
`zero_crossing_rate`, and `False` for `rms`.

From there it is arithmetic on NaN. The four frames overlapping sample 8000 come back NaN,
`trim` and `split` route through `_signal_to_frame_nonsilent`, which compares frame energy
against a threshold, and **every comparison against NaN is False**:

```python
float("nan") < 0.01   # False
float("nan") > 0.01   # False
```

So a NaN frame is never quiet and never loud. It passes no gate, no frame is marked
non-silent, and `trim` concludes the whole file is silence.

### Why it is worth fixing rather than documenting

`trim` and `split` are the standard silence-removal step at the front of most audio
pipelines, and this fails open in the worst direction: the caller receives an empty array
rather than an error, so a batch job drops files instead of failing on them. Non-finite
samples are not exotic in real corpora, coming from truncated decodes, failed resamples,
and division by a zero-energy normalisation window.

The one-line fix matches the surrounding code:

```python
def rms(*, y=None, S=None, frame_length=2048, hop_length=512, ...):
    if y is not None:
        util.valid_audio(y, mono=False)
```

Happy to open a PR with a regression test covering `rms`, `trim` and `split` if that would
be useful.

### Software versions

librosa 0.11.0, numpy 1.26.4, Python 3.12, macOS.

Kayvan-Zahiri:
Closes #2086.

`rms` is the only feature function that reaches `util.frame` without going through
`valid_audio`, so it never rejected non-finite input. `effects.trim` and
`effects.split` both route through it, and since NaN compares False against any
threshold, no frame was ever marked non-silent and they returned an empty result
for the whole signal.

```python
y = np.zeros(4096, dtype=np.float32); y[2048] = np.nan
librosa.effects.trim(y)[0].shape    # before: (0,)   after: ParameterError
```

## One behaviour change worth flagging before you merge

`valid_audio` also enforces the dtype, so this makes `rms` reject integer and
complex input that it previously accepted. That is a real change and I would
rather surface it than have it turn up later.

I checked whether it breaks a supported input or closes a hole, on `1.0.0rc0`
before this patch:

```
stft                rejects int16   ParameterError: Audio data must be floating-point
melspectrogram      rejects int16
mfcc                rejects int16
zero_crossing_rate  rejects int16
spectral_centroid   rejects int16
resample            rejects int16

feature.rms         ACCEPTS int16
effects.trim        ACCEPTS int16   (via rms)
effects.split       ACCEPTS int16   (via rms)
```

So integer support in `rms` came from the same missing check rather than from
intent, and the three functions that accept it are exactly the three affected
here. This aligns them with the rest of the library.

The caller most likely to notice is `scipy.io.wavfile.read`, which returns int16
for a 16-bit WAV. If you would rather not change dtype behaviour in a 1.0
release, the narrower version is a finiteness check alone:

```python
if not np.isfinite(y).all():
    raise ParameterError("Audio buffer is not finite everywhere")
```

That fixes the reported bug and keeps integer input working. Happy to switch to
it, I just think consistency with `valid_audio` is the better default and did not
want to make that call silently.

## Tests

Added to `tests/test_features.py`: `test_rms_nonfinite` parametrised over
nan/inf/-inf, and `test_effects_nonfinite` over `trim` and `split`.

```
tests/test_features.py   437 passed, 72 xfailed
tests/test_effects.py    264 passed
```

The wider suite shows 104 failures both with and without this patch, all
`ModuleNotFoundError: No module named 'resampy'` from my environment, so nothing
new.

The `S=` path is untouched.

### DIFF
--- a/librosa/feature/spectral.py
+++ b/librosa/feature/spectral.py
@@ -884,6 +884,8 @@ def rms(
     >>> plt.show()
     """
     if y is not None:
+        util.valid_audio(y)
+
         if center:
             padding = [(0, 0) for _ in range(y.ndim)]
             padding[-1] = (int(frame_length // 2), int(frame_length // 2))
--- a/tests/test_features.py
+++ b/tests/test_features.py
@@ -412,6 +412,26 @@ def test_rms_badshape():
     librosa.feature.rms(S=S, frame_length=100)
 
 
+@pytest.mark.parametrize("bad", [np.nan, np.inf, -np.inf])
+@pytest.mark.xfail(raises=librosa.ParameterError)
+def test_rms_nonfinite(bad):
+    # A single non-finite sample used to propagate silently: NaN compares
+    # False against every threshold, so effects.trim and effects.split
+    # concluded the whole signal was silence and returned nothing.
+    y = np.zeros(4096, dtype=np.float32)
+    y[2048] = bad
+    librosa.feature.rms(y=y)
+
+
+@pytest.mark.parametrize("effect", [librosa.effects.trim, librosa.effects.split])
+@pytest.mark.xfail(raises=librosa.ParameterError)
+def test_effects_nonfinite(effect):
+    # These route through rms, so they inherited the missing check.
+    y = np.zeros(4096, dtype=np.float32)
+    y[2048] = np.nan
+    effect(y)
+
+
 @pytest.fixture(params=[32, 16, 8, 4, 2], scope="module")
 def y_zcr(request):
     sr = 16384


bmcfee:
CI failures appear to be coming from a regression that popped up for us upstream over the weekend (noted over in #2084).  I don't think it should be a blocker for this one, but I'd like to get that fixed in a separate PR before merging this one.  That said, changes here look fine and should be good to merge once I sort out the other problem.  Thanks!

bmcfee:
Merged. Thanks again!

PR #643, opened 2017-11-09
Mpl display test fix

bmcfee:
#### What does this implement/fix? Explain your changes.

Matplotlib 2.1 switched to pytest from nosetests, which breaks the unit tests for the display module.

This PR is a short-term fix to pin the mpl version to 2.0.  The long-term solution will be to migrate to pytest as well, but that's a much bigger job than we can realistically accomplish before 0.6.

---
This change is [Reviewable](https://reviewable.io/reviews/librosa/librosa/643)

Issue #512, opened 2017-02-14
Links to old versions of documentation


bmcfee:
For archival purposes, the docs site should include links to older API versions.  Sphinx doesn't appear to make this totally easy to do, but we should be able to hack it pretty easily.

bmcfee:
https://robpol86.github.io/sphinxcontrib-versioning/#

PR #2089, opened 2026-08-15
Keep doctrees and reST sources out of the published docs

### LINKED ISSUE #2088: [Maint]: Reduce footprint of documentation site
The new documentation site produces a sizeable build artifact (310MB).  This isn't too much of a problem right now, though it could become a problem with subsequent versions as we deploy more large builds into the `doc` repo.

Probably this is unnecessary, as we could prune out the .doctrees and maybe even _sources.  This can always be done post-hoc by clobbering release assets and rebuilding the site.

Kayvan-Zahiri:
#### Reference Issue

Fixes #2088

Two separate reasons the built docs are larger than they need to be.

1. `sphinx-build` writes `.doctrees` into the output directory unless `-d` points somewhere else. `docs/Makefile` already passes `-d $(BUILDDIR)/doctrees`, but the three workflows that build the site and the release asset do not, so the doctrees end up inside the directory that gets zipped and published.

2. `html_copy_source` defaults to `True` even when `html_show_sourcelink` is `False`. `conf.py` already sets the sourcelink to False, so `_sources` is being shipped with nothing linking to it.

Measured from the git tree of `librosa/doc`, so these are the deployed sizes rather than estimates:

| version | total | `.doctrees` | `_sources` | prunable |
|---|---|---|---|---|
| dev | 468M | 224M | 78M | 65% |
| main | 468M | 224M | 78M | 65% |
| 1.0.0 | 468M | 224M | 78M | 65% |
| latest | 468M | 224M | 78M | 65% |
| 0.10.2 | 80M | 0M | 23M | 28% |
| 0.11.0 | 79M | 0M | 23M | 28% |
| **site** | **2030M** | **895M** | **357M** | **61%** |

0.10.2 and 0.11.0 carry no `.doctrees` at all, so that half is a recent change; `_sources` affects every version.

The change adds `-d build/doctrees` to the three workflows, where that path sits outside the directory being zipped or published, and sets `html_copy_source = False`. On a scratch sphinx project this moves the doctrees out and leaves `_sources` empty.

#### Any other comments?

Unrelated to this issue, but noticed while measuring: `latest` and `main` are stored in `librosa/doc` as full directory trees rather than symlinks, so each duplicates `1.0.0` and `dev` respectively. The workflow creates them with `ln -srf`, so something in the publish step is dereferencing them. Happy to open a separate issue if that is worth pursuing.

### DIFF
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -77,7 +77,7 @@ jobs:
         - name: Build documentation site
           shell: bash -l {0}
           run: |
-            sphinx-build -b html docs build/html
+            sphinx-build -b html -d build/doctrees docs build/html
             touch build/html/.nojekyll
 
         - name: Upload compiled documentation site
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Build documentations
         shell: bash -el {0}
         run: |
-          sphinx-build -b html docs build/html/dev
+          sphinx-build -b html -d build/doctrees docs build/html/dev
           cp docs/docsite-index-redirect.html build/html/index.html
 
 
--- a/.github/workflows/publish-doc-asset.yml
+++ b/.github/workflows/publish-doc-asset.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Build documentations
         shell: bash -el {0}
-        run: sphinx-build -b html docs build/html
+        run: sphinx-build -b html -d build/doctrees docs build/html
 
       - name: Zip and Upload Release Asset
         shell: bash -el {0}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -357,6 +357,10 @@ html_css_files = [
 ]
 html_show_sourcelink = False
 
+# Sourcelinks are disabled above, so nothing references the reST sources and
+# there is no reason to ship them.
+html_copy_source = False
+
 html_context = {
     "default_mode": "light"
 }
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -118,7 +118,7 @@ import librosa
 )
 plot_include_source = True
 plot_html_show_source_link = False
-plot_formats = [("png", 150), ("pdf", 150)]
+plot_formats = [("png", 150)]
 plot_html_show_formats = False
 
 from cycler import cycler


bmcfee:
Thanks for this!  Fix seems to work as expected.

I grabbed the build artifact out of the docs-ci run, and it does appear that the show source link is still there (and works).  This is actually what I would want, but doesn't seem entirely consistent with the config change here.

> Unrelated to this issue, but noticed while measuring: `latest` and `main` are stored in `librosa/doc` as full directory trees rather than symlinks, so each duplicates `1.0.0` and `dev` respectively. The workflow creates them with `ln -srf`, so something in the publish step is dereferencing them. Happy to open a separate issue if that is worth pursuing.

That's indeed weird - it didn't do that previously.  I wonder if it's something in the gh-pages action?

Kayvan-Zahiri:
That link is `sphinx.ext.viewcode`, and it's unrelated to `_sources`.

viewcode generates `_modules/**.html` from the Python source and points API entries at those pages. `html_copy_source` only controls whether the reST originals get copied into `_sources`, and `html_show_sourcelink` only controls the "Show Source" link pointing *there*, which `conf.py` already had off before this PR.

Checked on a scratch project with viewcode enabled and both settings off:

```
_sources files:                     0
_modules/:                          index.html, pkg.html
[source] link in rendered page:     present, pointing at _modules/pkg
```

So the links you want keep working, and nothing here touches them.

On the duplication, you guessed right, it's the gh-pages action. `peaceiris/actions-gh-pages` copies the publish directory with:

```ts
// src/git-utils.ts:47
await fs.promises.cp(src, dest, { dereference: true, force: true, recursive: true });
```

`dereference: true` follows symlinks and writes their targets out as real files, and it's hardcoded rather than exposed as an input. So `ln -srf build/html/$LATEST build/html/latest` lands in the published branch as a full copy of `1.0.0`, and likewise `main` from `dev`. That's roughly another 936M on top of the numbers in the description.

Happy to open a separate issue for that one.

bmcfee:
> `dereference: true` follows symlinks and writes their targets out as real files, and it's hardcoded rather than exposed as an input.

Ah that's annoying, and seems unfixable from our side.  Looking at the git blame, this appears to have crept in to the action about 3 months ago https://github.com/peaceiris/actions-gh-pages/commit/954f6bf8259a6185f366f5cf13baee63745e0f79

Probably the best fix here is to make a patch upstream that surfaces dereferencing as a parameter of the action.  I have no idea if such a thing would be merged, as it seems like there are a few open issues around resolving links that have been stale for a very long time.

Kayvan-Zahiri:
Worth correcting my own framing: the dereferencing isn't three months old.

That commit swapped `shelljs` for `node:fs`, and the line it replaced was already dereferencing:

```diff
-  cp('-RfL', [`${publishDir}/*`, `${publishDir}/.*`], destDir);
+  await copyDirContents(publishDir, destDir);   // fs.promises.cp(..., {dereference: true})
```

`-L` in `cp` means follow symlinks, so `dereference: true` is a faithful port rather than a behaviour change. `git log -S"-RfL"` puts its arrival in #512 on 2020-10-14, so the action has resolved symlinks for close to six years.

The proposal you have in mind also already exists: peaceiris/actions-gh-pages#643, "option to not follow symlinks when copying assets", opened 2021-08-10, six comments including a maintainer reply, still open. It cites the same `-L` switch.

On the odds of landing a patch there, the last several merged PRs are all Renovate dependency bumps, so I wouldn't plan around an upstream change.

That leaves the in-repo options, and none of them looks free:

- A redirect page for `latest` and `main`, like `docsite-index-redirect.html` already does for the site root, only redirects the index. Deep links such as `latest/generated/librosa.stft.html` would break, so it isn't a drop-in.
- Post-processing the published branch to rewrite the copies as git symlinks (mode 120000) fights `force_orphan: true`, which rebuilds the branch on every run.
- Doing nothing costs about 936M in the `doc` repo but nothing at serve time, since Pages serves either shape identically.

Given that, my instinct is a separate issue to track it rather than anything blocking this PR. Happy to open one, or to poke #643 upstream if you'd rather someone tried.

bmcfee:
Ah, right - we previously used a different action for this: https://github.com/librosa/librosa/blame/b19c8551cf02495283e2f969d821e21addd8badb/.github/workflows/docs.yml

Anyway, I agree with your assessment that punting to upstream is not going to work here.

I think that if the build size is substantially reduced, as this PR does, it's not so much of a problem to have dereferenced copies in the repo anyway.  We can probably trim out the `main` link to be replaced by a redirect, but that's not a huge deal.

bmcfee:
Note: we can also shave another ~12MB by removing pdf from the plot formats:

https://github.com/Kayvan-Zahiri/librosa/blob/4f691ac37ba46afd9c6512182acbed7d98f708f4/docs/conf.py#L121

I had this in here for latexpdf builds of the docsite, but that's really not necessary at this point.  I think with that change in, this should be good to merge.

Kayvan-Zahiri:
Done. Confirmed the saving against the published tree: 440 files, 12.3 MB per version directory, 49.4 MB across `dev`, `main`, `1.0.0` and `latest`.

Worth noting the HTML never referenced them. `plot_directive` renders `.{{ default_fmt }}`, which is `plot_formats[0]`, so `png`, and `plot_html_show_formats = False` suppresses the per-format download links. Every one of the 440 has a `.png` sibling of the same basename.

`make latexpdf` still builds, since the `.. only:: not html` branch uses a `.*` wildcard and falls back to the PNGs. Figures are raster there now.

bmcfee:
Thanks @Kayvan-Zahiri !  I've updated the 1.0.0 doc asset, so all changes should apply immediately after the next run of the doc build.

Issue #2094, opened 2026-08-21
feature.mfcc: expose `power_to_db` parameters, as proposed in #1734


isaacsight:
## Summary

This is a follow-up to #1734 rather than a new report. In that thread @bmcfee diagnosed the cause correctly and proposed the fix:

> This is basically another iteration on an issue identified in the multichannel PR, wherein the noise floor calculation became sensitive to information leaking across different channels.
>
> Probably what we ought to do here is expose some of the parameters to `power_to_db`, ie, so that you can explicitly set this in the call to mfcc rather than having to split it up into multiple steps.

The issue was closed three days later once the reporter's own problem was solved by the two-step workaround. The proposed change does not appear to have been made: in 1.0.0, `feature.mfcc` still takes no `top_db`, and the call is

```python
S = power_to_db(melspectrogram(y=y, sr=sr, norm=mel_norm, **kwargs))
```

`power_to_db` receives no arguments, so `top_db` stays at its default of `80.0`, and `**kwargs` is documented as going to `melspectrogram`. In 1.0.0 `power_to_db`'s `axes='auto'` resolves to `(-2, -1)` for 2-D input, so the reduction on line

```python
log_spec = np.maximum(log_spec, log_spec.max(axis=axes, keepdims=True) - top_db)
```

is over both mel and time — i.e. the whole clip for mono input.

I am opening this because I ran into it while building conformance vectors for MFCC, and the severity turned out to be larger than #1734 shows. Not a new diagnosis, just a measurement that might help prioritise the change that was already agreed.

## What the clamp costs

Self-contained, no downloads, librosa 1.0.0:

```python
import numpy as np, librosa

SR, HOP, WIN, NFFT = 16000, 160, 400, 512
rng = np.random.default_rng(0)

quiet = 0.001 * rng.standard_normal(HOP * 200 + WIN)   # 200 frames, ~2 s
kw = dict(sr=SR, n_mfcc=13, n_fft=NFFT, win_length=WIN, hop_length=HOP,
          center=False, n_mels=26)

alone = librosa.feature.mfcc(y=quiet, **kw)

print(f"{'appended burst':>16}{'max|diff|':>12}{'energy in c1..c12':>20}{'frames flat':>13}")
for amp in [0.0, 0.05, 0.9, 5.0, 50.0]:
    y = quiet if amp == 0 else np.concatenate([quiet, amp * rng.standard_normal(HOP * 100)])
[...]

PR #2095, opened 2026-08-21
Expose top_db on features that call power_to_db

isaacsight:
Addresses #2094, following the approach agreed there: surface the dB parameters while preserving default behaviour exactly.

## What this does

Adds `top_db` to the four public functions that call `power_to_db` without exposing it:

| function | call site |
|---|---|
| `feature.mfcc` | `power_to_db(melspectrogram(...))` |
| `feature.spectral_contrast` | `power_to_db(peak) - power_to_db(valley)` |
| `onset.onset_strength` | delegates to `onset_strength_multi` |
| `onset.onset_strength_multi` | `core.power_to_db(S)` |

`core.perceptual_weighting` also calls `power_to_db`, but it already forwards `**kwargs` there and accepts `top_db` today, so it is untouched.

In `spectral_contrast` the threshold is passed to **both** calls. `peak` and `valley` are thresholded against their own separate maxima, so honouring it in only one would change the difference.

## Default behaviour is unchanged

Verified bitwise against unmodified 1.0.0 on the same input, for all four functions:

```
function      max|stock - modified| at defaults
mfcc                                    0.0e+00
spectral_contrast                       0.0e+00
onset_strength                          0.0e+00
onset_strength_multi                    0.0e+00
```

There is also a test asserting that passing `top_db=80.0` explicitly reproduces omitting it, so a future change to the default cannot silently desynchronise the two paths.

## One thing I found after filing the issue

The issue was about `mfcc`, but `onset_strength` routes through the same clamp, and the downstream effect is larger than the feature-value change. With a quiet 120 bpm click track followed by a loud burst:

```
burst       onsets (truth 16)   tempo    envelope attenuation
none                       15   117.5                     1x
+53 dB                     15   117.5                    17x
+74 dB                     15    60.1                  1511x
+94 dB                      0     0.0                   10^6x
```

Beat tracking octave-errors at +74 dB and fails entirely at +94 dB. On the same signal, `top_db=None` recovers it:

```
beat_track with top_db=80.0  ->  tempo   0.0,  0 onsets
beat_track with top_db=None  ->  tempo 136.0, 10 onsets
```

That is why this covers the onset functions rather than `mfcc` alone.

## A deliberate omission, and a question

I did not add `ref` or `amin`.

`onset_strength` and `onset_strength_multi` already have a `ref` parameter, an `np.ndarray` holding the reference spectrogram for the flux computation. `power_to_db`'s `ref` is a scalar or callable setting the dB reference level. Adding the latter to those signatures would create a name collision of exactly the kind #1842 was about, so it did not seem like something to introduce quietly.

If you would like `ref` and `amin` surfaced too, I am happy to add them, but the naming for the onset functions is your call rather than mine. `top_db` is collision-free everywhere and is the parameter the issue was actually about, so this PR stops there.

## Tests

12 new tests in `tests/test_features.py`:

- `top_db=None` makes the computation frame-local for `mfcc`, `onset_strength`, `onset_strength_multi`
- the default is *not* frame-local, documenting the behaviour being opted out of
- passing the default explicitly equals omitting it, for all four
- `spectral_contrast` responds to the threshold, covering the two-call-site path
- negative `top_db` raises `ParameterError`

The frame-locality tests use `center=False` and a signal length that is an exact multiple of `hop_length`, so the compared frames are byte-identical between the two inputs and only what follows them differs. They assert `atol=1e-10` rather than exact equality: `stft` blocks long inputs via `MAX_MEM_BLOCK`, so accumulation order differs between the two signal lengths, which leaves a float64 residual measured at 8.9e-15 against a 283 difference when the threshold is active.

`pytest tests/test_features.py tests/test_onset.py tests/test_beat.py` gives 1149 passed, 98 xfailed. Four failures in `tests/test_multichannel.py` are `resampy` not being installed in my environment and reproduce identically on a clean checkout.

## Not included

No changelog entry, since the convention here needs the PR number. Happy to add one.

The streaming gallery example you suggested is included, in `docs/tutorials/03-advanced/plot_mfcc_stream.py`, modelled on `plot_pcen_stream.py`. It shows that streamed and whole-file MFCCs diverge at the default (max difference 267.7 on a 56 dB crest-factor signal) and agree to 4.8e-07 with `top_db=None`. Happy to split it into its own PR if you would rather keep this one to the API change.


isaacsight:
Thanks — four of these were correct and all four were mine. Fixed in e82547a.

The root cause of three was that I wrote the `top_db` docstring once and pasted it into four functions without adapting it to what each actually does. Each now describes its own behaviour.

**Peak is per-channel, not global.** Correct, and my wording was wrong. `power_to_db`'s `axes='auto'` resolves to `(-2, -1)`, so the reduction is over frequency and time within each leading channel. Verified: a quiet channel inside a stereo pair with a loud channel gives `max|d| = 0.000e+00` against the same channel computed alone. The docstrings now say "over the frequency and time axes, independently for each leading channel".

**`linear=True` ignores `top_db`.** Also correct, and this one I should have caught. `linear=True` takes the `peak - valley` branch and never calls `power_to_db`, so the parameter is inert there. Verified: `top_db=80` against `top_db=0` in linear mode gives `max|d| = 0.000e+00`. Now documented, and pinned by `test_spectral_contrast_linear_ignores_top_db` so the no-op stays deliberate rather than becoming silent.

I considered warning when `top_db` is set explicitly with `linear=True`, but since `top_db` has a default of 80.0 there is no way to distinguish an explicit value from the default without a sentinel, and that seemed like more machinery than the case warrants. Happy to add one if you would prefer.

**`top_db=None` does not make `onset_strength` frame-local.** Correct. Onset strength is a spectral flux, so it compares each frame with the one `lag` frames earlier, and `detrend` adds a further neighbour dependence. Removing the threshold removes the coupling to the loudest frame in the channel, which is the narrower claim the docstring now makes.

**The `spectral_contrast` assertion was too weak.** Correct — `not np.allclose` would have been satisfied by a regression forwarding `top_db` to `peak` but not `valley`. The test now pins `top_db=0`, where both spectra flatten to their own maxima and the contrast must collapse to a single constant, which is only reachable if the threshold got to both conversions. A second assertion pins it against an explicit reconstruction.

**On the scope point:** the PR description was stale rather than the file being unintended. @bmcfee suggested the streaming gallery example in #2094 and it was added in a later commit; I have updated the description. Glad to split it into a separate PR if that is preferred.

1150 tests pass, and defaults remain bitwise identical to unmodified 1.0.0 for all four entry points.

docs/tutorials/03-advanced/plot_mfcc_stream.py
```diff
@@ -0,0 +1,152 @@
...
+`librosa.power_to_db`, which by default thresholds the result at ``top_db``
+below its peak.  That peak is taken over the entire array it is given,
+including the time axis.  When you process a whole file at once, the peak is
+the loudest moment in the file.  When you process the same audio block by
+block, each block gets its own peak.
+
+The consequence is that MFCCs computed on a stream do not match MFCCs computed
+on the whole file, and neither is a function of the individual frame.  Passing
+``top_db=None`` disables the threshold and makes the computation frame-local,
+which is what you want when the features must not depend on their context.
```
bmcfee, line 30: This isn't actually the whole story.  The real source of the issue here is that the dB reference (defined by `ref=`) is computed dynamically as `ref=np.max`.  If you use a static reference, eg `ref=1`, then this shouldn't be a problem.

docs/tutorials/03-advanced/plot_mfcc_stream.py
```diff
@@ -0,0 +1,152 @@
...
+#####################################################################
+# Set up the block reader.  ``center=False`` is important here: it keeps
+# each block's frames aligned with the frames of the whole-file
+# computation, so the two are directly comparable.
+
+n_fft = 2048
+hop_length = 512
+n_mfcc = 13
+
+sr = librosa.get_samplerate(filename)
```
bmcfee: This step is actually no longer necessary: you can call `stream` with `sr=22050` and it will resample in place now.  I'd prefer that in this example since without specifying `sr`, `stream` now gives a future warning that shows up in the rendered doc.

librosa/feature/spectral.py
```diff
@@ -369,6 +369,7 @@ def spectral_contrast(
     fmin: float = 200.0,
     n_bands: int = 6,
     quantile: float = 0.02,
+    top_db: float | None = 80.0,
```
bmcfee, line 373: as noted above, `ref` for dB calculation should be exposed as well.

librosa/feature/spectral.py
```diff
@@ -1859,6 +1870,7 @@ def mfcc(
     norm: _DCTNorm | None = "ortho",
     lifter: float = 0,
     mel_norm: Literal["slaney"] | float | None = "slaney",
+    top_db: float | None = 80.0,
```
bmcfee, line 1884: Same comment as before.

docs/tutorials/03-advanced/plot_mfcc_stream.py
isaacsight: You're right that it isn't the whole story, and `ref` is now exposed on both functions.

One thing I want to put in front of you, because it changes what the example should say rather than just adding to it. `mfcc` calls `power_to_db(melspectrogram(...), top_db=top_db)` and never passes `ref`, so it takes the scalar branch at the default `ref=1.0`. In that configuration `top_db` is the one that binds — `ref` isn't dynamic unless a caller makes it so, and until this commit they had no way to: `mfcc`'s `**kwargs` forward to `melspectrogram`, not to `power_to_db`.

Measured on identical frames, comparing a quiet signal against the same frames followed by a loud burst:

```
default (ref=1.0, top_db=80)     max|d| = 2.501e+02
top_db=None, ref=1.0             max|d| = 8.882e-15    <- frame-local
top_db=None, ref=np.max          max|d| = 1.058e+03    <- ref re-couples
top_db=80,   ref=np.max          max|d| = 8.228e+02
```

So both reductions are real and independent, and which one binds depends on whether `ref` is callable. The example now says exactly that, and demonstrates the `top_db=None, ref=np.max` row so the reader sees the threshold switched off and the dependence still there.

If you'd rather the example lead with `ref` and treat `top_db` as the secondary case, say the word — I wrote it the other way round because the default configuration is the one most readers will hit first, but I don't have a strong view.

docs/tutorials/03-advanced/plot_mfcc_stream.py
isaacsight: Done — `stream` now takes `sr=22050` and the `get_samplerate` call is gone.

Verified the warning behaviour rather than assuming it:

```
stream() without sr : ['FutureWarning'] | first block (9728,)
stream(sr=22050)    : no warnings       | first block (9728,)
native sample rate  : 22050
```

The file is already at 22050, so the resample is a no-op and the blocks are unchanged — the only difference is the warning no longer renders into the doc.

librosa/feature/spectral.py
isaacsight: Added. `ref` is forwarded to both `power_to_db` calls, since peak and valley are converted separately.

Pinned by a test that turns on a property rather than just a difference: `spectral_contrast` is a difference of two dB conversions, so a **scalar** `ref` cancels exactly — measured `7.105e-15` for both `ref=100.0` and `ref=0.01` against `ref=1.0`. That can only hold if `ref` reached both calls; forwarding it to `peak` alone would shift the output by `-10 log10(ref)`. A callable `ref` does not cancel, and `ref=np.max` moves the output by `8.216`.

Also pinned that `linear=True` ignores `ref` for the same reason it ignores `top_db` — that branch never calls `power_to_db`.

librosa/feature/spectral.py
isaacsight: Added on `mfcc` too, defaulting to `1.0` so the default path is bitwise unchanged.

Pinned by the mechanism rather than by a diff: a scalar `ref` offsets the log-mel spectrogram by a constant, and the DCT of a constant lands entirely in coefficient 0. Measured with `top_db=None` to keep it linear — `ref=100.0` against `ref=1.0` gives `226.27` on coefficient 0 and `7.283e-14` across coefficients 1 and up.

Not added to `onset_strength` / `onset_strength_multi`: those already take `ref` as the reference spectrogram for the flux computation, which is a different quantity from `power_to_db`'s reference level, and quietly overloading the name looked like the thing #1842 was about. Happy to revisit under a different name if you want the dB reference reachable there as well.

bmcfee: > those already take `ref` as the reference spectrogram for the flux computation, which is a different quantity from `power_to_db`'s reference level, and quietly overloading the name looked like the thing #1842 was about.

Oof, yes, API normalization is an endless chore.  Thanks for catching it!  1842 was a similar but distinct issue.  I think the reasonable thing to do here is to do a deprecation rename of the `ref` argument in `onset_strength*`.  We have some machinery for this in the util.deprecation module, but we might want to modify it if the eventual plan is to change the behavior of `ref` rather than remove the argument entirely.  Either way, this is a distinct issue and behavior changes should not take effect until 1.1 or later.

docs/tutorials/03-advanced/plot_mfcc_stream.py
```diff
@@ -0,0 +1,185 @@
...
+whole_default = librosa.feature.mfcc(
+    y=y, sr=sr, n_mfcc=n_mfcc, n_fft=n_fft, hop_length=hop_length, center=False
+)
+stream_default = streamed_mfcc()
+
+#####################################################################
+# With the default ``top_db=80``, the two disagree.  Each block was
+# thresholded against its own loudest frame rather than against the
+# loudest frame in the file.
+n = min(whole_default.shape[1], stream_default.shape[1])
```
bmcfee, line 100: nitpick, i prefer to use `shape[-2]` for this kind of thing so that it generalizes better to multichannel inputs.

Likewise, `[..., :n]` instead of `[:, :n]` below.

tests/test_features.py
```diff
@@ -1209,4 +1209,179 @@ def test_hybrid_tempogram():
...
+@pytest.fixture(scope="module")
+def top_db_signals():
+    """A quiet passage, and the same passage followed by a loud burst.
+
+    center=False with a length that is an exact multiple of hop_length means
+    the compared frames are byte-identical between the two inputs; only what
+    follows them differs.
+    """
+    sr, hop, n_fft = 22050, 512, 2048
+    rng = np.random.default_rng(0)
```
bmcfee, line 1230: Please use the rng fixture instead of numpy's default rng.

tests/test_features.py
```diff
@@ -1209,4 +1209,179 @@ def test_hybrid_tempogram():
...
+def test_spectral_contrast_scalar_ref_cancels(top_db_signals):
+    """spectral_contrast is a difference of two dB conversions, so a scalar
+    `ref` cancels exactly.  This can only hold if `ref` reached both calls: a
+    regression forwarding it to `peak` alone would shift the output by
+    ``-10 log10(ref)``.
+    """
+    sr, hop, n_fft, quiet, loud, n = top_db_signals
+    kw = dict(y=loud, sr=sr, hop_length=hop, n_fft=n_fft, center=False)
+    base = librosa.feature.spectral_contrast(**kw)
+    for value in (100.0, 0.01):
```
bmcfee, line 1372: Instead of iterating over values in the test, it's better to raise them up as parameters to trigger separate tests.  This way, if one fails, it's easier to diagnose.

PR #2097, opened 2026-08-21
Forward `window` to stft and istft in effects.hpss, harmonic and percussive

### LINKED ISSUE #2096: effects.hpss/harmonic/percussive accept a `window` argument that is never forwarded to stft
`effects.hpss`, `effects.harmonic` and `effects.percussive` accept and document a `window` argument that has no effect. It is declared in the signature and never passed to `stft` or `istft`.

```python
y, sr = librosa.load(librosa.ex("trumpet"))

a = librosa.effects.hpss(y=y, window="hann")[0]
b = librosa.effects.hpss(y=y, window="blackman")[0]
np.abs(a - b).max()
# 0.0
```

Same for `harmonic` and `percussive`, and for `hamming`. For contrast, the argument is meaningful in the function these delegate to:

```python
np.abs(np.abs(librosa.stft(y, window="hann"))
     - np.abs(librosa.stft(y, window="blackman"))).max()
# 24.24
```

## Where it comes from

This looks like a small gap left by #1826, which closed #1804 by replacing `**kwargs` with explicit STFT parameters. Six were added to each of the three functions; five are forwarded.

| function | declared | forwarded to stft/istft |
|---|---|---|
| `effects.hpss` | `n_fft`, `hop_length`, `win_length`, `window`, `center`, `pad_mode` | all except `window` |
| `effects.harmonic` | same | all except `window` |
| `effects.percussive` | same | all except `window` |

In `effects.hpss` the call reads:

```python
stft = core.stft(
    y,
    n_fft=n_fft,
    hop_length=hop_length,
    win_length=win_length,
    center=center,
    pad_mode=pad_mode,
)
```

@lostanlen's enumeration in #1804 does list `window='hann'` among the STFT arguments, so the intent seems to have been to include it.

Because `window` is a named keyword-only parameter it is captured by name and cannot reach `**kwargs`, so there is currently no route by which a caller can set the analysis window for these three functions, and no error is raised when they try.

All three docstrings document it as:

> `window : str, tuple, number, function, or np.ndarray [shape=(n_fft,)]`
> Window specification. See `scipy.signal.get_window` for supported values.

## Suggested fix

Adding `window=window` to the `stft` call, and to the two `istft` calls in each function, appears to be all that is required. I have not checked whether `istft` should receive the same window as the forward transform in every case, though for perfect reconstruction it normally should.

Happy to open a PR, including a test that varies `window` and asserts the output moves, which is the check that would have caught this. I already have #2095 open against these files' neighbours; I can send this separately or fold it in, whichever you prefer.

Found while running an AST scan for named parameters that never appear in their function body, prompted by having made the same mistake myself in #2095 (`top_db` is inert when `spectral_contrast` is called with `linear=True`, which I have since documented and pinned).

Versions: librosa 1.0.0, numpy 2.5.2, Python 3.12, macOS.

isaacsight:
Fixes #2096.

`effects.hpss`, `effects.harmonic` and `effects.percussive` accept and document a `window` argument that never reached the transform. Seven call sites — one `core.stft` and one or two `core.istft` per function — now receive it.

Before:

```python
librosa.effects.hpss(y=y, window="hann")[0]
librosa.effects.hpss(y=y, window="blackman")[0]
# max|difference| = 0.0
```

After: `0.0234` for blackman, `0.0108` for hamming.

## Why `istft` is included and not just `stft`

Forwarding to the forward transform alone would leave analysis and synthesis on different windows, and `hpss` would stop reconstructing. That is pinned by a second test:

```
window=hann      max|(harm + perc) - y| = 0.0000
window=hamming   max|(harm + perc) - y| = 0.0000
window=blackman  max|(harm + perc) - y| = 0.0000
```

I did check whether a synthesis window differing from the analysis window is ever wanted here, and could not see a case for it in these three functions, since they are convenience wrappers around a single analyse-modify-resynthesise cycle. Happy to be told otherwise.

## Defaults

Unchanged, verified bitwise against unmodified 1.0.0 for all four outputs:

```
hpss_h        0.0e+00
hpss_p        0.0e+00
harmonic      0.0e+00
percussive    0.0e+00
```

## Tests

- `test_effects_window` — output differs across hann, hamming and blackman for all three functions. On its own this would also pass if `window` reached `stft` but not `istft`, which is why the next one accompanies it.
- `test_hpss_window_reconstruction` — `y_harm + y_perc` still sums to the input for each window.

321 pass in `tests/test_effects.py` and `tests/test_decompose.py`, 13 xfailed.

## On the linting question from #2096

> it's strange that ruff (or earlier linting actions) didn't catch this for us

It would have, but the rule is not enabled. `ARG` (flake8-unused-arguments) is absent from `select` in `pyproject.toml`, and ruff does not enable it by default.

I expected turning it on to be impractical, so I measured it rather than guessing:

```
$ ruff check --select ARG librosa/
Found 16 errors.
```

Three of those were the `window` arguments fixed here; `ruff check --select ARG librosa/effects.py` now passes. Sixteen across the whole package seems small enough to be worth a look, though I have not examined the other thirteen and some are likely deliberate — a `# noqa: ARG001` or a per-file ignore would cover those.

Left out of this PR entirely, since it is a repo-wide policy decision rather than part of the fix. Happy to open it as a separate issue if useful.

For what it is worth, this class of defect is why I ran an AST scan in the first place: a parameter added to a signature but never wired into the call still runs, still returns plausible values, and still passes every existing test, because no test varies it. A linter catches it in seconds. It has turned up in five projects I looked at this week, including a PR of my own.

No rush on review — enjoy the vacation.

### DIFF
--- a/librosa/effects.py
+++ b/librosa/effects.py
@@ -140,6 +140,7 @@ def hpss(
         n_fft=n_fft,
         hop_length=hop_length,
         win_length=win_length,
+        window=window,
         center=center,
         pad_mode=pad_mode,
     )
@@ -156,6 +157,7 @@ def hpss(
         n_fft=n_fft,
         hop_length=hop_length,
         win_length=win_length,
+        window=window,
         center=center,
         length=y.shape[-1],
     )
@@ -165,6 +167,7 @@ def hpss(
         n_fft=n_fft,
         hop_length=hop_length,
         win_length=win_length,
+        window=window,
         center=center,
         length=y.shape[-1],
     )
@@ -263,6 +266,7 @@ def harmonic(
         n_fft=n_fft,
         hop_length=hop_length,
         win_length=win_length,
+        window=window,
         center=center,
         pad_mode=pad_mode,
     )
@@ -279,6 +283,7 @@ def harmonic(
         n_fft=n_fft,
         hop_length=hop_length,
         win_length=win_length,
+        window=window,
         center=center,
         length=y.shape[-1],
     )
@@ -377,6 +382,7 @@ def percussive(
         n_fft=n_fft,
         hop_length=hop_length,
         win_length=win_length,
+        window=window,
         center=center,
         pad_mode=pad_mode,
     )
@@ -393,6 +399,7 @@ def percussive(
         n_fft=n_fft,
         hop_length=hop_length,
         win_length=win_length,
+        window=window,
         center=center,
         length=y.shape[-1],
     )
--- a/tests/test_effects.py
+++ b/tests/test_effects.py
@@ -201,6 +201,38 @@ def test_harmonic(ysr):
     assert np.allclose(yh1, yh2)
 
 
+@pytest.mark.parametrize(
+    "effect_fn",
+    [librosa.effects.hpss, librosa.effects.harmonic, librosa.effects.percussive],
+)
+def test_effects_window(ysr, effect_fn):
+    # The window must reach the STFT.  Asserting only that the outputs differ
+    # would also pass if it reached stft but not istft, so the reconstruction
+    # check below pins both.
+    y, sr = ysr
+
+    def first(result):
+        return result[0] if isinstance(result, tuple) else result
+
+    y_hann = first(effect_fn(y=y, window="hann"))
+    y_hamming = first(effect_fn(y=y, window="hamming"))
+    y_blackman = first(effect_fn(y=y, window="blackman"))
+
+    assert not np.allclose(y_hann, y_hamming)
+    assert not np.allclose(y_hann, y_blackman)
+
+
+@pytest.mark.parametrize("window", ["hann", "hamming", "blackman"])
+def test_hpss_window_reconstruction(ysr, window):
+    # Analysis and synthesis must use the same window, otherwise the harmonic
+    # and percussive parts no longer sum back to the input.
+    y, sr = ysr
+
+    y_harm, y_perc = librosa.effects.hpss(y=y, window=window)
+
+    assert np.allclose(y_harm + y_perc, y, atol=1e-6)
+
+
 @pytest.fixture(scope="module", params=[False, True], ids=["mono", "stereo"])
 def y_trim(request):
     # construct 5 seconds of stereo silence


bmcfee:
> Sixteen across the whole package seems small enough to be worth a look, though I have not examined the other thirteen and some are likely deliberate — a `# noqa: ARG001` or a per-file ignore would cover those.

I just gave it a look, and indeed, most of them are intentional and only used for API compatibility (e.g. matplotlib callbacks).

> Left out of this PR entirely, since it is a repo-wide policy decision rather than part of the fix. Happy to open it as a separate issue if useful.
> 
> For what it is worth, this class of defect is why I ran an AST scan in the first place: a parameter added to a signature but never wired into the call still runs, still returns plausible values, and still passes every existing test, because no test varies it. A linter catches it in seconds. It has turned up in five projects I looked at this week, including a PR of my own.

A separate issue/PR would be great, thanks for putting in the effort with this!

isaacsight:
Thanks for taking a look! I opened #2100 to track enabling unused-argument checks, with narrow exceptions for intentional API and callback compatibility cases.

I also checked the CI Docs failure on attempt 2: the documentation site builds successfully, but linkcheck reports a 403 from the external paper URL at arrow.tudublin.ie. The subsequent automated failure-comment step fails with “Resource not accessible by integration.” The code tests and lint checks passed. Would you prefer tracking those docs/workflow problems separately?

bmcfee:
> The subsequent automated failure-comment step fails with “Resource not accessible by integration.” The code tests and lint checks passed. Would you prefer tracking those docs/workflow problems separately?

Yes, let's take those separately.  I had been working on rolling a custom linkcheck reporting bot when I left for vacation, and hadn't quite finished working out the problems around permissioning.

bmcfee:
This LGTM, anything left here or are we good to merge?

isaacsight:
Thanks, Brian! I addressed Copilot’s remaining coverage suggestion in e582f43: the harmonic and percussive equivalence tests now compare against the corresponding HPSS component using Hann, Hamming, and Blackman windows.

Validation: 329 passed, 13 xfailed across tests/test_effects.py and tests/test_decompose.py. I also verified that removing window forwarding from either wrapper’s ISTFT makes all four of its non-default-window cases fail.

Nothing else left from my side—good to merge once CI finishes on the new commit. The docs/linkcheck issues remain separate as agreed. Thanks for reviewing!

bmcfee:
Thanks @isaacsight ! :shipit:

PR #2101, opened 2026-09-16
Enable Ruff ARG unused-argument checks and resolve findings

### LINKED ISSUE #2100: Enable Ruff unused-argument checks to catch silently ignored parameters
Following up on @bmcfee’s invitation in https://github.com/librosa/librosa/pull/2097#issuecomment-5621010992.

In #2096, `effects.hpss`, `effects.harmonic`, and `effects.percussive` accepted and documented `window` without forwarding it to the underlying transforms. #2097 fixes those call sites and adds regression coverage.

As reported in #2097, Ruff’s `ARG` rules were not enabled, and an explicit `ruff check --select ARG librosa/` found 16 unused arguments before that fix, including the three `window` arguments. That was the count at the time of the PR, rather than a fresh scan of main. @bmcfee reviewed the findings and noted that most remaining cases are intentional for API compatibility, such as matplotlib callbacks.

Proposal:

- Enable the appropriate Ruff `ARG` rules in the existing lint configuration and CI.
- Review the remaining findings individually; preserve parameters required for public API or callback compatibility.
- Add narrow suppressions with explanations for intentional unused arguments, rather than removing required parameters or broadly ignoring whole files.
- Keep this policy change separate from the audio fix in #2097.

The goal is to catch accidentally ignored parameters before they ship: code can return plausible output and pass existing tests even when a user-supplied setting has no effect. This check would complement behavioral tests; it would not prove that every parameter is used correctly.

Acceptance criteria: the existing lint job checks for unused arguments, intentional compatibility cases remain supported and documented, and the test suite continues to pass.

godking123:
Fixes #2100. Enables `ARG` in `pyproject.toml`, then resolves each finding individually. Note: the issue scoped this to `librosa/`, but I went ahead and resolved the ruff argument checks for /docs to be thorough.

$ ruff check --select ARG librosa/
Found 13 errors.

$ ruff check --select ARG docs/
Found 6 errors.

All fall into two categories:

**Suppressed (matplotlib/Sphinx calling convention requires the parameter):**

- `pos` in `TimeFormatter`, `ChromaFormatter`, `ChromaSvaraFormatter`, `ChromaFJSFormatter`, and `TonnetzFormatter` due to `Formatter.__call__(x, pos)` contract
- `pos` in `NoteFormatter`, `SvaraFormatter`, `FJSFormatter`, `LogHzFormatter`'s `_format_tick` method, forwarded by `AdaptiveFormatterBase.__call__` regardless of use
- `pos` in `__radian_formatter`, registered with `FuncFormatter`, same two-arg contract
- `gallery_conf`, `fname` in `reset_numpy_printoptions`, since sphinx-gallery's `reset_modules` callback always calls with this exact signature
- `app`, `what`, `name`, `options` in `skip_matplotlib_inherited`, since Sphinx's `autodoc-skip-member` event always calls with this exact six-argument signature; only `obj` and `skip` are actually used

Each gets `# noqa` (`ARG002` for methods, `ARG001` for functions) plus a comment naming the interface, e.g. `# pos is required by matplotlib's Formatter.__call__ interface and is unused in this method`.

**Removed (dead, not required by anything):**

- `unison` in `__coord_vqt_hz`, never used in the body, never forwarded to `core.interval_frequencies` (which has no such parameter), and already covered by `**_kwargs`
- `*args, **kwargs` in `_wrap` (`__float_window`), since the only call site (`__float_window(window)(len(sig))`) only ever passes `n`
- `gallery_conf`, `fname` params on `reset_mpl`, since it's an internal helper, not the registered callback (`reset_numpy_printoptions` is); it never used either parameter, only forwarded them
- `global plot_rcparams` in `reset_mpl`, since it's read-only, never assigned, so the `global` declaration was pointless

**Testing:** `ruff check librosa/` and `ruff check docs/` both 0 errors. `pytest`: 10877 passed, 0 failed.


librosa/display.py
```diff
@@ -2358,7 +2367,6 @@ def __coord_vqt_hz(
     bins_per_octave: int = 12,
     sr: float = 22050,
     intervals: str | Collection[float] | None = None,
-    unison: str | None = None,
```
bmcfee, line 2361: Good catch - this is probably just a copypasta from the vqt_fjs branch when I coded up the vqt_hz branch.

librosa/filters.py
```diff
@@ -375,7 +375,7 @@ def __float_window(window_spec):
 
     For integer-valued ``x``, there should be no change in behavior.
     """
-    def _wrap(n, *args, **kwargs):
+    def _wrap(n):
```
bmcfee: I see your point that the usage of `__float_window` is limited to not use the args/kwargs here.  However, I think the more correct fix is to propagate args and kwargs through to the `get_window` call below so that `_wrap` behaves as transparently as possible, even if we don't make full use of that functionality.

bmcfee (changes requested):
Looks basically fine to me, with the exception of the `_wrap` change in filters.py.  Thanks!

librosa/filters.py
godking123: That makes sense, it does look like the get_window function only accepts keyword arguments after the window and Nx variables, as seen by the function signature below.

`def get_window(window: _WindowSpec, Nx: int, *, fftbins: bool = True) -> np.ndarray:`

As such, I decided to only allow **kwargs in _wrap, since any positional args would result in a TypeError. This still lets fftbins pass through transparently, which was the main thing get_window accepts beyond window/Nx anyway.

bmcfee (approved):

bmcfee:
Thanks @godking123 !

